PR #2972, opened 2020-02-26
Fix build without boost serialization.

ricrogz:
The python wrapping of ScaffoldNetworks fails to build when Boost serialization is not available. The reason is that, when serialization is not available, the string constructor for `ScaffoldNetwork` isn't available either, but it is still specified in the python wrapper.

This patch removes the pickling capability from `ScaffoldNetwork` when boost serialization is disabled. I also split off the pickling test to a separate test file that does not get run when serialization is turned off (else, it will fail with a message saying serialization is not available for that class). I also (accidentally) yapf-formatted the original test file.

### DIFF
--- a/Code/GraphMol/ScaffoldNetwork/Wrap/testPickleScaffoldNetwork.py
+++ b/Code/GraphMol/ScaffoldNetwork/Wrap/testPickleScaffoldNetwork.py
@@ -1,0 +1,46 @@
+#
+# Copyright (C) 2019 Greg Landrum and T5 Informatics GmbH
+#  All Rights Reserved
+#
+#  This file is part of the RDKit.
+#  The contents are covered by the terms of the BSD license
+#  which is included in the file license.txt, found at the root
+#  of the RDKit source tree.
+#
+
+import pickle
+import unittest
+
+from rdkit import Chem
+from rdkit import RDConfig
+from rdkit import rdBase
+from rdkit.Chem.Scaffolds import rdScaffoldNetwork
+
+rdBase.DisableLog("rdApp.info")
+
+
+class TestScaffoldNetwork(unittest.TestCase):
+  """
+  RDKit must have been built with Boost serialization support
+  for this test to pass.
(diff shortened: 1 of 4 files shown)


Code/GraphMol/ScaffoldNetwork/Wrap/testPickleScaffoldNetwork.py
```diff
@@ -0,0 +1,46 @@
+#
```
greglandrum: Having this as a separate file is fine, but it could also have been handled using a `skipIf`

ricrogz: I thought about that, but then there is the problem of passing the value of the cmake RDK_USE_BOOST_SERIALIZATION variable to the python script. Last time I did something similar, it implied patching TestRunner.py to turn a command line argument into an env var, which was then picked up by the test...

I thought just splitting off the test to a new file was both easier and more proper.

greglandrum: Ah, right, there's no particularly easy way to detect on the Python side whether or not the support is there.
I have no objection to having the extra test file in this case.

bp-kelley:
For the other libraries, we have added a free function like:

```
RDKIT_SUBSTRUCTLIBRARY_EXPORT bool SubstructLibraryCanSerialize();
```

It isn't perfect, but we could add a function to RDBase that is more global

```
rdBase.HasBoostSerialization()
```

greglandrum:
> For the other libraries, we have added a free function like: ``` RDKIT_SUBSTRUCTLIBRARY_EXPORT bool SubstructLibraryCanSerialize(); ``` It isn't perfect, but we could add a function to RDBase that is more global ``` rdBase.HasBoostSerialization() ```

I like the idea of adding the global function. That's a good one for a separate PR.

PR #2987, opened 2020-03-06
Atomgraphfeat

thegodone:
This is an initial version of our 49 features using in classical Graphs neural networks atomic features.
it's 4x faster compare to python version

#### Any other comments?
We can decide to add more / or custom it in the future


bp-kelley:
@thegodone It looks like a lot of extra stuff got added here.   .DS_Store, file line endings.

Would it be possible to just have the stuff specific to AUTOFEAT?

thegodone:
I can make a new pull with the 5 main files.

PR #2995, opened 2020-03-10
Adding get list of Smiles augmentation

thegodone:
this PR is to simplify and speed up the smile augmentation process in Deep learning.

this Method return an unique list of smiles augmented strings directly to python.


thegodone:
I try another method (std::set)  10-15% faster than python as in this version.

thegodone:
Greg commits a nicer version!

PR #3003, opened 2020-03-11
Allow enhanced stereo to be used in substructure search

d-b-w:
Allows use of enhaced stereochemistry in substructure searching if `SubstructMatchParameters.useEnhancedStereo` is set.

The matching rules are pretty obnoxious, but a synopsis is (where < aka less than means that the left hand side is a substructure of the right hand side)

* achiral < everything, because an achiral query means ignore chirality in the match
* chiral < AND, because AND includes both the chiral molecule and another one
* chiral < OR, because OR includes either the chiral molecule or another one
* OR < AND, because AND includes whichever molecule the OR actually happens to mean.
* one group of two atoms < two groups of one atom, because the latter is 4 different diastereomers, and the former only two of the four.

When the test cases fail, the messages look like this:

    -------------------------------------------------------------------------------
    Enhanced stereochemistry
    AND and OR match their enantiomer
    -------------------------------------------------------------------------------
    /Users/wandschn/Documents/src/rdkit/Code/GraphMol/Substruct/catch_tests.cpp:216
    ...............................................................................

    /Users/wandschn/Documents/src/rdkit/Code/GraphMol/Substruct/catch_tests.cpp:218: FAILED:
    CHECK_THAT( *opposite_mol, IsSubstructOf(*mol_and, ps) )
    with expansion:
    CC[C@@H](F)[C@@H](C)O is not a substructure of CC[C@H](F)[C@H](C)O |&1:2,4|

    /Users/wandschn/Documents/src/rdkit/Code/GraphMol/Substruct/catch_tests.cpp:219: FAILED:
    CHECK_THAT( *opposite_mol, IsSubstructOf(*mol_or, ps) )
    with expansion:
    CC[C@@H](F)[C@@H](C)O is not a substructure of CC[C@H](F)[C@H](C)O |o1:2,4|


Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -15,11 +17,49 @@
 
 #include <GraphMol/RDKitBase.h>
 #include <GraphMol/SmilesParse/SmilesParse.h>
+#include <GraphMol/SmilesParse/SmilesWrite.h>
 #include <GraphMol/Substruct/SubstructMatch.h>
 
 using namespace RDKit;
 typedef std::tuple<std::string, std::string, size_t> matchCase;
 
+class _IsSubstructOf : public Catch::MatcherBase<const ROMol &> {
```
greglandrum: oooo, taking advantage of some of the extra stuff catch2 provides.
Nice!

Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -15,11 +17,49 @@
...
+  return _IsSubstructOf(m, ps);
+}
+
+static _IsSubstructOf IsSubstructOf(const ROMol &m) {
+  return _IsSubstructOf(m);
+}
+
+namespace Catch {
+template <>
+struct StringMaker<RDKit::RWMol> {
```
greglandrum: Why `RWMol` here?

Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -127,4 +167,72 @@ TEST_CASE("providing a final match function", "[substruct]") {
...
+TEST_CASE("Enhanced stereochemistry", "[substruct,StereoGroup]") {
+  // Chirality specifications.
+  // 1. An achiral molecule: CC(O)C(CC)F means unknown/all stereoisomers
+  // 2. A chiral molecule: C[C@H](O)[C@H](CC)F means 1 stereoisomer
+  // 3. A chiral molecule with an AND specifier: C[C@H](O)[C@H](CC)F |a1:1,3|
+  // means both stereoisomers
+  // 4. A chiral molecule with an OR specifier: C[C@H](O)[C@H](CC)F |o1:1,3|
+  // means one of the two stereoisomers
+  auto mol_achiral = "CC(O)C(CC)F"_smiles;
+  auto mol_chiral = "C[C@H](O)[C@H](CC)F"_smiles;
```
greglandrum: this set of tests should also include molecules that have one and both of the stereocenters flipped. Maybe:
```
auto mol_chiral_2flip = "C[C@@H](O)[C@@H](CC)F"_smiles;
auto mol_chiral_1flip = "C[C@H](O)[C@@H](CC)F"_smiles;
```
The behavior of these in the various test cases will also be informative

greglandrum: Ah, I see you have these below already, but I think it's clearer to move them up here and use them throughout the tests.

greglandrum:
I made a few comments inline, but a larger thing is that I'm not convinced the logic is correct.

I'm not sure I have this right, but here's my understanding of what the relationships should be:
image

The general ideas are:
1. Things not specified in the query aren't used in the matching. So the flat query structure is a substructure of everything here.
2.  The stereoisomers the query represents must be entirely contained within the stereoisomers of the target structure. 

Does this seem right?

Code/GraphMol/Substruct/catch_tests.cpp
d-b-w: `""_smiles` gives a `unique_ptr<RWMol>` so I was just blindly pasting. updating to `ROMol`.

Code/GraphMol/Substruct/catch_tests.cpp
d-b-w: 👍

Code/GraphMol/Substruct/catch_tests.cpp
d-b-w: good call.

d-b-w:
Greg - that _does_ seem right. I emailed outside of git, let's talk about this tomorrow.

Code/GraphMol/Substruct/catch_tests.cpp
d-b-w: requires an RWMol to get correct printing.

Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -127,4 +167,94 @@ TEST_CASE("providing a final match function", "[substruct]") {
...
+    CHECK_THAT(*mol_achiral, IsSubstructOf(*mol_chiral, ps));
+    CHECK_THAT(*mol_achiral, IsSubstructOf(*mol_and, ps));
+    CHECK_THAT(*mol_achiral, IsSubstructOf(*mol_or, ps));
+    CHECK_THAT(*mol_achiral, IsSubstructOf(*mol_absolute, ps));
+    CHECK_THAT(*mol_achiral, IsSubstructOf(*diastereomer, ps));
+  }
+  SECTION("chiral molecule is a substructure of AND or OR") {
+    CHECK_THAT(*mol_chiral, !IsSubstructOf(*mol_achiral, ps));
+    CHECK_THAT(*mol_chiral, IsSubstructOf(*mol_and, ps));
+    CHECK_THAT(*mol_chiral, IsSubstructOf(*mol_or, ps));
```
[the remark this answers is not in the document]
d-b-w: thanks, totally an oversight.

Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -127,4 +167,94 @@ TEST_CASE("providing a final match function", "[substruct]") {
...
+    CHECK_THAT(*opposite_mol, IsSubstructOf(*mol_and, ps));
+    CHECK_THAT(*opposite_mol, IsSubstructOf(*mol_or, ps));
+  }
+  SECTION("But not some arbitrary diastereomer") {
+    CHECK_THAT(*diastereomer, !IsSubstructOf(*mol_and, ps));
+    CHECK_THAT(*diastereomer, !IsSubstructOf(*mol_or, ps));
+  }
+  SECTION("Mixed stereo groups include single stereo groups") {
+    auto mol_mixed_or = "C[C@H](O)[C@H](CC)F |o1:1,o2:3|"_smiles;
+    std::cerr << "CHECK_THAT(*mol_mixed_or, !IsSubstructOf(*mol_or, ps));" << std::endl;
```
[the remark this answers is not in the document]
d-b-w: is it obvious that I got confused like 400x about which is the query and which is the superstructure ? :)

Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -127,4 +167,94 @@ TEST_CASE("providing a final match function", "[substruct]") {
...
+    // OR refers to two of the 4 molecules that mol_mixed_or
+    CHECK_THAT(*mol_or, IsSubstructOf(*mol_mixed_or, ps));
+
+    auto mol_mixed_or2 = "C[C@H](O)[C@@H](CC)F |o1:1,o2:3|"_smiles;
+    std::cerr << "CHECK_THAT(*mol_mixed_or2, !IsSubstructOf(*mol_or, ps));" << std::endl;
+    CHECK_THAT(*mol_mixed_or2, !IsSubstructOf(*mol_or, ps));
+    std::cerr << "CHECK_THAT(*mol_or, IsSubstructOf(*mol_mixed_or2, ps));" << std::endl;
+    CHECK_THAT(*mol_or, IsSubstructOf(*mol_mixed_or2, ps));
+
+    // I'm not sure about these ones, but they should be symmetric:
```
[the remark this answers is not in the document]
d-b-w: As soon as I wrote the `mol_mixed_or_and_abs2` test case, I was _pretty_ confident that this was the correct behavior.

Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -127,4 +167,94 @@ TEST_CASE("providing a final match function", "[substruct]") {
...
+    auto mol_mixed_or_and_abs = "C[C@H](O)[C@H](CC)F |o1:1|"_smiles;
+    CHECK_THAT(*mol_mixed_or_and_abs, !IsSubstructOf(*mol_or, ps));
+    CHECK_THAT(*mol_or, !IsSubstructOf(*mol_mixed_or_and_abs, ps));
+
+    auto mol_mixed_or_and_abs2 = "C[C@@H](O)[C@H](CC)F |o1:1|"_smiles;
+    CHECK_THAT(*mol_mixed_or_and_abs2, !IsSubstructOf(*mol_or, ps));
+    CHECK_THAT(*mol_or, !IsSubstructOf(*mol_mixed_or_and_abs, ps));
+  }
+  SECTION("It's OK to match part of a stereo group, though") {
```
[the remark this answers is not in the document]
d-b-w: Added:

    CHECK_THAT(*mol_and_partial, IsSubstructOf(*mol_and_long, ps));
    CHECK_THAT(*mol_or_partial, IsSubstructOf(*mol_or_long, ps));
    CHECK_THAT(*mol_or_partial, IsSubstructOf(*mol_and_long, ps));
    CHECK_THAT(*mol_and_partial, !IsSubstructOf(*mol_or_long, ps));

Code/GraphMol/Substruct/SubstructMatch.cpp
```diff
@@ -131,32 +216,46 @@ class MolMatchFinalCheckFunctor {
...
           static_cast<int>(countSwapsToInterconvert(moOrder, mOrder));
 
-      bool requireMatch = qPermCount % 2 == mPermCount % 2;
-      bool labelsMatch = qAt->getChiralTag() == mAt->getChiralTag();
+      const bool requireMatch = qPermCount % 2 == mPermCount % 2;
+      const bool labelsMatch = qAt->getChiralTag() == mAt->getChiralTag();
+      const bool matchOK = requireMatch == labelsMatch;
 
-      if (requireMatch != labelsMatch) {
+      // if this is not part of a stereogroup and doesn't match, return false
```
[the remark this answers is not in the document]
d-b-w: I don't think so, because it's OK for matchOK to be false if the atom is in a stereogroup. If it's in a stereogroup, what matters is that all atoms in the stereogroup are the same matchiness.  e.g. `C[C@@H](O)[C@@H](CC)F` and `"C[C@H](O)[C@H](CC)F` both match `C[C@H](O)[C@H](CC)F |&1:1,3|`, but `C[C@H](O)[C@@H](CC)F` doesn't.

d-b-w:
I think I've responded to most of the suggestions here. I also added a commit with some tweaks to the documentation.

Docs/Book/RDKit_Book.rst
```diff
@@ -1383,6 +1384,14 @@ The following table captures whether or not a substructure query (in the rows) m
 |      AND        |                 |                 |                 |                 |                 |                 |                 |
 +-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+-----------------+
 
+Substructure search using molecules with enhanced stereochemistry follows these rules (where substructure < superstructure):
```
greglandrum: This explanation is really helpful.

Code/GraphMol/Substruct/SubstructMatch.cpp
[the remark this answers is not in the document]
greglandrum: Right. Got it.

PR #3033, opened 2020-03-25
Fix/add patterns reset holders

bp-kelley:
There were two bugs:

If numThreads was set to 0, no fingerprints were made.  This caused the patterns not to actually be made which causes a seg fault when trying to explicitly get a fingerprint at an index.

```
lib.GetFpHolder().GetFingerprint(0)
```

would seg fault.

### DIFF
--- a/Code/GraphMol/SubstructLibrary/PatternFactory.cpp
+++ b/Code/GraphMol/SubstructLibrary/PatternFactory.cpp
@@ -9,6 +9,7 @@
 //
 #include "SubstructLibrary.h"
 #include "PatternFactory.h"
+#include <RDGeneral/RDThreads.h>
 
 #ifdef RDK_THREADSAFE_SSS
 #include <thread>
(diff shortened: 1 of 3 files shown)


Code/GraphMol/SubstructLibrary/PatternFactory.cpp
```diff
@@ -53,7 +56,11 @@ void addPatterns(SubstructLibrary &sslib, int numThreads) {
 #else
   fillPaterns(lib, fps, 0, sslib.size(), 1);
 #endif
+  if (ptr->size() != sslib.size()) {
+    throw ValueErrorException("Number of fingerprints generated not equal to current number of molecules");
+  }
   
   sslib.getFpHolder() = ptr;
+  sslib.resetHolders();
```
bp-kelley: bug #2

Code/GraphMol/SubstructLibrary/PatternFactory.cpp
```diff
@@ -31,6 +32,8 @@ void fillPatterns(const SubstructLibrary &slib,
   
 void addPatterns(SubstructLibrary &sslib, int numThreads) {
   PRECONDITION(sslib.getFpHolder().get() == nullptr, "Substruct library already has fingerprints");
+  numThreads = (int)getNumThreadsToUse(numThreads);
```
bp-kelley: bug #1

PR #3201, opened 2020-05-30
mol2 was not defined

autodataming:
mol2 was not defined


Docs/Book/Cookbook.rst
```diff
@@ -470,6 +470,7 @@ Create Fragments
    # Chem.FragmentOnSomeBonds() will fragment according to all permutations of numToBreak bonds at a time 
    # (numToBreak defaults to 1), and return tuple of molecules with numToBreak cuts applied. By default, 
    # addDummies=True, so empty valences are filled with dummy atoms:
+   mol2 = Chem.Mol(mol)
    mol2_f_tuple = Chem.FragmentOnSomeBonds(mol2, (0, 2, 4))
```
bp-kelley, line 474: We could also change mol2 to mol1 here, I don't want users to think they need to make a copy of the molecule before calling FragmentOnSomeBonds.

bp-kelley: Nice catch by the way!

PR #3205, opened 2020-06-03
Tautomer search

jones-gareth:

Enhancement to facilitate tautomer invariant substructure search.

- TautomerQuery class for matching tautomers against structures
- Modification to pattern fingerprints to allow screenout for tautomer queries
- Java and Python API


greglandrum:
@jones-gareth : I'm "a bit" puzzled about why the swig tests are failing in CI (they work fine on my linux box) but given that they fail for both linux and the mac, it seems like it's something real.
If you can't look into this let me know and I will try to reproduce at least the linux problems in a docker container.

jones-gareth:
@greglandrum : In the Java test I have used the __var__ declaration which is only valid for java 11. It looks like that is the problem.

```
    var mol = RWMol.MolFromSmiles("O=C1CCCCC1");
        ^
  symbol:   class var
  location: class TautomerQueryTests
```

I guess the test platform is using JDK 8 or 9.  I'll put in a new pull request this morning with fully typed variable declarations.  Hopefully that will fix it.

greglandrum:
@jones-gareth Yep, that's definitely the problem

Code/GraphMol/QueryOps.h
```diff
@@ -619,6 +624,8 @@ RDKIT_GRAPHMOL_EXPORT BOND_EQUALS_QUERY *makeBondOrderEqualsQuery(
     Bond::BondType what);
 //! returns a Query for unspecified SMARTS bonds
 RDKIT_GRAPHMOL_EXPORT BOND_EQUALS_QUERY *makeSingleOrAromaticBondQuery();
+//! returns a Query for tautomeric bonds
+BOND_EQUALS_QUERY *makeSingleOrDoubleOrAromaticBondQuery();
```
greglandrum, line 628: should be:
```suggestion
RDKIT_GRAPHMOL_EXPORT BOND_EQUALS_QUERY *makeSingleOrDoubleOrAromaticBondQuery();
```
This is the cause of the failing windows DLL CI builds

greglandrum:
@jones-gareth I will try and track down whatever problem is causing the windows DLL CI builds to fail tomorrow morning.

jones-gareth:
@greglandrum My local windows build is also failing- but it is a boost problem (what's new) and doesn't seem to be because of the new code.

greglandrum:
@jones-gareth : just did a PR against your branch that fixes the DLL problems

Code/GraphMol/TautomerQuery/TautomerQuery.cpp
```diff
@@ -0,0 +1,259 @@
...
+#endif
+      if (matchTautomer(mol, *tautomer, templateMatch, params)) {
+#ifdef VERBOSE
+        std::cout << "Got Match " << std::endl;
+#endif
+        matches.push_back(templateMatch);
+        if (matchingTautomers) {
+          matchingTautomers->push_back(tautomer);
+        }
+        if (matches.size() == params.maxMatches) goto searchFinished;
```
greglandrum, line 217: oh no... a goto?
As odd as it is to see one of these, particulary in "modern" C++, I can't think of a more straightforward solution to the problem you're solving. Introducing a new boolean solely for the purpose of avoiding this doesn't make a ton of sense.

Code/GraphMol/Fingerprints/Fingerprints.h
```diff
@@ -144,7 +144,8 @@ const unsigned int substructLayers = 0x07;
 RDKIT_FINGERPRINTS_EXPORT ExplicitBitVect *PatternFingerprintMol(
     const ROMol &mol, unsigned int fpSize = 2048,
     std::vector<unsigned int> *atomCounts = nullptr,
-    ExplicitBitVect *setOnlyBits = nullptr);
+    ExplicitBitVect *setOnlyBits = nullptr,
+    bool tautomericFingerprints = false);
```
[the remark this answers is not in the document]
jones-gareth: changed

Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -12,6 +12,7 @@
 #include <GraphMol/MolStandardize/FragmentCatalog/FragmentCatalogUtils.h>
 #include <GraphMol/SmilesParse/SmilesParse.h>
 #include <GraphMol/SmilesParse/SmilesWrite.h>
+#include <GraphMol/SmilesParse/SmartsWrite.h>
```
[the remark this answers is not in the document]
jones-gareth: fixed

Code/GraphMol/TautomerQuery/Wrap/rdTautomerQuery.cpp
```diff
@@ -0,0 +1,145 @@
...
+    auto docString =
+        "The Tautomer Query Class.\n\
+  Creates a query that enables structure search accounting for matching of\n\
+  Tautomeric forms\n";
+
+    python::class_<TautomerQuery, boost::noncopyable>(
+        "TautomerQuery", docString, python::no_init)
+        .def("__init__", python::make_constructor(createDefaultTautomerQuery))
+        .def("__init__", python::make_constructor(&TautomerQuery::fromMol))
+        .def("IsSubstructOf", tautomerIsSubstructOf,
```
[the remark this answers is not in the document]
jones-gareth: I've added bindings to use `SubstructWatchParameters`. Since the keyword functions ware already present, I have followed the example of `MolBundle` and provided both flavors.

Code/GraphMol/TautomerQuery/Wrap/rdTautomerQuery.cpp
```diff
@@ -0,0 +1,145 @@
...
+  std::vector<MatchVectType> matches;
+  std::vector<ROMOL_SPTR> matchingTautomers;
+  SubstructMatchParameters params;
+  params.uniquify = uniquify;
+  params.useChirality = useChirality;
+  params.useQueryQueryMatches = useQueryQueryMatches;
+  params.maxMatches = maxMatches;
+
+  {
+    NOGIL gil;
```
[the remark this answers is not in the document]
jones-gareth: I've done that

Code/GraphMol/TautomerQuery/TautomerQuery.cpp
```diff
@@ -0,0 +1,259 @@
...
+                          ? tautomerTransformFile
+                          : std::string(getenv("RDBASE")) +
+                                "/Data/MolStandardize/tautomerTransforms.in";
+  auto tautomerParams = std::unique_ptr<MolStandardize::TautomerCatalogParams>(
+      new MolStandardize::TautomerCatalogParams(tautomerFile));
+  MolStandardize::TautomerEnumerator tautomerEnumerator(
+      new MolStandardize::TautomerCatalog(tautomerParams.get()));
+  boost::dynamic_bitset<> modifiedAtomsBitSet(query.getNumAtoms());
+  boost::dynamic_bitset<> modifiedBondsBitSet(query.getNumBonds());
+  const auto t2 = tautomerEnumerator.enumerate(query);
```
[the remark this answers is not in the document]
jones-gareth: Fixed

Code/GraphMol/TautomerQuery/TautomerQuery.cpp
jones-gareth: Yes, I was surprised that "modern" C++ does not have labelled loops and this appears to be the best way to break out of an inner loop.

jones-gareth:
@greglandrum I agree that this is almost done.  I had to make some changes to Tautomer.cpp as there may be bond changes on tautomer sanitization (e.g. when the bond changes modify an aromatic ring), in addition to  those caused by the proton moving.
There is a last optimization that I'd like to to.  Currently, all the template matches are found before looking for tautomer matches.  I'd like to have a single functor that checked template matches as they are generated- that should improve boolean match tests in most cases.

greglandrum:
@jones-gareth this looks good to me. do you want to add the functor for checking template matches first or should we go ahead and merge this now and do that in a second pass?

jones-gareth:
May as well wait, rather than open another pull request. I won't be able to
get to it until middle of next week, so if that is an issue merge it now.

jones-gareth:
@greglandrum Does this still look good to you?  I have added a TautomerQueryMatcher class to TautomerQuery.cpp and use the extraFinalCheck mechanism in SubstructMatchParameters to check if one of the tautomers satisfy a template match.  If it is OK we can merge and close the request.

Code/GraphMol/TautomerQuery/TautomerQuery.h
```diff
@@ -35,7 +36,7 @@ class RDKIT_TAUTOMERQUERY_EXPORT TautomerQuery {
 
   // tests if a match to the template matches a specific tautomer
   bool matchTautomer(const ROMol &mol, const ROMol &tautomer,
-                     const MatchVectType &match,
+                     const std::vector<unsigned int> &match,
```
greglandrum, line 39: why did you switch from `MatchVectType` here? I know it's not ideal, but it's at least somewhat consistent.

jones-gareth: I'm using an extraFinalCheck method in the parameters for template substructure search to see if any of the tatutomers actually match.  The signature for that uses a vector rather than MatchVectType, so I changed it in matchTatutomer (the method is private and this is the only usage)

greglandrum: makes sense

greglandrum:
Thanks @jones-gareth !

PR #3298, opened 2020-07-20
ANI NN as ForceField Integration Forward Pass [WIP]

manangoel99:
[Note that this PR is not intended to be reviewed. It is a check of the new code against the CI system]

Implementation of ANI style neural networks as a Force Field to calculate the energy of the molecule
- Generating Atomic Environment Vectors from atomic positions
- Load pre-trained weights of the NN obtained from torchani into Eigen matrices
- Forward pass of AEV through the ensembled model to obtain atomic contributions whose sum is the total energy
- Built a python wrapper for the ANI ForceField
- Currently supports ANI-1x and ANI-1ccx


Code/GraphMol/Descriptors/AtomicEnvironmentVector.cpp
```diff
@@ -0,0 +1,845 @@
...
+  \param cutoff    A double value signifying cutoff distance
+
+  \return 2 dimensional array containing corresponding values computed by cutoff
+  function
+*/
+template <typename Derived>
+ArrayXXd CosineCutoff(ArrayBase<Derived> *distances, double cutoff) {
+  // Cosine cutoff function assuming all distances are less than the cutoff
+  PRECONDITION(cutoff > 0.0, "Cutoff must be greater than zero");
+  PRECONDITION(((*distances) <= cutoff).count() == distances->size(),
```
bp-kelley, line 36: Add a PRECONDITION for distances not being null
PRECONDITION(distances, "null distances passed in")

Code/GraphMol/Descriptors/AtomicEnvironmentVector.cpp
```diff
@@ -0,0 +1,845 @@
...
+               "All distances must be less than the cutoff");
+  PRECONDITION(distances != nullptr, "Array of distances is NULL");
```
bp-kelley, line 38: This needs to be above the dereference above.

Code/GraphMol/Descriptors/AtomicEnvironmentVector.cpp
```diff
@@ -0,0 +1,845 @@
...
+
+  return species;
+}
+
+VectorXi GenerateSpeciesVector(const int *atomNums, unsigned int numAtoms) {
+  PRECONDITION(atomNums != nullptr, "Array of atom types is NULL")
+  VectorXi species(numAtoms);
+
+  for (unsigned int i = 0; i < numAtoms; i++) {
+    switch (atomNums[i]) {
```
bp-kelley, line 81: This duplicates the switch statement from above, it seems we should have this logic in one place.

Code/GraphMol/Descriptors/AtomicEnvironmentVector.cpp
```diff
@@ -0,0 +1,845 @@
...
+  for (auto i = 0; i < upperTriag.rows(); i++) {
+    for (auto j = 0; j < upperTriag.cols(); j++) {
+      upperTriagFlattened(index) = upperTriag(i, j);
+      index++;
+    }
+  }
+
+  // Select indices in upperTriagFlattened from coordinates
+  ArrayXXd pairCoordinates(numPairs * 2, 3);
+  IndexSelect(&pairCoordinates, coordinates, upperTriagFlattened, 0);
```
bp-kelley, line 185: If IndexSelect used references IndexSelect(ArrayBase<Derived> &vector1, ...) this would be safer code and would remove the nullptr PRECONDITIONs.

Code/GraphMol/Descriptors/AtomicEnvironmentVector.cpp
```diff
@@ -0,0 +1,845 @@
...
+  return 0.5 * ((*distances) * (M_PI / cutoff)).cos() + 0.5;
+}
+
+VectorXi GenerateSpeciesVector(const ROMol &mol) {
+  // Generate atom species vector as mentioned in torchani
+  // H : 0
+  // C : 1
+  // N : 2
+  // O : 3
+  // All other atoms : -1
```
bp-kelley, line 48: All other atoms actually throw a ValueErrorException

PR #3327, opened 2020-08-07
Some work on TautomerEnumerator

ptosco:
Summary of changes in this PR:

* There was no way of passing `CleanupParameters` to the C++ `TautomerEnumerator` upon class instantiation. For this reason, the `maxTautomers` parameter, though present in the CleanupParameters structure, was not honored and was hardcoded to 1000. I have added a `TautomerEnumerator()` C++ constructor that allows passing an optional `CleanupParameters` object,
which defaults to the default `CleanupParameters`. The Python constructor already had this possibility, so no change required.

* I have added a `maxTransforms` parameter to `CleanupParameters` to enable early termination of the enumeration if a maximum number of transforms is reached. This option was originally present in the algorithm described by Sitzmann et al., and was indicated as a better termination criterion as it leads to more linear scaling of CPU time.

* I have added a callback (callable from both C++ and Python) in analogy to other potentially time consuming algorithms (e.g., `ResonanceMolSupplier` and `FMCS`) to allow monitoring/early termination.

* I have added two configurable parameters, i.e. `tautomerRemoveSp3Stereo` and `tautomerRemoveBondStereo`, which were present in the original algorithm described by Sitzmann et al. `tautomerRemoveSp3Stereo` controls whether chirality should be removed from `sp3` centers involved in tautomerism. A typical example would be L-aminoacids: as they can all exist as enolic tautomers, the canonical tautomer would come back as non-chiral, which is most likely undesirable, as it is well known that aminoacids are actually stereochemically stable as such because the enolic tautomer is very unfavorable, unless an activated ester is formed and subjected to acidic or alkaline conditions. In the original canonicalization workflow described by Sitzmann et al. `tautomerRemoveSp3Stereo` was set to `false`. For now I have kept it to `true` also in the RDKitfor backwards compatibility, but I think setting the default to `false` would be a good thing to avoid unintended racemization of stable chiral centers.
`tautomerRemoveBondStereo` has the same purpose for the stereochemistry of double bonds involved in tautomerism.
In this case the recommendation from Sitzmann et al. was to set this parameter to `true`, which is the current default also in the RDKit.

* I have fixed a bug where the same tautomer could appear multiple times among generated tautomers. E.g., the L-alanine canonical tautomer would appear twice, once with a chiral SMILES (the original entry), and once with a non-chiral SMILES (the one originating from the transformation to enol and then reverse transformation to keto form), both of which however point to identical and non-chiral molecules. To avoid this issue, now the SMILES key is updated in the `std::map` after `sp3` chirality information is removed from the respective tautomer.

* `TautomerEnumerator::enumerate()` now returns a `TautomerEnumeratorResult` object. This avoids having to pass pointers to `dynamic_bitset` in C++ (and empty lists that get filled with indices in Python, which are not very pythonic) to retrieve modified atoms and bonds. Additionally, a `status` parameter can be accessed to check whether the algorithm went to completion or terminated early due to some limit being reached or to cancellation by the callback.

* I have added back a missing transform that was commented out in `tautomerTransforms.in` as well as in the original MolVS codebase. Going back to the paper, I noticed that a double bond between C3 and S4 was missing in Rule 19 on page 529. Other than that, the SMIRKS in the paper was actually correct, and so was the SMARTS in `tautomerTransforms.in`, apart from a typo in the bonds column.

* I have done a general code cleanup and optimization; in particular:
- I have made smart pointers for molecules consistent throughout (all `ROMOL_SPTR`)
- It is not necessary to generate a vector of index pairs with the `detail::pairwise` function 
- It is not necessary to maintain two `std::map` objects (one for tautomers, one for kekulized forms) as they can be stored in a single `std::map`, thus halving the number of insertions
- For the purpose of canonicalization, it is not necessary to call `assignStereochemistry` on each tautomer, which is a quite expensive operation. In fact, E/Z and R/S descriptor calculation does not affect either the SMILES string generation (which only depends on parity or cis/trans isomerism, neither of which is affected by tautomerism) or the scoring, both of which concur in determining the tautomer ranking
- A smaller set of sanitizations can be used after the transformation have been applied to the kekulized molecule

The overall result of these optimizations is that canonicalization is now 25-30% faster as it used to be before my changes. 
I hope I haven't forgotten anything - please do get in touch if you have questions.


Code/GraphMol/MolStandardize/MolStandardize.cpp
```diff
@@ -155,14 +155,12 @@ std::vector<std::string> enumerateTautomerSmiles(
   cleanup(*mol, params);
   MolOps::sanitizeMol(*mol);
 
-  auto *tautparams = new TautomerCatalogParams(params.tautomerTransforms);
-  //	unsigned int ntautomers = tautparams->getNumTautomers();
-  TautomerEnumerator te(new TautomerCatalog(tautparams));
+  TautomerEnumerator te(params);
 
-  std::vector<ROMOL_SPTR> res = te.enumerate(*mol);
+  auto res = te.enumerate(*mol);
```
bp-kelley, line 160: This looks to be a breaking change. I know it will affect some of my code that uses the Tautomer enumerator.

Code/GraphMol/MolStandardize/Wrap/testMolStandardize.py
```diff
@@ -228,9 +229,9 @@ def test13Tautomers(self):
     ctaut = enumerator.Canonicalize(m)
     self.assertEqual(Chem.MolToSmiles(ctaut), "O=C1CCCCC1")
 
-    tauts = enumerator.Enumerate(m)
-    self.assertEqual(len(tauts), 2)
-    ctauts = list(sorted(Chem.MolToSmiles(x) for x in tauts))
+    taut_res = enumerator.Enumerate(m)
```
bp-kelley, line 232: In python land, you might be able to make "taut_res" itself iterable, this would perhaps prevent breaking code in python land.

You could also do the same in C++ land by subclassing std::vector<ROMOL_SPTR> but that has its own issues.

bp-kelley:
Take my "request" changes with a grain of salt.  I would like the Python code, at least, to be backwards compatible but that may not be possible/desirable here.

Code/GraphMol/MolStandardize/MolStandardize.cpp
ptosco: It is, and I really don't like it (I have already been yelled at by @greglandrum ;-) ), but adding another optional `TautomerEnumeratorStatus*` parameter to `TautomerEnumerator::enumerate()` is not a very exciting alternative either. I had thought of making `TautomerEnumeratorResult` iterable; as you support this idea too I'll give it a go tomorrow then we may decide what we dislike less. Thanks for reviewing!

greglandrum: I'm not fond of the backwards compatibility break either. My argument for why I think it's worth doing anyway is primarily that returning the results structures really is a lot cleaner and the enumeration code has only been out there for one release, so there hopefully won't be too much code that actually breaks.
Still, it's probably not that hard to shift me back to a: "let's not break backwards compatibility" position if you think it's important @bp-kelley

ptosco: I am working on a solution that I like much better: I hope you guys will like it too  :-)

ptosco:
@bp-kelley `TautomerEnumeratorResult` is now iterable under both C++ and Python.
In C++, if in your previous code you had something like:
```
for (auto t: te.enumerate()) {
  [...]
}
```
It will still work.
If you had explicit assignment to `std::vector` you will either need to switch to `auto` or change your code from
```
std::vector<ROMOL_SPTR> tautomers = te.enumerate();
```
to
```
std::vector<ROMOL_SPTR> tautomers = te.enumerate()();
```

While I was addressing your comments I started thinking of my own use cases where I need the tautomers as both `Mol` and SMILES strings, and I thought it was a waste to recompute all SMILES when they have already been computed by the algorithm.
This is why I decided to store the `SmilesTautomerMap` in `TautomerEnumeratorResult` to make it more versatile.
I also added a `TautomerEnumerator.PickCanonical()` Python wrapper for the use case of picking the canonical tautomer from a pre-generated iterable of `Mol`, which was only available from C++.

I think this constitutes a significant improvement from the first version of the PR; i hope you guys (@greglandrum, @bp-kelley) agree with me :-)

ptosco:
The macOS build timed out at the 184th/185 tests. I’d say that counts as a pass :-)

greglandrum:
@ptosco : it's going to take me a bit to get through this

greglandrum:
@jones-gareth since I know you're also interested in/paying attention to the tautomer enumeration code: do you have time to take a quick look at this and see if you agree with the changes Paolo is proposing?

Code/GraphMol/MolStandardize/MolStandardize.cpp
jones-gareth: The `TautomerEnumeratorResult` structure is definitely an improvement.  Now it is iterable it should at least not break code that uses auto.

One way to preserve backwards compatibility would be to name the rewritten `enumerate` method something like `enumerateToResult` then have

```cpp
std::vector<ROMOL_SPTR> TautomerEnumerator::enumerate(const ROMol &mol) {
  auto result = enumerateToResult(mol);
  return result.tautomers();
}
```

Code/GraphMol/MolStandardize/Tautomer.h
```diff
@@ -72,27 +281,58 @@ class RDKIT_MOLSTANDARDIZE_EXPORT TautomerEnumerator {
...
-
-  //! returns the canonical tautomer from a set of possible tautomers
-  /*!
-    Note that the canonical tautomer is very likely not the most stable tautomer
-    for any given conditions. The default scoring rules are designed to produce
-    "reasonable" tautomers, but the primary concern is that the results are
-    canonical: you always get the same canonical tautomer for a molecule
-    regardless of what the input tautomer or atom ordering were.
+  TautomerEnumeratorResult enumerate(const ROMol &mol,
+                                     bool reassignStereo = true) const;
```
jones-gareth, line 285: Looks to me that it would be good to add `reassignStereo`  to `CleanupParameters`

Code/GraphMol/MolStandardize/Tautomer.h
```diff
@@ -40,27 +43,233 @@ inline int scoreTautomer(const ROMol &mol) {
...
+
+ private:
+  void fillTautomersItVec() {
+    for (auto it = d_tautomers.begin(); it != d_tautomers.end(); ++it) {
+      d_tautomersItVec.push_back(it);
+    }
+  }
+  // the enumerated tautomers
+  SmilesTautomerMap d_tautomers;
+  // status of the enumeration: did it complete? did it hit a limit?
```
jones-gareth, line 162: Minor point, this comment belongs to `d_status` not `d_tautomersItVec`

jones-gareth:
@greglandrum I like these changes.  In addition to the algorithmic improvements the `TautomerEnumeratorResult` and `CleanupParameters` (though I guess I would call it `TautomerEnumeratorParameters`) structures are much nicer.  Not sure how important the signature change to enumerate() is, but it is certainly frustrating when you update a library and your code breaks.

Code/GraphMol/MolStandardize/Tautomer.h
ptosco: @jones-gareth Hi Gareth, the reason why I decided not to do that is that it really only affects the enumeration, not the canonicalisation (the canonical tautomer is reassigned stereo anyway). My thinking was that adding it to the `CleanupParameters` could be confusing. However I am happy to revisit this, particularly as it defaults to `true`. @greglandrum What are your thoughts on this?

Code/GraphMol/MolStandardize/Tautomer.h
ptosco: Sure, I'll fix that.

Code/GraphMol/MolStandardize/MolStandardize.cpp
ptosco: @jones-gareth That's certainly doable, even though it makes it less discoverable. @greglandrum Are you happy for me to go ahead with this change?

ptosco:
@jones-gareth `CleanupParameters` is also used by other `MolStandardize` bits (e.g., charge neutralization, metal disconnection, etc.) so I cannot rename it. This is why I prefixed tautomer-relevant settings with `tautomer`.

Code/GraphMol/MolStandardize/Tautomer.h
greglandrum: Since we have the other tautomer options in `CleanupParameters` I think it's consistent to include `tautomerReassignStereo` in `CleanupParameters`

ptosco: @greglandrum Thanks Greg, will do that and remove it from `enumerate()` then.

Code/GraphMol/MolStandardize/MolStandardize.cpp
ptosco: @greglandrum @bp-kelley What about Gareth's suggestion to call the new method `enumerateToResult()` and leave the old `enumerate()` method returning a `ROMOL_SPTR` to avoid breaking existing code?
It is worth noting that to fix existing broken C++ builds it is sufficient to replace `ROMOL_SPTR` with `auto` or replace `enumerate()` with `enumerate()()`/`enumerate().tautomes()`. I could add a `#pragma` in the code to make it more obvious when looking at failed CI logs.
Waiting for your opinion here. Thanks in advance!

greglandrum:
@ptosco Since the API change also makes me nervous I just did a PR against your branch that breaks the C++ API a "bit less" and allows anyone using the API from C++ to easily make their code work again without having to do real work.
Let me know what you think.

Code/GraphMol/MolStandardize/MolStandardize.cpp
ptosco: @greglandrum @bp-kelley OK, Greg provided a good solution to the problem. I am finalizing this PR now and this should be ready for approval soon.

ptosco:
@greglandrum This should be good to go once tests have passed.

greglandrum:
@bp-kelley : the current state of affairs with this PR is that python code should continue to work unless you're using the `modifiedAtoms` or `modifiedBonds` arguments or, for some reason, depending on the type of the result from `enumerate()`:
```
In [2]: m = Chem.MolFromSmiles('OC=Cc1[nH]cnc1')

In [3]: from rdkit.Chem.MolStandardize import rdMolStandardize

In [4]: te = rdMolStandardize.TautomerEnumerator()

In [5]: tauts = te.Enumerate(m)

In [6]: for m in tauts:
   ...:     print(Chem.MolToSmiles(m))
   ...: 
O=CCc1c[nH]cn1
O=CCc1cnc[nH]1
OC=Cc1c[nH]cn1
OC=Cc1cnc[nH]1
OCC=C1C=NC=N1

In [7]: Chem.MolToSmiles(tauts[0])
Out[7]: 'O=CCc1c[nH]cn1'

In [8]: type(tauts)
Out[8]: rdkit.Chem.MolStandardize.rdMolStandardize.TautomerEnumeratorResult
```
C++ code that either uses `auto` and iterates over the results *or* that uses the optional arguments for modifiedAtoms and modifiedBonds will continue to work.

I'm also not a fan of changing the API, but this feels like a pretty small change and we have the additional consolation that it's in code which was completed in the last release cycle, so at least we're not breaking a ton of code that's been around for years.

What's your take?

greglandrum (approved):
LGTM

ptosco:
@bp-kelley Hi Brian, sorry to bother. Would you have a bit of time to look at Greg's comment here:
https://github.com/rdkit/rdkit/pull/3327#issuecomment-694209972
This PR has been around for a while and it would be good to merge it now that it has gone through a few iterations. Thanks!

Code/GraphMol/MolStandardize/Tautomer.h
```diff
@@ -40,29 +43,247 @@ inline int scoreTautomer(const ROMol &mol) {
 }
 }  // namespace TautomerScoringFunctions
 
+enum TautomerEnumeratorStatus {
```
bp-kelley, line 46: I would prefer an enum class over a plain C enum

Code/GraphMol/MolStandardize/Tautomer.h
```diff
@@ -40,29 +43,247 @@ inline int scoreTautomer(const ROMol &mol) {
...
+    reference operator*() const { return d_it->second.tautomer; }
+    bool operator==(const const_iterator &other) const {
+      return (d_it == other.d_it);
+    }
+    bool operator!=(const const_iterator &other) const {
+      return !(*this == other);
+    }
+    const_iterator operator++(int) {
+      const_iterator copy(d_it);
+      ++*this;
```
bp-kelley, line 98: This doesn't look to do what is advertised.  In fact It is unclear that is happening.  You are returning the unincremented copy, but advancing the internal copy (and not returning it).

It appears that SmilesTautomerMap+=n is the same as SmilesTautomerMap+=1 which appears to be different than SmilesTautomerMap++;

Code/GraphMol/MolStandardize/Tautomer.h
```diff
@@ -40,29 +43,247 @@ inline int scoreTautomer(const ROMol &mol) {
...
+
+//! Contains results of tautomer enumeration
+class RDKIT_MOLSTANDARDIZE_EXPORT TautomerEnumeratorResult {
+  friend class TautomerEnumerator;
+
+ public:
+  class const_iterator {
+   public:
+    typedef ROMOL_SPTR value_type;
+    typedef std::ptrdiff_t difference_type;
```
bp-kelley, line 82: value_type, difference_type aren't used are they?

Code/GraphMol/MolStandardize/Tautomer.h
bp-kelley: If we really need random access, perhaps std::next is what you want.

If we don't need random access, I think instead of the const_iterator class we could just return

```
const SmilesTautomerMap::const_iterator TautomerEnumeratorResult::begin() {
  return d_tautomers.begin();
}
```

bp-kelley (changes requested):

bp-kelley:
@ptosco I like the changes but the const_iterator++(int) looks really suspicious.  Could you have a second look at this?

Code/GraphMol/MolStandardize/Tautomer.h
ptosco: Good point, done.

Code/GraphMol/MolStandardize/Tautomer.h
ptosco: I am pretty sure it does what is advertised.
`const_iterator &operator++()` is the standard pre-increment operator, which increments the iterator and returns itself.
`const_iterator operator++(int)` is the standard post-increment operator, The dummy `int` argument is needed to differentiate the post-increment operator from the pre-increment one, but it is not actually used: https://en.cppreference.com/w/cpp/language/operator_incdec
The post-increment operator makes a copy of `this` first, as the iterator should not change in the returned copy.
Then, it calls the pre-increment operator on `this`, which actually increments the iterator.
Finally, it returns the copy with the original iterator. The net result is that the iterator is increment, but the non-incremented version of the iterator is returned, which is the expected behavior of the post-increment operator.
I cannot see anything wrong with it.

The `+=` operator is not supported at all, as this is not a random access iterator.
It used to be a simple `input_iterator`; as the underlying `std::map::const_iterator` supports bidirectional iterators, I added the pre- and post-decrement operator and made it a `bidirectional_iterator`, and added relevant test to prove that it does what it is supposed to do.

We cannot simply return `d_tautomers.begin()` as then we would be iterating on key, value pairs, where key is `std::string` and value is `Tautomer`. Instead, to maintain backwards compatibility we need to iterate on `ROMOL_SPTR`.

I hope the above makes sense to you.

Code/GraphMol/MolStandardize/Tautomer.h
ptosco: They aren't, but they are part of the `std::iterator_traits` that should be defined for every iterator:
http://www.cplusplus.com/reference/iterator/iterator_traits/
This ensures, for example, that `std::distance` can be used on this iterator; I have just added tests for that.

Code/GraphMol/MolStandardize/Tautomer.h
bp-kelley: @ptosco - ah I had forgotten that ++(int) returns the current value and then applies the postfix.  The (int) was to differentiate this from the postfix with no return.  Obvious, eh? :)

bp-kelley: What I'm confused by is that ++(int) essentially does post fix on:

```
++(*this);
```
which is
```
++d_it->second.tautomer
```

and ++() does
```
++d_it;
```

which are not the same thing.

ptosco: Nope, that's not what happens.
Doing:
```
++*this;
```
is equivalent to doing:
```
operator++();
```
i.e., the post-increment operator calls the pre-increment one after having copied itself in order to return a non-incremented copy of itself.
You can see that very clearly adding some debugging `std::cerr`s:
```
    const_iterator operator++(int) {
      const_iterator copy(d_it);
      std::cerr << "I am the post-increment operator and now I'll call my pre-increment mate" << std::endl;
      ++*this;
      std::cerr << "I am the post-increment operator and I have just called my pre-increment mate" << std::endl;
      return copy;
    }
    const_iterator &operator++() {
      std::cerr << "I am the pre-increment operator" << std::endl;
      ++d_it;
      return *this;
    }
```
Then when the post-increment operator is called you see the following:
```
1) testTautomerEnumeratorResult_const_iterator calling post-increment
I am the post-increment operator and now I'll call my pre-increment mate
I am the pre-increment operator
I am the post-increment operator and I have just called my pre-increment mate
```

bp-kelley: Got it, I was reading

```
 ++*(*this)
```

not

```
 ++(*this)
```

That's some confusing syntax.  Perhaps

```
this->operator++();
```

Would be easier to read.

bp-kelley: In any case, sorry about my slow comprehension here.

ptosco: @bp-kelley I have applied your suggested change. I guess we are good to merge once tests have passed.

PR #3345, opened 2020-08-18
Fix Cookbook formatting and added 4 new examples

vfscalfani:
## What does this implement/fix? Explain your changes.

I fixed a variety of typos, formatting, references, and links in the Cookbook. I then added four new examples, including:

1. hybridization types and counts
2. sidechain/core enumeration
3. reaction fingerprints and similarity calculation
4. reordering atoms by canonical rank

All doctests passed with RDKit 2020.03.3 and sphinx 3.1.0 (maybe not relevant since you check automatically...)

#### Any other comments?

Probably a good idea for someone to review my code comments, if possible, in the reordering atoms examples. I was not familiar with this and so had to infer what was going on. I think I got it right though.

Thanks
Vin


Docs/Book/Cookbook.rst
```diff
@@ -502,7 +555,71 @@ Create Fragments
...
+   from rdkit import Chem
+   from rdkit.Chem import Draw
+
+.. testcode::
+   
+   core = Chem.MolFromSmiles('*c1c(C)cccc1(O)')
+   chain = Chem.MolFromSmiles('CN*')
+   rxn = AllChem.ReactionFromSmarts('[c:1][#0].[#0][*:2]>>[c:1]-[*:2]')
+   ps = rxn.RunReactants((core,chain))
+   print(Chem.MolToSmiles(ps[0][0],isomericSmiles=True))
```
christopheryoung, line 580: I wonder if this leaves `ps` a little mysterious. Is `ps[0]` something worth printing out in some way? Can the internal structure of `ps` be hinted at in the recipe description?

d-b-w: Maybe it would help to change the variable name to:

    product_lists = rxn.RunReactants((core,chain))
    print(Chem.MolToSmiles(product_lists[0][0]))

(also `isomericSmiles=True` is the default)

christopheryoung: What about:
```
descriptive_name = product_lists[0]
other_descriptive_name = descriptive_name[0]
print(Chem.MolToSmiles(other_descriptive_name))
```
where of course we choose actually descriptive names for the variables?

vfscalfani: Thanks for the fast review and feedback! I actually just found a similar example in the API docs, and I really like this format below. Or at least it makes the most sense to me. 

```
# core is '*c1c(C)cccc1(O)'
# chain is 'CN*'
rxn = AllChem.ReactionFromSmarts('[c:1][#0].[#0][*:2]>>[c:1]-[*:2]')
reacts = (Chem.MolFromSmiles('*c1c(C)cccc1(O)'),Chem.MolFromSmiles('CN*'))
products = rxn.RunReactants(reacts) # tuple
len(products)
```
1

`len(products[0])`

1

`print(Chem.MolToSmiles(products[0][0])) # [0][0] to index out the rdchem mol object`

'CNc1c(C)cccc1O'


I'll go ahead and update the PR with this in a few minutes, but I'm happy to revise again if you disagree.

christopheryoung: My thought was that it might be nice to pull `products[0]` and `products[0][0]` out into variables since that gives us an opportunity to name them before we print the second value. Whether that makes sense depends on whether the intermediate names would be obvious already to someone reading the cookbook. Not sure if that's so or not.

PR #3407, opened 2020-09-12
Github3391

DavidACosgrove:
#### Reference Issue
Fixes 3391

in drawMolecules, the min and max font sizes weren't being set in a timely manner, which meant that the scale wasn't being set correctly.


DavidACosgrove:
Deleting this as it was a merge failure.  Have replaced with a new PR.

PR #3408, opened 2020-09-14
ANI implementation based on that of Manan Goel

jminuse:
#### Reference Issue
Related to #3298 , #3348 

This code implements the ANI models, currently ANI-1x and ANI-1ccx, in rdkit. The models are added as a ForceFieldContrib subclass, so they work with the existing ForceField code, including geometry optimization.

This is a fixed version of a previous rdkit pull request to add ANI neural networks as a force field, as seen here: https://github.com/rdkit/rdkit/pull/3348
This previous pull request had accurate energies, but did not have working gradients, so it used finite differences to approximate the gradients. The new version with analytical gradients is considerably faster.

#### Any other comments?

1) For parameter loading, this implementation reads Eigen matrices from files using EigenSerializer, which it adds in the directory Numerics/EigenSerializer/. I'm not sure this is the most economical way to do this: it seems like a lot of code to add for this simple functionality.

2) As implemented here, model metaparameters are constants, rather than being read from the provided files. I did this because all current ANI variants (including the unreleased ANI-2x and the proprietary Schrodinger-ANI) use the same metaparameters, and because the rdkit-provided files (AEVParams/) do not actually contain all of the metaparameters needed, making the flexibility of reading from files both unnecessary and insufficient. (The missing metaparameters are the radial and angular cutoffs and the radial and inner-product prefactors.)


greglandrum:
Hi @jminuse, thanks for the contribution.
It looks like the failures in the CI builds are real, could you please take a look at those?

greglandrum:
@manangoel99 : could you please take a look at this?

manangoel99:
@greglandrum I'll take a look

jminuse:
With the second commit, these tests now pass on my local build (Mac).

bp-kelley:
@greglandrum @jminuse I can run valgrind on the cpp test if you need it, I'll be back from kid drop off in an hour or so.

bp-kelley:
I'm running valgrind now,.

Code/ForceField/ANI/AtomicContrib.cpp
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  // load parameters by model number in ensemble, by element, and by layer
+  int model_i = 0;
+  while (true) {
+    // get ensembleSize by counting files in param dir
+    std::string paramFile =
+        paramDir + "/model" + std::to_string(model_i) + ".bin";
+    FILE *fp = fopen(paramFile.c_str(), "rb");  // check that file exists
+    fclose(fp);
```
bp-kelley, line 103: You could be calling fclose on a null ptr, move this after the break.

bp-kelley (changes requested):

Code/ForceField/ANI/AtomicContrib.cpp
jminuse: Good point. I was worried about leaving it open, but of course if it's null then it's not open by definition.

jminuse: Moving fclose() seems to have fixed the Ubuntu test failures. Nice catch!

If I'm reading the Azure output correctly, the Mac test has now failed in the build step, and not at a file I recognize.

```
[ 91%] Linking CXX executable moldraw2DRxnTest1
clang-10: error: unable to execute command: Segmentation fault: 11
clang-10: error: linker command failed due to signal (use -v to see invocation)
make[2]: *** [Code/GraphMol/MolDraw2D/moldraw2DRxnTest1] Error 254
make[1]: *** [Code/GraphMol/MolDraw2D/CMakeFiles/moldraw2DRxnTest1.dir/all] Error 2
```

My Mac build process prints this perhaps-related but non-fatal message:

```
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: /Users/stevenso/Downloads/rdkit/manangoel/rdkit/lib/libRDKitMolDraw2D_static.a(MolDraw2DJS.cpp.o) has no symbols
```

In case it matters, my cmake command is:
```
cmake -DPYTHON_INCLUDE_DIR=$PYROOT/include/python3.8/  \
  -DRDK_BUILD_AVALON_SUPPORT=OFF \
  -DRDK_BUILD_CAIRO_SUPPORT=ON \
  -DRDK_BUILD_INCHI_SUPPORT=ON \
  -DBoost_NO_BOOST_CMAKE=ON ..
```

bp-kelley: Whoa, that's a clang linker segment fault.  I think that's a random error...

jminuse: Is there a way to retry the build without pushing an empty commit?

lilleswing: @bp-kelley anything we can do to test if it is a random error?

bp-kelley: There might be a way to restart a build, but since we squash pushing a commit isn’t too bad.

isayev:
Hey, everyone! ANI co-author is here.  So exciting to see our methods in RDKit! Thanks, @jminuse, this is **AWESOME**! 

BTW, ANI-2x model is available for some time now. In fact, we have a Model Zoo repo https://github.com/aiqm/ani-model-zoo
It would be nice to be able to update models as they appear.

Quick side note. For future compatibility, is there a way to read SAE and AEV parameters directly from the model file instead of hardcoded? Those are technically hyperparameters and could change any new release with the data.

bp-kelley:
@greglandrum restarted the osx  build, if osx doesn't pass, I don't mind if we merge after greg and I have one more once over.

jminuse:
I am curious about the file format in which the ANI parameters are stored. What is the standard it conforms to? The files have extension ".bin" and seem to consist of a binary format with interleaved std::string labels and std::vector data. This requires a specialized function deserializeAll (in EigenSerializer.cpp) to read. Currently, this function makes explicit reference to weights and biases, making it neural network specific - if we include it in rdkit/Code/Numerics, it should be at least generalized a bit.

jminuse:
I have modified serializeAll() and deserializeAll() in EigenSerializer, from ANI-specific functions into general-purpose functions for reading files of this format (interleaved std::string labels and Eigen matrices). The Azure tests now pass on all platforms except MacOS_10_13_x64, which gives the following:
	 29 - ANIForceFieldCatchTest (SEGFAULT)
	 30 - pyForceFieldANI (SEGFAULT)
	118 - ANIForceFieldBuilderTest (SEGFAULT)
My local Mac build does not reproduce these failures, even if I `git clone` an entirely fresh repo and build from scratch. Given this, I may push a few more test commits in order to run the tests on Azure.

jminuse:
> It's bz2 archive ;)

Sorry, I did not mean your original parameter files (.nnf, .wparam and .bparam) - those are fine. I meant the ones that ship with rdkit and that @manangoel99 used (.bin files).

jminuse:
@greglandrum maybe you can weigh in here. It makes sense to use a binary parameter file for fast loading, and StackOverflow seems to agree that Boost does not have a built-in portable binary serializer. Therefore, making a binary format with alternating string labels and Eigen matrices, and adding code to read/write it, is one reasonable solution. But I am not familiar enough with Boost or rdkit to assert that this is not duplicative of something already present.

manangoel99:
Hey guys, apologies for the late response.
@jminuse Thanks for implementing the back prop. I really want to know if you had to figure out the derivatives manually or did tapenade work out? 
Another thing is that portable_binary_archive files in ```Code/Numerics/EigenSerializer``` were taken from https://www.boost.org/doc/libs/1_67_0/libs/serialization/example/portable_binary_archive.hpp and I believed it would make it portable to windows as well but that didn't seem to work so I think we can look in that direction or maybe read/write the weights into a more popular object like a 2D STL vector which might have more support and convert that to an ```Eigen::MatrixXf``` object. This would add some latency though but might make it more portable.
Eigen also supports broadcasting to a certain extent so I think that can be done in a couple of places in the forward prop and back prop. I think that might improve the speed even further. Please let me know if I can help implement some of this!

thegodone:
Hi all,

@jminuse <https://github.com/jminuse> you've mentioned it's faster with
gradients.

Can you give more details on how faster is it ?

very good contribution by the way.

BR

Guillaume

Le lun. 21 sept. 2020 à 07:57, Manan Goel <notifications@github.com> a
écrit :

jminuse:
> Can you give more details on how faster is it ?

@thegodone, the time for geometry optimization of a methane molecule is reduced from ~5s to ~0.5s. As the system grows, the speedup grows, because I removed a scaling factor of N_atoms^2 from the run time. One factor of N_atoms comes from replacing finite differences (calculates energy 3*N_atoms times) with analytic gradients (takes about as long as 3 energy evaluations). Another factor of N_atoms comes from calculating the AEV for the molecule once, rather than once for each atom.

> I really want to know if you had to figure out the derivatives manually or did tapenade work out?

@manangoel99, for the AEV Jacobian, I used Tapenade C-to-C automatic differentiation by simplifying the AEV code down to a single angular slot (removing loops and control flow) before using Tapenade to get its gradient in C. Then I cleaned that angular gradient code up and merged it back into the main code (with loops over atoms and hyperparameters) using the chain rule, sum rule, etc.

I wrote the matrix backprop with the help of these sources, which lay it out well in matrix notation:

https://towardsdatascience.com/linear-algebra-cheat-sheet-for-deep-learning-cd67aba4526c

http://www.gatsby.ucl.ac.uk/teaching/courses/sntn/sntn-2017/resources/Matrix_derivatives_cribsheet.pdf

jminuse:
> Eigen also supports broadcasting to a certain extent so I think that can be done in a couple of places in the forward prop and back prop. I think that might improve the speed even further. Please let me know if I can help implement some of this!

@manangoel99 please do, and benchmark it!

In my hands, Eigen ops were slower than plain C++ loops. I suspect Eigen's implicit temporary matrices caused the slowdown (see links below). But good Eigen code, perhaps using Eigen::Map, might solve this problem and get more benefit from Eigen.

http://eigen.tuxfamily.org/dox/TopicWritingEfficientProductExpression.html

https://stackoverflow.com/questions/21402426/c-eigen-matrix-operations-vs-memory-allocation-performance

Code/GraphMol/ForceFieldHelpers/ANI/Builder.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "Builder.h"
+
+namespace RDKit {
+namespace ANI {
+using namespace ForceFields::ANI;
+namespace Tools {
+void addANIContribs(const ROMol &mol, ForceFields::ForceField *field,
+                    std::string model, int confId) {
+  PRECONDITION(field, "bad ForceField");
+  auto conf = mol.getConformer(confId);
```
d-b-w, line 24: `conf` doesn't seem to be used here. Would you please remove it from this function if it's not needed?

Code/GraphMol/ForceFieldHelpers/ANI/ANI.h
```diff
@@ -0,0 +1,56 @@
...
+//! Convenience function for optimizing a molecule using ANI
+/*
+  \param mol          the molecule to use
+  \param maxIters     the maximum number of force-field iterations
+  \param model        Param dir, such as one in /Code/ForceField/ANI/Params/
+  \param confId       the optional conformer id, if this isn't provided, the
+  molecule's default confId will be used.
+
+  \return a pair with:
+     first: 0 if the optimization converged, 1 if more iterations are required.
```
d-b-w, line 21: Maybe `converged` should be a `bool`, then, if there are only two allowed values?

    std::pair<bool, double> ANIOptimizeMolecule(ROMol &mol, std::string model,
                                           int confId = -1,
                                           int maxIters = 1000);

Call like (C++17):

    auto [converged, energy] =  ANIOptimizeMolecule(mol, "ANI-1x");

Code/Numerics/EigenSerializer/portable_binary_iarchive.cpp
```diff
@@ -0,0 +1,112 @@
+/////////1/////////2/////////3/////////4/////////5/////////6/////////7/////////8
+// portable_binary_iarchive.cpp
+
+// (C) Copyright 2002 Robert Ramey - http://www.rrsd.com .
+// Use, modification and distribution is subject to the Boost Software
+// License, Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+//  See http://www.boost.org for updates, documentation, and revision history.
```
d-b-w, line 9: yeah... Can we include boost's version, then?

d-b-w (changes requested):
Some basic code maintainability/stability requests below.

I guess my big question is about forking the files from boost.

Code/Numerics/EigenSerializer/portable_binary_archive.hpp
```diff
@@ -0,0 +1,45 @@
+#ifndef PORTABLE_BINARY_ARCHIVE_HPP
+#define PORTABLE_BINARY_ARCHIVE_HPP
+
+// (C) Copyright 2002 Robert Ramey - http://www.rrsd.com .
```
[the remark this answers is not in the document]
jminuse: This is what I was asking @manangoel99 - I don't understand the choice to include this code. I will see if I can remove it.

Code/Numerics/EigenSerializer/portable_binary_archive.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <boost/cstdint.hpp>
+#include <boost/static_assert.hpp>
+
+#include <climits>
+#if CHAR_BIT != 8
+#error This code assumes an eight-bit byte.
+#endif
+
+#include <boost/archive/basic_archive.hpp>
+#include <boost/detail/endian.hpp>
```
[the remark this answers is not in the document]
jminuse: I believe these files have been copy-and-pasted from here:
https://www.boost.org/doc/libs/1_74_0/libs/serialization/example/
As suggested here: https://stackoverflow.com/questions/8135685/boostarchivebinary-i-oarchive-portability

These includes are found in the original file: https://www.boost.org/doc/libs/1_74_0/libs/serialization/example/portable_binary_archive.hpp

I don't know why these files aren't simply included in boost, but apparently they aren't.

Code/GraphMol/ForceFieldHelpers/Wrap/rdForceFields.cpp
```diff
@@ -481,4 +517,58 @@ RETURNS: a list of (not_converged, energy) 2-tuples. \n\
...
               docString.c_str());
+#ifdef RDK_BUILD_ANI
+//   docString =
+//       "Uses ANI NN to optimize a molecule's structure\n\n\
+//  \n\
+//  ARGUMENTS:\n\n\
+//     - mol : the molecule of interest\n\
+//     - maxIters : the maximum number of iterations (defaults to 200)\n\
+//     - confId : indicates which conformer to optimize\n\
+//     - model : param dir, such as one in /Code/ForceField/ANI/Params/\n\
```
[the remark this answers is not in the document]
jminuse: An enum would be moderately helpful to provide compile-time checking for users misspelling the model names. But, I think it's important that users can specify their own parameter dirs. That's how I am using Schrodinger-ANI with this build, for instance, and experimenting with different ensembles by changing the model files in the directory.

Code/Numerics/EigenSerializer/portable_binary_iarchive.cpp
jminuse: This code is definitely on the Boost website (https://www.boost.org/doc/libs/1_74_0/libs/serialization/example/portable_binary_iarchive.cpp) but I'm not sure it is shipped with Boost. It seems to be just an example of how to use Boost.

Code/GraphMol/ForceFieldHelpers/ANI/ANI.h
jminuse: I agree in principle, but, pair<int, double> pattern is how the analogous methods work for the other force fields. See std::pair<int, double> MMFFOptimizeMolecule in ForceFieldHelpers/MMFF/MMFF.h, and std::pair<int, double> UFFOptimizeMolecule in ForceFieldHelpers/UFF/UFF.h

lilleswing:
@here is this good to go now?

greglandrum:

PR #3436, opened 2020-09-25
Fixes #3430

### LINKED ISSUE #3430: Tautomers with endocyclic double bonds should be preferred over exocyclic ones
Currently the tautomer scoring function does not seem to discriminate between tautomers with endocyclic double bonds vs exocyclic ones. In fact, both are aromatic according to the RDKit aromaticity model, but I still would prefer the endocyclic double bonds to score higher. After discussion with Greg we agreed this should be corrected.
This gist shows an example of the above:
https://gist.github.com/ptosco/20b06985cd8830d5e549165f6b9fc969

ptosco:
Turns out this was a very simple fix. There was already a rule to penalize exocyclic nitrogen bonds from aromatics to nitrogen, which however was conflicting with another rule addressing guanidines, so the net effect of the rules was zero and the exocyclic double bonded ended up with exactly the same score as the exocyclic one.
Changing the guanidine pattern to only apply to aliphatic C fixes the issue.
I had to change a single unit test where before my fix tautomers had the same score and the exocyclic tautomer was chosen as the canonical one based on lexical order:
image
After my fix the endocyclic tautomer scores higher, as expected.:
image
This is also the tautomer predicted to be the most stable by both UNICON (https://doi.org/10.1021/acs.jcim.6b00069) and MoKa (http://dx.doi.org/10.1021/ci800340j).

### DIFF
--- a/Code/GraphMol/MolStandardize/Tautomer.cpp
+++ b/Code/GraphMol/MolStandardize/Tautomer.cpp
@@ -115,7 +115,7 @@ int scoreSubstructs(const ROMol &mol) {
       {"C=hetero", "[C]=[!#1;!#6]", 1},
       {"aromatic C = exocyclic N", "[c]=!@[N]", -1},
       {"methyl", "[CX4H3]", 1},
-      {"guanidine terminal=N", "[#7][#6](=[NR0])[#7H0]", 1},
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
it's great that this was so simple.
Thanks for the fix

PR #3437, opened 2020-09-25
Code modernization and an optimization

ptosco:
- code modernization
- removed unnecessary `O(n)` generation of `const` set

### DIFF
--- a/Code/GraphMol/RGroupDecomposition/RGroupCore.h
+++ b/Code/GraphMol/RGroupDecomposition/RGroupCore.h
@@ -1,0 +1,41 @@
+//
+//  Copyright (C) 2020 Novartis Institutes for BioMedical Research
+//
+//   @@ All Rights Reserved @@
+//  This file is part of the RDKit.
+//  The contents are covered by the terms of the BSD license
+//  which is included in the file license.txt, found at the root
+//  of the RDKit source tree.
+//
+#ifndef RGROUP_CORE
+#define RGROUP_CORE
+
+#include "../RDKitBase.h"
+#include "RGroupUtils.h"
+
+namespace RDKit {
+
+//! RCore is the core common to a series of molecules
+struct RCore {
+  boost::shared_ptr<RWMol> core;
+  boost::shared_ptr<RWMol> labelledCore;
+  std::set<int> core_atoms_with_user_labels;
+  RCore(){};
+  RCore(const RWMol &c, bool onlyMatchAtRGroups = false) : core(new RWMol(c)) {
+    if (onlyMatchAtRGroups) {
+      findIndicesWithRLabel();
+    }
+  }
+  void findIndicesWithRLabel() {
+    // First find all the core atoms that have user
+    //  label and but their indices into core_atoms_with_user_labels
+    for (const auto atom : core->atoms()) {
+      if (atom->hasProp(RLABEL)) {
+        core_atoms_with_user_labels.insert(atom->getIdx());
+      }
+    }
+  }
+};
+
+}  // namespace RDKit
+#endif
--- a/Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
+++ b/Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
@@ -23,7 +23,7 @@ struct RGroupDecompData {
   // matches[mol_idx] == vector of potential matches
   std::map<int, RCore> cores;
   std::map<std::string, int> newCores;  // new "cores" found along the way
-  int newCoreLabel;
+  int newCoreLabel = EMPTY_CORE_LABEL;
   RGroupDecompositionParameters params;
 
   std::vector<std::vector<RGroupMatch>> matches;
@@ -37,20 +37,14 @@ struct RGroupDecompData {
 
   RGroupDecompData(const RWMol &inputCore,
                    RGroupDecompositionParameters inputParams)
-      : cores(),
-        newCores(),
-        newCoreLabel(EMPTY_CORE_LABEL),
-        params(std::move(inputParams)) {
+      : params(std::move(inputParams)) {
     cores[0] = RCore(inputCore);
     prepareCores();
   }
 
   RGroupDecompData(const std::vector<ROMOL_SPTR> &inputCores,
                    RGroupDecompositionParameters inputParams)
-      : cores(),
-        newCores(),
-        newCoreLabel(EMPTY_CORE_LABEL),
-        params(std::move(inputParams)) {
+      : params(std::move(inputParams)) {
     for (size_t i = 0; i < inputCores.size(); ++i) {
       cores[i] = RCore(*inputCores[i]);
     }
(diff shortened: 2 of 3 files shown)


ptosco:
@bp-kelley I have grouped all the cosmetic changes in this PR. It would be super-cool if you could review and merge this in, so then I can move on to the serious stuff that you raised comments on. Thanks!

bp-kelley:
@ptosco Thanks for doing breaking up the changes, it will make the review easier.

I wish we had been using C++11 when I started this, it makes a lot of things easier.  I probably don't have time today, but it should go quickly tomorrow morning.

ptosco:
@bp-kelley Great, thanks. This review should be a easy - most changes are trivial, the only relevant one is the once-and-for-all initialization of the `core_atoms_with_user_labels` set - but that is also quite straightforward. Have a nice weekend!

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
```diff
@@ -20,7 +21,7 @@ namespace RDKit
 {
 struct RGroupDecompData {
   // matches[mol_idx] == vector of potential matches
-  std::map<int, RWMol> cores;
+  std::map<int, RCore> cores;
```
d-b-w, line 24: could this be a vector, instead? It looks like it's always contiguous. Then you'd be able to remove so much `.second`

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
```diff
@@ -52,20 +52,21 @@ struct RGroupDecompData {
         newCoreLabel(EMPTY_CORE_LABEL),
```
d-b-w, line 52: I don't think that you need to default initialize `cores` and `newCores`, and if you used 

    int newCoreLabel = EMPTY_CORE_LABEL;

in the definition, you wouldn't need to define that, either. (see CPP core guidelines, here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers)

Code/GraphMol/RGroupDecomposition/RGroupCore.h
```diff
@@ -0,0 +1,41 @@
...
+#include "../RDKitBase.h"
+#include "RGroupUtils.h"
+
+namespace RDKit {
+
+//! RCore is the core common to a series of molecules
+struct RCore {
+  boost::shared_ptr<RWMol> core;
+  boost::shared_ptr<RWMol> labelledCore;
+  std::set<int> core_atoms_with_user_labels;
```
d-b-w, line 22: Have you considered using the more performant `std::unordered_set` here? It's basically my default unless I know I want ordered iteration, or if I can guarantee that the set will always be very small.

d-b-w (commented):
Sorry to spam the review - I couldn't resist, given the title :/

Great stuff!

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
ptosco: Not without major refactoring, as `core_idx` can be negative, as I learnt trying your suggestion.

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
ptosco: Good point, I had noticed the initializers in passing but then I forgot to remove them.

Code/GraphMol/RGroupDecomposition/RGroupCore.h
ptosco: This `std::set` is always going to be very small, so I don't think it would be make a big difference.

ptosco: @d-b-w Thanks for your comments!

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
bp-kelley: The explicitly aren’t consecutive as some may come from users as well,  R1...R3...R7,  pesky users :). Also, the negative versus positive indicate which are user defined and which need to be added.

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
```diff
@@ -435,8 +424,8 @@ struct RGroupDecompData {
     //  [m1_permutation_idx,  m2_permutation_idx, m3_permutation_idx]
 
     while (iterator.next()) {
-      if (count > N) {
-        throw ValueErrorException("Next did not finish");
+      if (count > iterator.maxPermutations) {
+        throw ValueErrorException("next() did not finish");
```
bp-kelley, line 428: note to myself, I always disliked this exception, it isn't meaningful to the user and really is a runtime condition.

Code/GraphMol/RGroupDecomposition/RGroupCore.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef RGROUP_CORE
+#define RGROUP_CORE
+
+#include "../RDKitBase.h"
+#include "RGroupUtils.h"
+
+namespace RDKit {
+
+//! RCore is the core common to a series of molecules
+struct RCore {
```
bp-kelley, line 19: this makes the bookkeeping easier, wish I had thought of it when I did my optimization pass.

Code/GraphMol/RGroupDecomposition/RGroupCore.h
bp-kelley: btw - from experience, the major speed cost is in the permutations and the scoring function.  I doubt you'll see most other optimizations.

bp-kelley (approved):
This is a nice foundation for the algorithm changes.

Even though I wrote the code, I'm going to have to reorient myself, there are a few ways to add RCore's now that look a tad suspicious to me (or I forgot the logic) for instance.

bp-kelley:
@ptosco - once this is =fully approved let's do a quick PR where we run clang-format before we do the algorithm review.

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
```diff
@@ -31,41 +32,35 @@ struct RGroupDecompData {
...
   }
 
   void prepareCores() {
-    size_t idx = 0;
-    for (auto coreIt = cores.begin(); coreIt != cores.end(); ++coreIt, ++idx) {
-      RWMol *alignCore = coreIt->first ? &cores[0] : nullptr;
-      CHECK_INVARIANT(params.prepareCore(coreIt->second, alignCore),
+    for (auto &core : cores) {
+      RWMol *alignCore = core.first ? cores[0].core.get() : nullptr;
+      CHECK_INVARIANT(params.prepareCore(*core.second.core, alignCore),
```
bp-kelley, line 58: There is maybe a case for moving this logic into RCore

```
RCore::prepareCore (const &RGroupDecompositionParameters params, const RWMol& alignCore)
```

or perphaps prepareCore should be:

```
RGroupDecompositionParameters::prepareCore(RCore &, const RWMol &alignCore)
```

ptosco: @bp-kelley Hi Brian, thanks for reviewing this. As soon as this is merged I will move on to the remainder of the PR in the other branch (@greglandrum maybe you wish to have a quick look before merging?).

PR #3459, opened 2020-10-03
Add Cookbook entries 30-32

vfscalfani:
This pull request adds 3 new entries to the cookbook:

30: Identifying E/Z Stereochemistry
31: Selecting the largest fragment from a molecule
32: Including stereochemistry annotations in a 2D depiction. 

Thanks,
Vin


greglandrum (approved):
I think you will want to update the first two of these after the new release is out (there's a new, and more accurate, way of getting CIP labels, which Ricardo and I will talk about on Thursday), but I'm fine merging as is since what's there certainly isn't wrong

vfscalfani:
@greglandrum thanks, I made a note and will do that.

PR #3466, opened 2020-10-05
BlockLogs was reenabling all logs, not just the ones that were disabled

bp-kelley:
Simple fix to BlockLogs, it was turning on all logs when it went out of scope, this just turns on the logs that were turned off.


greglandrum (approved):
I guess this change must have gone in after I did the big clang-tidy run to make sure squiggly brackets were around every `if` and `for`

PR #3469, opened 2020-10-07
ROMol: add inline impl for common getNumAtoms call

jinpan:

This change improves the throughput of many common rdkit operations on my setup (x86_64 debian buster / clang 7 / 8-core cpu):

The number of `RDKit::MolToSmiles` calls/s is about 2.8% higher
The number of `RDKit::SmilesToMol` calls/s is about 5.8% higher
The number of `RDKit::MolToInchiKey` calls/s is about 2.0% higher


greglandrum (approved):
LGTM

PR #3476, opened 2020-10-09
Removed missing headers

ptosco:
These missing headers are currently causing the build to fail.


ptosco:
It looks like the CI consistently chokes on the Java build.

ptosco:
It looks like the problem exists only in my build tree.

PR #3477, opened 2020-10-09
Fixes typo that causes the build to fail

ptosco:
OK, there was actually a problem, but it was just a typo.


d-b-w:
ha, I just did this too. Looks right to me.

PR #3478, opened 2020-10-09
Cmake config improvements

rvianello:
This PR extends the functionality of the RDKit cmake config file to automatically manage finding some of its dependencies and to make the toolkit's build options available to the application's cmake script. Also, targets are exported in the RDKit:: namespace.

#### Any other comments?
There's surely space for further refinements, but I would be happy to have this version available from the upstream master.


greglandrum (approved):
I'm not cmake-literate enough to be able to review this one for real, but it looks fine

rvianello:
thanks!

PR #3502, opened 2020-10-16
Added support for isotopic Hs to TautomerEnumerator

ptosco:
Small PR to add, as anticipated, support for implicit isotopic hydrogens to the `TautomerEnumerator`.


greglandrum (changes requested):
I started to review this, but then realized it's easier to just do a PR against yours. :-)
The major suggestion there is that I think the `_isotopicHs` property should be removed from atoms which end up not having any Hs on them.

greglandrum (approved):

PR #3508, opened 2020-10-17
Fix python invalid escape sequences

ricrogz:
Sometimes, when working with python some Deprecation Warnings about invalid escape sequences show up:

```
rdkit\Chem\EnumerateStereoisomers.py:165: DeprecationWarning: invalid escape sequence \[
    """ returns a generator that yields possible stereoisomers for a molecule
```
```
rdkit\ML\Descriptors\MoleculeDescriptors.py:109: DeprecationWarning: invalid escape sequence \ 
    doc = re.sub('\ *\n\ *', ' ', doc)
```

Those aren't a problem (though someone mentioned that in a future python release these warnings might turn into errors), although they are sometimes an annoyance.

I thought I might get them fixed so that the warnings stop showing up. Also, along the way, I noticed my editor trimmed whitespace at the end of the lines of the files I edited.

### DIFF
--- a/Code/GraphMol/Wrap/test_data/do_smiles.bomb.py
+++ b/Code/GraphMol/Wrap/test_data/do_smiles.bomb.py
@@ -1,7 +1,7 @@
 
 
 import re
-splitExpr = re.compile('[\t\ ]')
+splitExpr = re.compile('[\t ]')
 
 from Chem import rdmol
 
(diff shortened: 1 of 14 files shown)


ricrogz:
Oh, I forgot: I'm not familiar with the InChi syntax, so I hope I didn't mess those up.

greglandrum (approved):

Issue #3511, opened 2020-10-19
Traceback when pickling ROMol after BCUT descriptors are calculated


d-b-w:
I'm using the 2020.09beta (commit sha 73d0f05ac7181fac28f19c64279ab2d63e349c0e) under Python3. This is a custom build of RDKit; I'm not using conda.

I ran this script:

    import pickle
    from rdkit import Chem
    from rdkit.Chem import Descriptors
    from rdkit.ML.Descriptors.MoleculeDescriptors import MolecularDescriptorCalculator

    mol = Chem.MolFromSmiles('C')
    descriptors = [name for name, fxn in Descriptors.descList]
    calculator = MolecularDescriptorCalculator(descriptors)
    calculator.CalcDescriptors(mol)
    pickle.dumps(mol)

The pickle step fails with this traceback:

    Traceback (most recent call last):
    File "pickle_described_mol.py", line 10, in <module>
        pickle.dumps(mol)
    RuntimeError: Pickling of "rdkit.rdBase._vectd" instances is not enabled (http://www.boost.org/libs/python/doc/v2/pickle.html)

The proximate cause of this issue was inclusion of the BCUT descriptors in the default descriptor list (https://github.com/rdkit/rdkit/pull/3178), but I guess there is a deeper cause.

PR #3512, opened 2020-10-20
Fix AUTOCORR2D descriptors

ricrogz:
Yesterday, looking at #3511, I became curious about the AUTOCORR2D descriptors: I noticed that while BCUT2D ones left over a cache (with a non-Python type), there wasn't one for the AUTOCORR2D descriptors.

Looking a bit further, I noticed these had to be explicitly set up... using a function that was missing an underscore (`setupDescriptors()`). From there, it was an interesting trip to find that the setup mechanism wasn't working... and neither was the test that was supposed to catch it, since it was over indented so that it was an inner function of the previous test, and therefore, never ran.

This is my attempt to fix it

### DIFF
--- a/rdkit/Chem/UnitTestDescriptors.py
+++ b/rdkit/Chem/UnitTestDescriptors.py
@@ -165,8 +165,10 @@ class TestCase(unittest.TestCase):
   def testVectorDescriptorsInDescList(self):
     # First try only bcuts should exist
     descriptors = set([n for n,_ in Descriptors.descList])
-    names = set(["BCUT2D_%s"%i for i in ('MWHI',"MWLOW","CHGHI","CHGLO",
-                                 "LOGPHI","LOGPLOW","MRHI","MRLOW")])
+    names = set([
+      "BCUT2D_%s" % i
+      for i in ('MWHI', "MWLOW", "CHGHI", "CHGLO", "LOGPHI", "LOGPLOW", "MRHI", "MRLOW")
+    ])
     self.assertEqual(descriptors.intersection(names), names)
 
     Descriptors.setupAUTOCorrDescriptors()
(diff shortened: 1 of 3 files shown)


greglandrum (approved):
LGTM, but @bp-kelley should probably also confirm

rdkit/Chem/UnitTestDescriptors.py
```diff
@@ -162,18 +162,19 @@ def testVectorDescriptors(self):
...
-      Descriptors.setupAUTOCORRDescriptors()
-      descriptors2 = set([n for n,_ in Descriptors.descList])
-      self.assertEqual(descriptors2.intersection(descriptors), descriptors)
-
-      names = ["AUTOCORR2D_%s"%str(i+1) for i in range(192)]
-      self.assertEqual(descriptors.intersection(names), names)
+  def testVectorDescriptorsInDescList(self):
+    # First try only bcuts should exist
+    descriptors = set([n for n,_ in Descriptors.descList])
+    names = set(["BCUT2D_%s"%i for i in ('MWHI',"MWLOW","CHGHI","CHGLO",
```
bp-kelley, line 168: This is code breaking it looks like

bp-kelley (changes requested):

rdkit/Chem/UnitTestDescriptors.py
bp-kelley: As it changes the function names in the descriptor list.

ricrogz: Ok, reformatting it.

bp-kelley: Thanks, I probably shouldn't have copied the AUTOCORR format, but that's what I did :)

bp-kelley: @ricrogz - sorry I misread the change!  It was fine.

ricrogz: No problem! I wasn't quite sure whether you complained about the formatting, which was ugly, or whether the names of the descriptors were wrong... which was part of the problem here.

PR #3516, opened 2020-10-21
[Contrib] Initial commit NIBR 2019 substructure filters

NadineSchneider:
This implements a revised set of 444 SMARTS substructure filters to annotate, flag and exclude compounds. These filters were revised and developed in the course of the NIBR screening deck re-design. More details on this work can be found in a recent publication: Schuffenhauer, A. et al. Evolution of Novartis' small molecule screening deck design, J. Med. Chem. (2020), DOI. https://dx.doi.org/10.1021/acs.jmedchem.0c01332


greglandrum (approved):
This looks great. Thanks @NadineSchneider!

Do you think that it would make sense for us to directly integrate these into the C++ FilterCatalog implementation so that they are straightforward for anyone to use?

NadineSchneider:
Many thanks for the feedback @greglandrum! 

It would be nice if we could integrate them into the Contrib for now (we added that link to the paper :)) and add them in addition at some point in the C++ FilterCatalog. What do you think?

bp-kelley:
@NadineSchneider @greglandrum Once they are merged I can add them to the filter catalog along with pat Walters collections as well.

This is great work!

NadineSchneider:
Hi @bp-kelley , thanks for the feedback! We actually had a look at Pat's collection and included most of them in our review. Please let me know when you have more questions or need help.

PR #3518, opened 2020-10-21
Remove accidentally included boost header

ricrogz:
I just noticed that in #3083 I accidentally pulled a patched header from boost into rdkit (and nobody noticed!). This removes this header.

I was using boost 1.68 to build rdkit, and I remember I had a memory error that looked like the one described in https://github.com/boostorg/serialization/issues/104. I used this patched header for the memtesting (which did actually make the bug go away), but never intended to check it into repo. I still don't know how that happened.


d-b-w (approved):
Without this commit, programs that are linked to both RDKit and that also use boost::serialization crash with errors like:

    *** Error in `.....': free(): invalid pointer: 0x00007fdd922f00b0 ***

greglandrum (approved):

PR #3530, opened 2020-10-29
Fixes #3492 (through a different route)

### LINKED ISSUE #3492: a question about atom coordinates after embedding

ptosco:
This PR addresses #3492 through a different route.
The outcome is the same, but the strategy used is quite different.
Less `custodian_and_ward` - I have even removed it from the `GetAtoms`, `Getbonds` and `GetConformers` methods, such that the attempt to access atoms if the underlying molecule has been deleted will fail earlier than before with a `Underlying ROMol object deleted` Python `RuntimeError`.
As a result, the performance impact in this `getconformers2` branch compared to `master` is much smaller than with the `getconformers` branch:

<h2><code>master</code> branch</h2>

```
%%timeit
for m in mols:
    for i in range(m.GetNumAtoms()):
        assert(m.GetAtoms()[i].GetOwningMol())
6.3 s ± 33.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for m in mols:
    atoms = m.GetAtoms()
    for i in range(m.GetNumAtoms()):
        assert(atoms[i].GetOwningMol())
4.1 s ± 15.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for m in mols:
    for a in m.GetAtoms():
        assert(a.GetOwningMol())
6.72 s ± 33.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

<h2><code>getconformers2</code> branch, i.e. this PR</h2>

```
%%timeit
for m in mols:
    for i in range(m.GetNumAtoms()):
        assert(m.GetAtoms()[i].GetOwningMol())
6.83 s ± 41.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for m in mols:
    atoms = m.GetAtoms()
    for i in range(m.GetNumAtoms()):
        assert(atoms[i].GetOwningMol())
4.23 s ± 35.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for m in mols:
    for a in m.GetAtoms():
        assert(a.GetOwningMol())
7.01 s ± 56.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

<h2><code>getconformers</code> branch</h2>

```
%%timeit
for m in mols:
    for i in range(m.GetNumAtoms()):
        assert(m.GetAtoms()[i].GetOwningMol())
7.95 s ± 180 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for m in mols:
    atoms = m.GetAtoms()
    for i in range(m.GetNumAtoms()):
        assert(atoms[i].GetOwningMol())
4.73 s ± 211 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
for m in mols:
    for a in m.GetAtoms():
        assert(a.GetOwningMol())
7.51 s ± 40.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

We could push this approach even further making `Atom`, `Bond` and `Conformer` copy constructors register themselves as children of the parent `ROMol`, such that if people do something like:

```
mol = Chem.MolFromSmiles("CCC")
atoms = [a for a in mol.GetAtoms()]
del mol
atoms[0].GetOwningMol()
```

they don't get a segfault (this is the only case that still fails also in all 3 branches), because atoms could be dis-owned when the `ROMol` is deleted.
However, this would add some overhead (but only when atoms or bonds or conformers are copied).

Before going any further, I'd like to check @greglandrum and @bp-kelley opinion on this POC.
If you like it, I'll fix the threading issues and add relevant tests (you do not need the threading support until this is only used from Python, so it's OK as it is for this POC).
And let me know what you think of the further improvement I proposed.

Thanks!

### DIFF
--- a/Code/GraphMol/ROMol.cpp
+++ b/Code/GraphMol/ROMol.cpp
@@ -56,7 +56,7 @@ void ROMol::destroy() {
   d_stereo_groups.clear();
 };
 
-ROMol::ROMol(const std::string &pickle) : RDProps() {
+ROMol::ROMol(const std::string &pickle) : RDProps(), RDParent() {
   initMol();
   numBonds = 0;
   MolPickler::molFromPickle(pickle, *this);
--- a/Code/GraphMol/ROMol.h
+++ b/Code/GraphMol/ROMol.h
@@ -31,6 +31,7 @@
 // our stuff
 #include <RDGeneral/types.h>
 #include <RDGeneral/RDProps.h>
+#include <RDGeneral/RDParentChild.h>
 #include "Atom.h"
 #include "Bond.h"
 #include "Conformer.h"
@@ -168,7 +169,7 @@ struct CXXBondIterator {
   CXXBondIter end() { return {graph, vend}; }
 };
 
-class RDKIT_GRAPHMOL_EXPORT ROMol : public RDProps {
+class RDKIT_GRAPHMOL_EXPORT ROMol : public RDProps, public RDParent {
  public:
   friend class MolPickler;
   friend class RWMol;
@@ -267,7 +268,7 @@ class RDKIT_GRAPHMOL_EXPORT ROMol : public RDProps {
     return {&d_graph};
   }
 
-  ROMol() : RDProps() { initMol(); }
+  ROMol() : RDProps(), RDParent() { initMol(); }
 
   //! copy constructor with a twist
   /*!
@@ -281,7 +282,7 @@ class RDKIT_GRAPHMOL_EXPORT ROMol : public RDProps {
          the specified conformer from \c other.
   */
   ROMol(const ROMol &other, bool quickCopy = false, int confId = -1)
-      : RDProps() {
+      : RDProps(), RDParent() {
     dp_ringInfo = nullptr;
     initFromOther(other, quickCopy, confId);
     numBonds = rdcast<unsigned int>(boost::num_edges(d_graph));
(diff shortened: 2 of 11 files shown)


ptosco:
Closing as #3527 was accepted.

PR #3538, opened 2020-11-01
A couple of fixes to the build system

ptosco:
A couple of trivial fixes to the build system:
- add `#ifdef` to enable building `ScaffoldNetwork` with no Boost serialization support
- do not look for `Threads` if not needed


greglandrum (approved):
The first change makes sense to me but the second is now redundant.

ptosco:
@greglandrum I have resolved the conflict - should now be good to go.

PR #3557, opened 2020-11-08
A few improvements to SubstructLibrary

ptosco:
* Enable `SubstructLibrary` to build without threading support
There were a few typos and small glitches which currently prevented `SubstructLibrary` from building without threading support.

* Enforce consistency across single- and multi-threaded runs
Due to the way parallelization was implemented, multi-threaded runs were reporting different indices depending on the number of threads they were running on. In general, it is desirable that results do not depend on degree of parallelization. This PR enforces consistency with negligible (if any) performance impact

* Improve performance on single-threaded runs avoiding overhead of spawning threads

* Consolidate `internalCountMatches` and `internalGetMatches` into one function to reduce code duplication

The notebook below shows that enforcing consistency across single- and multi-threaded runs has basically no performance impact.

<h1><code>substructlib</code> branch</h1>

```
import rdkit
rdkit.__version__
'2021.03.1dev1'
import sys
import os
import gzip
import pickle
from rdkit import Chem
from rdkit.Chem import rdSubstructLibrary
from IPython.display import clear_output
Load ChEMBL27, compute fingerprints and trusted SMILES and pickle the resulting SubstructLibrary:

if (not os.path.exists("/home/toscopa1/pkl/chembl_substructLibrary.pkl")):
    holder = rdSubstructLibrary.CachedTrustedSmilesMolHolder()
    fps = rdSubstructLibrary.PatternHolder()

    with gzip.open("/home/toscopa1/sdf/chembl_27.sdf.gz", "rb") as hnd:
        suppl = Chem.ForwardSDMolSupplier(hnd)
        for i, m in enumerate(suppl):
            if (not i % 1000):
                clear_output(True)
                print(i)
            sys.stdout.flush()
            if (not m):
                continue
            try:
                smi = Chem.MolToSmiles(m)
            except Exception as e:
                continue
            fp = Chem.PatternFingerprint(m)
            holder.AddSmiles(smi)
            fps.AddFingerprint(fp)

    library = rdSubstructLibrary.SubstructLibrary(holder, fps)

    with open("/home/toscopa1/pkl/chembl_substructLibrary.pkl", "wb") as hnd:
        pickle.dump(library, hnd)
else:
    with open("/home/toscopa1/pkl/chembl_substructLibrary.pkl", "rb") as hnd:
        library = pickle.load(hnd)
len(library)
1941410
query = Chem.MolFromSmiles("c1cc2cccnc2cc1")
```

Number of quinoline hits in ChEMBL27:

```
total_matches = library.CountMatches(query)
total_matches
92237
```

Benchmark how long it takes to retrieve all hits on 1, 2, 4, 8 threads:

```
%%timeit
indices = library.GetMatches(query, numThreads=1, maxResults=-1)
5.54 s ± 54.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=2, maxResults=-1)
3.16 s ± 85.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=4, maxResults=-1)
1.78 s ± 42.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=8, maxResults=-1)
1.05 s ± 32.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Make sure that all runs retrieve all results and with the same indices:

```
indices_1_thread = library.GetMatches(query, numThreads=1, maxResults=-1)
len(indices_1_thread)
92237

indices_2_thread = library.GetMatches(query, numThreads=2, maxResults=-1)
len(indices_2_thread)
92237

indices_4_thread = library.GetMatches(query, numThreads=4, maxResults=-1)
len(indices_4_thread)
92237

indices_8_thread = library.GetMatches(query, numThreads=8, maxResults=-1)
len(indices_8_thread)
92237

differences_1_2 = [(i, indices_1_thread[i], indices_2_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_2_thread[i]]
len(differences_1_2)
0

differences_1_4 = [(i, indices_1_thread[i], indices_4_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_4_thread[i]]
len(differences_1_4)
0

differences_1_8 = [(i, indices_1_thread[i], indices_8_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_8_thread[i]]
len(differences_1_8)
0
```

Now repeat the tests setting maxResults to 99% of the theoretical maximum:

```
percent99_matches = int(0.99 * total_matches)
percent99_matches
91314

%%timeit
indices = library.GetMatches(query, numThreads=1, maxResults=percent99_matches)
5.47 s ± 47.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=2, maxResults=percent99_matches)
3.08 s ± 55.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=4, maxResults=percent99_matches)
1.81 s ± 56.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=8, maxResults=percent99_matches)
1.08 s ± 48.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

indices_1_thread = library.GetMatches(query, numThreads=1, maxResults=percent99_matches)
len(indices_1_thread)
91314

indices_2_thread = library.GetMatches(query, numThreads=2, maxResults=percent99_matches)
len(indices_2_thread)
91314

indices_4_thread = library.GetMatches(query, numThreads=4, maxResults=percent99_matches)
len(indices_4_thread)
91314

indices_8_thread = library.GetMatches(query, numThreads=8, maxResults=percent99_matches)
len(indices_8_thread)
91314

differences_1_2 = [(i, indices_1_thread[i], indices_2_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_2_thread[i]]
len(differences_1_2)
0

differences_1_4 = [(i, indices_1_thread[i], indices_4_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_4_thread[i]]
len(differences_1_4)
0

differences_1_8 = [(i, indices_1_thread[i], indices_8_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_8_thread[i]]
len(differences_1_8)
0
```

All good.

<h1><code>master</code> branch</h1>

```
import rdkit
rdkit.__version__
'2020.03.6'
import sys
import os
import gzip
import pickle
from rdkit import Chem
from rdkit.Chem import rdSubstructLibrary
from IPython.display import clear_output
if (not os.path.exists("/home/toscopa1/pkl/chembl_substructLibrary.pkl")):
    holder = rdSubstructLibrary.CachedTrustedSmilesMolHolder()
    fps = rdSubstructLibrary.PatternHolder()

    with gzip.open("/home/toscopa1/sdf/chembl_27.sdf.gz", "rb") as hnd:
        suppl = Chem.ForwardSDMolSupplier(hnd)
        for i, m in enumerate(suppl):
            clear_output(True)
            print(i)
            sys.stdout.flush()
            if (not m):
                continue
            try:
                smi = Chem.MolToSmiles(m)
            except Exception as e:
                continue
            fp = Chem.PatternFingerprint(m)
            holder.AddSmiles(smi)
            fps.AddFingerprint(fp)

    library = rdSubstructLibrary.SubstructLibrary(holder, fps)

    with open("/home/toscopa1/pkl/chembl_substructLibrary.pkl", "wb") as hnd:
        pickle.dump(library, hnd)
else:
    with open("/home/toscopa1/pkl/chembl_substructLibrary.pkl", "rb") as hnd:
        library = pickle.load(hnd)
len(library)
1941410
query = Chem.MolFromSmiles("c1cc2cccnc2cc1")
query

total_matches = library.CountMatches(query)
total_matches
92237
```

Benchmark how long it takes to retrieve all hits on 1, 2, 4, 8 threads:

```
%%timeit
indices = library.GetMatches(query, numThreads=1, maxResults=-1)
6.38 s ± 129 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=2, maxResults=-1)
3.11 s ± 48.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=4, maxResults=-1)
1.83 s ± 62.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=8, maxResults=-1)
1.07 s ± 40.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Make sure that all runs retrieve all results and with the same indices. Here we need to add a sorted as the original C++ code did not include a sort:

```
indices_1_thread = sorted(library.GetMatches(query, numThreads=1, maxResults=-1))
len(indices_1_thread)
92237

indices_2_thread = sorted(library.GetMatches(query, numThreads=2, maxResults=-1))
len(indices_2_thread)
92237

indices_4_thread = sorted(library.GetMatches(query, numThreads=4, maxResults=-1))
len(indices_4_thread)
92237

indices_8_thread = sorted(library.GetMatches(query, numThreads=8, maxResults=-1))
len(indices_8_thread)
92237

differences_1_2 = [(i, indices_1_thread[i], indices_2_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_2_thread[i]]
len(differences_1_2)
0

differences_1_4 = [(i, indices_1_thread[i], indices_4_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_4_thread[i]]
len(differences_1_4)
0

differences_1_8 = [(i, indices_1_thread[i], indices_8_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_8_thread[i]]
len(differences_1_8)
0
```

Now repeat the tests setting maxResults to 99% of the theoretical maximum:

```
percent99_matches = int(0.99 * total_matches)
percent99_matches
91314

%%timeit
indices = library.GetMatches(query, numThreads=1, maxResults=percent99_matches)
6.22 s ± 347 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=2, maxResults=percent99_matches)
3.49 s ± 233 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=4, maxResults=percent99_matches)
1.86 s ± 33.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
indices = library.GetMatches(query, numThreads=8, maxResults=percent99_matches)
1.04 s ± 15.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

indices_1_thread = sorted(library.GetMatches(query, numThreads=1, maxResults=percent99_matches))
len(indices_1_thread)
91314

indices_2_thread = sorted(library.GetMatches(query, numThreads=2, maxResults=percent99_matches))
len(indices_2_thread)
91314

indices_4_thread = sorted(library.GetMatches(query, numThreads=4, maxResults=percent99_matches))
len(indices_4_thread)
91314

indices_8_thread = sorted(library.GetMatches(query, numThreads=8, maxResults=percent99_matches))
len(indices_8_thread)
91314
```

In this case we have increasing differences in retrieved indices with varying numbers of threads:

```
differences_1_2 = [(i, indices_1_thread[i], indices_2_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_2_thread[i]]
len(differences_1_2)
6

differences_1_4 = [(i, indices_1_thread[i], indices_4_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_4_thread[i]]
len(differences_1_4)
1317

differences_1_8 = [(i, indices_1_thread[i], indices_8_thread[i]) for i in range(len(indices_1_thread)) if indices_1_thread[i] != indices_8_thread[i]]
len(differences_1_8)
2559
```


Code/GraphMol/FileParsers/SDMolSupplier.cpp
```diff
@@ -267,6 +267,7 @@ unsigned int SDMolSupplier::length() {
     // now remember to set the stream to the last position we want to read
     dp_inStream->clear();
     dp_inStream->seekg(d_molpos[d_last]);
+    df_end = false;
```
bp-kelley, line 270: This doesn’t belong in this pr

ptosco: @bp-kelley Yeah, not sure what I messed up in the previous push. I have re-pushed and now it's OK.

bp-kelley:
@ptosco can we split apart the bug fix?  I have some comments about the original design and the implementation, but I don't want to halt the bug fix.

ptosco:
@bp-kelley Done, the bug fix is now in its own PR (#3558).

bp-kelley:
@ptosco Here's my thoughts.  The original design was explicitly not consistent between threads for the sake of code complexity.  The thinking behind this was that because there was no easy way to "continue" a search, it really didn't matter.

This is still a bit true, since while you could take the maximum return index and make a new substructlibrary with only the needed elementsf, I would rather keep the inconsistency and have a state object that allows continuation:

```
state = sslib.SubstructLibraryState()
while state:
         for match in sslib.GetMatches(query, state=state):
               ...
```

This would supply a way of "fetching more" from, say,  a web server and essentially making a hit list.  I would rather put the complexity here and ignore the single-multithread consistency.  If this means we need consistency, then great but I don't think we do.

What do you think?

bp-kelley:
btw I also have a key holder implementation I've never pushed that holds a "title" for the mols based on @jones-gareth implementation (from two years ago!  Ug, I've been slow on this one)

ptosco:
@bp-kelley @greglandrum I think consistency is indeed important. In the current implementation a user gets different results depending on the amount of resources used for the search, which I think is undesirable. Of course all the <i>N</i> retrieved molecules are valid matches for the query, but one might wonder "hey, I did the exact same search yesterday with the same query on the same library and I got different results, why is that?", and the answer "because you ran it on a beefier machine" is not the one I would like to hear... Given that the (slightly) greater complexity does not result in a noticeable impact on performance I really would go for it. Getting different results at each run depending on available resources is something that I would avoid if there is an inexpensive way to do so.

bp-kelley:
@ptosco, how about a two pronged approach.  I think continuations are more important and let me see if I can add that to the consistency you've baked in.

One question, you've converted the vector reference to a pointer, I was very unclear why you did this.  There is no performance advantage and the code is clearer with a ref.  Did I miss something?

ptosco:
@bp-kelley The change only affects `internalGetMatches` which is an internal, non-user-exposed function, so the pointer ugliness is completely hidden away. By doing so `internalGetMatches` can be used also for counting by passing a `nullptr`, and `internalMatchCounter` is not needed anymore, thus removing some code duplication/complexity.

bp-kelley:
Ah, I missed that, fair enough.

I found my original state/consistency code, I'm going to resurrect it and
put it up as an independent PR so we can take the good parts from both if
possible.

bp-kelley:
@ptosco It looks like both our solutions ended up being somewhat similar

The only real difference was that I put the datastructures into the Bit struct

```
struct Bits {
  const ExplicitBitVect *queryBits;
  const FPHolderBase *fps;
  bool recursionPossible;
  bool useChirality;
  bool useQueryQueryMatches;
  std::atomic<int> counter;
  std::vector<int> screened;
```

screened[threadIdx] held the # mols screened.  Unlike you, I didn't use this to make the results consistent, but rather to use to continue for the next batch.

I can do the same with your solution though, so I'm happy to accept this implementation after I give it a few more passes.

It does have a worse case behavior though, but if someone manages to pack their data into the thread stride, that's on them.  If it's a real problem we can change to blocks not strides.

Code/GraphMol/SubstructLibrary/SubstructLibrary.cpp
```diff
@@ -126,144 +123,156 @@ void SubSearcher(const ROMol &in_query, const Bits &bits,
...
+           ++thread_group_idx) {
+        if (endIdxVect[thread_group_idx] >= maxEndIdx) {
+          continue;
+        }
+        // need to use boost::ref otherwise things are passed by value
+        thread_group.emplace_back(std::async(
+            std::launch::async, SubSearcher, std::ref(query), bits,
+            std::ref(mols), endIdxVect[thread_group_idx] + numThreads,
+            std::ref(maxEndIdx), numThreads, needs_rings,
+            std::ref(counterVect[thread_group_idx]), -1,
```
bp-kelley, line 230: It might be more elegant to have the thread return the # screened/ # matched via

```
  std::future<unsigned int> num_matched = fut.get();
```

bp-kelley: Or maybe std::pair

Code/GraphMol/SubstructLibrary/SubstructLibrary.cpp
```diff
@@ -126,144 +123,156 @@ void SubSearcher(const ROMol &in_query, const Bits &bits,
...
+    }
+    int thread_group_idx;
+    for (thread_group_idx = 0; thread_group_idx < numThreads;
+         ++thread_group_idx) {
+      // need to use boost::ref otherwise things are passed by value
+      thread_group.emplace_back(
+          std::async(std::launch::async, SubSearcher, std::ref(query), bits,
+                     std::ref(mols), startIdx + thread_group_idx,
+                     std::ref(endIdxVect[thread_group_idx]), numThreads,
+                     needs_rings, std::ref(counterVect[thread_group_idx]),
```
bp-kelley, line 196: We might be able to clean the code up a tad if we introduce a thread barrier and share the counterVect, but that can be a future optimization.

bp-kelley (changes requested):

bp-kelley:
Only one required change (sign casting) necessary.  Looks good to me, glad it was easy to slot in.  I also tested this below func to make sure we weren't skipping any mols by accident.  I can add it later or do a PR or you can paste it in if you want.

```
void testMaxResultsAllSameNumThreads() {
  BOOST_LOG(rdErrorLog) << "-------------------------------------" << std::endl;
  BOOST_LOG(rdErrorLog) << "   Results do not depend on numThreads (all same) "
                        << std::endl;

  auto *mols = new MolHolder();
  auto *fps = new PatternHolder();
  boost::shared_ptr<MolHolder> mols_ptr(mols);
  boost::shared_ptr<PatternHolder> fps_ptr(fps);

  SubstructLibrary ssslib(mols_ptr, fps_ptr);
  boost::logging::disable_logs("rdApp.error");
  auto mol = "N"_smiles;
    for(int i=0; i< 999; ++i) {
        ssslib.addMol(*mol);
  }

  boost::logging::enable_logs("rdApp.error");
  std::vector<std::vector<unsigned int>> resVect;
  ROMOL_SPTR query(SmartsToMol("N"));
  TEST_ASSERT(query);
  for (auto numThreads : {1, 2, 4, 8}) {
    resVect.emplace_back(
        ssslib.getMatches(*query, true, false, false, numThreads));
    TEST_ASSERT(resVect.back().size() == 999);
  }
  for (auto it = resVect.begin() + 1; it != resVect.end(); ++it) {
    TEST_ASSERT(resVect.front().size() == it->size());
    for (size_t i = 0; i < resVect.front().size(); ++i) {
      TEST_ASSERT(resVect.front().at(i) == it->at(i));
    }
  }
  size_t results60 = resVect.front().size() * 0.6;
  size_t results99 = resVect.front().size() * 0.99;
  for (auto maxRes : {results60, results99}) {
    std::vector<std::vector<unsigned int>> resVectPartial;
    for (auto numThreads : {1, 2, 4, 8}) {
      resVectPartial.emplace_back(
          ssslib.getMatches(*query, true, false, false, numThreads, maxRes));
    }
    for (auto it = resVectPartial.begin(); it != resVectPartial.end(); ++it) {
      TEST_ASSERT(it->size() == maxRes);
      for (size_t i = 0; i < maxRes; ++i) {
        TEST_ASSERT(resVect.front().at(i) == it->at(i));
      }
    }
  }
}
```

Code/GraphMol/SubstructLibrary/SubstructLibrary.cpp
```diff
@@ -126,144 +123,156 @@ void SubSearcher(const ROMol &in_query, const Bits &bits,
...
+    }
+    for (auto &fut : thread_group) {
+      fut.get();
+    }
+    for (thread_group_idx = 0; thread_group_idx < numThreads;
+         ++thread_group_idx) {
+      if (idxs) {
+        idxs->insert(
+            idxs->end(),
+            std::make_move_iterator(internal_results[thread_group_idx].begin()),
```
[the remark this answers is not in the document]
ptosco: Absolutely right, I stripped the `std::make_move_iterator`.

ptosco:
@bp-kelley Thanks for your review - I have already added your test.

bp-kelley:
@ptosco phew!  Thanks!

PR #3564, opened 2020-11-10
Minor tweaks to SubstructLibrary

ptosco:
This PR contains minor trivial tweaks to `SubstructLibrary`:

- added some missing `const` keywords
- added an `addFingerprint` overload to allow passing pointers.
This is useful to add a pattern fingerprint with non-default number of bits and pass ownership to the `SubstructLibrary`: `PatternFingerprintMol` returns an `ExplicitBitVect *`, and in the absence of this overload an unnecessary copy would be generated
- added a test


Code/GraphMol/SubstructLibrary/SubstructLibrary.h
```diff
@@ -249,10 +249,15 @@ class RDKIT_SUBSTRUCTLIBRARY_EXPORT FPHolderBase {
...
   }
 
+  //! Adds a raw bit vector pointer to the fingerprinter (which takes ownership)
+  unsigned int addFingerprint(ExplicitBitVect *v) {
+    fps.push_back(v);
+    return rdcast<unsigned int>(fps.size() - 1);
+  }
+
   //! Adds a raw bit vector to the fingerprinter
   unsigned int addFingerprint(const ExplicitBitVect &v) {
```
bp-kelley, line 259: This needs a big caveat emptor here.  The fingerprint needs to be exactly the same as the one generated by makeFingerprint or bad things may happen at run time.

bp-kelley (approved):

bp-kelley:
@ptosco, yeah those functions should definitely be const.  

The lack of an addFingerprint was a design decision to prevent conflicts in fingerprint types.  I.e. you need to define the fingerprinting class before you can add fps.  I can see, though, that this might be slow when adding molecules.  If this is the reason, perhaps a threaded loader would be more appropriate.

After the comment/doc update I'll merge this.  But I'm still a bit worried that we are going to open ourselves up to confusion with the addFingerprint and users who don't overload the fingerprint generator.

ptosco:
@bp-kelley Note that I have only added an overload to `addFingerprint` - the function was already there. I can definitely add the comment but this specific PR does not affect the ability to add raw fingerprints, which was already there. Am I wrong?
The threaded mol loader is on my to-do list - I'll create an issue and assign it to myself. Are you OK with this?

bp-kelley:
@ptosco I knew I was reviewing myself when I wrote it!

Before we write the threaded mol loader, I would like to think about having a standard pattern for doing this in general.  I.e. should we exploit the threaded mol reader?  If we just have a list of smiles to add, I think that can be independent and probably pretty easy to implement.

I.e. mols.reserve(size), pattern.reserve(size) -> threaded fill.

ptosco:
@bp-kelley Good point about the threaded mol loader - we definitely want to make it as general as possible.
About `addFingerprint`, I have had another thought to address your concern: what about changing

```
  virtual ExplicitBitVect *makeFingerprint(const ROMol &m) const {
    return PatternFingerprintMol(m, 2048);
  }
```

into

```
  virtual ExplicitBitVect *makeFingerprint(const ROMol &m, unsigned int numBits = 2048) const {
    return PatternFingerprintMol(m, numBits);
  }
```

That would actually serve the same purpose and would be much safer. Agreed?

bp-kelley:
I think this belongs in the constructor actually since it would need to be serialized.

```
class RDKIT_SUBSTRUCTLIBRARY_EXPORT PatternHolder : public FPHolderBase {
 unsigned int numbits;
 public:
  PatternHolder() : numbits(2048) {}
  PatternHolder(unsigned int numbits) : numbits(numbits) {}
  //! Caller owns the vector!
  virtual ExplicitBitVect *makeFingerprint(const ROMol &m) const {
    return PatternFingerprintMol(m, numbits);
  }
};
```

If we make the variable one a separate class, actually, we can preserve backwards compatibility.

ptosco:
@bp-kelley Right, we do want to serialize the number of bits. I hope what I have committed makes sense - the bed is calling me aloud.

ptosco:
OK, I have messed up something with Java wrappers. I'll build them locally and sort this out.

bp-kelley:
Get the tests working then I'll push a PR to your fork for the serialization.  Then I'll go back to your stable sort PR!

ptosco:
@bp-kelley This should be good now.

ptosco:
@bp-kelley yeah, I left out a small, insignificant detail: the serialization itself :-D  mind you, I said I was sleepy.
Don’t spend time on this, I’ll do it tomorrow, and add a test. Save your energies for the stable sort thing :-)

bp-kelley:
It should be easy to make the serialization use the old when numbits is 2048 so we don’t break backwards compatibility.

----
Brian Kelley

ptosco:
@bp-kelley I think this should be good now.

Code/GraphMol/SubstructLibrary/SubstructLibrary.h
```diff
@@ -286,6 +295,20 @@ class RDKIT_SUBSTRUCTLIBRARY_EXPORT PatternHolder : public FPHolderBase {
   }
 };
 
+//! Uses the pattern fingerprinter with a user-defined number of bits to rule out matches
+class RDKIT_SUBSTRUCTLIBRARY_EXPORT PatternNumBitsHolder : public FPHolderBase {
```
bp-kelley, line 299: I'd rather not make two classes, let me see if I can get it to work with only one.

bp-kelley (changes requested):

Code/GraphMol/SubstructLibrary/SubstructLibrary.h
ptosco: @bp-kelley That’s doable, but I think you need to bump the version number if you wish to keep backwards compatibility when deserialising older `PatternHolder`s (I wasn’t sure how you felt about that). Or (less elegant) catch the exception when trying to read `numBits` past the archive end when deserialising older objects.

ptosco:
@bp-kelley Sorry, I had not realized this was being blocked by a conflict.
I have resolved the conflicts and got rid of `PatternNumBitsHolder` as per your suggestion.
Instead, I leveraged the serialization version to do the right thing at deserialization time to keep backwards compatibility. I have added a test for that that uses a pickle generated with the version 1 serializer and it works OK.

ptosco:
@bp-kelley the `Unsupported version` is caused by the fact that the pickle I have included as backward-compatibility test uses a very recent version of Boost which apparently is incompatible with the one used by the Azure build server. I will generate a pickle with the same Boost version used by the server now - that should fix the problem.

ptosco:
@bp-kelley I think this is ready to go now.

Code/GraphMol/SubstructLibrary/SubstructLibrarySerialization.h
```diff
@@ -179,12 +184,12 @@ void load(Archive &ar, RDKit::SubstructLibrary &slib,
 }  // end namespace serialization
 }  // end namespace boost
 
-BOOST_CLASS_VERSION(RDKit::MolHolder, 1);
```
bp-kelley, line 182: I think you only need to change the pattern holder version.

bp-kelley (changes requested):

bp-kelley:
@ptosco Looks good, let's verify if you need to bump all the serialization versions, but after that I'll merge.

ptosco:
@bp-kelley Thanks Brian. You can indeed bump only the `PatternHolder` version, I bumped all of them just for consistency, but I'll revert others to `1` if you prefer so.

bp-kelley:
Since they didn’t change let’s do this.  Is there any way to write a version 1 pickle if there isn’t a changed numbits by the way?

ptosco:
@bp-kelley No, but that's a good idea - I'll implement it now and add it to this PR. This will also avoid having to maintain the V1 `pkl` file for testing.

ptosco:
@bp-kelley the serialization version is a compile time constant, so it cannot be set at run time.
Instead, I have reverted to version 1 for all including `PatternHolder`. I write `numBits` only if `!= 2048`, and read it from the archive only if it actually exists. However, this has one major drawback: older RDKit version may read new pickles and ignore the number of bits. This is fine if the library had been generated with 2048 bits, otherwise everything will be screwed up. Instead, bumping version to 2 will avoid this.
Let me know which version you prefer. If you prefer the former, I will just set all versions to 1 but `PatternHolder` to 2 and revert this last change.

ptosco:
@bp-kelley the failures in the tests are due to upstream issues in `master` (`molzip`, etc.). Looks like the `master` branch build is partly broken.

bp-kelley:
Yep, I had a bad commit, it should be reverted.

ptosco:
@bp-kelley The mac test failure is on the multiThreadedMolSupplier - nothing to do with this PR.
Let mw know if you like my last commit (description here: https://github.com/rdkit/rdkit/pull/3564#issuecomment-740784439).
I don't like it, because there is a danger that an old RDKit version will read 1024-bit pickles as if they were 2048-bit.
Instead, the version 2 format would not be read by older RDKit, but the version 1 format will still be read by newer RDKit, which I think is the most important thing.
If you agree with me, I'll revert the last commit and go back to having `PattternHolder` on version 2 and all the rest on version 1.

PR #3565, opened 2020-11-11
R group symmetry

jones-gareth:

This pull request is a work in progress and is not yet ready for merge.

Implements two improvements to RGroupDecomposition symmetry handling

1. A fast scalable scoring function based on fingerprints that may be used instead of the current scoring function.
2. A GA to perform stochastic sampling that may be used as an alternative to the current exhaustive search.

I've put in the pull request now so that I can confirm that this is something you want in the RDKit and also to get feedback that I'm putting it together properly (and any other helpful feedback..)

#### Fingerprint scoring function

Rgroups are fingerprinted using either Morgan fingerprints or topological fingerprints.  The attachment point(s) is replaced by Boron prior to fingerprinting.

The score (to be minimized) for a permutation of structure matches is the mean fingerprint rms variance over all rgroup labels.  For each rgroup label the rms fingerprint variance is the root of the sum of the variances for each fingerprint bit. So, if a permutation has 100 rgroups at label 1, there are 100 associated fingerprints and we determine the variance for each fingerprint bit position.  For example, if half the fingerprints have bit 1 set then the variance for bit 1 is 0.25, if all fingerprints have bit 2 set then the variance for bit 2 is 0, and so on.  The more similar the rgroup fingerprints the smaller the score.

The cool thing about this score is the the variance for a fingerprint bit can be calculated from the bit count/number of bits set (because binary fingerprints only have 0 or 1 values):
```cpp
    double ss = bitCount - (bitCount * bitCount) / (double)numberFingerprints;
    double variancePerBit = ss / (double)numberFingerprints;
```
When you add (or remove) a compound match to an existing permutation you can quickly calculate the score by adding (or subtracting) the fingerprint bits for the new compound and  adjusting the number of fingerprints.  The scoring function is scalable for Brian's greedy chunks search as you just keep the bit counts from the last pruned permutation and add new compounds to that. It also makes for an extremely efficient GA mutation operator.

#### GA

I have a simple implementation of a GA in C++.  I wasn't sure where to put it, but it's in the External directory.  Because I use it in another project, I'd like to keep separate from the main RDKit if possible.

The GA takes too long with the current scoring function, but works quite well with the fingerprint function.  I've been testing on the rg_easy example from Brian's multiple cores notebook.  The GA finds better scoring solutions than GreedyChunks, but takes a lot longer and the GreedyChunks works quite well for this data.  I need to find a good dataset that does not work for the Greedy algorithm to test the GA on.  If anyone has a suitable dataset?

#### Performance

On my laptop- decomposition using the rg_easy dataset from Brian's multiple core notebook (4300 compounds and 9 cores).

| Search Strategy | Score Method | Execution Time | Variance Score |
| ----------------- | ---------------  | ---------------- | -----------------|
|  GreedyChunks  | Brian's             |  46m                |     NA                |
|  GreedyChunks | Fingerprint Variance|   47s         |     -1.65             |
| GA                    | Fingerprint Variance|  37m         |     -1.58             |

In summary, variance scoring is ~60X faster than the original scoring method for this dataset.  The GA method takes a lot longer, but gets a better score.

#### TODO

Much more testing. Example notebooks.  Anything else?


bp-kelley:
@jones-gareth - this is awesome!  I think it may come in pretty handy for free-wilson type algorithms.

Code/GraphMol/RGroupDecomposition/RGroupGa.cpp
```diff
@@ -0,0 +1,263 @@
...
+std::string timeInfo(const chrono::system_clock::time_point start) {
+  auto now = chrono::high_resolution_clock::now();
+  chrono::duration<double, milli> milliseconds = now - start;
+  auto seconds = milliseconds.count() / 1000.0;
+  auto format = boost::format("Time %7.2f") % seconds;
+  return format.str();
+}
+
+vector<vector<size_t>> RGroupGa::run() {
+  auto startTime = std::chrono::high_resolution_clock::now();
```
bp-kelley, line 185: high_resolution_clock::now returns a 
```
std::chrono::time_point<std::chrono::high_resolution_clock>
```
not a std::chrono:time_point, I wonder if this is causing build failures on win/osx?

bp-kelley (changes requested):

bp-kelley:
@ptosco You should look at this as well now that we have an RGroup team.  This is pretty slick.

External/GA/ga/BinaryStringChromosomePolicy.cpp
```diff
@@ -0,0 +1,26 @@
+/*
```
greglandrum, line 1: As a logistics point: these files should all have a standard copyright at the top so that it's clear they're part of the RDKit. Something like:
```
//
//  Copyright (C) 2020 Gareth Jones
//
//   @@ All Rights Reserved @@
//  This file is part of the RDKit.
//  The contents are covered by the terms of the BSD license
//  of the RDKit source tree.
//
```

greglandrum:
Agreed that this sounds really cool. I'm definitely psyched to have it included.

Code/GraphMol/RGroupDecomposition/RGroupGa.cpp
jones-gareth: I'll look into moving to std::clock(), which seems a lot simpler.

External/GA/ga/BinaryStringChromosomePolicy.cpp
jones-gareth: I'll add the headers

jones-gareth: Since I'm doing this as part of Glysade is 
```
//  Copyright (C) 2020 Gareth Jones, Glysade LLC
```
ok?

bp-kelley: Indeed!  I assigned mine to Novartis back in the day.

greglandrum: > Since I'm doing this as part of Glysade is
> 
> ```
> //  Copyright (C) 2020 Gareth Jones, Glysade LLC
> ```
> 
> ok?

Of course. I was just making a suggestion

Code/GraphMol/RGroupDecomposition/RGroupGa.cpp
jones-gareth: Moving to `std::clock` has fixed this issue.  All builds are successful now.

External/GA/ga/BinaryStringChromosomePolicy.cpp
jones-gareth: All new files have the standard header.

bp-kelley:
@jones-gareth awesome!  I particularly interested in seeing how this affects some of our more tricky free-wilson analysis.

Code/GraphMol/RGroupDecomposition/RGroupFingerprintScore.cpp
```diff
@@ -0,0 +1,358 @@
...
+    RWMol mol(*rgroupData->combinedMol);
+    for (auto atom : mol.atoms()) {
+      // replace attachment atom by Boron
+      // TODO- Handle multiple attachments differently?
+      if (atom->getAtomicNum() == 0) {
+        atom->setAtomicNum(5);
+        if (atom->getIsotope() > 0) atom->setIsotope(0);
+      }
+    }
+    MolOps::sanitizeMol(mol);
```
bp-kelley, line 38: Will this always sanitize?

bp-kelley (changes requested):

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
```diff
@@ -472,13 +553,19 @@ struct RGroupDecompData {
...
             largest_heavy_counts = heavy_counts;
             best_permutation = tied_permutation;
-          } else if (heavy_counts == largest_heavy_counts) {
-            size_t perm_value = iterator.value();
-            if (perm_value < min_perm_value) {
-              min_perm_value = perm_value;
-              best_permutation = tied_permutation;
-            }
           }
+          // commented out as min_perm_value and perm_value are unsigned
```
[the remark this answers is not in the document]
jones-gareth: @bp-kelley  I guess I can't see how perm_value can ever be less than min_perm_value.  Both variables are unsigned  and min_perm_value is initialized to zero.  min_perm_value is only updated when `perm_value < min_perm_value` which can never happen.  So, I think it is dead code.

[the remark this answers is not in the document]
ptosco: @jones-gareth Yes, I realized that a few days ago while I was working on another issues. There are other bugs in that code; this fix will be part of another PR that I should be able to finalize today or tomorrow as soon as I find the time to write unit tests for it.

Code/GraphMol/RGroupDecomposition/RGroupFingerprintScore.cpp
jones-gareth: @bp-kelley I haven't seen it fail, and I can't think of how it would, but I have wrapped it in a try/catch in the latest commit.

jones-gareth:
@bp-kelley (and perhaps @ptosco).  In my latest commit I have tidied up the rgroup decomp notebooks in the Docs directory that were not working for me.
In that process I found  two bugs in the relabeling code (where you set positive labels to the negative unindexed labels), for when there are user rgroup labels and unindexed labels in the decomposition.  The bugs result in duplicate rgroup labels.
In the first one, if there is a user rgroup label that is not present in the decomposition that label can be assigned to an unindexed label.
In the second bug you need multiple cores with at least one different user label.  In that case a user label in the second core that is not present in the first core can get assigned to an unindexed label in the first core.
I added a test `testMutipleCoreRelabellingIssues` which reproduces the behavior which caused the bug- I noticed it originally as the `getRGroupsAsColumns` method was returning columns with different sizes.
I have patched the relabel routines and everything seems to be working, but it is neither elegant or efficient, so you should maybe take a look.
I hope this makes sense- I'm not sure I have explained it well.

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
```diff
@@ -43,7 +43,7 @@
 #include <utility>
 #include <vector>
 
-// #define DEBUG
+#define VERBOSE
```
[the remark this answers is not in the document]
jones-gareth: @bp-kelley I've removed those defines and am using the boost loggers.

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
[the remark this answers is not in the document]
jones-gareth: @ptosco I have just also fixed a couple of bugs in the core rgroup code (see my other comment).  Maybe we could time our changes so that only one of us is working on the code at the same time and we don't have to rebase our branches.

bp-kelley:
I think there was a regression in the label let at some point, thanks!   I might pull these fixes and make a separate commit for bookkeeping if that’s ok.

jones-gareth:
@bp-kelley Makes sense.

jones-gareth:
The failing check is due to a checksum error for Avalon tools in the cartridge build.  

```
2020-12-01T20:19:14.8361366Z   The md5 checksum for
2020-12-01T20:19:14.8362682Z   /home/vsts/work/1/s/External/AvalonTools/AvalonToolkit_1.2.0.source.tar is
2020-12-01T20:19:14.8363572Z   incorrect; expected: 092a94f421873f038aa67d4a6cc8cb54, found:
2020-12-01T20:19:14.8364472Z   b1caaf10d85eb3b2aa4f7b8c1b0aca6b
```

jones-gareth:
The latest commit fixes 3 additional bugs that affect multi-core decomp.

#### First 2 bugs:

Firstly, in `RGroupDecompData.h`, removing hydrogens (lines 152-177). `labelCores` maps labels to cores.  It is only updated if:

```cpp
        if (labelCores.find(core_idx) == labelCores.end()) {
```

The problem is that label cores is not indexed by core_idx but by rgoup label.  This error can result in a rgroup mistakenly identified as containing only hyrodgens and removed.

I've replaced the if statement by a set and statement:

```cpp
      std::set<int> coresVisited;
....
       if (coresVisited.find(core_idx) == coresVisited.end()) {
          coresVisited.insert(core_idx);
```

The second bug is in `RGroupDecompParams.cpp::prepareCore` line 206-222.

Unindexed rgoups are assigned an index based on a offset and atom index:

```cpp
 if (setLabel(atom, indexOffset - atom->getIdx(), foundLabels, maxLabel,
                   relabel, Labelling::INDEX_LABELS)) {
        nextOffset++;
```

Before processing the next core, the index offset is then adjusted by the number of new rgroup labels added:
```cpp
  indexOffset -= nextOffset;
```
Which is incorrect and can lead to cores sharing labels which should be different.

The bug is fixed by:
```cpp
  indexOffset -= core.getNumAtoms();
```

I have added 2 new tests,` testUnprocessedMapping` and  `testCoresLabelledProperly` to validate fixing these.

#### Broken Tests

These 2  bug fixes broke tests `testSDFGRoupMultiCoreNoneShouldMatch` and `testMultiCorePreLabelled`.  I think that it is actually the specs for those tests that are broken.  I've created a notebook- `RGroupDecomposition-Tests.ipynb` in Docs/notebooks for the data used in those tests.

For test `testSDFGRoupMultiCoreNoneShouldMatch` the expected decomp is missing the hydroxy in  compound 2.  The test now includes that in the decomp, matching R1 as expected. 

Similar fixes exist for `testMultiCorePreLabelled` (the F in the second compound is now included in the decomp), but I have additional questions about this test.

#### Issues with test testMultiCorePreLabelled and 3rd bug

@ptosco:- I think the specification of `testMultiCorePreLabelled` is wrong.  The expected decomp for `labels=RGroupLabels.MDLRGroupLabels` is different from the expected decomp with `labels=RGroupLabels.AutoDetect` for cores with MDL rgroup labels.  However, I believe that those 2 decomps ought to be identical (I expect any toolkit user will feel the same)- any difference would surely indicate that `AutoDetect` has failed.

I think the difference in the cases is due to this in `RGroupDecompParams.cpp`, line 107:

```cpp
    if (labels != AutoDetect) {
      mcsParams.AtomTyper = rgdAtomCompare;
      mcsParams.CompareFunctionsUserData = &autoLabels;
    }
```

I believe this is a bug (the 3rd bug) and the if statement should be:

```
if (autoLabels != AutoDetect) {
```

With that in place the decomps for the expected results for different combinations of labels and MCS parameters are identical (since the MCS does not alter rgroup labels).

However, this caused some other tests to fail and brought me to yet another issue in `RGroupDecompParams::rgdAtomCompare`:

```cpp
 atom1HasLabel |= a1->hasProp(RLABEL);
  atom2HasLabel |= a2->hasProp(RLABEL);
```

These lines will cause issues when `AtomIndexLabels` is set and the label is negative.  In which case the function will evaluate to false- as the rgroups in the second core (the core that isn't the template) are only set after the MCS call. Fixed by:

```cpp
if (a1->hasProp(RLABEL)) {
    auto label = a1->getProp<int>(RLABEL);
    atom1HasLabel |= label > 0;
  }
  if (a2->hasProp(RLABEL)) {
    auto label = a2->getProp<int>(RLABEL);
    atom2HasLabel |= label > 0;
  }
```

An aside- I don't think that the use of `RelabelDuplicateLabels` in `testMultiCorePreLabelled` has any effect- that setting appears to be only used for internal labels.

#### Rgroups for wildcard atoms

In both the `testSDFGRoupMultiCoreNoneShouldMatch` and `testMultiCorePreLabelled` some of the patterns returned for rgroups matching wildcards seem pretty odd- though (as far as I can see) the groups do all map onto the core. So `c(:[*:1]):[*:1]` may be better as `c[*:1]` and the first `c(c:[*:5]):[*:4]` should be `c[*:4]` and the second `c[*:5]` (see  `RGroupDecomposition-Tests.ipynb`) .  However, that might be better done as a separate PR. 
@bp-kelley you may have some thoughts on this.

-----
Phew!

jones-gareth:
Looks like there is also a Python test `test_multicore_prelabelled` that also needs to be fixed.  I'll do that on the next commit, unless anybody disagrees with my reasoning on th C++ `testMultiCorePreLabelled`.

ptosco:
@jones-gareth The change in `RGroupDecompParams.cpp` that you reference is the fix for the problem described here:

https://github.com/rdkit/rdkit/pull/3497
https://gist.github.com/ptosco/7b6603f1bf665262c1d6022a2c2fccfa

If you revert that change, that use case will break again. We may introduce a different `RGroupDecompositionParameters` parameter rather than tying that behavior to `RGroupLabels.MDLRGroupLabels` if you think it would be more intuitive, but please do not revert those changes to the MCS code as they are needed to address the use case above.
The specification for the `testMultiCorePreLabelled` test are indeed correct.

bp-kelley:
@jones-gareth good catch on the multiple core issue.  This one totally escaped me a year(+) ago.

Let's do this in two passes, I'll help verify the tests and do a PR into your branch, then we can do a second pass with @ptosco's branch.  Does this sound ok?

jones-gareth:
> 
> 
> @jones-gareth The change in `RGroupDecompParams.cpp` that you reference is the fix for the problem described here:
> 
> #3497
> https://gist.github.com/ptosco/7b6603f1bf665262c1d6022a2c2fccfa
> 
> If you revert that change, that use case will break again. We may introduce a different `RGroupDecompositionParameters` parameter rather than tying that behavior to `RGroupLabels.MDLRGroupLabels` if you think it would be more intuitive, but please do not revert those changes to the MCS code as they are needed to address the use case above.
> The specifications for the `testMultiCorePreLabelled` test are indeed correct.

@ptosco Thanks for the gist, that is a good illustration of the problem.  The changes here do **not** break that use case.  Whether `alignment` is `MCS` or `NoAlignment` and  `labels` is `MDLRGroupLabels` or `AutoDetect` the decomposition is the same:

```cpp
std::vector<std::string> expectedRows{
      "Core:O=C(c1cncn1[*:2])[*:1] R1:CN[*:1] R2:CC[*:2]",
      "Core:*1:*c2*cc([*:2])*c2nc1[*:1] R1:F[*:1] R2:Br[*:2]"};
}
```
Which is the same as the correct decomposition in cell 27 of your gist.  The bug you mention in cell 17 is now fixed.

jones-gareth:
@bp-kelley Sounds like a good plan.  I've fixed the failing python rgroup tests.  For some reason the automated tests don't seem to have run.

greglandrum:
> @bp-kelley Sounds like a good plan. I've fixed the failing python rgroup tests. For some reason the automated tests don't seem to have run.

Yeah, there was a problem with the CI builds yesterday. That should be cleared up now.

greglandrum:
Since there seem to be a bunch of moving parts here: would you guys please let me know when you're all happy with it and I should review this? Thanks! :-)

jones-gareth:
> 
> 
> Since there seem to be a bunch of moving parts here: would you guys please let me know when you're all happy with it and I should review this? Thanks! :-)

@greglandrum I am ready for you to review.  The fingerprint variance scoring for r group decomposition worked well, but the GA did not appear to provide a significant improvement to the greedy algorithm (small improvements were observed with a large CPU cost). Unfortunately, the PR has been cluttered by bug fixes to the core RGD code.

jones-gareth:
@bp-kelley, @greglandrum -ready for you to review. I currently have no more changes planned and @ptosco's recent PR is merged in.

bp-kelley:
Great, I’m waiting on the molzip or for some automated tests and am working on the structures Greg sent out.

bp-kelley:
@jones-gareth @ptosco @greglandrum 

Hi all!  I'm almost done with the automation testing using the mol zip functionality to ensure we end up with the same molecule we started with when recombining r-groups.

Then I'll add some of the hand-curated groups that greg and I have been looking at.

One thing I noticed is that I really don't like how I do atom replacements in rings.  For example:

```
core query: c1cc*c1
Res: Core:c1cc:[*:1]:c1 R1:o(:[*:1]):[*:1]" 
```

indicates that the replacement in *:1 is an oxygen, however this is really hard to deduce from the r group.

I propose a change to

```
Core:c1cc:[*:1]:c1 R1:[o:1]
```

Which we would indicate in the spec to be an atom replacement, not really an r-group. 

This isn't anything we need to do anytime soon, if agreed I'll add it as an issue.

Hoping to get this done in the next few days, it's awesome work.

jones-gareth:
@bp-kelley 

I've noticed the issue with wildcards.  There is this example from *testMultiCorePreLabelled*.


| Core | R1 | R2 | R3 | R4 | R5 | R6 |
| -- | -- | -- | -- | -- | -- | -- |
| Core:c1c([*:2])[*:3]c2nc([*:6])[*:5]:[*:4]c2[*:1]1 | R1:c(:[*:1]):[*:1] | R2:Br[*:2] | R3:n(:[*:3]):[*:3] | R4:c(c:[*:5]):[*:4] | R5:c(c:[*:5]):[*:4] | R6:F[*:6] |

R4 and R5 are sort of correct, but it's pretty confusing (and the  2 atom overlay with the core seems like a bug): `R4:[c:4]` and `R5:[c:5]` would be much better.

Is this an issue that is just to do with rings- or would we get the same behavior with wildcards in core sidechains? I agree that it should be separate issue.

I really like the idea of reconstituting the original molecule from the core and rgroups as a basic test.

Code/GraphMol/RGroupDecomposition/RGroupDecomp.h
```diff
@@ -15,6 +15,10 @@
 #include <GraphMol/Substruct/SubstructMatch.h>
 #include <chrono>
 
+#ifdef RDK_BUILD_CPP_TESTS
```
greglandrum, line 18: Does this really need to be in the main header file if it's only used in the tests?
Can you just include the declaration in the tests themselves?

jones-gareth: This is a forward declaration for (Line 130)
```cpp
#ifdef RDK_BUILD_CPP_TESTS
  friend void ::testCoresLabelledProperly();
#endif
```

I guess I don't know how to move the friend declaration to the test class and it won't compile without the forward declaration.

I wanted to be able to test the internal structure of the private RGroup data structures- it was hard to catch labeling errors otherwise.

The friend declaration seems like an anti-pattern, but I didn't want to include creating a public API in this PR.

greglandrum: Ah right. 
I have an alternate proposal for solving this - creating a new test file for internals and doing `#define private public` before including RGroupDecomp.h in that file - that doesn't require this modification or changing the global CMakeLists.txt.
Will do a PR against your branch with that suggestion.

greglandrum:
FYI: when I try this on my linux box I get an occasional segfault in `testGaBatch()`
I unfortunately can't get it to happen in the debugger (nor can I get valgrind to complain), but if I allow it generate a core file I get the following:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f088bc9177a in GapeGa::GaOperation<RDKit::RGroupDecompositionChromosome>::getWeight (this=0x0)
    at /scratch/RDKit_git/Code/GraphMol/RGroupDecomposition/../../../External/GA/ga/GaOperation.h:43
43	  double getWeight() const { return weight; }
[Current thread is 1 (Thread 0x7f08889f1700 (LWP 507123))]
(gdb) where
#0  0x00007f088bc9177a in GapeGa::GaOperation<RDKit::RGroupDecompositionChromosome>::getWeight (this=0x0)
    at /scratch/RDKit_git/Code/GraphMol/RGroupDecomposition/../../../External/GA/ga/GaOperation.h:43
#1  0x00007f088bc8cd7e in GapeGa::LinkedPopLinearSel<RDKit::RGroupDecompositionChromosome, RDKit::RGroupGa>::LinkedPopLinearSel (
    this=0x7f08889f04d0, populationPolicy_=...)
    at /scratch/RDKit_git/Code/GraphMol/RGroupDecomposition/../../../External/GA/ga/LinkedPopLinearSel.h:111
#2  0x00007f088bc865e1 in RDKit::RGroupGa::run (this=0x7ffc31ab1b20, runNumber=3)
    at /scratch/RDKit_git/Code/GraphMol/RGroupDecomposition/RGroupGa.cpp:246
#3  0x00007f088bca65bb in std::__invoke_impl<RDKit::GaResult, RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> (__f=
    @0x5609f3800d98: (RDKit::GaResult (RDKit::RGroupGa::*)(RDKit::RGroupGa * const, int)) 0x7f088bc86576 <RDKit::RGroupGa::run(int)>, 
    __t=@0x5609f3800d90: 0x7ffc31ab1b20) at /usr/include/c++/9/bits/invoke.h:73
#4  0x00007f088bca63aa in std::__invoke<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> (__fn=
    @0x5609f3800d98: (RDKit::GaResult (RDKit::RGroupGa::*)(RDKit::RGroupGa * const, int)) 0x7f088bc86576 <RDKit::RGroupGa::run(int)>)
    at /usr/include/c++/9/bits/invoke.h:96
#5  0x00007f088bca61b7 in std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >::_M_invoke<0ul, 1ul, 2ul> (this=0x5609f3800d88) at /usr/include/c++/9/thread:244
#6  0x00007f088bca5f3e in std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >::operator()
    (this=0x5609f3800d88) at /usr/include/c++/9/thread:251
#7  0x00007f088bca5af9 in std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<RDKit::GaResult>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::operator() (this=0x7f08889f0d90) at /usr/include/c++/9/future:1339
#8  0x00007f088bca569c in std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<RDKit::GaResult>, std::__future_base::_Result_base::_Deleter>, std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult> >::_M_invoke(std::_Any_data const&) (__functor=...) at /usr/include/c++/9/bits/std_function.h:286
#9  0x00007f088bc8ae7e in std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>::operator()() const (this=0x7f08889f0d90) at /usr/include/c++/9/bits/std_function.h:688
#10 0x00007f088bc895a6 in std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*) (this=0x5609f3800d50, __f=0x7f08889f0d90, __did_set=0x7f08889f0cef)
    at /usr/include/c++/9/future:561
#11 0x00007f088bc93bc7 in std::__invoke_impl<void, void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::__invoke_memfun_deref, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) (__f=
    @0x7f08889f0d30: (void (std::__future_base::_State_baseV2::*)(std::__future_base::_State_baseV2 * const, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter>()> *, bool *)) 0x7f088bc8956c <std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*)>, 
    __t=@0x7f08889f0d00: 0x5609f3800d50) at /usr/include/c++/9/bits/invoke.h:73
#12 0x00007f088bc8f320 in std::__invoke<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) (
    __fn=
    @0x7f08889f0d30: (void (std::__future_base::_State_baseV2::*)(std::__future_base::_State_baseV2 * const, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter>()> *, bool *)) 0x7f088bc8956c <std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*)>)
    at /usr/include/c++/9/bits/invoke.h:95
#13 0x00007f088bc8ac0a in std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#1}::operator()() const (this=0x7f08889f0c70) at /usr/include/c++/9/mutex:671
#14 0x00007f088bc8ac41 in std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&--Type <RET> for more, q to quit, c to continue without paging--
&, bool*&&)::{lambda()#2}::operator()() const (this=0x0) at /usr/include/c++/9/mutex:676
#15 0x00007f088bc8ac56 in std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&)::{lambda()#2}::_FUN() () at /usr/include/c++/9/mutex:676
#16 0x00007f088a2d647f in __pthread_once_slow (once_control=0x5609f3800d68, init_routine=0x7f088ada8b80 <__once_proxy>)
    at pthread_once.c:116
#17 0x00007f088bc851f6 in __gthread_once (__once=0x5609f3800d68, __func=0x7f088ada8b80 <__once_proxy>)
    at /usr/include/x86_64-linux-gnu/c++/9/bits/gthr-default.h:700
#18 0x00007f088bc8ad17 in std::call_once<void (std::__future_base::_State_baseV2::*)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*), std::__future_base::_State_baseV2*&&, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*&&, bool*&&) (__once=..., __f=
    @0x7f08889f0d30: (void (std::__future_base::_State_baseV2::*)(std::__future_base::_State_baseV2 * const, std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter>()> *, bool *)) 0x7f088bc8956c <std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>*, bool*)>)
    at /usr/include/c++/9/mutex:683
#19 0x00007f088bc89324 in std::__future_base::_State_baseV2::_M_set_result(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>, bool) (this=0x5609f3800d50, __res=..., __ignore_failure=false)
    at /usr/include/c++/9/future:401
#20 0x00007f088bca4f44 in std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}::operator()() const (this=0x5609f3800d50) at /usr/include/c++/9/future:1662
#21 0x00007f088bca86a2 in std::__invoke_impl<void, std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}>(std::__invoke_other, std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}&&) (__f=...)
    at /usr/include/c++/9/bits/invoke.h:60
#22 0x00007f088bca8637 in std::__invoke<std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}>(std::__invoke_result&&, (std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}&&)...) (__fn=...)
    at /usr/include/c++/9/bits/invoke.h:95
#23 0x00007f088bca84e4 in std::thread::_Invoker<std::tuple<std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}> >::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x5609f3983238)
    at /usr/include/c++/9/thread:244
#24 0x00007f088bca80e7 in std::thread::_Invoker<std::tuple<std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}> >::operator()() (this=0x5609f3983238)
    at /usr/include/c++/9/thread:251
#25 0x00007f088bca692e in std::thread::_State_impl<std::thread::_Invoker<std::tuple<std::__future_base::_Async_state_impl<std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >, RDKit::GaResult>::_Async_state_impl(std::thread::_Invoker<std::tuple<RDKit::GaResult (RDKit::RGroupGa::*)(int), RDKit::RGroupGa*, int> >&&)::{lambda()#1}> > >::_M_run() (this=0x5609f3983230)
    at /usr/include/c++/9/thread:195
#26 0x00007f088ada9d84 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#27 0x00007f088a2cd609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#28 0x00007f088aa97293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

Code/GraphMol/RGroupDecomposition/RGroupDecomp.h
jones-gareth: Good idea.

jones-gareth:
@greglandrum haven't seen that one even though I develop on Linux.  I'll see if I can figure out what is happening from the stack trace.

greglandrum:
> @greglandrum haven't seen that one even though I develop on Linux. I'll see if I can figure out what is happening from the stack trace.

It’s also clearly not happening with the CI builds.

It’s one of those Heisenbugs that doesn’t happen all the time, but once it starts remains reproducible. I only saw it consistently with a debug build. With a normal build I would get it occasionally if I ran all the tests using `ctest -j4`, but that didn’t seem super reproducible.

bp-kelley:
@jones-gareth I did a PR against your repo with the "rebuilding"  molecules where possible.  Hopefully I did that right.  I'll follow up with some more examples.

jones-gareth:
> 
> 
> > @greglandrum haven't seen that one even though I develop on Linux. I'll see if I can figure out what is happening from the stack trace.
> 
> It’s also clearly not happening with the CI builds.
> 
> It’s one of those Heisenbugs that doesn’t happen all the time, but once it starts remains reproducible. I only saw it consistently with a debug build. With a normal build I would get it occasionally if I ran all the tests using `ctest -j4`, but that didn’t seem super reproducible.

I've just pushed what I think is the fix.

jones-gareth:
> 
> 
> @jones-gareth I did a PR against your repo with the "rebuilding" molecules where possible. Hopefully I did that right. I'll follow up with some more examples.

@bp-kelley looking at your PR, there are a lot of changes.  Would it make sense to finish reviewing my changes here before merging in your PR?

bp-kelley:
Just merge with the master branch, I’ll do a new pr.  It will cut down the changes.

jones-gareth:
> 
> 
> Just merge with the master branch, I’ll do a new pr. It will cut down the changes.

Merged with master

Code/GraphMol/RGroupDecomposition/RGroupScore.h
```diff
@@ -13,6 +13,7 @@
 #include "RGroupMatch.h"
 #include <vector>
 #include <set>
+#include <boost/serialization/vector.hpp>
```
greglandrum, line 16: It's not clear why this is here.
If it's necessary, it should at least be behind `#ifdef RDK_USE_BOOST_SERIALIZATION`

greglandrum (changes requested):
I'm going to assume that you and @bp-kelley have the substantial bits of this covered. So this focuses on smaller/formatting bits.

Code/GraphMol/RGroupDecomposition/RGroupGa.cpp
```diff
@@ -0,0 +1,357 @@
...
+    fitness = fingerprintVarianceGroupScore(labelsToVarianceData);
+    // assert(fitness == recalculateScore());
+  } else {
+    fitness = rGroupData.score(permutation, &labelsToVarianceData);
+  }
+  return fitness;
+}
+
+double RGroupDecompositionChromosome::recalculateScore() {
+  std::cerr << "Recalculating score" << std::endl;
```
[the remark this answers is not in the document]
jones-gareth: Deleted that line

Code/GraphMol/RGroupDecomposition/RGroupGa.cpp
```diff
@@ -0,0 +1,357 @@
...
+#include <limits>
+#include <future>
+#include "RGroupGa.h"
+#include "RGroupDecompData.h"
+#include "RGroupDecomp.h"
+#include "../../../External/GA/util/Util.h"
+
+namespace RDKit {
+
+using namespace std;
```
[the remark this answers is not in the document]
jones-gareth: Removed namespace declaration

Code/GraphMol/RGroupDecomposition/RGroupUtils.cpp
```diff
@@ -86,6 +86,7 @@ bool setLabel(Atom *atom, int label, std::set<int> &labels, int &maxLabel,
       }
     }
 
+    BOOST_LOG(rdDebugLog) << "setting label on atom " << atom->getIdx() << " to " << label << std::endl;
```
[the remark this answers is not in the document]
jones-gareth: I've deleted that log statement

Code/GraphMol/RGroupDecomposition/RGroupScore.h
jones-gareth: I've removed it.  Not sure how it got there- my guess is I selected the wrong vector during autocomplete.

greglandrum:
> I've just pushed what I think is the fix.

Just tried a few times with this version and the problem no longer occurs for me

greglandrum (approved):
LGTM
I'm fine with merging this whenever you guys are happy with the technical side

bp-kelley:
@jones-gareth @ptosco @greglandrum I'm pretty happy with this.  I reserve the right to do another cleanup pass as the code is getting pretty complicated at this point, but I would love to merge this.

It's working pretty ace with my free Wilson stuff.

bp-kelley (approved):

jones-gareth:
> It's working pretty ace with my free Wilson stuff.

@bp-kelley I'd be interested in finding out more about your Free Wilson implementation. We're likely to be working on that in the future.

greglandrum:
@ptosco : any objections to me merging?

ptosco:
@greglandrum Please go ahead and merge - this is a great piece of work and it should be made available ASAP so it can be enjoyed and tested by the wider community.

bp-kelley:
@jones-gareth Thanks, this is awesome!

jones-gareth:
@bp-kelley @greglandrum @ptosco Thanks everybody for your help, this was a lot more involved than I originally anticipated.  I'm sure it is not the end of the story!

PR #3573, opened 2020-11-18
Fixes #3540

### LINKED ISSUE #3540: Drawing artifacts in draw_to_canvas_with_offset
**Configuration:**

WASM version of RDKit as of Nov 1 2020 from: https://rdkit.org/rdkitjs/beta/GettingStartedInJS.html (v2020.09)
Chrome Version 86.0.4240.111 (Official Build) (64-bit)
Microsoft Windows [Version 10.0.18363.1139]

**Description:**

`draw_to_canvas_with_offset` brings a drawing artifact in the top-left canvas corner.

This is what we have prior to drawing the molecule to the canvas:

```
var canvas = document.getElementById("draw-canvas-2");
var ctx = canvas.getContext("2d");
ctx.fillStyle = "#00FF00";
ctx.fillRect(0,0,300,300);
```
image

Here is what happens if we add a molecule drawing to the canvas:

```
var smiles = "CC(=O)Oc1ccccc1C(=O)O";
var mol = Module.get_mol(smiles);
var canvas = document.getElementById("draw-canvas-2");
var ctx = canvas.getContext("2d");
ctx.fillStyle = "#00FF00";
ctx.fillRect(0,0,300,300);
mol.draw_to_canvas_with_offset(canvas, 100, -100, 150, 100);
```
image

There is a clear artifact rectangle in the top-left, which is sized same as the molecule render.

ptosco:
The `clearDrawing()` methods were not taking the `offset` into account when clearing the background prior to drawing the molecule. This PR fixes that.

### DIFF
--- a/Code/GraphMol/MolDraw2D/MolDraw2DCairo.cpp
+++ b/Code/GraphMol/MolDraw2D/MolDraw2DCairo.cpp
@@ -167,7 +167,7 @@ void MolDraw2DCairo::drawPolygon(const std::vector<Point2D> &cds) {
 void MolDraw2DCairo::clearDrawing() {
   PRECONDITION(dp_cr, "no draw context");
   setColour(drawOptions().backgroundColour);
-  cairo_rectangle(dp_cr, 0, 0, width(), height());
+  cairo_rectangle(dp_cr, offset().x, offset().y, width(), height());
   cairo_fill(dp_cr);
 }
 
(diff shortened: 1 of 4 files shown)


bp-kelley (approved):

greglandrum (approved):
You could theoretically add a test for this by using MolDraw2DSVG and making sure that the output doesn't have the `x='0' y='0'` pattern in it (at least I think it's that simple).

I don't think it's necessary, so if you want to go ahead and merge as is, I'm fine with it.

PR #3588, opened 2020-11-26
update documentation, include reference to Debian Package Tracker

nbehrnd:
#### Reference Issue
The description of installing RDKit in Linux mentions Ubuntu, but not Debian.  Sometimes, the version packaged in Debian's repositories (branch testing) is ahead of RDKit's version in Ubuntu releases (especially Ubuntu LTS) releases.

The change now additionally points to a site listing which version was made available as .deb in either branch stable / testing / unstable and offers contact to the debichem maintainers.

#### Any other comments?
I'm aware a (mini)conda environment may offer using a more recent version of RDKit, than by installing a .deb package.

### DIFF
--- a/Docs/Book/Install.md
+++ b/Docs/Book/Install.md
@@ -173,13 +173,14 @@ If you are trying to use multiple installations of PostgreSQL in different envir
 
 ### Installation from repositories
 
-#### Ubuntu 12.04 and later
+#### Ubuntu 12.04 and later, Debian
 
-Thanks to the efforts of the Debichem team, RDKit is available via the Ubuntu repositories. To install:
+Thanks to the efforts of the Debichem team, RDKit is available via the Ubuntu and Debian repositories. To install:
 
 ```shellsession
 $ sudo apt-get install python-rdkit librdkit1 rdkit-data
 ```
+The current status of packaging RDKit for the branches of Debian is published on https://tracker.debian.org/pkg/rdkit
 
 #### Fedora, CentOS, and RHEL
 


UnixJunkie:
Is this the python3-ready version of rdkit that you mention?

UnixJunkie:
I don't think people will care unless this is the rdkit version which is recent enough to work with Python3.
And I guess this one is not available on Ubuntu 12.04, since it is not even available on 18.04.

nbehrnd:
Thank you for providing the hint, because my initial interest was
only to hint to Debichem's moderated site, but not about changing the
actual instruction by sudo apt-get install.

Occasionally facing an old installation of Xubuntu 18.04.4 LTS
(bionic), the present command indeed would fetch an old version of
RDKit for Python 2.  In said installation, this is RDKit 2016.03.5
(based on the release by October 10, 2016) within Python 2.7.17
(November 2019).

On the other hand -- just checked -- a current Debian 10 (bullseye /
sid (testing)) currently provides RDKit 2020.03.4 (i.e., July 2,
2020) within Python 3.8.6 (Sep 25, 2020) if calling for
python3-rdkit.  At least in the mirror used to update, python-rdkit
is gone for months, so there, there is no successful `import rdkit`
within Python 2.7.18 (April 20, 2020).

According to Ubuntu's searchable package index, the current LTS
20.04 (focal) equally transitioned to python3-rdkit only, too and
offer RDKit 2019.09.1 (i.e., October 25, 2019).

UnixJunkie:
The install instructions for Debian/Ubuntu could be updated, but you would need
to say from which debian/ubuntu version users get a Python3-ready rdkit installed via apt.
In effect, installing via apt is a breeze compared to conda (much faster; much less pain for the sysadmin).
Apt also provides a system-wide install, which is interesting on servers/shared computers.

nbehrnd:
After retraction of this pull request, I filed a new PR addressing
this on Friday, Nov 27th.  For now, a decision by RDKit's maintainers
about this suggest is pending.

For one, the added instruction covers updated .deb based Linuxes such
as current Ubuntu 20.04 LTS and Debian 10 / testing; the later tested
to work successfully (python3-rdkit / Python 3.8.6).  For two, the
earlier line is retained to offer an installation aiming deployment
in Python 2.7.17 in an senior setup like Xubuntu 18.04.4 LTS.

As for the comfort the installation with .deb packages offer, I agree
with you; RDKit then is only a few keystrokes away.  Beside this, I
however would like to add conservation of disk space and band width
fetching / maintaining the independent installation to the balance.

To use a recently corrected feature in RDKit, I installed RDKit
within Miniconda, the light-weight subset of Anaconda.  RDKit
delivered the results the expected way.  But (e.g, for such a test)
there is a difference to either

a) fetch and install three .debs worth 47.6 MB debichem repackaged
RDKit in addition to an already working and apt-get maintained
installation of Python3, .or.

b) to accommodate about 1.1 GB+ to for Miniconda plus RDKit plus some
larger libraries (e.g., matplotlib, pandas) because Miniconda's
package system basically starts as an independent installation and
adds up (e.g., updating Miniconda).  Even if -- possibly slightly
different only -- the same libraries are just a few folders away to
assist «the other» Python to work.

It is a conceptual nuisance, not a problem related per se to RDKit,
nor Anaconda, nor debichem.

PR #3663, opened 2020-12-19
Make the scope control for Qt more idiomatic

d-b-w:
I was looking at some internal examples, and I think that a
context manager is the idiomatic way to control lifetimes
of the QPainter and QImage.

I ran the tests and can reproduce some terrible lifetime
crashes if I don't use the context manager (and don't
use Greg's attribute trick).


ZontaNicola:
I don't have enough experience with pyqt, but this looks good!

greglandrum (approved):

PR #3684, opened 2021-01-02
Update build instructions for Linux

sirbiscuit:
Numpy and matplotlib are not part of the base installation of miniconda
(anymore?). Add them to the list of packages that should be installed
using conda.

Specify the boost version (boost==1.67.0), because installing the most
recent version raises errors during "make" (at the moment).

Instruct the user to create a conda activate.d and deactivate.d script
to configure the path variables (LD_LIBRARIES_PATH, PYTHONPATH,
RDBASE).

Simplify the cmake instruction, because specifying BOOST_ROOT and
PYTHON_NUMPY_INCLUDE_PATH is not necessary anymore (in this setup).

#### Any other comments?

A script in env_name/etc/conda/activate.d is executed automatically when a conda environment (here: env_name) is activated. The (suggested) installation instructions use this mechanism to configure the environment variables RDBASE, LD_LIBRARY_PATH and PYTHONPATH. From my point of view, this is a reasonable way to install a local RDKit directory in a conda environment, but it looks extremely complicated. Feel free to suggest a different solution!

### DIFF
--- a/Docs/Book/Install.md
+++ b/Docs/Book/Install.md
@@ -88,37 +88,73 @@ introduced in more recent macOS versions.
 
 The following commands will create a development environment for Linux x86_64 and Python 3.
 
-Start by downloading the latest anaconda installer from [Anaconda](https://www.anaconda.com/download/#linux) and install it. Then, install the required packages:
+First, install conda [as described previously](#how-to-get-conda), create a new conda environment and activate it:
 
-```
-bash Anaconda3-5.2.0-x86_64.sh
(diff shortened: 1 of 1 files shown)


Polydynamical (approved):

greglandrum (commented):
Thanks for taking a look at this! I think we should modernize a bit further: switch to conda-forge and update the boost version to 1.76. 

There are examples in the azure devops folder for how to do this.

If you don't want to take that on, let me know and I can do a PR against your branch

PR #3691, opened 2021-01-04
Remove rdmolops link dependency on MolDraw2D

ricrogz:
Even after #3659, the rdmolops Python wrapper gets linked to Qt on linux because of MolDraw2D.

To prevent this, and since rdmolops doesn't need to be linked against MolDraw2D, I'm removing the link dependency.


ricrogz:
My bad: I looked at the wrong file... rdmolops **does** require MolDraw2D.

PR #3735, opened 2021-01-16
PandasTools.LoadSDF -- read performance 

Vladimber:
Hello!
I think that the PandasTools.LoadSDF code can be optimized to avoid iterative deletion of properties when reading the file (by default embedProps = False). Instead it will be nice to have a method that resets the entire property map.
The changes are rather small, while the speedup in reading large files can be significant (about 25% for a file with only ten thousand compounds with four properties).
I will be glad to hear your opinion on this proposal.


Vladimber:
My bad, found serious issues with this solution.

PR #3743, opened 2021-01-18
Improve consistency in R-group tables involving symmetric groups

ptosco:
I noticed that R-group tables involving symmetric $-groups can be disappointing when the `NoAlignment` flag is set.

Take this example: 2 pre-labelled cores, 2 molecules.
The two cores are pre-labelled with 3 R groups each; `R1` and `R3`, which label symmetric positions, are intentionally inverted.

```
from rdkit import Chem
from rdkit.Chem.rdRGroupDecomposition import RGroupDecompositionParameters, RGroupDecomposition, RGroupCoreAlignment
import pandas as pd
from IPython.core.display import HTML

def show(df):
    return HTML(df.to_html(notebook=True))

def groups_to_df(groups, mols, include_core=False, redraw_sidechains=False):
    """ add the molecule to the dataframe 
    """
    cols = ['Mol', 'Id']+list(groups.keys())
    if redraw_sidechains:
        for k,vl in groups.items():
            if k=='Core':
                continue
            for i,v in enumerate(vl):
                vl[i] = Chem.RemoveHs(v)
                rdDepictor.Compute2DCoords(vl[i])


    if not include_core:
        cols.remove('Core')
        del groups['Core']
    groups['Mol'] = mols
    groups['Id'] = [m.GetProp("_Name") for m in mols]
    frame = pd.DataFrame(groups,columns=cols)
    PandasTools.ChangeMoleculeRendering(frame)
    return frame

core1 = Chem.MolFromSmiles('[cH:1]1[cH:2][cH:3]ccc1')
core1
```
image

```
core2 = Chem.MolFromSmiles('[cH:3]1[cH:2][cH:1]cnc1')
core2
```
image

Now I add 2 molecules which bear the same `R1` and `R3` groups; `mol1` matches the first core and `mol2` matches the second core.

```
mol1 = Chem.MolFromSmiles("c1(CO)c(F)c(CN)ccc1")
mol1.SetProp("_Name", "mol1")
mol1
```
image

```
mol2 = Chem.MolFromSmiles("c1(CO)c(Cl)c(CN)cnc1")
mol2.SetProp("_Name", "mol2")
mol2
```
image

Now I do an R-group decomposition setting `alignment` to `RGroupCoreAlignment.NoAlignment`. This means that the two cores will not be MCS-aligned.
Both `mol1` and `mol2` can match their respective cores in two different ways, given the `R1` and `R3` symmetry.
I would expect that, of the 4 possible equi-scoring permutations, the permutation where the CH<sub>2</sub>OH and CH<sub>2</sub>NH<sub>2</sub> R-groups are both `R1` or `R3` is chosen.
However, this does not happen:

```
ps1 = RGroupDecompositionParameters()
ps1.removeHydrogensPostMatch = True
ps1.onlyMatchAtRGroups = True
ps1.alignment = RGroupCoreAlignment.NoAlignment
rgd1 = RGroupDecomposition([core1, core2], ps1)

rgd1.Add(mol1)
0

rgd1.Add(mol2)
1

rgd1.Process()
True

show(groups_to_df(rgd1.GetRGroupsAsColumns(), [mol1, mol2], include_core=True))
```
image

The reason is that currently the tie-breaking rule only attempts to keep hydrogen R-groups together, but does nothing about identical R-groups. This is usually masked by the fact that MCS alignment of cores takes care of symmetry. But in cases where I do not want to do MCS alignment of cores because I have carefully pre-labelled cores whose labels I do not want to alter, I have to live with this disappointing R-group table.

This PR introduces a simple fix to address this problem and introduce better consistency in R-group tables regardless of the `alignment` setting, i.e.:
image

However, as matching is set to `RGroupMatching.GreedyChunks`, I would expect that the permutation


greglandrum:
@bp-kelley are you going to have time to take a look at this one?

bp-kelley:
@greglandrum @ptosco I've been mulling this over and I admit to fact that the complexity of the code is making me nervous.  This appears to be a regression included by the tie breaking code, previously identical r-groups were being considered and what I'm confused about is where there was a tie in the first place.

ptosco:
@bp-kelley Not sure what you mean by "previously". I have just tested RDKit 2020.03.6, before any change to RGD was made by me or Gareth, and the problem was already there:

```
import rdkit
rdkit.__version__
'2020.03.6'

[...]

ps1 = RGroupDecompositionParameters()
ps1.removeHydrogensPostMatch = True
ps1.onlyMatchAtRGroups = True
# This was needed in older RDKit versions
ps1.alignment = getattr(RGroupCoreAlignment, "None")
rgd1 = RGroupDecomposition([core1, core2], ps1)

rgd1.Add(mol1)
0

rgd1.Add(mol2)
1

rgd1.Process()
True

show(groups_to_df(rgd1.GetRGroupsAsColumns(), [mol1, mol2], include_core=True))
```

image

My impression is that this went unnoticed as most users kept the MCS `alignment` default. Using MCS, the problem does not appear.

ptosco:
Also, I think the proposed fix is quite neat. I am not convinced that simply adding another tie-breaking criterion on top of the existing ones adds that much complexity to the already quite scary RGD mix...

bp-kelley:
By previously I'm just confused why the alignment and non-alignment make a difference here, the scoring function given all symmetry matches should have figured this out.  I'm worried that the alignment is actually not fixing the issue.

bp-kelley:
i.e. I'm confused why a fix is needed.  I'll do a debug dump just to see, I don't have any technical problems with the fix, but I'd rather not have it be necessary.

bp-kelley:
@ptosco I've allocated a chunk of time Friday to satisfy my curiosity.

ptosco:
@bp-kelley I mumbled a bit on your words and I figured out that it's just a matter of removing isotopic labels before comparing SMILES, as they will be different if MCS alignment did not take place, but this does not mean substituents are not the same.
So here's a much neater fix that I guess you are going to like :-)

bp-kelley:
There's been a regression in the scoring function (or maybe it was always there), when comparing smiles strings it looks like isotopes are getting mixed into the r-groups when using NoAlign

```
Label: 3
3 00 [2*]C([H])([H])N([H])[H] count:1
3 00 *C([H])([H])N([H])[H] count:1
    lscore^2 0: 0.25    lscore^2 1: 0.0625Max Linker Matches :0
Increment: 0.3125 Linker_Increment: 1
increment*linkerIncrement: 0.3125
Score = 2.1875
Final Score = 2.1875
```

And aren't being considered a match.  I think this is the root cause.

bp-kelley:
It's the opposite actually, the isotope labels are forced to be the same during mcs alignment
```
Label: 3
3 00 [2*]C([H])([H])O[H] count:1
3 00 [2*]C([H])([H])O[H] count:2
    lscore^2 0: 1Max Linker Matches :0
Increment: 1 Linker_Increment: 1
increment*linkerIncrement: 1
Score = 3.1875
Final Score = 3.1875
```
We can fix this, I think, by removing or aligning the isotopes labels prior to scoring step when using NoAlign.

ptosco:
This is basically what my latest revision of the PR does. i don't think we need the isotope labels in the `smilesVect` at all.

ptosco:
I think it's always been there but no-one noticed as `NoAlign` is seldom used (if you think, no one before Riccardo noticed that `RGroupAlignemnt.None` was hardly accessible from Python)

Code/GraphMol/RGroupDecomposition/RGroupData.h
```diff
@@ -59,7 +59,10 @@ struct RGroupData {
               std::inserter(attachments, attachments.end()));
 
     mols.push_back(newMol);
-    smilesVect.emplace_back(MolToSmiles(*newMol, true));
+    static const std::regex remove_isotopes_regex("\\[\\d*\\*\\]");
+    std::string smiles_no_isotopes = std::regex_replace(
+        MolToSmiles(*newMol, true), remove_isotopes_regex, "*");
+    smilesVect.push_back(std::move(smiles_no_isotopes));
```
bp-kelley, line 65: For some reason, I just don't like the regular expression usage, but I'm going to ignore my feelings for now and profile the change.

Let's make it easier on copy elision and make this:

```
smilesVect.push_back(
  std::regex_replace(MolToSmiles(*newMol, true), remove_isotopes_regex, "*")
);
```

bp-kelley (changes requested):

Code/GraphMol/RGroupDecomposition/RGroupData.h
bp-kelley: By profile the change I mean see if this is faster than removing the isotope from the atom before making the smiles string :)

ptosco: I broke it in two just for clarity - with the `std::move` the two constructs should be equivalent once compiled. Anyway, I am happy to do the change.
I don't particularly fancy the regex either, it just seems the simplest way to get job done.

bp-kelley: And I'm glad we came to the same conclusions, if the scoring function couldn't align these, there was something seriously wrong there.

This probably fixes some edge cases with multiple core replacements as well :)

ptosco: Done.

ptosco: > And I'm glad we came to the same conclusions, if the scoring function couldn't align these, there was something seriously wrong there.

Absolutely! Glad that you insisted as it prompted me to reconsider my initial patch.

bp-kelley:
When the tests pass, I'll merge.

ptosco:
Thanks Brian!

PR #3751, opened 2021-01-19
Addresses #3740

### LINKED ISSUE #3740: Test suite failures if eigen3 is not available
**Describe the bug**

The testsuite testcase #85 fails like this for me:

```
 85/179 Test  #85: pyMolDescriptors .......................***Failed    0.52 sec
  File "/build/rdkit-vngW4C/rdkit-202009.1/Code/GraphMol/Descriptors/Wrap/testMolDescriptors.py", line 5, in <module>
    from rdkit.Chem import rdMolDescriptors as rdMD, Descriptors
  File "/build/rdkit-vngW4C/rdkit-202009.1/obj-x86_64-linux-gnu/rdkit/Chem/Descriptors.py", line 222, in <module>
    _du.VectorDescriptorWrapper(_rdMolDescriptors.BCUT2D, names=names, version="1.0.0", namespace=locals())
AttributeError: module 'rdkit.Chem.rdMolDescriptors' has no attribute 'BCUT2D'
```

**To Reproduce**

Run the test suite

**Expected behavior**

Testsuite testcase #85 succeeds.

 - RDKit version: 202009.1
 - OS: Debian unstable
 - Python version (if relevant): 3.9.1
 - Are you using conda? no
 - If you are not using conda: how did you install the RDKit? from source

ricrogz:
Resolves #3740 

This adds some checks to skip setting up and testing some descriptors that are not available if Eigen has not been found by CMake.

I highly recommend enabling the "Hide whitespace changes" option when looking at this patch, since it seems the files I changed were not formatted before my changes, and applying yapf to them made a lot of changes.


greglandrum (approved):
Changes look good and I think they make sense for at least the rest of the 2020.09 release cycle.

Maybe it's worth having a conversation about whether or not we should add eigen3 as a requirement for the RDKit? It's header only, so it doesn't seem like a burden on other developers and it could simplify things on the RDKit side.

PR #3767, opened 2021-01-27
R group match any issue

jones-gareth:

Fixes issues round RGD where rgroups may be attached to any atom and wildcard/query atoms are present in the core.

There is an example [here](https://gist.github.com/jones-gareth/9e97d5ce9981d83a8189849146b00193) where the decomposition is obviously wrong. This is a more extreme example of the issue seen in test testMultiCorePreLabelled where 
we see `R4:c(c:[*:5]):[*:4] R5:c(c:[*:5]):[*:4]`, which should really be something like `R4c[*:4] R5:c[*:5]`. 

There are also plenty of other examples in the tests that show decompositions that, while they may not be strictly wrong, are pretty confusing.  For example, test_github3402 has `'R1:c1ccc(C([*:1])=[*:1])cc1` which should be `'R1':c1ccc(C[*:1])cc1`.

The code that was responsible for these issues also causes an apparently unrelated issue.  In testRingMatching3 core `*1***[*:1]1` decomposes `c1cocc1CCl` to give `Core:c1co([*:2])cc1[*:1] R1:ClC[*:1]`.  The original core is lost and replaced by a new core of non-query atoms, even though the original core clearly matches the structure.

The changes in this CR address these issues and implements the [suggestion](https://github.com/rdkit/rdkit/pull/3565#issuecomment-749622868) from @bp-kelley for labeling attachments to core query atoms that are not terminal user specified rgroups.

- The original example now is [correct](https://gist.github.com/jones-gareth/6e500c86150f7c968b40485326ba5ce0). It has been added as a test.
- The groups in testMultiCorePreLabelled are now `R4:[H][c:4] R5:[H][c:5]`.
- The example from testRingMatching3 is now `Core:[*:1]1[*:2][*:3][*:4][*:5]1 R1:[H][c:1] R2:[o:2] R3:[H][c:3]`
- test_github3402 is now `R1: c1ccc([C:1])cc1`
- There are plenty more examples in the PR
- I've updated `Docs/Notebooks/RGroupDecomposition-Tests.ipynb` with the new labeling scheme and additional tests

#### Any other comments?

The main fix is in `RCore::replaceCoreDummiesWithMolMatches` to prevent errors in the fragments when `MolOps::getMolFrags` is called from `RGroupDecomposition::add`.  I believe that the  bit of code that creates new cores is no longer needed and indicative of an error.  But I'm too chicken to delete it. Most of the other changes are bookkeeping to handle the replacements properly.

Handling of hydrogens connected to the query atom replacement is tricky, but I think has been done correctly.  It leads to rgroups like `R5: [H][C:5].[H][C:5]` (for a query atom in the core that matches an sp3 carbon with 2 hydrogens in the structure)

I did see one issue- the decomp for  testRingMatching3 puts the largest RGroup containing halogens on R5 - an atom index rgroup.  Ideally, all the halogens should be in R1 - the user defined R Group.  This is (presumably) a separate issue round ranking of duplicate scoring decompositions so I haven't looked at it here. If you agree I'll add it as an issue.


Code/GraphMol/RGroupDecomposition/RGroupCore.h
```diff
@@ -36,9 +39,10 @@ struct RCore {
       }
     }
   }
+
   // Return a copy of core where dummy atoms are replaced by
   // the atomic number of the respective matching atom in mol
-  std::unique_ptr<ROMol> replaceCoreDummiesWithMolMatches(
+  std::unique_ptr<ROMol> replaceCoreDummiesWithMolMatchesOriginal(
```
ptosco, line 45: Please feel free to remove `replaceCoreDummiesWithMolMatchesOriginal` as your version is much more comprehensive than the one I had originally put in place to fix a problem I had found.

Code/GraphMol/RGroupDecomposition/RGroupCore.h
```diff
@@ -49,6 +53,47 @@ struct RCore {
...
+    std::unique_ptr<RWMol> coreReplacedDummies(new RWMol(*core));
+    for (const auto &p : match) {
+      auto atom = coreReplacedDummies->getAtomWithIdx(p.first);
+      if (isAnyAtomWithMultipleNeighborsOrNotUserRLabel(*atom)) {
+        auto molAtom = mol.getAtomWithIdx(p.second);
+        if (atom->hasQuery()) {
+          auto atomicNumber = molAtom->getAtomicNum();
+          auto newAtom = new Atom(atomicNumber);
+          coreReplacedDummies->replaceAtom(p.first, newAtom, false, true);
+          delete newAtom;
```
ptosco, line 70:
```suggestion
          Atom newAtom(atomicNumber);
          coreReplacedDummies->replaceAtom(p.first, &newAtom, false, true);
```

Code/GraphMol/RGroupDecomposition/RGroupCore.h
```diff
@@ -49,6 +53,47 @@ struct RCore {
...
+    }
+
+    for (auto bond : coreReplacedDummies->bonds()) {
+      if (bond->hasQuery()) {
+        const auto molBond =
+            mol.getBondBetweenAtoms(match[bond->getBeginAtomIdx()].second,
+                                    match[bond->getEndAtomIdx()].second);
+        auto newBond = new Bond(molBond->getBondType());
+        coreReplacedDummies->replaceBond(bond->getIdx(), newBond, true);
+        delete newBond;
```
ptosco, line 86:
```suggestion
        Bond newBond(molBond->getBondType());
        coreReplacedDummies->replaceBond(bond->getIdx(), &newBond, true);
```

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
```diff
@@ -221,11 +242,19 @@ int RGroupDecomposition::add(const ROMol &inmol) {
...
+            std::cerr << "Fragment " << i << " " << MolToSmiles(*newMol)
+                      << std::endl;
+#endif
           }
         } else {
           // special case, only one fragment
           if (fragments.size() == 1) {  // need to make a new core
             // remove the sidechains
+            
+            // GJ I think if we ever get here that it's really an error and I believe
```
ptosco, line 255: If
1) this code was probed by unit tests previously and it is not anymore after your changes
2) those unit tests still pass after your changes
I'd be in favor of deleting it: the RGD code is already sufficiently complicated :-)

Code/GraphMol/RGroupDecomposition/RGroupUtils.cpp
```diff
@@ -87,13 +86,23 @@ bool setLabel(Atom *atom, int label, std::set<int> &labels, int &maxLabel,
...
     labels.insert(label);
     maxLabel = (std::max)(maxLabel, label + 1);
     return true;
   }
   return false;
 }
 
+bool isAnyAtomWithMultipleNeighborsOrNotUserRLabel(const Atom &atom) {
+  if (atom.getAtomicNum()) return false;
+  if (atom.getDegree() > 1) return true;
```
ptosco, line 99: The comment in the header reads:
```
Returns true if the core atom is a dummy atom with a single connection and a user defined rgroup label
```
But the function seems to return `true`:
- if it is a dummy atom with multiple connections
- if it is a dummy atom with a single connection and no user-defined R-label

Code/GraphMol/RGroupDecomposition/RGroupCore.h
jones-gareth: Removed.

Code/GraphMol/RGroupDecomposition/RGroupCore.h
jones-gareth: Changed

Code/GraphMol/RGroupDecomposition/RGroupUtils.cpp
jones-gareth: You're correct.  I rewrote that function a few times and forgot to update the header.  I've put in a correct comment.

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
jones-gareth: This code path was used by tests such as testRingMatching3.  It is no longer used by any of the tests.  I'll leave it in for now and see if @greglandrum has any thought on this, since I think he wrote the code.

bp-kelley:
@jones-gareth I've allocated a chunk of time Friday to go over this.  Thanks for looking into this.

I might try to add the [o:1] spec to molzip as well and then document it ;)

Code/GraphMol/RGroupDecomposition/RGroupCore.h
jones-gareth: Changed

bp-kelley:
@jones-gareth I'm a bit conflicted here, I think the behavior is suboptimal but essentially correct.  It appears to be extracting the largest portion it possibly can.  I think this should be the default behavior if the query atoms **aren't** in a ring, but I think this is a case where we need to have a lot of examples so I'll need an extra day or so to assemble them.

jones-gareth:
> 
> 
> @jones-gareth I'm a bit conflicted here, I think the behavior is suboptimal but essentially correct. It appears to be extracting the largest portion it possibly can. I think this should be the default behavior if the query atoms **aren't** in a ring, but I think this is a case where we need to have a lot of examples so I'll need an extra day or so to assemble them.

My take is that an rgroup should only overlap with the core at the point of attachment, regardless of the number of query atoms in the core.  If the core contains a user specified attachment point (as in my notebook example) then the rgroup should stop there and not extract further.

I feel that this should also be true for acyclic portions of the sketched core.  If a user has sketched some acyclic shrubbery or even an acyclic "core" it's hard to explain or interpret rgroups overlapping with the input structure.  Perhaps we need yet another option?

One thing I have thought of doing is outputting the core (when converting to rows or columns) with the core query atoms and bonds substituted by the structure matches.

bp-kelley:
@jones-gareth I think I can get behind that logic and we should output the core as found in the molecule in these cases.

jones-gareth:
@bp-kelley Should I add the matched core output to this CR, or would it be better to open a follow on PR.

bp-kelley:
@jones-gareth let's fix this one, I'll add the new spec to molzip and we can add "use extracted core" as a new feature request.

greglandrum:
I would like to (strongly) suggest that we impose a moratorium on merging further changes to the R-Group Decomposition code until we've got better tests in place.
Please take a look at #3778 
Unless you violently disagree, please convert this to a draft and we can come back to it after we're all happy that the code is being tested properly.

greglandrum (changes requested):
I would like to (strongly) suggest that we impose a moratorium on merging further changes to the R-Group Decomposition code until we've got better tests in place.
Please take a look at #3778
Unless you violently disagree, please convert this to a draft and we can come back to it after we're all happy that the code is being tested properly.

bp-kelley:
@jones-gareth, @greglandrum  also found this one which I sort of knew about.

```
m = Chem.MolFromSmiles('C1CC2NC12')
scaff = Chem.MolFromSmiles('C1([*:1])C([*:2])CC1')
rdRGroupDecomposition.RGroupDecompose([scaff],[m],asSmiles=True)[0]
```
Core:

image

R1

image

Which is pretty wrong for a few reasons as is the one:

```
m = Chem.MolFromSmiles('C1CC2NC12')
scaff = Chem.MolFromSmiles('C1([*:1])CCC1')
rows = rdRGroupDecomposition.RGroupDecompose([scaff],[m],asSmiles=False)[0]
```

Core:

image

R1:

image

jones-gareth:
> 
> 
> @jones-gareth, @greglandrum also found this one which I sort of knew about.
> 

@bp-kelley Thanks, I'll take a look, I'm thinking it is something to do with the symmetry

bp-kelley:
@jones-gareth I don't think it is a symmetry issue, it's just that if R1 and R2 are specified, when you substructure match then simply can't match the same atom.  So one matches the N and ones matches a hydrogen and it all falls apart.

If you use a "labeled" core without RGroups it all works great.

```
m = Chem.MolFromSmiles('C1CC2NC12')
scaff = Chem.MolFromSmarts('[C:1]1[C:2]CC1')
rows = rdRGroupDecomposition.RGroupDecompose([scaff],[m],asSmiles=False)[0]
```

Core:

image

R1

image

R2

image

I'm wondering if this is the real solution for these cases, just move the RGroup label to the core and remove the dummy atom...

jones-gareth:
@bp-kelley Okay, I understand now.  I think your suggestion is correct.  At least I can't see another way.  Shall I take a stab at adding to this CR?

greglandrum:
@bp-kelley @jones-gareth : should I wait until the set of changes you just discussed are in before I review this?

jones-gareth:
@greglandrum I will another look at this before you review.  I need to rebase on master and see how to integrate with the new notebook test visualization.  It would also be nice to include a fix for the cyclo-butane issue raised by @bp-kelley in this CR, but I don't know how feasible that will be.  Should have time over the next few days.

jones-gareth:
I have discarded the original [suggestion](https://github.com/rdkit/rdkit/pull/3565#issuecomment-749622868) of putting the matches to query core atoms on the sidechains and replaced by putting the matches on the output core.  So, when a core includes dummy or query atoms that are not rgroups then the core output from getRGroupsAsRows/Columns will contain the matches for those query atoms.

The example that I used at the top of the pull request looks like this:

Capture

(see also https://gist.github.com/jones-gareth/e62493ae37020f131b996b375460529a for an updated notebook).

When preparing the core for matching, user rgroups that are dummy atoms of degree 1 are moved to their neighbor while the decomposition is run and reattached during relabeling for output.  This fixes the error in the cyclopropane example.

Capture

I've added a test called `testSingleAtomBridge`.  This fix also means that there is a (correct) match in the now confusingly named `testSDFGRoupMultiCoreNoneShouldMatch` test. The current code is still susceptible to the original error if there is more that one rgroup attached to a core atom.  If you agree, I'll create an issue for that.

I've restricted identifying dummy atoms as user defined rgroups to those cases where the dummy has degree 1.  So,  `n1**c(*)*c1` has one user specified rgroup,  and `n1*****1` has none (when previously it had 5 user specified rgroups).  The original, looser, specification was making some of the test decompositions really ugly and didn't seem consistent with how people sketch rgroups.  I can change it back or (better) make it an option if you don't like it.

I've updated `Docs/Notebooks/RGroupDecomposition-Tests.ipynb` with the changes so you can see what the current test specs  look like.

I had thought of moving some of the representative decompositions from `testRGroupDeomp.cpp` to Greg's JSON catch test.  After thinking about it, I believe it would make sense to move all predicted decompositions into JSON format.  This is a bit of work and probably belongs in a new PR.  If you agree I'll add it as an issue.

bp-kelley:
@jones-gareth Looks good!  I have been having trouble getting in-the-wild r group decompositions, I might steal some from some patents as a stop gap or try an make some more like the cubane.

greglandrum:
> The current code is still susceptible to the original error if there is more that one rgroup attached to a core atom. If you agree, I'll create an issue for that.

Please do so that this doesn't get lost.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
```diff
@@ -473,6 +473,13 @@ ROMol *replaceCore(const ROMol &mol, const ROMol &core,
       // add the bonds now, after we've finished the loop over neighbors:
       for (auto &newBond : newBonds) {
         newMol->addBond(newBond, true);
+        auto beginAtom = newBond->getBeginAtom();
```
greglandrum, line 476: I like this addition, but I think it's worth putting this whole block inside a check to see if the molecule actually has any conformers.

greglandrum:
unfortunately now that I've started (but not finished) a review, github won't let me make single comments on the code anymore.

I don't like the changes in output that show up in test_rgroups.py at line 193.
Previous output:
```
            'Core': ['C(N(O[*:6])[*:5])[*:1]'],
            'R1': ['Cl[*:1]'],
            'R5': ['N[*:5]'],
            'R6': ['O[*:6]']
```
new output:
```
           'Core': ['C(N(O[*:6])[*:5])[*:1]'],
            'R1': ['Cl[*:1].[H][*:1].[H][*:1]'],
            'R5': ['N[*:5]'],
            'R6': ['O[*:6]']
```
It's not necessarily a blocker, but I think this change makes the results significantly uglier and less useful without post-processing (to get rid of those extraneous Hs in R1)

Given that almost anyone who's going to use the RGD code is going to need to do that post-processing before using the results, it seems like it would be nice to handle it on our side.

ptosco:
I have a couple of problems with the new results reported by `testRingMatching3`.
When I run it with the default parameters results (I know the test was switched to the new scoring method, but I think most users would run a simple case like this with default parameters and expect it to work) are quite disappointing:
image

Also with the prescribed `FingerprintVariance` scoring I find the output confusing to a chemists' eye:
image

I personally found the previous way to present results with R-groups much more intuitive:
image

Am I missing the point?

jones-gareth:
> unfortunately now that I've started (but not finished) a review, github won't let me make single comments on the code anymore.
> 
> I don't like the changes in output that show up in test_rgroups.py at line 193.
> Previous output:
> 
> ```
>             'Core': ['C(N(O[*:6])[*:5])[*:1]'],
>             'R1': ['Cl[*:1]'],
>             'R5': ['N[*:5]'],
>             'R6': ['O[*:6]']
> ```
> 
> new output:
> 
> ```
>            'Core': ['C(N(O[*:6])[*:5])[*:1]'],
>             'R1': ['Cl[*:1].[H][*:1].[H][*:1]'],
>             'R5': ['N[*:5]'],
>             'R6': ['O[*:6]']
> ```
> 
> It's not necessarily a blocker, but I think this change makes the results significantly uglier and less useful without post-processing (to get rid of those extraneous Hs in R1)
> 
> Given that almost anyone who's going to use the RGD code is going to need to do that post-processing before using the results, it seems like it would be nice to handle it on our side.

I agree that this is uglier.  It is is due to the fix made to decompose the cubane structure.  Labelled rgoups now behave in the same way as indexed/non-labelled rgroups.  Whenever you do a decomp with onlyMatchAtRGroups = False, you see these hydrogens filling valance on unlabeled atoms and now you're seeing it on the labeled groups.  I think it would be pretty trivial to remove the extra protons, and can do it in this CR if everyone agrees and is comfortable with changing the current behavior for unlabeled rgroups when onlyMatchAtRgroups = False.

One other significant change from this handling of labelled rgroups is that a single rgroup will now match multiple connections to the neighboring atom:

Capture

Previously (with onlyMatchAtRGroups = True) the core would not match the molecule.  Personally, I prefer the new behavior.

jones-gareth:
> I have a couple of problems with the new results reported by `testRingMatching3`.
> When I run it with the default parameters (I know the test was switched to the new scoring method, but I think most users would run a simple case like this with default parameters and expect it to work), results are quite disappointing:

.....

@ptosco you bring up a couple of issues with the testRingMatching3, to which I add a third.

1. The default parameters no longer work.
2. Sometimes the rgroup is on ring, sometimes it is off ring.
3. (mine) The cores do not always Kekulize

Recall that this test system is unusual- in the master branch the original core `*1***[*:1]1`-doesn't match each molecule (because of a bug) and a new core with all atoms typed is created for each input structure. The bug is now fixed and the core that matches each molecule now comprises query/dummy atoms.

1. The default parameters don't give a good decomposition.  I've noted this in the code- the default scoring function prefers a poor decomposition:
```
R-4 mol 1 [3*][H] mol 2 [3*][H] mol 3 [3*][H]
R-3 mol 2 [2*][H]
R-2 mol 1 [1*][H] mol 2 [1*][H] mol 3 [1*][H]
R-1 mol 1 *[H] mol 2 *[H] mol 3 *[H]
R1 mol 1 [4*]C([H])([H])Cl mol 2 [4*]C([H])([H])I mol 3 [4*]C([H])([H])F
Score 3.26235
R-4 mol 1 [3*][H] mol 2 [3*][H] mol 3 [3*][H]
R-3 mol 1 [2*]C([H])([H])Cl mol 2 [2*][H]
R-2 mol 1 [1*][H] mol 2 [1*][H] mol 3 [1*][H]
R-1 mol 1 *[H] mol 2 *[H] mol 3 *[H]
R1 mol 2 [4*]C([H])([H])I mol 3 [4*]C([H])([H])F
Score 3.27778
```
The first decomp has all halogens on R1, but the second flawed decomp has a higher score.  I don't really understand the default scoring code and would prefer to raise this as an issue and have somebody who has worked on that code take a look at it.

2. When are rgoups on or off rings?  This branch and master will add a dummy atom to a typed (atomic number > 0) atom that has an rgroup label and transfer the label to the rgroup.   If the user rgroup label is on a query atom it stays there.  In this case the R1 label is on a query atom and thus stays in the ring.  In the current master branch the original core does not match, the new core has all atoms typed and the rgroup is put off ring.  So, in master, the groups being off ring is an artifact from the matching bug.

I'm in agreement with you that it is better if the rgroup is off ring, but I didn't do this so to be consistent with the current logic.  Personally, I'd like to see the rgroups off ring and I can add that to this CR, if you're in agreement.

3.  Finally, I can't seem to get the cores in this system to Kekulize.  @greglandrum - I was hoping you can help here.  I think the code in `RCore::coreWithMatches` is responsible.  It takes the original query atoms and bonds and sets matching types.  That seems to work fine, but the resulting structure won't Kekulize.

Code/GraphMol/RGroupDecomposition/RGroupCore.h
```diff
@@ -36,19 +39,84 @@ struct RCore {
...
+      auto coreAtom = finalCore->getAtomWithIdx(atomIdx);
+      auto templateAtom = coreReplacedDummies.getAtomWithIdx(atomIdx);
+      auto unlabelledCoreAtom = core->getAtomWithIdx(atomIdx);
+      if (coreAtom->getAtomicNum() == 0 && templateAtom->getAtomicNum() > 0 &&
+          isAnyAtomWithMultipleNeighborsOrNotUserRLabel(*unlabelledCoreAtom)) {
+        replaceDummyAtom(*finalCore, *coreAtom, *templateAtom);
+        finalCore->getAtomWithIdx(atomIdx)->setNoImplicit(true);
+      }
+    }
+
```
jones-gareth, line 106: @greglandrum - Setting the bonds here works, but the resulting structure wont Kekulize.

greglandrum:
> 
> I agree that this is uglier. It is is due to the fix made to decompose the cubane structure. Labeled rgoups now behave in the same way as indexed/non-labeled rgroups. Whenever you do a decomp with onlyMatchAtRGroups = False, you see these hydrogens filling valance on unlabeled atoms and now you're seeing it on the labeled groups. I think it would be pretty trivial to remove the extra protons, and can do it in this CR if everyone agrees and is comfortable with changing the current behavior for unlabeled rgroups.

That's fine with me.

> One other significant change from this handling of labelled rgroups is that a single labeled rgroup will now match multiple connections to the neighboring atom:
> 
> Capture
> 
> Previously (with onlyMatchAtRGroups = True) the core would not match the molecule. Personally, I prefer the new behavior and it is consistent with how unlabeled rgroups behave.

I believe that's also ok.

What would these changes mean for chiral centers like in this example?
```
In [13]: ms = [Chem.MolFromSmiles(x) for x in ['C1CO[C@H]1C','C1CO[C@]1(F)C','C1CO[C@@H]1C']]

In [14]: core = Chem.MolFromSmiles('C1COC1[*:1]')

In [15]: rdRGroupDecomposition.RGroupDecompose([core],ms,asSmiles=True)
Out[15]: 
([{'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'},
  {'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': 'F[*:2]'},
  {'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'}],
 [])

In [17]: chiral_core = Chem.MolFromSmiles('C1CO[C@H]1[*:1]')

In [18]: rdRGroupDecomposition.RGroupDecompose([chiral_core],ms,asSmiles=True)
Out[18]: 
([{'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'},
  {'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'F[*:1]', 'R2': 'C[*:2]'},
  {'Core': 'C1CC([*:1])([*:2])O1', 'R1': '[H][*:1]', 'R2': 'C[*:2]'}],
 [])
```
Am I going to need to provide two explicit R-groups in order to get the stereo resolved properly?

jones-gareth:
> What would these changes mean for chiral centers like in this example?
> 
> ```
> In [13]: ms = [Chem.MolFromSmiles(x) for x in ['C1CO[C@H]1C','C1CO[C@]1(F)C','C1CO[C@@H]1C']]
> 
> In [14]: core = Chem.MolFromSmiles('C1COC1[*:1]')
> 
> In [15]: rdRGroupDecomposition.RGroupDecompose([core],ms,asSmiles=True)
> Out[15]: 
> ([{'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'},
>   {'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': 'F[*:2]'},
>   {'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'}],
>  [])
> 
> In [17]: chiral_core = Chem.MolFromSmiles('C1CO[C@H]1[*:1]')
> 
> In [18]: rdRGroupDecomposition.RGroupDecompose([chiral_core],ms,asSmiles=True)
> Out[18]: 
> ([{'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'},
>   {'Core': 'C1CC([*:1])([*:2])O1', 'R1': 'F[*:1]', 'R2': 'C[*:2]'},
>   {'Core': 'C1CC([*:1])([*:2])O1', 'R1': '[H][*:1]', 'R2': 'C[*:2]'}],
>  [])
> ```
> 
> Am I going to need to provide two explicit R-groups in order to get the stereo resolved properly?

Yes, looks like you will need to provide 2 r groups.  Not sure how to resolve that- we could not move the rgroup to the neighbor if the neighbor is a chiral center, but then the behavior would be inconsistent with non-chiral centers and unlabeled rgroups on chiral centers..

```
In [6]: chiral_core = Chem.MolFromSmiles('C1CO[C@H]1[*:1]')

In [7]:  rdRGroupDecomposition.RGroupDecompose([chiral_core],ms,asSmiles=True)
Out[7]:
([{'Core': 'C1C[C@@H]([*:1])O1', 'R1': 'C[*:1].[H][*:1]'},
  {'Core': 'C1C[C@@H]([*:1])O1', 'R1': 'C[*:1].F[*:1]'},
  {'Core': 'C1C[C@@H]([*:1])O1', 'R1': 'C[*:1].[H][*:1]'}],
 [])

In [8]: chiral_core = Chem.MolFromSmiles('C1CO[C@]1([*:1])([*:2])')

In [9]:  rdRGroupDecomposition.RGroupDecompose([chiral_core],ms,asSmiles=True)
Out[9]:
([{'Core': 'C1C[C@@]([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'},
  {'Core': 'C1C[C@@]([*:1])([*:2])O1', 'R1': 'F[*:1]', 'R2': 'C[*:2]'},
  {'Core': 'C1C[C@@]([*:1])([*:2])O1', 'R1': '[H][*:1]', 'R2': 'C[*:2]'}],
 [])
```


```

**Comment by greglandrum** [comment]
> Yes, looks like you will need to provide 2 r groups. Not sure how to resolve that- we could not move the rgroup to the neighbor if the neighbor is a chiral center, but then the behavior would be inconsistent with non-chiral centers and unlabeled rgroups on chiral centers..
> 
> ```
> In [6]: chiral_core = Chem.MolFromSmiles('C1CO[C@H]1[*:1]')
> 
> In [7]:  rdRGroupDecomposition.RGroupDecompose([chiral_core],ms,asSmiles=True)
> Out[7]:
> ([{'Core': 'C1C[C@@H]([*:1])O1', 'R1': 'C[*:1].[H][*:1]'},
>   {'Core': 'C1C[C@@H]([*:1])O1', 'R1': 'C[*:1].F[*:1]'},
>   {'Core': 'C1C[C@@H]([*:1])O1', 'R1': 'C[*:1].[H][*:1]'}],
>  [])
> 
> In [8]: chiral_core = Chem.MolFromSmiles('C1CO[C@]1([*:1])([*:2])')
> 
> In [9]:  rdRGroupDecomposition.RGroupDecompose([chiral_core],ms,asSmiles=True)
> Out[9]:
> ([{'Core': 'C1C[C@@]([*:1])([*:2])O1', 'R1': 'C[*:1]', 'R2': '[H][*:2]'},
>   {'Core': 'C1C[C@@]([*:1])([*:2])O1', 'R1': 'F[*:1]', 'R2': 'C[*:2]'},
>   {'Core': 'C1C[C@@]([*:1])([*:2])O1', 'R1': '[H][*:1]', 'R2': 'C[*:2]'}],
>  [])
> ```

I think this is fine. It's different from what was happening before, but the whole thing remains internally consistent.

**Comment by bp-kelley** [comment]
I think some of these changes make it hard to reason with unless there is a specification for recombination.  Without this, we can't do simple things like free-wilson.

molzip should be able to reconstruct any decomposition and these rules won't currently work.  I'd be happier if we codified the rules made the recombination code and enforced use.  I'm happy if this is two steps/PRs.

I admit to being uncomfortable about this:



as we now have no way for the user to easily specify that -R means.  To most chemists, it means exactly one heavy substituent, otherwise you can't decompose primary/secondary/tertiary amines.

**Comment by ptosco** [comment]
I am with @bp-kelley. I have a feeling that this PR introduces too many changes at once to be able to figure out all their implications.
I am a also uncomfortable with some of the more disruptive changes.
I have extracted some bits which are certainly an improvement and not subject to debate into #3877, that I am just about to update and finalize with a couple more tests and a Python wrapper.
I would propose to extract the "cubane fix" as well into a separate PR if it can be done independently from the other behavior changes.

**Comment by jones-gareth** [comment]
> I think some of these changes make it hard to reason with unless there is a specification for recombination. Without this, we can't do simple things like free-wilson.
> 
> molzip should be able to reconstruct any decomposition and these rules won't currently work. I'd be happier if we codified the rules made the recombination code and enforced use. I'm happy if this is two steps/PRs.
> 
> I admit to being uncomfortable about this:
> 
> 
> 
> as we now have no way for the user to easily specify what -R means. To most chemists, it means exactly one heavy substituent, otherwise you can't decompose primary/secondary/tertiary amines.

I've been trying to understand the different between the user-defined rgroups and the unlabelled rgroups.  If the user-defined rgroup means exactly one heavy substituent, I can work with that.

If I were to revert  the PR so that each user defined rgroup matches only one heavy substituent would that address your concerns for now?  We can add in the recombination code once that is ready.

Unfortunately, this would mean the cubane fix would no longer work.  I think that will now require matching the core without the rgroups then matching the bonds to the R groups separetely.

**Comment by jones-gareth** [comment]
> I am with @bp-kelley. I have a feeling that this PR introduces too many changes at once to be able to figure out all their implications.
> I am a also uncomfortable with some of the more disruptive changes.
> I have extracted some bits which are certainly an improvement and not subject to debate into #3877, that I am just about to update and finalize with a couple more tests and a Python wrapper.
> I would propose to extract the "cubane fix" as well into a separate PR if it can be done independently from the other behavior changes.

While it makes sense to put the dummy atom coordinates in a separate PR, it won't be possible for the "cubane fix" as that is the source of Brian's issue that a user defined rgroup should match a single heavy substituent.

Could you provide more detail round the disruptive changes that you are uncomfortable with?  I plan to make sure that each user defined rgroup matches a single heavy substituent and think I addressed your issues round testRingMatching3.  

I'm not sure I want to abandon this PR yet!

**Comment by jones-gareth** [comment]
@ptosco @bp-kelley I guess that either or both of you maintain the default rgroup scoring function.  I didn't get any feedback about this behavior:

```
R-4 mol 1 [3*][H] mol 2 [3*][H] mol 3 [3*][H]
R-3 mol 2 [2*][H]
R-2 mol 1 [1*][H] mol 2 [1*][H] mol 3 [1*][H]
R-1 mol 1 *[H] mol 2 *[H] mol 3 *[H]
R1 mol 1 [4*]C([H])([H])Cl mol 2 [4*]C([H])([H])I mol 3 [4*]C([H])([H])F
Score 3.26235
R-4 mol 1 [3*][H] mol 2 [3*][H] mol 3 [3*][H]
R-3 mol 1 [2*]C([H])([H])Cl mol 2 [2*][H]
R-2 mol 1 [1*][H] mol 2 [1*][H] mol 3 [1*][H]
R-1 mol 1 *[H] mol 2 *[H] mol 3 *[H]
R1 mol 2 [4*]C([H])([H])I mol 3 [4*]C([H])([H])F
Score 3.27778
```

Should I just add it as an issue?

**Comment by ptosco** [comment]
> @ptosco @bp-kelley I guess that either or both of you maintain the default rgroup scoring function. I didn't get any feedback about this behavior:

At the moment I only get this behavior with the PR, I don't get it with `master`, so I wouldn't raise a separate issue.
I'll make some time later to look more in-depth into `testRingMatching3` and then get back to you.

**Comment by jones-gareth** [comment]
> > @ptosco @bp-kelley I guess that either or both of you maintain the default rgroup scoring function. I didn't get any feedback about this behavior:
> 
> At the moment I only get this behavior with the PR, I don't get it with `master`, so I wouldn't raise a separate issue.
> I'll make some time later to look more in-depth into `testRingMatching3` and then get back to you.

I'm confident that if you had those groups in master you would get those scores- `testRingMatching3` in master has a different set of rgroups, so you don't see the issue.  I can look at putting together an internals test to illustrate.

**Comment by greglandrum** [comment]
@bp-kelley, @ptosco, @jones-gareth : it feels like this interface is not the right one to be trying to resolve these questions. At the moment I'm not even sure we all agree on what the questions are, but it certainly feels like that we aren't getting closer to converging on answers.

Regardless of what comes out of this, the PR is going to be changing the current behavior of the RGD code. I think that's fine given that the current behavior around many of these edge cases is something that just happened, and not necessarily the best possible solution. So how about we put together a notebook or some other document which contains the situations that we have questions about, capture what we think *should* happen, and then figure out which pieces of this PR get us closer to that. I don't expect that the PR itself will get us all the way there. I will capture whatever we produce and ensure that it ends up in the documentation in one form or another.

Does that work for you?

**Comment by jones-gareth** [comment]
>  So how about we put together a notebook or some other document which contains the situations that we have questions about, capture what we think _should_ happen, and then figure out which pieces of this PR get us closer to that. I don't expect that the PR itself will get us all the way there. I will capture whatever we produce and ensure that it ends up in the documentation in one form or another.
> 
> Does that work for you?

Makes sense.  I have now got this PR into a state that reflects how I would like the RGD code to work. Today I tweaked it so that

- All extraneous hydrogens are removed from R groups
- User labelled R groups match only a single heavy substituent
- Core labels are all off ring

Aside from the cubane test, which is now failing, all the examples in `Docs/Notebooks/RGroupDecomposition-Tests.ipynb` look how I want them to be.  For example, `testMatchOnAnyAtom` (the original focus of the PR):

Capture

`testSDFGRoupMultiCoreNoneShouldMatch`, which is much better without extraneous hydrogens:

Capture

And `testRingMatching3` which, core Kekulization aside, has the decomposition I prefer:

Capture

I've made a lot of changes here to the RGD rules in master but these and the other examples in  `Docs/Notebooks/RGroupDecomposition-Tests.ipynb` show how I'd like RGD to work.  I'm sure the rest of you have your own opinions, and I'm looking forward to building a consensus on RGD behavior.

**Comment by jones-gareth** [comment]
The latest commit has a couple of failing checks, but they don't seem related to the PR.  OSX is failing on and windows is failing on testMultithreadedMolSupplier and on of the Windows builds failed on testMultithreadedMolSupplier.

**Comment by ptosco** [comment]
@jones-gareth Thanks Gareth for your work. I am going to build your PR locally today and review it. @jones-gareth @bp-kelley @greglandrum Once we are all reasonably happy, It would be good to approve and merge this PR quickly so we have something stable to work on and further tweak/fix residual issues before release.

**Comment by ptosco** [comment]
@jones-gareth The results in `RGroupDecomposition-Tests.ipynb` look good to me.
It is clearly a big improvement to replace dummies with real atoms taken from the cores - makes the cores prettier and the R groups more understandable.
I'll work on the kekulization issue - I think I can fix it. I have had similar issues when postprocessing cores and I managed to solve them. I'll make a PR against your branch as soon as I have a fix.

**Comment by greglandrum** [comment]
I've been looking at the output of the current version and it mostly looks good.
I particularly like the way you've cleaned up the handling of wildcard atoms in the core - I understood why it was the way it was, but it always felt weird.

I did find one thing that I'm not particular fond of which I think is likely an oversight/bug. Here's a gist from running the new code on an old RDKit blog post:
https://gist.github.com/greglandrum/0afd4f5cc9194432acd85bf22261ed8a

This is the specific example:
```
mols = [Chem.MolFromSmiles(smi) for smi in 'c1c(F)cccn1 c1c(Cl)c(C)ccn1 c1c(F)cncn1 c1c(F)c(C)ccn1'.split()]
sma_scaffold = Chem.MolFromSmarts('c:1:c(-[*:1]):c(-[*:2]):*:c:n:1')
groups,unmatched = rdRGroupDecomposition.RGroupDecompose([sma_scaffold],mols,asSmiles=False,asRows=False)
PandasTools.RGroupDecompositionToFrame(groups,mols,include_core=True)
```
and here's the output:
image

I would say that R3 should not be there in this case: the only instances contain either an H or nothing

I could imagine that this isn't directly connected to this PR, so we can file a separate issue and get to it post-merge if @jones-gareth thinks that makes sense.

**Comment by bp-kelley** [comment]
@greglandrum we initially separated nothing from something for did this for sulfurs which can be S or SH.  We can perhaps special case this, but I had thought it was important enough.

@jones-gareth This is looking pretty good.   I'm also morbidly pleased that answering these questions aren't easy.

Two proposal:  

1) Perhaps we could annotate the core atom that are different in case we want to highlight them in the core drawings.

2) I still don't like




Perhaps the rendering/smiles could be

image


This would be easier to implement in molzip.  No idea how this affects hydrogens though.

**Comment by bp-kelley** [comment]
@greglandrum I also thought previous the core atom was weird, and actually have in my notes "extract real core so it can be sorted".   So I think this is a big win.

I'm going to try some edge cases with this PR.

**Comment by greglandrum** [comment]
> @greglandrum we initially separated nothing from something for sulfurs which can be S or SH. We can perhaps special case this, but I had thought it was important enough.

@bp-kelley do you have an example where this happens ?

> 
> @jones-gareth This is looking pretty good. I'm also morbidly pleased that answering these questions aren't easy.
> 
> 2. I still don't like
> 
> 
> 
> Perhaps the rendering/smiles could be
> 
> image
> 
> This would be easier to implement in molzip. No idea how this affects hydrogens though.

I can see how that's easier for molzip, but I would not want to explain to a chemist why the stupid software is producing that output.

-greg

**Comment by bp-kelley** [comment]
@greglandrum fair enough, thought I see them as two sides of the same coin, conventions matter.

@jones-gareth the labelled versus unlabelled is simple.  Labels come from the user and need to be honored.  Unlabelled are matches that come along for the ride but weren't specified.  If a user only wants to see substitutions at user-labelled groups this is where it is important, otherwise it's just book keeping.  We could say that core atoms can match multiple substituents and that would at least define the problem space.

So the previous definition for a labelled RGroup was:

1)

image

matchOnlyAtRGroups=true would match only one substituent (otherwise the matching code gets crazy unless we perform some tricks as in this PR)

2)

image

with matchOnlyAtRGroups=true would match multiple substituents.  This is also the difference between labelling a core atom and using a proper RGroup.  It was basically saying, "I don't care about exact # RGroups just stuff attached to this atom."

From what I'm understanding, we are changing the behavior of (1) to be more like (2) and this is the sticking point with myself and @ptosco.

I'm willing to go forwards with this definition, but I believe we are redefining the definition of r-group to be one or more substituents.  So what happens with:

image

I'm going to find out :)

My only other comment was that this is now appears to be a different definition from RGroups used in reactions schemas.

**Comment by jones-gareth** [comment]
There are still (at least!) a couple of outstanding issues in this PR that I need to fix:

`testGeminalRGroups` requires that onlyMatchAtRGroups = true, otherwise it ignores the user-speciifed r-groups.  The same thing  (or something similar) is happening here when onlyMatchAtRGroups = false:

Capture

There is also the issue in the  default scoring function for `testRingMatching3`.

I'll check these issues out in the next few days, but be aware of them if you're planning to hammer the code.

@bp-kelley The test system about no longer matches if onlyMatchAtRGroups = true, as we agreed that a user defined R group would match only one substituent containing heavy atoms.

**Comment by bp-kelley** [comment]
I can look at the default scoring, it doesn't make sense to me as matching Hs should be preferred.  There is no logic to say Halogens should be hanlded the same so we rely on the surrounding matches.

**Comment by jones-gareth** [comment]
> @greglandrum fair enough, thought I see them as two sides of the same coin, conventions matter.
> 
> @jones-gareth the labelled versus unlabelled is simple. Labels come from the user and need to be honored. Unlabelled are matches that come along for the ride but weren't specified. If a user only wants to see substitutions at user-labelled groups this is where it is important, otherwise it's just book keeping. We could say that core atoms can match multiple substituents and that would at least define the problem space.
> 
> So the previous definition for a labelled RGroup was:
> 
> 
> image
> 
> matchOnlyAtRGroups=true would match only one substituent (otherwise the matching code gets crazy unless we perform some tricks as in this PR)
> 
> 
> image
> 
> with matchOnlyAtRGroups=true would match multiple substituents. This is also the difference between labelling a core atom and using a proper RGroup. It was basically saying, "I don't care about exact # RGroups just stuff attached to this atom."
> 
> From what I'm understanding, we are changing the behavior of (1) to be more like (2) and this is the sticking point with myself and @ptosco.
> 
> I'm willing to go forwards with this definition, but I believe we are redefining the definition of r-group to be one or more substituents. So what happens with:
> 
> image
> 
> I'm going to find out :)
> 
> My only other comment was that this is now appears to be a different definition from RGroups used in reactions schemas.

@bp-kelley I am on board with your R group definition!  I guess I wasn't clear enough, but I changed the PR based on the comments that you and @ptosco made about user labelled R-groups.  The code here is almost there. So, with onlyMatchOnRGroups = true this works:

image

But this as the core does not:

Capture (Custom) (1)

So those 2 cases work as expected.

unfortunately this does not:

Capture (Custom)

So I have got the first and third R group case correct, but the second is incorrect.  Give me a little bit or time and I'll get that fixed.

**Comment by jones-gareth** [comment]
> I can look at the default scoring, it doesn't make sense to me as matching Hs should be preferred. There is no logic to say Halogens should be handled the same so we rely on the surrounding matches.

Maybe  just let me have a quick look at it first- my focus yesterday was on getting a notebook together that reflected the decompositions that I wanted to see.  I think in the process I've introduced a couple of bugs that (hopefully) will be easy enough to fix.

**Comment by jones-gareth** [comment]
> @jones-gareth The results in `RGroupDecomposition-Tests.ipynb` look good to me.
> It is clearly a big improvement to replace dummies with real atoms taken from the cores - makes the cores prettier and the R groups more understandable.
> I'll work on the kekulization issue - I think I can fix it. I have had similar issues when postprocessing cores and I managed to solve them. I'll make a PR against your branch as soon as I have a fix.

It would be cool if you can fix that- I think I can see why it is not working in most of the cases (incorrect valance), but I don't know how to fix it.

**Comment by bp-kelley** [comment]
I would kekulize the original molecule and extract the core from that.

**Comment by bp-kelley** [comment]
By original molecule, I mean the one the core matched.

**Comment by jones-gareth** [comment]
The latest commit has the user labelling working as expected and `testGeminalRGroups` works matching on any rgroup.  Unfortunately, this is only true when using the fingerprint variance scoring function (more on this below).

I've added a new test `testUserMatchTypes`, to illustrate the different possibilities that @bp-kelley outlined:

Capture (Custom) (1)
Capture (Custom) (2)
Capture (Custom) (3)

To get this to work I had to change the tie resolution software.  The problem was that this decomp was preferred to the correct one:

Capture (Custom)

Both this and the correct decomp have one unlabeled R group.  The tie braking code selects on the smallest count of unlabeled r groups, but does not maximize the number of labelled R groups.  I've changed it so it now first filters on the largest number of labelled groups.

A similar thing was happening in the scoring functions.  I could see why that might happen with the fingerprint variance scoring function.  It just selected compound matches such that the R group structures are similar- it knew nothing about if the R group were labeled.  I have added a correction to the fingerprint score that uses the frequency of unmatched labeled R groups, to discourage the selection of permutations that don't match user labels.  The default scoring function probably needs something similar as it (seems to me it) can exhibit the same behavior.  When adding compounds, it may make sense just to throw out core matches to unlabeled R groups if we get core matches to labeled R groups.

I got rid of the fingerprint distance metric as I can't see any good reason to use it instead of the variance score and I don't want to maintain it (it's not in any release branch).

When I get a chance I'll try and put together some test examples for when the default scoring function fails.

**Comment by jones-gareth** [comment]
> @jones-gareth This is looking pretty good. I'm also morbidly pleased that answering these questions aren't easy.
> 
> Two proposal:
> 
> 1. Perhaps we could annotate the core atom that are different in case we want to highlight them in the core drawings.

It there a particular property you would use for that?

**Comment by bp-kelley** [comment]
@greglandrum @jones-gareth

This is looking really good.  The "real core" extraction is exactly the right thing to do.  Interestingly, this allows me to use query and multiple scaffolds in my free Wilson implementation which is awesome.  I'm running some tests tonight and tomorrow with a bunch of in-house and chembl decompositions.

Let's use the "match types" in the docs!

For the atom property, we could try setting an `common_properties::atomNote` string, this will g
et rendered in the depiction.  I'm not sure what the note would be though...  perhaps just "*" or something like that.  What do you think greg?

**Comment by bp-kelley** [comment]
There are some debugging logs that need to be turned off:

```
[21:21:50] Exhaustive or GreedyChunks process, best score 4.0155 permutation size 10
```

**Comment by greglandrum** [comment]
> 
> Let's use the "match types" in the docs!
> 
> For the atom property, we could try setting an `common_properties::atomNote` string, this will get rendered in the depiction. I'm not sure what the note would be though... perhaps just "*" or something like that. What do you think greg?

I'm not thrilled about setting the atomNote property from within core RDKit code. It's not clear to me in the first place why it's generally useful to distinguish between user-provided R-groups and those which are automatically assigned. The user knows which labels they drew and can easily tell this on their own. If there's an application which needs something like this, it's straightforward enough to implement the "which r groups were in the scaffold I provided and what did I get back from the RGD code" logic.

**Comment by greglandrum** [comment]
I think the results from this look good and we should merge it RSN and file bug reports for things which are outstanding.
I know @ptosco did a PR against @jones-gareth's fork with some kekulization-related fixes, 

@jones-gareth, @bp-kelley, @ptosco : is there anything left to do here which you think is essential and not further enhancement or fixing non-critical bugs?

I still would like to see the issue where R groups which contain only Hs or nothing are not removed:
image
but I'm happy to file a bug on that one.

**Comment by bp-kelley** [comment]
@greglandrum I was referring to identifying scaffold changes from query atoms.  We used to mark them with [N:1] but don't anymore.  However, we do supply the cores so I can be convinced to wait on this.

**Comment by ptosco** [comment]
> I still would like to see a fix for the issue where R groups which contain only Hs or nothing are not removed, but I'm happy to file a bug on that one.

This is now filed here:
https://github.com/rdkit/rdkit/issues/3895

**Comment by greglandrum** [comment]
One of the behaviors of the current code which I find to be a complete mystery is the handling of R groups which are degree two (or higher) in the query. Here's a super simple example which demonstrates the point:
```
cores = [Chem.MolFromSmiles("c1ccc[1*]1")]
smi = ["c1cocc1", "c1c[nH]cc1", "c1cN(CC)cc1"]
mols = [Chem.MolFromSmiles(s) for s in smi]
groups,unmatched = rdRGroupDecomposition.RGroupDecompose(cores,mols,asSmiles=False,asRows=False)
PandasTools.RGroupDecompositionToFrame(groups,mols,include_core=True)
```
The output is:
image

I understand that this is intentional, but it is inconsistent with what the rest of the RGD code does: replace the specified R group with a specific atom and then indicate an attachment point to the rest of the core. I find it confusing and surprising (the kind of surprising which leads to people asking hard-to-answer questions).

Two requests:
1. I don't think that having labelled R groups with degree higher than one is going to be a common thing, any objection to us adding an option 'allowNonTerminalRGroups`. If this is set to False then a core with a labelled R group which has degree >1 will generate an error; otherwise it does what currently happens here. The default would be False and it could be enabled to allow the new behavior for people who explicitly want that.
2. Could someone please document this behavior, with an example and text explaining why it's doing what it's doing so that I can add it to the documentation somewhere (not that most people will read that, but at least we'll have something to point to when questions come up).
 
For fun, here's an instance which is even worse, in this case the core isn't updated to reflect the positive charge on the nitrogen. I will file a post-merge bug for this:
```
cores = [Chem.MolFromSmiles("C1C[1*]2CCC1OC2")]
smi = ["CC12CCC(CC1)OC2", "C[N+]12CCC(CC1)OC2",]
mols = [Chem.MolFromSmiles(s) for s in smi]
groups,unmatched = rdRGroupDecomposition.RGroupDecompose(cores,mols,asSmiles=False,asRows=False)
PandasTools.RGroupDecompositionToFrame(groups,mols,include_core=True)
```
image

**Comment by ptosco** [comment]
I agree the handling of labelled R-groups with degree>1 should be done differently and that the current behavior is very surprising and confusing. But given this is not something that is likely to occur often, I'd handle this after this PR is merged. I am in favor of adding the flag that Greg suggest if people agree this is the cheapest and most expedite solution to merge this PR quickly.
@greglandrum Regarding the formal charge bug that you raise, that is already addressed by the PR I made against Gareth's branch, so it is not a problem:
image

**Comment by greglandrum** [comment]
NOTE: I just updated my comment above to include another example with a simple dummy atom in the core.

**Comment by jones-gareth** [comment]
> I think the results from this look good and we should merge it RSN and file bug reports for things which are outstanding.
> I know @ptosco did a PR against @jones-gareth's fork with some kekulization-related fixes,
> 
> @jones-gareth, @bp-kelley, @ptosco : is there anything left to do here which you think is essential and not further enhancement or fixing non-critical bugs?
> 
> I still would like to see a fix for the issue where R groups which contain only Hs or nothing are not removed, but I'm happy to file a bug on that one.

I think the fix for removing R groups that contain only Hs or nothing is pretty trivial, but it will require refactoring of the tests.  Personally, I like to draw a line under this one.

I do need to check and merge @ptosco 's PR.  I can do that later today.

**Comment by ptosco** [comment]
@jones-gareth I already have that fix and the modified test results in a branch. I can submit a PR once this is merged.

**Comment by jones-gareth** [comment]
> 1. I don't think that having labelled R groups with degree higher than one is going to be a common thing, any objection to us adding an option 'allowNonTerminalRGroups`. If this is set to False then a core with a labelled R group which has degree >1 will generate an error; otherwise it does what currently happens here. The default would be False and it could be enabled to allow the new behavior for people who explicitly want that.
> 2. Could someone please document this behavior, with an example and text explaining why it's doing what it's doing so that I can add it to the documentation somewhere (not that most people will read that, but at least we'll have something to point to when questions come up).
> 

I don't find the behavior too odd.  I think there is an ambiguity between the `*` as an attachment point or as any atom in the ring.  In your example it does appear that the user might want an attachment point in the  ring, so the current behavior is confusing.  However, if you have a core like `*1****[*:1]1`, I think the way the code behaves is good- matching the entire ring and putting a separate attachment point on the last atom.  

In the current code when `*` is degree one it is an explicit attachment point, when it is degree 2 it behaves like any other atom (in terms of R group attachment).

**Comment by jones-gareth** [comment]
> @jones-gareth I already have that fix and the modified test results in a branch. I can submit a PR once this is merged.

Cool, I misunderstood.

**Comment by bp-kelley** [comment]
@jones-gareth @greglandrum @ptosco I do think a lot of the user-confusion is the distinction I made about specified and non specified RGroups.  Part of my vision was to be able to "circle" a scaffold in a sketcher and without any extra effort get decent RGroups out.

Perhaps this was misguided, but I have found it really useful.

**Comment by bp-kelley** [comment]
I had a conversation with Greg and he convinced me of his assessment.  I'm probably late to the understanding, but the conclusion was ring rgroups and multi-degree rgroups are invalid unless specifically allows.  So queries like

(1)

image

will be rejected as matches unless explicitly allowed.  

(2) The following will be accepted:

image

I believe that the results of (1) and (2) should be identical I think has been the focus of this PR.

@greglandrum Is this a good summary?  Did I miss anything?

**Comment by greglandrum** [comment]
@bp-kelley Yeah, I think that's more or less what I was saying.
To restate: I see the use case for allowing user-specified R groups with more than one bond to them, and I agree that the way this PR handles those situation is better than what currently happens on master, but I think the whole situation is confusing enough that we should by default treat those situations as an error. We can add the option `RGroupDecompositionParameters::allowNonTerminalRGroups` which allows these to be used, but that should default to False. 

Two examples Brian and I talked through:
image

**Comment by jones-gareth** [comment]
> I had a conversation with Greg and he convinced me of his assessment. I'm probably late to the understanding, but the conclusion was ring rgroups and multi-degree rgroups are invalid unless specifically allowed. So queries like
> 
> (1)
> 
> image
> 
> will be rejected as matches unless explicitly allowed.
> 
> (2) The following will be accepted:
> 
> image
> 

That all makes sense.  I can tweak the PR to include the `RGroupDecompositionParameters::allowNonTerminalRGroups` option to allow R groups of more than one degree.


> I believe that the results of (1) and (2) should be identical I think has been the focus of this PR.
> 

This is not entirely correct in (1) R1 can match multiple sidechains with heavy atom, while in (2) R1 can only match a single sidechain with heavy atoms.

I have one question:

Without the option this is not allowed- `*1****[*:1]1`. I assume that the heavy atom equivalent is OK - `c1cccc[c:1]1`?

It's really good to get a full understanding of all the R group definitions.

**Comment by greglandrum** [comment]
> 
> Without the option this is not allowed- `*1****[*:1]1`. I assume that the heavy atom equivalent is OK - `c1cccc[c:1]1`?
> 

I'm not clear what that would do. Is that intended to be equivalent to `c1ccccc1[*:1]`?

> It's really good to get a full understanding of all the R group definitions.

oh yes. and then to document the hell out of all of it. :-)

**Comment by jones-gareth** [comment]
> . We can add the option `RGroupDecompositionParameters::allowNonTerminalRGroups` which allows these to be used, but that should default to False.
> 

I can add that option today.  What is the best what of handling that error (non terminal rgroup without the option)? I was thinking of raising ValueErrorException since this would be an example of malformed input.

**Comment by greglandrum** [comment]
> > . We can add the option `RGroupDecompositionParameters::allowNonTerminalRGroups` which allows these to be used, but that should default to False.
> 
> I can add that option today. What is the best what of handling that error (non terminal rgroup without the option)? I was thinking of raising ValueErrorException since this would be an example of malformed input.

Yeah, I think `ValueErrorException` with an informative message (perhaps to rdDebugLog) explaining how to fix it?

**Comment by jones-gareth** [comment]
> > I assume that the heavy atom equivalent is OK - `c1cccc[c:1]1`?
> 
> I'm not clear what that would do. Is that intended to be equivalent to `c1ccccc1[*:1]`?
> 

Not quite- this is the reasoning from Brian- 

> So the previous definition for a labelled RGroup was:
> 
> 
> image
> 
> would match only one substituent
> 
> 
> image
> 
> would match multiple substituents. This is also the difference between labelling a core atom and using a proper RGroup. It was basically saying, "I don't care about exact # RGroups just stuff attached to this atom."
> 

Now I think about it, if you want to use this feature I think you should have to set `RGroupDecompositionParameters::allowNonTerminalRGroups`, because it's pretty much  the same thing as a query atom of degree 2 with a R group label.

**Comment by greglandrum** [comment]
> > > I assume that the heavy atom equivalent is OK - `c1cccc[c:1]1`?
> > 
> > 
> > I'm not clear what that would do. Is that intended to be equivalent to `c1ccccc1[*:1]`?
> 
> Not quite- this is the reasoning from Brian-
> 
> > for a labelled RGroup was:
> > image
> > would match only one substituent
> > image
> > would match multiple substituents. This is also the difference between labelling a core atom and using a proper RGroup. It was basically saying, "I don't care about exact # RGroups just stuff attached to this atom."
> 
> Now I think about it, if you want to use this feature I think you should have to set `RGroupDecompositionParameters::allowNonTerminalRGroups`, because it's pretty much the same thing as a query atom of degree 2 with a R group label.

I don't follow that at all and think the only thing that should count as an R group is a dummy atom with either an isotope label or an atom map number.

**Comment by jones-gareth** [comment]
> I don't follow that at all and think the only thing that should count as an R group is a dummy atom with either an isotope label or an atom map number.

Yes, I feel it is going to be hard to explain to users that you can have rgroup labels on typed atoms.  In these cases the labelled atom behaves in the same way as an unlabeled R group, so if anyone needs this behavior perhaps they should be using `onlyMatchAtRGroups = False` instead.  The only place this occurs in the test suite is in `testUserMatchTypes` which I just wrote to validate Brian's definitions.  It would be simple to disallow this.

This functionality has been in RGD for a while, so one issue is that people may be using it.

**Comment by jones-gareth** [comment]
This latest commit has these fixes:

- Support for the `allowNonTerminalRGroups` parameter
- R groups that contain H or nothing are removed (assuming `removeAllHydrogenRGroups` is set)
- R group labels on atoms that are not dummy atoms are ignored (this is trivial to revert if we want to support)

**Comment by greglandrum** [comment]
@jones-gareth I'm going to look through results again carefully (hopefully) this morning.
Here's one thing I noticed which is new and, I think, wrong:
```
In [15]: cores = [Chem.MolFromSmiles("C1CCC1[*:1]")]
    ...: smi = ["C1CCC1(N)(O)"]
    ...: mols = [Chem.MolFromSmiles(s) for s in smi]
    ...: groups,unmatched = rdRGroupDecomposition.RGroupDecompose(cores,mols,asSmiles=True,asRows=True)
[05:05:40] Preparing core 0
[05:05:40] Exhaustive or GreedyChunks process, best score 6 permutation size 1

In [16]: groups
Out[16]: [{'Core': 'C1CC([*:2])C1[*:1]', 'R2': 'N[*:2].O[*:2]'}]
```
I believe the output should be:
```
[{'Core': 'C1CCC1[*:1]', 'R1': 'N[*:1].O[*:1]'}]
```
2020.09.5 gets this one right.

FWIW: this looks like it could perhaps be related to #3896?

jones-gareth:
> @jones-gareth I'm going to look through results again carefully (hopefully) this morning.
> Here's one thing I noticed which is introduced by this PR and, I think, wrong:
> 
> ```
> In [16]: groups
> Out[16]: [{'Core': 'C1CC([*:2])C1[*:1]', 'R2': 'N[*:2].O[*:2]'}]
> ```

I'm seeing this decomp on Linux:

Capture (Custom)

Which I believe is correct.

> 
> I believe the output should be:
> 
> ```
> [{'Core': 'C1CCC1[*:1]', 'R1': 'N[*:1].O[*:1]'}]
> ```
> 

I don't think we want this decomposition as it has two heavy substituents connected to 1 labeled R group which we decided was a bad idea (in this PR about 5 days ago).

However, I been getting decompositions like one you show when using the default scoring function (although I'm not seeing it  in this case).  I think I know why as I also saw the same thing using fingerprint variance scoring, which I fixed.  Does using `options.scoreMethod = RGroupScore.FingerprintVariance` fix your decomp?  I've had to do that for some of the tests.

I plan to put together some test data for default scoring function failures that I'm seeing.

greglandrum:
> > @jones-gareth I'm going to look through results again carefully (hopefully) this morning.
> > Here's one thing I noticed which is introduced by this PR and, I think, wrong:
> > ```
> > In [16]: groups
> > Out[16]: [{'Core': 'C1CC([*:2])C1[*:1]', 'R2': 'N[*:2].O[*:2]'}]
> > ```
> 
> I'm seeing this decomp on Linux:
> 
> Capture (Custom)
> 
> Which I believe is correct.
> 
> > I believe the output should be:
> > ```
> > [{'Core': 'C1CCC1[*:1]', 'R1': 'N[*:1].O[*:1]'}]
> > ```
> 
> I don't think we want this decomposition as it has two heavy substituents connected to 1 labeled R group which we decided was a bad idea (in this PR about 5 days ago).

:-)
If we can add the "extra" R group, I'm that much happier.
Your solution, which I also get when I set the scoring to FingerprintVariance, looks great.

Here's a snippet using the current version of your branch:
image

The results look good with FingerprintVariance, and wrong (not just sub-optimal) when using the default scoring method.

jones-gareth:
@ptosco The changes you made look really good.  All the Kekulization issues (in `testRingMatching*`) are fixed and the removal of unused rgroups is a nice touch.

My latest commit has test spec changes to account for Paolo' additions.  I also increased the weight of the penalty term in the fingerprint variance score. Doing this fixes issue #3896, but doesn't affect any of the tests.

greglandrum:
@jones-gareth : this looks good to me.
I just did a PR against your branch to get rid of most of the output to `rdDebugLog` and switch the rest to `rdWarningLog`, but otherwise I think it's ready to go.

greglandrum:
Since I just merged another PR to master which created a merge conflict, I'll also resolve that in my PR

greglandrum (approved):
LGTM

ptosco (approved):
FYI: I am working on some residual kekulization issues I have found yesterday testing on an internal dataset. These issues affect R-groups added to cores on aromatic nitrogens when `onlyMatchAtRGroups=false` and there are no query atoms in the core, so if you spot one of those don't worry too much as a fix will be coming soon after this one is merged.

greglandrum:
Thanks for sticking with this one @jones-gareth!

bp-kelley:
@jones-gareth already used it in “production” for some free Wilson work!

PR #3783, opened 2021-02-03
Fix pillow error in IPythonConsole.py

skearnes:
Follow-up to #3690.

After importing IPythonConsole, regular PIL image rendering fails with:

```
FormatterWarning: Exception in image/png formatter: 'Image' object has no attribute 'text'
```

The current pillow [documentation](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.info) suggests that `text` should be replaced with `info`.

I've tested this with Google-internal colab; will run additional tests with Jupyter soon...


greglandrum (approved):
LGTM

PR #3856, opened 2021-02-25
Fixes #3852

### LINKED ISSUE #3852: MinimalLib: Bonds are parallel in SVG but not on an HTML5 Canvas
**Describe the bug**
In certain rotations, bonds drawn in two parallel lines have these lines actually not parallel.

**To Reproduce**
It can be seen directly on https://rdkit.org/rdkitjs/beta/GettingStartedInJS.html. Produce an SVG, then check the corresponding Canvas drawing, to notice the parallel lines (in the top bond) are slightly not parallel, while many others are actually parallel:

image

It is seen more strongly in other occasions:

image

**Expected behavior**
Parallel lines remain parallel.

Chrome Version 88.0.4324.182 (Official Build) (64-bit)
RDKit Minimal Lib from https://rdkit.org/rdkitjs/beta/GettingStartedInJS.html.

**Additional context**
It may be some rounding error involved in producing this behaviour.

ptosco:
This PR addresses and fixes #3852:
- `width` should not be implicitly cast to `unsigned int` (I have also fixed the same issue in a few places in the `svg` and `cairo` drawers)
- coordinates should not be rounded ahead of using them in context, as the context will do that with interpolation/antialiasing when generating the bitmap

The differences are particularly evident for the first molecule; look in particular at the carbonyl double bond and at the benzene double bonds. The connections between lines used to have a small "angle" where there was a color change, and now they don't anymore, and all bonds are now nicely parallel, as we rely on antialiasing to do the interpolation and the smoothing rather than rounding pixels to integer upfront.

Before | After
--- | ---
image | image

Before | After
--- | ---
image | image

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawTextJS.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawTextJS.cpp
@@ -34,7 +34,7 @@ void DrawTextJS::drawChar(char c, const Point2D &cds) {
   std::string col = DrawColourToSVG(colour());
   context_.set("font", font);
   context_.set("fillStyle", col);
-  context_.call<void>("fillText", txt, std::round(cds.x), std::round(cds.y));
+  context_.call<void>("fillText", txt, cds.x, cds.y);
 }
 
 // ****************************************************************************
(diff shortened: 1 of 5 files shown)


ptosco:
I'll check what's up with the failing test - I hadn't even done a C++ build.

ptosco:
OK, this is now ready for review.

greglandrum (approved):

PR #3935, opened 2021-03-15
Fix MolDraw2DQt exports

ricrogz:
Enabling Qt support breaks the windows build: since we split off support to a separate library, some of the components of MolDraw2D now need to be exported, so that they can be used from MolDraw2DQt. Also, the export macro for `MolDraw2DQt` is wrong (in the CMakeLists file relevant to the Qt directory we define `RDKIT_MOLDRAW2DQT_BUILD`, but not `RDKIT_MOLDRAW2D_BUILD`)

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawText.h
+++ b/Code/GraphMol/MolDraw2D/DrawText.h
@@ -17,6 +17,7 @@
 #include <string>
 #include <vector>
 
+#include <RDGeneral/export.h>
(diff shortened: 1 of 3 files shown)


ricrogz:
Oh, it seems there are some more problems... I'll update with more.

ricrogz:
My bad -- the other problems are a different issue. I'll post separately.

greglandrum (approved):

PR #3969, opened 2021-03-23
Do not add unnecessary R-labels (and an optimization)

ptosco:
- use a `boost::dynamic_bitset` rather than a `std::set` for lookups
I replaced a `std::set` which was being used for lookups with a `dynamic_bitset` as it is more efficient. We consistently use `dynamic_bitset` for this purpose all over the RGD code, but apparently we missed this one.

- do not add unnecessary R-labels
While testing the RGD code on some internal datasets I ran into two cases that I believe were yielding suboptimal results when `onlyMatchAtRGroups=false`, because unnecessary R labels were being dynamically added.
Both are addressed by the new `testDoNotAddUnnecessaryRLabels()` and I will illustrate them below.

1. There is a single core
image
while the molecules are
image
Before this PR, the code would not make a difference between assigning a H substituent vs a heavy atom substituent to user-defined or dynamically added labels.
The scoring function, when evaluating the various permutations, would then choose the permutation which adds the smallest number of labels, and so would choose the solution that yields only R1 and R2, with zero added labels.
However, when running with `NoSymmetrization`, in the situation above you end up with three labels (R1, R2, R3), since a single permutation is retained:

```
{'Core': ['c1ncc([*:3])c([*:2])c1[*:1]',
  'c1ncc([*:3])c([*:2])c1[*:1]',
  'c1ncc([*:3])c([*:2])c1[*:1]',
  'c1ncc([*:3])c([*:2])c1[*:1]'],
 'R1': ['F[*:1]', 'Cl[*:1]', 'O[*:1]', '[H][*:1]'],
 'R2': ['[H][*:2]', 'C[*:2]', '[H][*:2]', 'C[*:2]'],
 'R3': ['[H][*:3]', '[H][*:3]', '[H][*:3]', 'F[*:3]']}
```

There is a simple fix for this, which is filtering out all degenerate matches that would require adding a label whenever there is the possibility to choose an alternative degenerate match that does not require adding a label, because it accommodates all heavy substituents into existing labels, rather than assigning one of the existing labels to H and adding another label for the other heavy substituent. This fix has a negligible impact on performance.

2. There are two cores:
image
while the molecules are
image
Even though the first molecule matches the first core and does not require adding any label, before this PR both molecules would be associated to the second core:

```
{'Core': ['c1cc([*:1])c([*:3])cc1[*:4]', 'c1cc([*:1])c([*:3])cc1[*:4]'],
 'R1': ['Cl[*:1]', 'Br[*:1]'],
 'R3': ['C[*:3]', 'C[*:3]'],
 'R4': ['F[*:4]', '[H][*:4]']}
```

This happens because the first core already matches (with 2 additional labels), and the second core is never even tried, thus yielding a rather unexpected result, as labels should only be added when there is no better alternative.
After this PR, the expected result is generated:

```
{'Core': ['Fc1ccc([*:2])c([*:1])c1', 'c1ccc([*:3])c([*:1])c1'],
 'R1': ['C[*:1]', 'C[*:1]'],
 'R2': ['Cl[*:2]', ''],
 'R3': ['', 'Br[*:3]']}
```

where the first molecule is associated to the first core, and a single additional label is added to the second core to accommodate the second molecule.
I do realize that this fix requires probing all available cores rather than just picking the first that matches so is more expensive, but I believe getting the expected result is preferable to getting a suboptimal result, though a bit faster.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -17,6 +17,13 @@
   (#3696)
 - There have been numerous changes to the RGroup Decomposition code which change
   the results. (#3767)
+- In RGroup Decomposition, when onlyMatchAtRGroups is set to false, each molecule
+  is now decomposed based on the first scaffold which adds/uses the least number
+  of non-user-provided R labels, rather than the first scaffold tout court.
+  Among other things, this allows the code to provide the same results for both
+  onlyMatchAtRGroups=true and onlyMatchAtRGroups=false when suitable scaffolds
+  are provided without requiring the user to get overly concerned about the
+  input ordering of the scaffolds. (#3969)
 - There have been numerous changes to `GenerateDepictionMatching2DStructure()` (#3811)
 - Setting the kekuleSmiles argument (doKekule in C++) to MolToSmiles will now
   cause the molecule to be kekulized before SMILES generation. Note that this
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -18,8 +18,9 @@
 - There have been numerous changes to the RGroup Decomposition code which change
   the results. (#3767)
 - In RGroup Decomposition, when onlyMatchAtRGroups is set to false, each molecule
-  is now decomposed based on the first scaffold which adds/uses the least number
-  of non-user-provided R labels, rather than the first scaffold tout court.
+  is now decomposed based on the first matching scaffold which adds/uses the
+  least number of non-user-provided R labels, rather than simply the first
+  matching scaffold.
   Among other things, this allows the code to provide the same results for both
   onlyMatchAtRGroups=true and onlyMatchAtRGroups=false when suitable scaffolds
   are provided without requiring the user to get overly concerned about the
--- a/Code/GraphMol/RGroupDecomposition/RGroupCore.h
+++ b/Code/GraphMol/RGroupDecomposition/RGroupCore.h
@@ -24,37 +24,32 @@ struct RCore {
   boost::shared_ptr<RWMol> core;
   boost::shared_ptr<RWMol> labelledCore;
 
-  // A list of user indices for when onlyMatchAtRGroups = True
-  std::set<int> core_atoms_with_user_labels;
+  // Bitset: indices corresponding to atoms bearing user-defined labels are 1
+  boost::dynamic_bitset<> core_atoms_with_user_labels;
   // Number of user labelled rgroups in the core
   size_t numberUserRGroups = 0;
   RCore(){};
-  RCore(const RWMol &c, bool onlyMatchAtRGroups = false) : core(new RWMol(c)) {
-    // Remove this from constructor if the create new core path can be removed from RGroupDecomposition::add
-    if (onlyMatchAtRGroups) {
-      findIndicesWithRLabel();
-    }
+  RCore(const RWMol &c) : core(new RWMol(c)) {
+    findIndicesWithRLabel();
     countUserRGroups();
   }
 
+  inline bool isCoreAtomUserLabelled(int idx) const {
+    return core_atoms_with_user_labels.test(idx);
+  }
+
   void countUserRGroups() {
-    numberUserRGroups = 0;
-    for (const auto atom : core->atoms()) {
-      int label;
-      if (atom->getPropIfPresent(RLABEL, label)) {
-        if (label > 0) {
-          ++numberUserRGroups;
-        }
-      }
-    }
+    numberUserRGroups = core_atoms_with_user_labels.count();
   }
 
   void findIndicesWithRLabel() {
-    // First find all the core atoms that have user
-    //  label and put their indices into core_atoms_with_user_labels
+    // Find all the core atoms that have user
+    // label and set their indices to 1 into core_atoms_with_user_labels
+    core_atoms_with_user_labels.resize(core->getNumAtoms());
     for (const auto atom : core->atoms()) {
-      if (atom->hasProp(RLABEL)) {
-        core_atoms_with_user_labels.insert(atom->getIdx());
+      int label;
+      if (atom->getPropIfPresent(RLABEL, label) && label > 0) {
+        core_atoms_with_user_labels.set(atom->getIdx());
       }
     }
   }
--- a/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
+++ b/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
@@ -161,6 +161,9 @@ int RGroupDecomposition::add(const ROMol &inmol) {
         global_min_heavy_nbrs = min_heavy_nbrs;
         rcore = &core.second;
         core_idx = core.first;
+        if (global_min_heavy_nbrs == 0) {
+          break;
+        }
       }
     } else if (!tmatches_filtered.empty()) {
       rcore = &core.second;
--- a/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
+++ b/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
@@ -89,8 +89,12 @@ int RGroupDecomposition::add(const ROMol &inmol) {
   int core_idx = 0;
   const RCore *rcore = nullptr;
   std::vector<MatchVectType> tmatches;
+  std::vector<MatchVectType> tmatches_filtered;
 
-  // Find the first matching core.
+  // Find the first matching core (onlyMatchAtRGroups)
+  // or the first core that requires the smallest number
+  // of newly added labels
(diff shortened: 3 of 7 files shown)


greglandrum:
> This happens because the first core already matches (with 2 additional labels), and the second core is never even tried, thus yielding a rather unexpected result, as labels should only be added when there is no better alternative.
After this PR, the expected result is generated:
```
{'Core': ['Fc1ccc([*:2])c([*:1])c1', 'c1ccc([*:3])c([*:1])c1'],
 'R1': ['C[*:1]', 'C[*:1]'],
 'R2': ['Cl[*:2]', ''],
 'R3': ['', 'Br[*:3]']}
```
> where the first molecule is associated to the first core, and a single additional label is added to the second core to accommodate the second molecule.
> I do realize that this fix requires probing all available cores rather than just picking the first that matches so is more expensive, but I believe getting the expected result is preferable to getting a suboptimal result, though a bit faster.

Assuming I understand this correctly, I disagree: I think being able to say: "the cores are matched in the order provided and the code takes the first one that matches" is really helpful and makes the results more predictable. It's pretty straightforward for client code to reorder the cores to get the behavior you're looking for with this behavior change.

greglandrum:
Fixing the behavior in your bullet point 1., on the other hand, makes sense to me.

ptosco:
> Assuming I understand this correctly, I disagree: I think being able to say: "the cores are matched in the order provided and the code takes the first one that matches" is really helpful and makes the results more predictable. It's pretty straightforward for client code to reorder the cores to get the behavior you're looking for with this change.

@greglandrum Thanks for the review. I see the source of your concern, but my change does not impact on the core ordering - the order is still taken into account (and I do agree this is very important). Before the end of the morning I will come up with a better example (that I will include as a test case) that explains how the change works and how I think this improves the previous behavior.

ptosco:
@greglandrum I have added a test to better illustrate my point and I will explain it below.
The algorithm still honors the ordering of cores. It will look for the first core in the order provided that matches the molecule without requiring adding labels. I think this is exactly the behavior that we want.
I have put together another example to illustrate my point.
Say that we have 3 cores, in this order:
image

And these 5 molecules:
image

Note that I have moved the core with fluorine at the first place, as you suggested.
The current `master` code will now yield these results, which I still think are suboptimal (I would be tempted to say "wrong"):
image

The first molecule now correctly picks the first core, as it now the first choice.
However, the second molecule picks the second core, because it is the first that matches in the order provided. As a result, R3 and R4 are added.
To avoid this, I could have inverted the order of the second and third core:
image

but this would cause the third molecule to also pick the disubstituted core, while the monosubstituted core, being last, would never be picked:
image

This means that I cannot achieve an optimal RGD by simply reordering the cores.

Instead, with the code in my PR, I can use this core ordering
image

and get the results that I expect, with a single added R label rather than two, which are unnecessary:
image

In summary, this is how the new algorithm works:
* cores are tried in the provided order
* for each core, the number of additional labels required to accommodate the molecule is recorded
* as soon as a core that requires no additional labels to accommodate the molecule is found, this core is accepted as no better core can be found
* After all cores are tried, the first core in the provided order that requires the smallest number of additional R labels is picked

I hope this make sense to you and you agree this is the most desirable behavior.

ptosco:
Here's a couple of benchmarks I ran on an internal dataset with 25 cores and 7805 molecules.
The current PR speeds up RGD with `onlyMatchAtRGroups=True` by a factor >2. I believe most of the gain comes from early pruning of unnecessary symmetries (point 1 above).

### Current `master` timings with `onlyMatchAtRGroups=True`
```
%%timeit
params = RGroupDecompositionParameters()
params.onlyMatchAtRGroups = True
params.alignment = RGroupCoreAlignment.NoAlignment
RDLogger.DisableLog("rdApp.warning")
rgd = RGroupDecomposition(cores_r_groups, params)
n_accepted = 0
for m in mols:
    if (rgd.Add(m) != -1):
        n_accepted += 1
rgd.Process()
RDLogger.EnableLog("rdApp.warning")
print("n_accepted", n_accepted)

n_accepted 4818
20.3 s ± 486 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Current `master` timings with `onlyMatchAtRGroups=False`
```
%%time
params = RGroupDecompositionParameters()
params.onlyMatchAtRGroups = False
params.alignment = RGroupCoreAlignment.NoAlignment
RDLogger.DisableLog("rdApp.warning")
rgd = RGroupDecomposition(cores_r_groups, params)
n_accepted = 0
for m in mols:
    if (rgd.Add(m) != -1):
        n_accepted += 1
rgd.Process()
RDLogger.EnableLog("rdApp.warning")
print("n_accepted", n_accepted)

n_accepted 5062
1min 59s ± 1.43 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### This PR timings with `onlyMatchAtRGroups=True`
```
%%timeit
params = RGroupDecompositionParameters()
params.onlyMatchAtRGroups = True
params.alignment = RGroupCoreAlignment.NoAlignment
RDLogger.DisableLog("rdApp.warning")
rgd = RGroupDecomposition(cores_r_groups, params)
n_accepted = 0
for m in mols:
    if (rgd.Add(m) != -1):
        n_accepted += 1
rgd.Process()
RDLogger.EnableLog("rdApp.warning")
print("n_accepted", n_accepted)

n_accepted 4818
21.2 s ± 253 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### This PR timings with `onlyMatchAtRGroups=False`
```
%%time
params = RGroupDecompositionParameters()
params.onlyMatchAtRGroups = False
params.alignment = RGroupCoreAlignment.NoAlignment
RDLogger.DisableLog("rdApp.warning")
rgd = RGroupDecomposition(cores_r_groups, params)
n_accepted = 0
for m in mols:
    if (rgd.Add(m) != -1):
        n_accepted += 1
rgd.Process()
RDLogger.EnableLog("rdApp.warning")
print("n_accepted", n_accepted)

n_accepted 5062
48.5 s ± 1.31 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

greglandrum:
Ok, Paolo and I just talked offline and I think I understand the goal here. I would express it a bit differently than Paolo did, so let me try that and he can check whether or not he agrees with my statement.

The current behavior (on master and in the previous releases) of the RGD code is to use the first scaffold which matches a molecule. The second part of this PR changes that so that the code uses the first scaffold which adds/uses the least number of  non-user-provided R groups. Among other things, this allows the code to provide the same results for both `onlyMatchAtRGroups =true` and `onlyMatchAtRGroups=false` when suitable scaffolds are present without requiring the user to get overly concerned about the ordering of the scaffolds.

ptosco:
Correct, that summarizes it nicely. I have tested this scenario on an internal dataset with 25 cores and 7805 molecules.

When I run it with `onlyMatchAtRGroups=true` 4818 molecules are accepted.
When I run it with `onlyMatchAtRGroups=false` 5062 molecules are accepted.

With this PR in place, if I compare the 4818 molecules which are present in both RGDs, they are identical in terms of associated cores and extracted R-groups, without having to change the ordering of the cores between the two runs, which is where I wanted to be,

Instead, with the current `master` code, I end up with 55 of the 4818 common molecules being associated to a different core and R-groups in the RGD run with `onlyMatchAtRGroups=false` compared to the former.

greglandrum (approved):
LGTM
@jones-gareth : do you agree with the behavior change being proposed?

jones-gareth:
> 
> 
> LGTM
> @jones-gareth : do you agree with the behavior change being proposed?

These changes are good- they both favor user defined R groups which is what we want. 

The first change of pruning suboptimal matches from single cores would have prevented many of the scoring function issues in the previous CRs (though it's good they're fixed!).  Could give considerable speed up for symmetric cores.

The new behavior for multiple cores makes sense.  I've never come across that issue as I've normally used multiple cores for matching different generic scaffolds.  Assuming we're all happy with the change this LGTM.

bp-kelley:
It makes total sense to use the best matching scaffold.  I think we were
just being greedy before.

ReleaseNotes.md
```diff
@@ -17,6 +17,13 @@
   (#3696)
 - There have been numerous changes to the RGroup Decomposition code which change
   the results. (#3767)
+- In RGroup Decomposition, when onlyMatchAtRGroups is set to false, each molecule
+  is now decomposed based on the first scaffold which adds/uses the least number
+  of non-user-provided R labels, rather than the first scaffold tout court.
```
greglandrum, line 22: “tout court”?

ptosco: I thought that was international, but maybe not. Changed :)

PR #3971, opened 2021-03-24
Add return codes and make RGroupDecomp less verbose

bp-kelley:
Makes two return codes for RGroupDecomp.add

  -1 no matching cores
  -2 sidechain with no attachment points

Moves rdWarning to rdDebug


greglandrum (changes requested):
Change makes sense and looks good, but please add a test and comment in the release notes about the backwards incompatible change.

greglandrum (approved):

PR #4002, opened 2021-04-04
RGD - a fix for the cubane issue (single target atom matches 2 user R group attachments)

jones-gareth:
#### Issue

Fixes the cubane example mentioned in #3767.  This is the situation where the molecule to be decomposed has a single atom that matches two user-defined R groups.

Given this molecule:

Capture2

And any of these cores:

Capture2

You now get this (correct) RGD:

Capture2

The following changes have been made:

- Added a `matchingMol` field to `RCore` that is the core with terminal R groups removed.
- `matchingMol` is used to initially match a molecule to the core then all combinations of the terminal R groups are evaluated for compatible matches.
- exposure of the `MolMatchFinalCheckFunctor` so that matches handle chirality properly.
- more changes to `RCore` to check that `onlyMatchAtRGroups` true is handled properly
- changes to `removeCore` in `ChemTransforms`  to handle multiple core atoms mapped to a single molecule atom
- Created `RGroupCore.cpp` as the header file was getting large.


bp-kelley:
There are a lot of changes here not related to the issue, it's kind of hard to debug.  It would be helpful if you highlight the minimal changes that are needed for review.

jones-gareth:
> There are a lot of changes here not related to the issue, it's kind of hard to debug. It would be helpful if you highlight the minimal changes that are needed for review.

`RCore::buildMatchingMol` removes terminal R groups from the core to create a molecule used for matching- *matchingMol*

`RCore::matchTerminalUserRGroups` takes a *matchingMol* match and enumerates all possible matches including terminal R groups.

In `chemTransforms.cpp`, `RDKit::replaceCore` has changes so that two or more query atoms in the core can match to the same molecule atom.

In `RGroupDecomposition::add` there are changes for `onlyMatchAtRGroups`.

bp-kelley:
@greglandrum @jones-gareth I'll finish this up today, but it looks like conflicts have popped up in the interim.

jones-gareth:
> @greglandrum @jones-gareth I'll finish this up today, but it looks like conflicts have popped up in the interim.

Just fixed the conflicts.  Hopefully the tests will pass.

jones-gareth:
@greglandrum, @bp-kelley I hate to hassle you, but do you have a timeline for review?

greglandrum:
> @greglandrum @jones-gareth I'll finish this up today, but it looks like conflicts have popped up in the interim.

I will take a look tomorrow morning

bp-kelley:
All my tests are passing (most using in-house free wilison tests), but I'm
taking a lot longer to vet the extra reorrganization changes.  I'll see if
I can understand these tonight as well.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
```diff
@@ -51,6 +51,18 @@ void updateSubMolConfs(const ROMol &mol, RWMol &res,
     res.addConformer(newConf, false);
   }
 }
+
+struct SideChainMapping {
+  int molIndex;
+  int coreIndex = -1;
+  bool useMatch;
```
bp-kelley, line 58: useMatch could be set false here, or coreIndex should be unset, you override it in the constructor anyway.

bp-kelley:
There are just too many changes for me to properly review, I think I follow the logic though.

  1. make a matching molecule with no Rgroups
  2. do the matching, reassemble the "rgroups" keeping track of single atoms such as the one in cubane
  3. Alter deleteSubstructs so that two or more query atoms can match the same atom (I'm very confused by this, it seems to violate what a substructure means).

What would help me immensely if we could do a diff to the reorganized code to isolate the changes.  Otherwise, my review will basically be validation.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
```diff
@@ -51,6 +51,18 @@ void updateSubMolConfs(const ROMol &mol, RWMol &res,
...
+
+  SideChainMapping(int molIndex)
+      : molIndex(molIndex), coreIndex(-1), useMatch(false) {}
+  SideChainMapping(int molIndex, int coreIndex, bool useMatch)
+      : molIndex(molIndex), coreIndex(coreIndex), useMatch(useMatch) {}
+};
+
 }  // namespace
 
 ROMol *deleteSubstructs(const ROMol &mol, const ROMol &query, bool onlyFrags,
```
bp-kelley, line 68: There need to be tests for this new behavior I think.

bp-kelley (changes requested):

jones-gareth:
>     3. Alter deleteSubstructs so that two or more query atoms can match the same atom (I'm very confused by this, it seems to violate what a substructure means).

It's actually replaceCore that has the the changes  (the github formatting makes it look like deleteSubstructs).  The changes account for the fact that two core atoms (in this case 2 R group attachment points) can map to a single structure atom.
> 
> 
> What would help me immensely if we could do a diff to the reorganized code to isolate the changes. Otherwise, my review will basically be validation.

How does what you want differ from the "Files changed" tab? I'm happy to try to put together a diff of what you want, but I don't understand your requirements.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
jones-gareth: The only time the new behavior is activated is when RGD has two attachment points mapped to a single atom.  That is tested in testSingleAtomBridge in testRGroupDecomp.cpp.  I think that testRGroupDecomp.cpp is the right place to test the new behavior.  Admittedly, I only have a single test, though it does check 5 decompositions.

My bigger concern is that somehow I have broken the existing behavior of replaceCore- hopefully the existing tests protect against this.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
jones-gareth: I'll unset coreIndex

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
```diff
@@ -347,8 +359,13 @@ ROMol *replaceCore(const ROMol &mol, const ROMol &core,
                    const MatchVectType &matchV, bool replaceDummies,
                    bool labelByIndex, bool requireDummyMatch) {
   unsigned int origNumAtoms = mol.getNumAtoms();
+  std::vector<std::pair<int, SideChainMapping>> matches;
+  matches.reserve(origNumAtoms);
+
   std::vector<int> matchingIndices(origNumAtoms, -1);
   std::vector<int> allIndices(origNumAtoms, -1);
+  boost::dynamic_bitset<> molAtomsMapped(origNumAtoms);
+  std::set<int> multipleMappedMolAtoms;
```
greglandrum, line 368: It's probably better to use a dynamic_bitset than an std::set for this "flag the interesting items and then loop over items and check if they are interesting" pattern. 
The code ends up being a tiny bit simpler and the bitset alternative has been faster every time I've tested it.

I assume this would end up being different if the potential length of the bitset were huge, but that's not going to be the case with atom indices.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
```diff
@@ -358,11 +375,45 @@ ROMol *replaceCore(const ROMol &mol, const ROMol &core,
...
+    for (const auto &match : matches) {
+      const auto &mappingInfo = match.second;
+      if (multipleMappedMolAtoms.find(mappingInfo.molIndex) !=
+          multipleMappedMolAtoms.end()) {
+        auto coreAtom = core.getAtomWithIdx(mappingInfo.coreIndex);
+        CHECK_INVARIANT(
+            coreAtom->getDegree() == 1,
+            "Multiple core atoms match a mol atom, but one of the core "
+            "atoms has degree > 1 ");
+        auto coreNeighborIdx = *core.getAtomNeighbors(coreAtom).first;
```
greglandrum, line 403: This relies on the boost::graph returning atom indices as vertex descriptors.
It currently works (and probably shows up elsewhere in the code) but would break if we were to make changes to the underlying graph representation.
It would be safer to do:
```
auto coreNeighborIdx = core[*core.getAtomNeighbors(coreAtom).first]->getIdx();
```

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
```diff
@@ -358,11 +375,45 @@ ROMol *replaceCore(const ROMol &mol, const ROMol &core,
...
+        auto molNeighborIdx =
+            std::find_if(matchV.cbegin(), matchV.cend(),
+                         [coreNeighborIdx](std::pair<int, int> p) {
+                           return p.first == static_cast<int>(coreNeighborIdx);
+                         })
+                ->second;
+        if (molNeighborIdx > -1) {
+          auto connectingBond =
```
greglandrum, line 411: Since the next line could potentially seg fault, please add:
```
auto connectingBond =
    mol.getBondBetweenAtoms(mappingInfo.molIndex, molNeighborIdx);
CHECK_INVARIANT(connectingBond,"expected bond in molecule not found");
```

Code/GraphMol/RGroupDecomposition/RGroupCore.cpp
```diff
@@ -0,0 +1,365 @@
...
+  std::vector<std::vector<int>> s = {{}};
+  for (const auto &u : v) {
+    std::vector<std::vector<int>> r;
+    for (const auto &x : s) {
+      for (const auto y : u) {
+        r.push_back(x);
+        r.back().push_back(y);
+      }
+    }
+    s = move(r);
```
greglandrum, line 21: I think this should be `std::move(r)`

Code/GraphMol/RGroupDecomposition/RGroupCore.cpp
```diff
@@ -0,0 +1,365 @@
...
+
+namespace RDKit {
+
+namespace {
+// From answer 12 in
+// https://stackoverflow.com/questions/5279051/how-can-i-create-cartesian-product-of-vector-of-vectors
+// by anumi
+static std::vector<std::vector<int>> cartesianProduct(
+    const std::vector<std::vector<int>> &v) {
+  std::vector<std::vector<int>> s = {{}};
```
greglandrum, line 12: The empty initializer list isn't necessary.

Code/GraphMol/Substruct/SubstructMatch.h
```diff
@@ -206,6 +209,21 @@ inline unsigned int SubstructMatch(ResonanceMolSupplier &resMolSupplier,
   return matchVect.size();
 };
 
+class RDKIT_SUBSTRUCTMATCH_EXPORT MolMatchFinalCheckFunctor {
```
greglandrum, line 212: Since this is moving into the public API, it should pick up some documentation. Perhaps:
```
//! Class used as a final step to confirm whether or not a given atom->atom mapping is a valid
//! substructure match.
```

greglandrum (changes requested):
I didn't spend  a lot of time in the RGD changes themselves since I'm assuming that your new tests cover those and that  Brian is looking at them.

The non-RGD changes mostly look good; I made a few suggestions/comments.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
greglandrum: > My bigger concern is that somehow I have broken the existing behavior of replaceCore- hopefully the existing tests protect against this.

I think it's fine to rely on this.

greglandrum: But I agree with @bp-kelley that there should be a test in ChemTransforms for the new behavior.
It's good to also have the tests in the RGD code that demonstrate that your changes work there, but having more of a unit-test-like test here would also be good.

bp-kelley:
@jones-gareth because the code in the headers was moved to a cpp, the diffs show basically everything is changed and it is hard to focus on the changes relevant to this PR.  If you did the reorganization prior to the fixes, I can just change the diff targets, but as it is it is really hard to isolate the specific changes.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
jones-gareth: Changed

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
jones-gareth: Changed

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
jones-gareth: Changed

Code/GraphMol/RGroupDecomposition/RGroupCore.cpp
jones-gareth: Changed- though I'm not sure how it worked originally!

Code/GraphMol/RGroupDecomposition/RGroupCore.cpp
jones-gareth: I think it is- otherwise this line will do nothing (as s is empty) and the product will always be empty:

```cpp
 for (const auto &x : s) {
```

Code/GraphMol/Substruct/SubstructMatch.h
jones-gareth: Done

jones-gareth:
> @jones-gareth because the code in the headers was moved to a cpp, the diffs show basically everything is changed and it is hard to focus on the changes relevant to this PR. If you did the reorganization prior to the fixes, I can just change the diff targets, but as it is it is really hard to isolate the specific changes.

@bp-kelley I have moved existing functions from RGroupCore.cpp back into RGroupCore.h.  For the most part RGroupCore.h has old code and RGroupCore.cpp has the new stuff.  Once we get done with this, I can do a quick CR to tidy up those files.

Code/GraphMol/ChemTransforms/ChemTransforms.cpp
jones-gareth: I'll put together a ChemTransforms test

Code/GraphMol/RGroupDecomposition/RGroupCore.cpp
greglandrum: Yeah, that's embarrassing. Sorry for the noise.

greglandrum (approved):
LGTM

greglandrum:
@bp-kelley : I think this looks good, so please merge whenever you've approved.

bp-kelley:
@jones-gareth I finally had time to fully understand the PR.  I do think we should write down the new matching strategy (matchingMol and then RGroup identification) somewhere as I *know* I will forget this at some point and the complexity is non trivial.  I'm going to add this as an issue just so I can move forwards.

PR #4069, opened 2021-04-21
Added function to explicitly delete RDKit mol in JSMol.

DavidACosgrove:
#### Reference Issue

Addresses discussion point #4066

This adds a delete_mol function to JSMol to free the underlying RWMol.

#### Any other comments?
I found that when processing large lists of molecules in a loop, the garbage collector didn't free the memory fast enough, and I got an OOM error.  This solved it.
Tested by building with supplied Dockerfile and run on a MacBook under Catalina, with Safari and Firefox.


ptosco:
@DavidACosgrove Hi David, maybe I am misunderstanding the problem but wouldn't calling `delete()` on the `mol` object on the JS side solve the OOM issue? `delete()` _must_ be called on all C++ object to avoid leaking memory:
https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#memory-management
Calling `delete()` on the JS side should have the same effect as resetting the `unique_ptr` on the C++ side.

PR #4104, opened 2021-05-05
Normalize line endings in source code files

ptosco:
The context for this PR is that I have noticed that commit `f829c877d8fbbebb1f6b59983e4bc3ad563be9b8` on 15 Apr introduce a number of changes in line endings (from `LF` to `CR+LF`).

As a consequence, when I tried to merge the current `master` into my local branch that still was on `LF` endings, I got huge "bogus" conflicts caused by these changes in line endings.

While I can fix the problem at my end by running `unix2dos` on the conflicting files ahead of merging, I thought it would be good to set a standard to avoid these issues in the future,

Git provides a mechanism to always convert line endings to the native endings of the machine where the repo is cloned:
https://docs.github.com/en/github/getting-started-with-github/configuring-git-to-handle-line-endings

As there is no guarantee that developers has the same settings in their own `git` configuration, enforcing `auto` settings for all source code files in the `.gitattributes` file should provide a future-proof way of handling the problem.

Happy to shouted at if this is not a good idea :-)


greglandrum (approved):
LGTM

ptosco:
@greglandrum I have fixed conflicts again; if the tests pass could you please merge? Otherwise soon there will be more conflicts, I am afraid. Thanks!

PR #4143, opened 2021-05-16
Run clang-format against header files

e-kwsm:
In this PR header files are prettified by clang-format as there are a bunch of cpp files that are not formatted as well.
Also, unnecessary trailing commas are removed.

#### Any other comments?
~~Some of modification of comments are not included because doxygen blocks are broken.~~ Fixed.


bp-kelley:
At one point Greg (@greglandrum) and I talked about running clang-format on the entire codebase after a release, so my suggestion is to run clang-format on everything.  This will make reviews easier in my opinion.

greglandrum:
@bp-kelley, @e-kwsm : I think this makes sense, but we should be sure that all outstanding C++ PRs have been merged before doing it. This will otherwise make life extremely difficult for those contributors.

bp-kelley:
Agreed.

greglandrum:
Thanks @e-kwsm : I will go ahead and merge this after the CI builds finish.

greglandrum:
There was never going to be a really good time to merge this so...

PR #4189, opened 2021-05-30
Comments added to RGD core matching

jones-gareth:
#### Reference Issue

Addresses #4168 

Added comments to explain matching strategy for when cores contain user defined R groups.

Also moved function body code from RGroupCore.h to RGroupCore.cpp


greglandrum (approved):
LGTM

PR #4211, opened 2021-06-05
Fix memory leak with deleting array

bp-kelley:
Fixes delete called on array


greglandrum (approved):
nice one.
did you run valgrind across the RGD code?

bp-kelley:
@greglandrum the Xcode analyzer actually found this one, I caught it by accident.

PR #4347, opened 2021-07-21
Two MinGW build fixes and one MSVC build fix

ptosco:
- fixes MinGW build. Some pre-processor directives were (relatively) recently
  changed and caused the MinGW build to break. This PR fixes the problem
- make sure that supplier `istream` is opened in binary mode or `istream::tellg()` will
  report wrong offsets on MinGW builds (MSVC will make no difference) and cause
  `SmilesMolSupplier` read faillures
- avoid that in MSVC conda DLL builds the `libboost_*.lib` static
  libraries are picked by `cmake` instead of the `boost_*.lib` import libraries
  as they live in the same directory by defining `BOOST_ALL_DYN_LINK`
  I have noticed this in a Python 3.9 build against the latest Boost 1.73
  available in conda.

You may ignore the huge diff in the two FMCS tests; they are due to having applied `clang-format`. The only relevant change in each of the files is:
```
#ifdef _WIN32
#include <windows.h>
#endif
```

### DIFF
--- a/Code/GraphMol/FileParsers/GeneralFileReader.h
+++ b/Code/GraphMol/FileParsers/GeneralFileReader.h
@@ -104,7 +104,7 @@ std::unique_ptr<MolSupplier> getSupplier(const std::string& path,
 
   std::istream* strm;
   if (compressionFormat.empty()) {
-    strm = new std::ifstream(path.c_str());
+    strm = new std::ifstream(path.c_str(), std::ios::in | std::ios::binary);
   } else {
     strm = new gzstream(path);
   }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,6 +234,9 @@ endif(MSVC)
 if(MSVC)
   # FIX: do we still need this?
   target_compile_definitions(rdkit_base INTERFACE "-DBOOST_ALL_NO_LIB")
+  if(RDK_INSTALL_DLLS_MSVC)
+    target_compile_definitions(rdkit_base INTERFACE "-DBOOST_ALL_DYN_LINK")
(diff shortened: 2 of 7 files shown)


Code/GraphMol/FileParsers/GeneralFileReader.h
```diff
@@ -104,7 +104,7 @@ std::unique_ptr<MolSupplier> getSupplier(const std::string& path,
 
   std::istream* strm;
   if (compressionFormat.empty()) {
-    strm = new std::ifstream(path.c_str());
+    strm = new std::ifstream(path.c_str(), std::ios::in | std::ios::binary);
```
bp-kelley, line 107: Does opening in binary mode have any other side effects?  I think this assumes the file is ASCII which I think is a fair assumption except when reading UTF-8 csv files.

ptosco: @bp-kelley It won't affect the encoding - the low-level C functions will read bytes in both cases.
`rb` is identical to `r` for POSIX systems:
https://www.gnu.org/software/libc/manual/html_node/Opening-Streams.html
It only affects non-POSIX systems like Windows, and its scope is only limited to how newlines are handled:
http://c-faq.com/stdio/textvsbinary.html
As for why MinGW `ftell` does strange things on files opened in text mode, I found this:
https://lists.gnu.org/archive/html/bug-gnulib/2008-03/msg00256.html
As the offsets are only used to jump to the right position in the file with `fseek` and newlines are stripped anyway by the tokenizer, I think this is safe.

greglandrum (approved):
LGTM
and it doesn't break my windows java wrapper builds, which was my biggest concern

PR #4356, opened 2021-07-23
Three more Windows build fixes

ptosco:
- on Windows when using MSVC it is not possible to build at the same time static and dynamic libraries as on *nix, because of the `dllimport`/`dllexport` keywords.
  For this reason, we do not append the `_static` suffix to static libraries as it is not necessary, since the two kinds of library never coexist.
  The same should happen when using MinGW compilers, as `dllimport`/`dllexport` are platform-specific, not compiler-specific.
  However, this is not the case at the moment, so the MinGW static build is broken.
  This PR fixes this issue.
- The C# wrapper build on Windows under MinGW is broken as the `CMakeLists.txt` file assumes that MSVC is used under Windows. This PR fixes this.
- Under Windows, `-DSWIGWORDSIZE32` need to be defined also in 64-bit builds. This is because of these definitions:
  https://github.com/swig/swig/blob/90cdbee6a69d13b39d734083b9f91069533b0d7b/Lib/swigarch.i#L50
  Given that under Windows `LONG_MAX == INT_MAX`, the SWIG build breaks if `-DSWIGWORDSIZE64` is defined.

```
$ grep -wE 'INT_MAX|LONG_MAX' /mingw64/x86_64-w64-mingw32/include/limits.h
#define INT_MAX 2147483647
#define LONG_MAX 2147483647L

$ grep -wE 'INT_MAX|LONG_MAX' /m/Microsoft\ Visual\ Studio/2019/Community/VC/Tools/MSVC/14.16.27023/include/limits.h
#define INT_MAX       2147483647    // maximum (signed) int value
#define LONG_MAX      2147483647L   // maximum (signed) long value
```


greglandrum (dismissed):
Changes look fine to me except for the SWIG changes, which I don't understand.
The current SWIG wrappers most definitely do build with MSVC on Windows.

greglandrum (commented):
I'm still wondering about the necessity of the SWIG changes

ptosco:
@greglandrum Let me double check tonight - maybe it’s a MinGW-only thing.

ptosco:
@greglandrum Yes, I have just confirmed it's a MinGW only problem. For now I have reverted the change, over the next few days I'll investigate further and if need be submit another PR to fix the MinGS build.

greglandrum (approved):

PR #4426, opened 2021-08-20
Addresses #4425

### LINKED ISSUE #4425: "to-Python converter already registered; second conversion method ignored." warnings issued at import
Observed on Release 2020_03_3 on CentOS7 linux and also in current master on Ubuntu 21.04:

When running:
```
from rdkit.Chem import rdTautomerQuery
```
or 
```
from rdkit.Chem import rdTautomerQuery
```
I get warnings like:
```
<frozen importlib._bootstrap>:228: RuntimeWarning: to-Python converter for std::vector<unsigned long, std::allocator<unsigned long> > already registered; second conversion method ignored.
```

I have been looking around our wrapping code, and I have noticed that in some places we have a check in place to prevent trying to re-register converters for types we already have, but it seems these checks are not carried out in all places where they should.

I'll post a PR soon.

ricrogz:
Closes #4425 

It took me a while to find the converters responsible of triggering the warnings. And while looking for them, I made some refactoring:

- added a templated function to Wrap.h that makes the check to see if a converter for a specific type has already been set up.
- added a check based on this function to `RegisterVectorConverter()`
- also added checks to other converter registrations for types that look like generic enough as to be potentially used somewhere else.
- Replaced several definitions of converters for vectors of different types with `RegisterVectorConverter()`, to benefit for the check and reduce redundant code.
- remove some unused boost::python headers.

And I think that's all I did. The rest is clang-format getting over excited about code that apparently hasn't been formatted in a while...

### DIFF
--- a/Code/GraphMol/Abbreviations/Wrap/rdAbbreviations.cpp
+++ b/Code/GraphMol/Abbreviations/Wrap/rdAbbreviations.cpp
@@ -9,7 +9,6 @@
 //
 
 #include <RDBoost/python.h>
-#include <boost/python/suite/indexing/vector_indexing_suite.hpp>
 #include <GraphMol/GraphMol.h>
 #include <RDBoost/Wrap.h>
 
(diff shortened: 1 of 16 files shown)


greglandrum (approved):
LGTM
Thanks!

PR #4450, opened 2021-08-26
Addresses #4449

### LINKED ISSUE #4449: Building with static dependencies breaks CMake exports
Recently, I had to build and link RDKit & dependencies into another project as static libs. The other project used CMake, so I tried to import RDKit's exports to configure the project, and found a problem that broke the build of the other project:

When we use external builds of maeparser and coordgen, RDKit's mechanism to build static libraries is breaking the export target. Looking into rdkit-targets.cmake (generated under `rdkit/build/CMakeFiles/Export/lib/cmake/rdkit`), it contains the following (I added line breaks after each semicolon, to make it more readable):

```
set_target_properties(RDKit::Depictor_static PROPERTIES
  INTERFACE_LINK_LIBRARIES "RDKit::RDGeometryLib_static;
RDKit::RDGeneral_static;
RDKit::MolAlign_static;
/tmp/deps/coordgen-install/lib/libcoordgen.a_static;
/tmp/deps/maeparser-install/lib/libmaeparser.a_static;
Boost::system;
Boost::iostreams;
Boost::system;
Boost::iostreams;
RDKit::SubstructMatch_static;
RDKit::GraphMol_static;
RDKit::RDGeneral_static;
RDKit::rdkit_base"
)

[...]

set_target_properties(RDKit::FileParsers_static PROPERTIES
  INTERFACE_LINK_LIBRARIES "RDKit::RDGeometryLib_static;
RDKit::RDGeneral_static;
RDKit::Depictor_static;
RDKit::SmilesParse_static;
RDKit::GraphMol_static;
/tmp/deps/maeparser-install/lib/libmaeparser.a_static;
Boost::system;
Boost::iostreams;
Boost::system;
Boost::iostreams;
/usr/lib/x86_64-linux-gnu/libz.so;
RDKit::rdkit_base"
)
```

The problem here is that neither `/tmp/deps/coordgen-install/lib/libcoordgen.a_static` nor `/tmp/deps/maeparser-install/lib/libmaeparser.a_static` exist: the actual lib filenames (`libcoordgen.a` and `libmaeparser.a`) have been added a "_static" suffix). So, importing this CMake targets file will cause the parent project to fail building, since we won't be able to link against these libraries.

I know where this comes from, and I think I can improve/fix it, so I'll post a PR soon.

Also, there's a second issue, which is that having explicit file names in the CMake export target prevents it from being relocatable, but that's a different issue (the exports will work, as long as the explicitly mentioned dependencies aren't moved).

ricrogz:
Fixes #4449 

Currently, we use a list of hardcoded patterns to distinguish which link dependencies shouldn't get the "_static" suffix added. This list doesn't contain any patterns matching maeparser or coordgen. The simplest solution would be to add them, but I think we can do better than this.

In my changes, I replace this mechanism with a different check: if a link dependency is an explicit file path or a target imported from a CMake package (`find_package()` sets the `IMPORTED` property to `TRUE`), then we use it as is. In any other case, we add the "_static" suffix.

The more intuitive approach, which is checking whether RDKit created the target is not possible: we do not create the targets in sequence, so that when we specify link dependencies we might not yet have defined the targets we are linking against, so we can't check anything about them.

Also, directly using `target_link_libraries()` to add the dependencies instead of creating a `rdk_static_link_libraries` list avoids adding redundant/unrequired dependencies to some libraries: macros work in the same scope of the file that invokes  them, so if we call `rdkit_library()` twice in the same CMakeLists.txt, the second one will find `rdk_static_link_libraries` as the first left it, because we never clean it up.

### DIFF
--- a/Code/cmake/Modules/RDKitUtils.cmake
+++ b/Code/cmake/Modules/RDKitUtils.cmake
@@ -55,13 +55,24 @@ macro(rdkit_library)
       add_library(${RDKLIB_NAME}_static ${RDKLIB_SOURCES})
 
       foreach(linkLib ${RDKLIB_LINK_LIBRARIES})
-        if(${linkLib} MATCHES "^(Boost)|(Thread)|(boost)|^(optimized)|^(debug)|(libz)")
-          set(rdk_static_link_libraries "${rdk_static_link_libraries}${linkLib};")
-        else()
-          set(rdk_static_link_libraries "${rdk_static_link_libraries}${linkLib}_static;")
+        if(TARGET "${linkLib}")
+          get_target_property(linkLib_IMPORTED "${linkLib}" IMPORTED)
+          if (linkLib_IMPORTED)
+            # linkLib is an imported target: use it as-is
+            target_link_libraries(${RDKLIB_NAME}_static PUBLIC "${linkLib}")
+            continue()
+          endif()
+        elseif(EXISTS "${linkLib}")
+          # linkLib is a file, so keep it as-is
+          target_link_libraries(${RDKLIB_NAME}_static PUBLIC "${linkLib}")
+          continue()
         endif()
+
+        # We haven't seen linkLib yet. This probably means it is a target
+        # we will be creating at some point (if not, then we are missing a find_package).
+        # Add the "_static" suffix to link against its static variant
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

PR #4466, opened 2021-09-01
Force import of wrapped exceptions

### LINKED ISSUE #4465: DataStruct vectors leak when iterating
PLEASE DELETE THIS SECTION AFTER YOU HAVE READ IT
If you have a question about the RDKit or aren't sure if what you're seeing is the right behavior please use the Discussions tab above (https://github.com/rdkit/rdkit/discussions) instead of posting it here.

Please help us help you: We normally need the information below in order to answer questions or understand bug reports. If you do not provide the information requested we may not be able to help you and will probably close the issue.

**Describe the bug**
The Vectors defined in DataStructs leak memory when using pythonic iterators.

This appears to be caused by a leak in the C++ IndexErrorException that isn't properly wrapped.
The solution appears to be a simple import of rdkit.rdBase which defines these properly

**To Reproduce**
```
from rdkit.DataStructs import SparseBitVect
# if you add the following there are no leaks
#from rdkit import rdBase
import random
s = SparseBitVect(2048)
bv = SparseBitVect(2048)
for i in range(100):
    s.SetBit(random.randint(0,2047))

while 1:
    l = list(s)
```

**Expected behavior**
No memory leaks

bp-kelley:
#### Reference Issue

Fixes #4465

When importing any rdkit module, first importing rdBase, this defines the proper exceptions and prevents them from leaking.

### DIFF
--- a/rdkit/__init__.py
+++ b/rdkit/__init__.py
@@ -1,3 +1,6 @@
+# Need to import rdBase to properly wrap exceptions                                                                                                         #  otherwise they will leak memory 
+from . import rdBase
+
 try:
   from .rdBase import rdkitVersion as __version__
 except ImportError:


bp-kelley:
@greglandrum there is very little chance that this caused the windows error on RGroupDecomp :)

greglandrum (approved):

PR #4497, opened 2021-09-10
Fixes #4496

### LINKED ISSUE #4496: OR queries involving aromatic atoms cannot be drawn
**Describe the bug**
A query molecule containing OR aromatic query atoms cannot be drawn.
The bug is caused by a small oversight in #3914

**To Reproduce**
```
from rdkit import Chem

d2d = rdMolDraw2D.MolDraw2DCairo(200, 200)
m = Chem.MolFromSmarts("[c,n]1[c,n][c,n][c,n][c,n][c,n]1")
m = rdMolDraw2D.PrepareMolForDrawing(m, kekulize=False, wedgeBonds=False)
d2d.DrawMolecule(m)
[17:52:22]

****
Pre-condition Violation
Atomic number not found
Violation occurred on line 97 in file /scratch/toscopa1/src/rdkit/Code/GraphMol/PeriodicTable.h
Failed Expression: atomicNumber < byanum.size()
****

  File "/home/toscopa1/py/query_atom_bug.py", line 8, in <module>
    d2d.DrawMolecule(m)
        Atomic number not found
        Violation occurred on line 97 in file Code/GraphMol/PeriodicTable.h
        Failed Expression: atomicNumber < byanum.size()
        RDKIT: 2021.09.1pre
        BOOST: 1_72
```

**Expected behavior**
It should be possible to draw such a molecule.

 - RDKit version: 2021.03

ptosco:
#### Reference Issue
Fixes #4496

This fixes a bug introduced by #3914. Aromatic query atoms get 1000 _added_ to their atomic number, not subtracted. So the check for aromatic OR queries in `getAtomListQueryVals` was never triggered, and anyway it would be applied in the wrong direction. This went unnoticed as there was no explicit test for aromatics; I have now added two.


ptosco:
I don’t think the failure in the Java build is related to this PR.

ricrogz:
> I don’t think the failure in the Java build is related to this PR.

Sorry, I think it is: you are updating the CMake file for the java wrappers. Your change allows CMake to use SWIG 4.0, which is currently not supported and produces the casting erros that caused the Java build to fail.

ptosco:
@ricrogz Ah, you are right: I had committed by mistake that file to my local `master`, so it wasn't showing up in my local `diff`. Good catch, thanks!

ricrogz:
No problem!

greglandrum (approved):
LGTM

Issue #4558, opened 2021-09-25
GetSubstructMatches() loops at 43690 iterations.


ricrogz:
[The attached solvent.sdf.gz file](https://github.com/rdkit/rdkit/files/7228907/solvent.sdf.gz) contains a large number of atoms (84525 atoms in 28175 water molecules with explicit Hydrogen atoms).

Running the following script should successfully match and return all the water molecules. `uniquify` is intentionally disabled to allow a quick execution of the query, so we expect the script to print that it found 56350 matches (since the query is symmetric, it can match the same molecule as H-O-H' and also as H'-O-H).
```
from rdkit import Chem
from itertools import groupby


def all_equal(iterable):
    g = groupby(iterable)
    return next(g, True) and not next(g, False)


mol = Chem.MolFromMolFile('solvent.sdf', sanitize=False, removeHs=False)
print(f'num atoms: {mol.GetNumAtoms()}')

q = Chem.MolFromSmarts('[H]O[H]')
max_sz = 500000
w = mol.GetSubstructMatches(q, uniquify=False, maxMatches=max_sz)

print(len(w))
```
On my computer, and with a Debug build, the script returns after 7 seconds, but it surprisingly prints:
```
500000
```
Since there's only 28175 water molecules, at least one of them must have been returned multiple (more than 2) times. Notice that we got the maximum number of matches we allowed. And even if we further increase this number, we always get returned the maximum number of matches, up to the point where the call becomes too slow to to wait for it to return, and I abort the experiment.

Digging a bit more, I tried to see if there was some kind of pattern in the molecules that were being returned. And there was. I added this to the script:
```
for i in range(20):  # Check the first 20 mols
    needle = w[i]
    idxs = [i]
    pos = i
    
    # Find the indexes of repeated molecules
    try:
        for _ in range(len(w)):
            pos = w.index(needle, pos + 1)
            idxs.append(pos)
    except ValueError:
        pass

    # Try to find a pattern in the repetitions
    deltas = [b - a for a, b in zip(idxs, idxs[1:])]
    equal = all_equal(deltas)
    print(f'indexes: {idxs}')
    print(f'deltas : {deltas}')
    print(f'all equal: {equal}')
    print('')
```

Which prints:
```
indexes: [0, 43690, 87380, 131070, 174760, 218450, 262140, 305830, 349520, 393210, 436900, 480590]
[...]

PR #4560, opened 2021-09-25
Improve performance of removing substruct/tautomer duplicates

ricrogz:
~This leverages `std::lower_bound()` and sorting to reduce the number of searches to determine duplicates, and `std::list` and `std::list::insert()`  to reduce the number of vector reallocations to improve performance. It also use `std::move()` to avoid copies, although this seems to have a smaller effect.~

The result of these changes is that the my test script for issue #4558 (56350 matches in total, half of them, 28175, are unique) went from taking ~4 minutes to uniquify results, all  the way down to 15 seconds.

I haven't tested the Tautomer one, I just copied over the changes from the one for Substructure matching.

### DIFF
--- a/Code/GraphMol/Substruct/SubstructUtils.cpp
+++ b/Code/GraphMol/Substruct/SubstructUtils.cpp
@@ -8,6 +8,7 @@
 //  of the RDKit source tree.
 //
 #include "SubstructUtils.h"
+#include <set>
 #include <RDGeneral/utils.h>
 #include <GraphMol/RDKitBase.h>
 #include <GraphMol/RDKitQueries.h>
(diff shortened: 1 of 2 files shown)


ricrogz:
Even better: using `std::set` makes it even faster: now my test runs in 1 second!

ricrogz:
Using `std::set` further improves performance: I can go up to 250k watermolecules (750k atoms in total) and run the search with `uniquify=True`. This takes about 90 s... and about 22Gb of memory. I can't go much higher than that before the script runs out of memory and gets killed.

`boost::unorered_set` further improves performance: the same test takes about 20 seconds, but, sadly, `boost::dynamic_bitsets` can only be hashed in the most recent versions of boost, which complicates the implementation quite a bit.

greglandrum (approved):
Wow... nice!

PR #4657, opened 2021-10-29
Fix some compile-time warnings in the postgres cartridge code

rvianello:
This pull request is intended to address a few warnings in the cartridge code, due to some unused variables, one missing inclusion directive and a few integer-to/from-pointer conversions in the fmcs code.

#### Any other comments?
In the process I added a test case exercising the overloaded fmcs aggregator for mol values.


greglandrum (approved):
LGTM

PR #4859, opened 2022-01-06
Clean-up Python 4815 - P1.1: Chem\AtomPairs

IchiruTake:
This is the first trial in breaking down the large PR 4815 #4815

### DIFF
--- a/rdkit/Chem/AtomPairs/Torsions.py
+++ b/rdkit/Chem/AtomPairs/Torsions.py
@@ -49,18 +49,18 @@ def pyScorePath(mol, path, size, atomCodes=None):
   1
 
   >>> m = Chem.MolFromSmiles('C=CC(=O)O')
-  >>> c1 = Utils.GetAtomCode(m.GetAtomWithIdx(0),1)
-  >>> c2 = Utils.GetAtomCode(m.GetAtomWithIdx(1),2)
-  >>> c3 = Utils.GetAtomCode(m.GetAtomWithIdx(2),2)
-  >>> c4 = Utils.GetAtomCode(m.GetAtomWithIdx(4),1)
+  >>> c1 = Utils.GetAtomCode(m.GetAtomWithIdx(0), 1)
+  >>> c2 = Utils.GetAtomCode(m.GetAtomWithIdx(1), 2)
+  >>> c3 = Utils.GetAtomCode(m.GetAtomWithIdx(2), 2)
+  >>> c4 = Utils.GetAtomCode(m.GetAtomWithIdx(4), 1)
   >>> t = c1 | (c2 << rdMolDescriptors.AtomPairsParameters.codeSize) | (c3 << (rdMolDescriptors.AtomPairsParameters.codeSize * 2)) | (c4 << (rdMolDescriptors.AtomPairsParameters.codeSize * 3))
   >>> pyScorePath(m, (0, 1, 2, 4), 4) == t
   1
 
-  """
+  """ 
   codes = [None] * size
   for i in range(size):
-    if i in (0, size - 1):
+    if i == 0 or i == size - 1:
       sub = 1
     else:
       sub = 2
@@ -99,7 +99,7 @@ def ExplainPathScore(score, size=4):
 
   Again, it's order independent:
 
-  >>> score=pyScorePath(m, (2, 1, 0), 3)
+  >>> score = pyScorePath(m, (2, 1, 0), 3)
   >>> ExplainPathScore(score, 3)
   (('C', 1, 0), ('C', 2, 1), ('C', 1, 1))
 
@@ -137,15 +137,16 @@ def ExplainPathScore(score, size=4):
   (('O', 1, 0), ('O', 2, 0), ('O', 2, 0), ('O', 1, 0))
 
   """
-  codeMask = (1 << rdMolDescriptors.AtomPairsParameters.codeSize) - 1
+  codeSize: int = rdMolDescriptors.AtomPairsParameters.codeSize
+  codeMask = (1 << codeSize) - 1
   res = [None] * size
   for i in range(size):
-    if i in (0, size - 1):
+    if i == 0 or i == size - 1:
       sub = 1
     else:
       sub = 2
     code = score & codeMask
-    score = score >> rdMolDescriptors.AtomPairsParameters.codeSize
+    score = score >> codeSize
     symb, nBranch, nPi = Utils.ExplainAtomCode(code)
     res[i] = (symb, nBranch + sub, nPi)
   return tuple(res)
--- a/rdkit/Chem/AtomPairs/Torsions.py
+++ b/rdkit/Chem/AtomPairs/Torsions.py
@@ -140,7 +140,10 @@ def ExplainPathScore(score, size=4):
   codeMask = (1 << rdMolDescriptors.AtomPairsParameters.codeSize) - 1
   res = [None] * size
   for i in range(size):
-    sub = 1 if i in (0, size - 1) else 2
+    if i in (0, size - 1):
+      sub = 1
+    else:
+      sub = 2
     code = score & codeMask
     score = score >> rdMolDescriptors.AtomPairsParameters.codeSize
     symb, nBranch, nPi = Utils.ExplainAtomCode(code)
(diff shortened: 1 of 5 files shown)


rdkit/Chem/AtomPairs/Torsions.py
```diff
@@ -54,33 +54,27 @@ def pyScorePath(mol, path, size, atomCodes=None):
...
 
   # "canonicalize" the code vector:
-  beg = 0
-  end = len(codes) - 1
-  while (beg < end):
-    if codes[beg] > codes[end]:
-      codes.reverse()
-      break
-    elif codes[beg] == codes[end]:
+  beg, end = 0, len(codes) - 1
```
gedeck, line 70: I'm not a fan of converting multiple assignments into a single one using tuple unpacking. 

My rationale for this is that it makes reading the code harder as it requires extra attention to associate the individual parts with each other. Looking at the old code, this is immediately clear. 

Code readability is key for a shared code base and I therefore discourage the use of this. @greglandrum, as the owner of the code base, your decision.

rdkit/Chem/AtomPairs/Torsions.py
```diff
@@ -54,33 +54,27 @@ def pyScorePath(mol, path, size, atomCodes=None):
...
+  while beg < end:
+    if codes[beg] == codes[end]:
       beg += 1
       end -= 1
     else:
+      if codes[beg] > codes[end]:
+        codes.reverse()
       break
   accum = 0
   for i in range(size):
```
gedeck, line 80: In the following loop
```
  accum = 0
  for i in range(size):
    accum |= codes[i] << (rdMolDescriptors.AtomPairsParameters.codeSize * i)
```
could use `enumerate` here
```
  accum = 0
  for i, code in enumerate(codes):
    accum |= code << (rdMolDescriptors.AtomPairsParameters.codeSize * i)
```
Also looks like `rdMolDescriptors.AtomPairsParameters.codeSize` is a constant, so could be moved out of the loop. This might be a micro-optimisation that isn't worth doing though, but would make it clearer what is being done here.
```
  accum = 0
  codeSize = rdMolDescriptors.AtomPairsParameters.codeSize
  for i, code in enumerate(codes):
    accum |= code << (codeSize * i)
```

rdkit/Chem/AtomPairs/Torsions.py
```diff
@@ -93,71 +87,75 @@ def ExplainPathScore(score, size=4):
...
   res = []
   for k, v in iv.GetNonzeroElements().items():
     res.extend([k] * v)
   res.sort()
+  """
+  nonZeroElements = GetTopologicalTorsionFingerprint(mol, targetSize).GetNonzeroElements()
+  resDict = sorted(nonZeroElements.items())
+  res = []
+  for k, v in resDict:
+    res.extend([k] * v)
```
gedeck, line 158: The name `resDict` gives the impression that we have a dictionary. It's however a list of tuples. You could call it `frequencies` instead.

rdkit/Chem/AtomPairs/Utils.py
```diff
@@ -153,7 +153,7 @@ def NumPiElectrons(atom):
         # unsaturations (valence - degree):
         res = atom.GetExplicitValence() - atom.GetNumExplicitHs()
         if res < atom.GetDegree():
-            raise ValueError("explicit valence exceeds atom degree")
+            raise ValueError("The explicit valence exceeds atom degree")
```
gedeck, line 156: The same error message is used in the C++ code (Code/GraphMol/Fingerprints/FingerprintUtil.cpp). 

Please keep messages unchanged. When working on a shared code base, it's better to adjust to its conventions

gedeck (changes requested):

rdkit/Chem/AtomPairs/Torsions.py
greglandrum: I agree with @gedeck.
I think assigning from a tuple directly is great, something like:
```
a, b = function_returning_a_tuple()
```
but the form you have here is difficult to read. I'd prefer:
```
beg = 0
end = len(codes) - 1
```

rdkit/Chem/AtomPairs/Torsions.py
```diff
@@ -54,33 +54,27 @@ def pyScorePath(mol, path, size, atomCodes=None):
...
   1
 
   """
   codes = [None] * size
   for i in range(size):
-    if i == 0 or i == (size - 1):
-      sub = 1
-    else:
-      sub = 2
+    sub = 1 if i in (0, size - 1) else 2
```
greglandrum, line 63: I personally find python's ternary syntax terrible, so I prefer the original form of this (a lot)
But since @gedeck didn't complain, I'm not going to insist.

Why do you think that this is better than what was previously there?

rdkit/Chem/AtomPairs/Torsions.py
```diff
@@ -93,71 +87,75 @@ def ExplainPathScore(score, size=4):
...
 
   """
   codeMask = (1 << rdMolDescriptors.AtomPairsParameters.codeSize) - 1
   res = [None] * size
   for i in range(size):
-    if i == 0 or i == (size - 1):
-      sub = 1
-    else:
-      sub = 2
+    sub = 1 if i in (0, size - 1) else 2
```
greglandrum, line 137: Same comment about the ternary operator. I would prefer that we leave the code as it was unless there's a reason for the change

greglandrum (changes requested):
I agree with @gedeck 's suggestions and would also request that the use of the ternary operator be removed unless there's a good reason for it.

greglandrum:
@IchiruTake I realized that I didn't really explain my comments about the ternary operator here or in the other PR and that I should do that since this is your first time contributing. I'm really happy that you're interested in contributing and want to try and be sure you understand why some of the feedback on the PR was given. I will try to do that here.

Changing this if/else code:
```
if a:
  foo = 1
else:
  foo = 2
```
to this form (using the ternary operator):
```
foo = 1 if a else 2
```
is a stylistic change. It doesn't affect performance or correctness, it's just coding style. Since there's no objective "right" answer to what the best coding style is, it ends up differing from person to person. Many projects will try and maintain a consistent style for the code in their project. It's not always easy, and we're not always successful, but we do try to do that with the RDKit.

I would prefer not to make this particular stylistic change in the RDKit because I find it less readable, and readability is really important for people who are coming to the code later in order to understand what it's doing or fix bugs.

Does that help explain the comments?

rdkit/Chem/AtomPairs/Torsions.py
```diff
@@ -93,71 +93,70 @@ def ExplainPathScore(score, size=4):
...
+  >>> score=pyScorePath(m, (0, 1, 2, 3), 4)
+  >>> ExplainPathScore(score, 4)
   (('O', 1, 0), ('O', 2, 0), ('O', 2, 0), ('O', 1, 0))
 
   """
   codeMask = (1 << rdMolDescriptors.AtomPairsParameters.codeSize) - 1
   res = [None] * size
   for i in range(size):
-    if i == 0 or i == (size - 1):
+    if i in (0, size - 1):
```
gedeck, line 143: I'm not sure if I find this easier to read. Intuitively, I look at this and interpret it as an interval.

gedeck (approved):

gedeck:
When you are ready, it's best to let us know by re-requesting review again. It then shows up in our task list. 

image

rdkit/Chem/AtomPairs/Torsions.py
greglandrum: Yeah, I also don't find this form easier to read.

Not that it's relevant here, but as an FYI this form is also considerably slower:
```
In [1]: def f1(size,nreps):
   ...:     for i in range(nreps):
   ...:         accum = 0
   ...:         for j in range(size):
   ...:             if j in (0,size-1):
   ...:                 accum += j
   ...: 

In [2]: def f2(size,nreps):
   ...:     for i in range(nreps):
   ...:         accum = 0
   ...:         for j in range(size):
   ...:             if j ==0 or j == size-1:
   ...:                 accum += j
   ...: 
   ...: 

In [3]: %timeit f1(100,1000)
4.86 ms ± 41.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit f2(100,1000)
3.72 ms ± 11.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

```

greglandrum: @IchiruTake : do you have a reason for preferring the `in` form here instead of using the two equality checks?

IchiruTake:
I think it may come from the optimization came from Python as mentioned in
the PR 4815. Similar behaviour for operator `and`

Vào Th 3, 18 thg 1, 2022 vào lúc 20:33 Greg Landrum <
***@***.***> đã viết:

greglandrum (approved):

PR #4867, opened 2022-01-08
Fix python reentry in filtercatalog threaded runs

bp-kelley:
This makes a change to allow thread initializers to be called when the filter catalog threaded runner spawns threads.

In C++ land, this does nothing.

In python land, this allows us to have a PyGILStateHolder initialized to allow reentry into pure python code.

### DIFF
--- a/Code/GraphMol/FilterCatalog/ThreadedFilterCatalogRunner.h
+++ b/Code/GraphMol/FilterCatalog/ThreadedFilterCatalogRunner.h
@@ -1,0 +1,81 @@
+//  Copyright (c) 2022 Brian P Kelley
+//  All rights reserved.
+//
+//  This file is part of the RDKit.
+//  The contents are covered by the terms of the BSD license
+//  which is included in the file license.txt, found at the root
+//  of the RDKit source tree.
+//
+
+#ifndef THREADED_FILTERCATALOG_RUNNER
+#define THREADED_FILTERCATALOG_RUNNER
+#include "FilterCatalog.h"
+#include "Filters.h"
+#include "FilterMatchers.h"
+#include <GraphMol/SmilesParse/SmilesParse.h>
+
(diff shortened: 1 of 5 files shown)


Code/GraphMol/FilterCatalog/ThreadedFilterCatalogRunner.h
```diff
@@ -0,0 +1,81 @@
...
+      new FilterCatalogEntry("no valid RDKit molecule",
+                             boost::shared_ptr<FilterMatcherBase>()));
+  return bad_smiles;
+}
+template<class ThreadInitializer>
+void CatalogSearcher(
+    const FilterCatalog &fc, const std::vector<std::string> &smiles,
+    std::vector<std::vector<FilterCatalog::CONST_SENTRY>> &results, int start,
+    int numThreads) {
+  ThreadInitializer t;
```
bp-kelley, line 35: This is the real change here which required moving to a header file from the .cpp file

Code/GraphMol/FilterCatalog/Wrap/rough_test.py
```diff
@@ -543,6 +543,41 @@ def testThreadedRunner(self):
...
+      def __init__(self, minMw, maxMw):
+        FilterCatalog.FilterMatcher.__init__(self, "MW violation")
+        self.minMw = minMw
+        self.maxMw = maxMw
+
+      def IsValid(self):
+        return True
+
+      def HasMatch(self, mol):
+        mw = rdMolDescriptors.CalcExactMolWt(mol)
```
bp-kelley, line 559: The filter catalog runner was hanging on any call to the subclassed MWFilter.

This also checks the logging if python logging is enabled.

bp-kelley:
Found a better solution...

PR #4891, opened 2022-01-17
Clean-up Python #4815 - P1.3: Chem\Draw

IchiruTake:

---

Updated Changes:
- [New]: Clean-up unused import in IPythonConsole.py
- [New]: Clean-up unused import in UnitTestSimilarityMaps.py
- [Q&A]: In line 434, `symbolLength += 1 + len(str(nHs))` confused me as the number of hydrogen for every atoms is limited at most to one digit only, usually from 0 to 4 (probably at most 7). The number of hydrogen which is larger than 10 simply needs a further checking, I guess. So, the 'symbolLength' variable should be incremented by 2 rather than 1 + len(str(nHs)).

### DIFF
--- a/rdkit/Chem/Draw/IPythonConsole.py
+++ b/rdkit/Chem/Draw/IPythonConsole.py
@@ -7,21 +7,16 @@
 #  which is included in the file license.txt, found at the root
 #  of the RDKit source tree.
 #
-from IPython.display import SVG
-import numpy
-import warnings
-import uuid
-import json
-import os
 import base64
 import copy
-from io import BytesIO, StringIO
-from rdkit.Chem.Draw import rdMolDraw2D
-from rdkit.Chem import Draw
-from rdkit.Chem import rdchem, rdChemReactions
-from rdkit import Chem
-import sys
+import warnings
+from io import BytesIO
+
 import IPython
+from IPython.display import SVG
+from rdkit import Chem
+from rdkit.Chem import Draw, rdchem, rdChemReactions
+from rdkit.Chem.Draw import rdMolDraw2D
 
 if IPython.release.version < '0.11':
   raise ImportError('this module requires at least v0.11 of IPython')
@@ -75,15 +70,13 @@ def drawMol3D(m, view=None, confId=-1, drawAs=None, bgColor=None, size=None):
   if view is None:
     view = py3Dmol.view(width=size[0], height=size[1])
   view.removeAllModels()
+  
   try:
-    iter(m)
-  except TypeError:
-    addMolToView(m, view, confId, drawAs)
-  else:
-    ms = m
+    ms = iter(m)
     for m in ms:
       addMolToView(m, view, confId, drawAs)
-
+  except TypeError:
+    addMolToView(m, view, confId, drawAs)
   view.setBackgroundColor(bgColor)
   view.zoomTo()
   return view.show()
@@ -112,6 +105,7 @@ def _toHTML(mol):
     nm = mol.GetProp('_Name')
   else:
     nm = ''
+    
   res = []
   if not ipython_useSVG:
     png = Draw._moltoimg(mol, molSize, [], nm, returnPNG=True, drawOptions=drawOptions)
@@ -122,7 +116,7 @@ def _toHTML(mol):
     res.append(f'<tr><td colspan=2 style="text-align:center">{svg}</td></tr>')
 
   for i,(pn, pv) in enumerate(props.items()):
-    if ipython_maxProperties>=0 and i>= ipython_maxProperties:
+    if ipython_maxProperties >= 0 and i >= ipython_maxProperties:
       res.append('<tr><td colspan=2 style="text-align:center">Property list truncated.<br />Increase IPythonConsole.ipython_maxProperties (or set it to -1) to see more properties.</td></tr>')
       break
     res.append(
@@ -172,7 +166,6 @@ def _toMolBundlePNG(bundle):
     fn = _MolsToGridImageSaved
   else:
     fn = Draw.MolsToGridImage
-
   return fn(bundle, subImgSize=molSize, drawOptions=drawOptions, useSVG=False, returnPNG=True)
 
 
@@ -237,6 +230,7 @@ def ShowMols(mols, maxMols=50, **kwargs):
     fn = _MolsToGridImageSaved
   else:
     fn = Draw.MolsToGridImage
+    
   if len(mols) > maxMols:
     warnings.warn(
       "Truncating the list of molecules to be displayed to %d. Change the maxMols value to display more."
@@ -247,14 +241,13 @@ def ShowMols(mols, maxMols=50, **kwargs):
         kwargs[prop] = kwargs[prop][:maxMols]
   if not "drawOptions" in kwargs:
     kwargs["drawOptions"] = drawOptions
+  
   res = fn(mols, **kwargs)
   if kwargs['useSVG']:
     return SVG(res)
-  else:
-    if kwargs['returnPNG']:
-      return display.Image(data=res, format='png')
-    else:
-      return res
+  if kwargs['returnPNG']:
+    return display.Image(data=res, format='png')
+  return res
 
 
 ShowMols.__doc__ = Draw.MolsToGridImage.__doc__
@@ -263,23 +256,23 @@ ShowMols.__doc__ = Draw.MolsToGridImage.__doc__
 def _DrawBit(fn, *args, **kwargs):
   if 'useSVG' not in kwargs:
     kwargs['useSVG'] = ipython_useSVG
+  
   res = fn(*args, **kwargs)
   if kwargs['useSVG']:
     return SVG(res)
-  else:
-    sio = BytesIO(res)
-    return Image.open(sio)
+  sio = BytesIO(res)
+  return Image.open(sio)
 
 
 def _DrawBits(fn, *args, **kwargs):
   if 'useSVG' not in kwargs:
     kwargs['useSVG'] = ipython_useSVG
+
   res = fn(*args, **kwargs)
   if kwargs['useSVG']:
     return SVG(res)
-  else:
-    sio = BytesIO(res)
-    return Image.open(sio)
+  sio = BytesIO(res)
+  return Image.open(sio)
 
 
 _DrawMorganBitSaved = None
--- a/rdkit/Chem/Draw/MolDrawing.py
+++ b/rdkit/Chem/Draw/MolDrawing.py
@@ -425,16 +425,15 @@ class MolDrawing(object):
           symbolLength = len(base)
           
           nHs = 0
-          hs = ''
           if not atom.HasQuery():
             nHs = atom.GetTotalNumHs()
-          if nHs > 0:
-            if nHs > 1:
-              hs = 'H<sub>%d</sub>' % nHs
-              symbolLength += 1 + len(str(nHs))
-            else:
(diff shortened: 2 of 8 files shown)


rdkit/Chem/Draw/IPythonConsole.py
```diff
@@ -75,15 +70,13 @@ def drawMol3D(m, view=None, confId=-1, drawAs=None, bgColor=None, size=None):
...
-  except TypeError:
-    addMolToView(m, view, confId, drawAs)
-  else:
-    ms = m
+    ms = iter(m)
     for m in ms:
       addMolToView(m, view, confId, drawAs)
-
+  except TypeError:
+    addMolToView(m, view, confId, drawAs)
```
gedeck, line 79: I wonder if this is a Python 2 left over. We should be able to write this now as:
```
from collections.abc import Iterable

if isinstance(m, Iterable):
  for mol in m:
    addMolToView(mol, view, confId, drawAs)
else:
  addMolToView(m, view, confId, drawAs)
```
or 
```
if not isinstance(m, Iterable):
  m = [m]
for mol in m:
  addMolToView(mol, view, confId, drawAs)
```
I checked that a Mol object is not an iterable.

rdkit/Chem/Draw/MolDrawing.py
```diff
@@ -416,27 +416,26 @@ def AddMol(self, mol, centerIt=True, molTrans=None, drawingTrans=None, highlight
...
           if nHs > 0:
             if nHs > 1:
               hs = 'H<sub>%d</sub>' % nHs
               symbolLength += 1 + len(str(nHs))
             else:
               hs = 'H'
               symbolLength += 1
-          else:
-            hs = ''
+
```
gedeck, line 438: You can convert the `if-if-else` block into a `if-elif` block and make the block clearer
```
          nHs = 0
          if not atom.HasQuery():
            nHs = atom.GetTotalNumHs()
          hs = ''
          if nHs > 1:
            hs = 'H<sub>%d</sub>' % nHs
            symbolLength += 1 + len(str(nHs))
          elif nHs == 1:
            hs = 'H'
            symbolLength += 1
```
I moved the hs assignment before the `if-elif` block as it belongs to it. First part deals with `nHs` the second part with `hs`.

rdkit/Chem/Draw/SimilarityMaps.py
```diff
@@ -53,6 +53,10 @@
 from rdkit.Chem import rdDepictor
 from rdkit.Chem import rdMolDescriptors as rdMD
 
+def _CleanFpInfoAttr_(mol):
+  if hasattr(mol, '_fpInfo'):
+    delattr(mol, '_fpInfo')
```
gedeck, line 58: should probably be called `_DeleteFpInfoAttr` because it removes it. Cleaning gives the impression it is still there.

I would also move this to the end of the function `GetAtomicWeightsForFingerprint`.

rdkit/Chem/Draw/SimilarityMaps.py
```diff
@@ -98,18 +97,14 @@ def GetAtomicWeightsForModel(probeMol, fpFunction, predictionFunction):
...
+
   # loop over atoms
-  weights = []
-  for atomId in range(probeMol.GetNumAtoms()):
-    newFP = fpFunction(probeMol, atomId)
-    newProba = predictionFunction(newFP)
-    weights.append(baseProba - newProba)
-  if hasattr(probeMol, '_fpInfo'):
-    delattr(probeMol, '_fpInfo')
+  weights = [baseProba - predictionFunction(fpFunction(probeMol, atomId)) for atomId in range(probeMol.GetNumAtoms())]
```
gedeck, line 105: I prefer writing long list comprehensions like this over two lines
```
weights = [baseProba - predictionFunction(fpFunction(probeMol, atomId)) 
           for atomId in range(probeMol.GetNumAtoms())]
```

rdkit/Chem/Draw/SimilarityMaps.py
```diff
@@ -121,12 +116,12 @@ def GetStandardizedWeights(weights):
...
       weights -- the list with the atomic weights
     """
-  tmp = [math.fabs(w) for w in weights]
-  currentMax = max(tmp)
-  if currentMax > 0:
-    return [w / currentMax for w in weights], currentMax
-  else:
-    return weights, currentMax
+
+  maxAbsWeight = max([math.fabs(w) for w in weights])
```
gedeck, line 120: You don't need the square brackets here. 
```
maxAbsWeight = max(math.fabs(w) for w in weights)
```

rdkit/Chem/Draw/aggCanvas.py
```diff
@@ -20,8 +20,7 @@
 
 
 def convertColor(color):
-  color = (int(color[0] * 255), int(color[1] * 255), int(color[2] * 255))
-  return color
+  return int(color[0] * 255), int(color[1] * 255), int(color[2] * 255)
```
gedeck, line 23: Add the round brackets to make it clearer that a tuple is returned.

rdkit/Chem/Draw/aggCanvas.py
```diff
@@ -137,18 +136,17 @@ def addCanvasText(self, text, pos, font, color=(0, 0, 0), **kwargs):
...
         dPos = [pos[0] + offset, pos[1] - h / 2.]
       else:
-        dPos = [pos[0] - w / 2 + offset, pos[1] - h / 2.]
+        dPos = [pos[0] - w / 2. + offset, pos[1] - h / 2.]
 
       if supH:
         dPos[1] += supH
       for txt, fmt, tw, th in dblocks:
-        tPos = dPos[:]
+        tPos = dPos.copy()
```
gedeck, line 149: Keep the old slice. If that worked before it means that a copy is not required.

gedeck (changes requested):

rdkit/Chem/Draw/aggCanvas.py
IchiruTake: `dPos[:]` means you are making a copy, so these you are completely identical. In new Python , list.copy() is faster
https://stackoverflow.com/questions/2612802/list-changes-unexpectedly-after-assignment-why-is-this-and-how-can-i-prevent-it
https://www.python.org/dev/peps/pep-0590/

rdkit/Chem/Draw/SimilarityMaps.py
IchiruTake: Agree ! I would fix it in the next patch

rdkit/Chem/Draw/IPythonConsole.py
IchiruTake: I don't agree with it. The main reason is that why we have to import some new modules to check whether it is iterable or not, or add to a list and push it to the function. I think that you can make another PR to fix it

rdkit/Chem/Draw/MolDrawing.py
IchiruTake: Agree ! I would fix it in the next patch

rdkit/Chem/Draw/aggCanvas.py
gedeck: Ok - good to know.

rdkit/Chem/Draw/SimilarityMaps.py
IchiruTake: If that so, there is a tons of places use this style. But I simply do not have time to look over them

gedeck: That's ok. Just don't do that in your changes.

rdkit/Chem/Draw/aggCanvas.py
IchiruTake: I don't agree with it. 
The main reason came from the input as we already defined it as RGB color with three inputs.
Secondly, this function will unlikely to be changed in the future. Looking from the code, the input is the normalized color which have been divided by 255 previously. Python simply understand this type of return is always a tuple regardless of output

gedeck: I'm not talking about Python, I'm talking about someone looking at the code. People are reading code.

rdkit/Chem/Draw/SimilarityMaps.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thanks you for your review

rdkit/Chem/Draw/aggCanvas.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thanks you for your review

rdkit/Chem/Draw/SimilarityMaps.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. I would change the name as requested. Thanks you for your review.

rdkit/Chem/Draw/IPythonConsole.py
IchiruTake: I am so sorry about that, but this would be unchanged. Thanks you for your review

gedeck: ok

rdkit/Chem/Draw/MolDrawing.py
```diff
@@ -416,27 +416,25 @@ def AddMol(self, mol, centerIt=True, molTrans=None, drawingTrans=None, highlight
           symbolLength = len(symbol)
         else:
           base = atom.GetSymbol()
-          if (base == 'H' and (iso == 2 or iso == 3) and
-              self.drawingOptions.atomLabelDeuteriumTritium):
+          if base == 'H' and (iso == 2 or iso == 3) and self.drawingOptions.atomLabelDeuteriumTritium:
```
gedeck, line 419: I comment on this just for the record. Long lines are ok if you have a large monitor and I also aim to avoid breaking if statements to two lines. However, in a collaborative code base, it's better to write code that doesn't get too wide. For example, I tend to work most of the time just on my laptop. 

Anyway, it's better to consider one of the automated formatting tools to make the code more consistent. I've used yapf on this code base a long time ago, but we didn't implement an automated process. This has nothing to do with this PR though.

gedeck (approved):

greglandrum (changes requested):
a few small changes requested

rdkit/Chem/Draw/aggCanvas.py
```diff
@@ -137,18 +134,17 @@ def addCanvasText(self, text, pos, font, color=(0, 0, 0), **kwargs):
...
       h += supH + subH
       offset = w * pos[2]
-      dPos = [pos[0] - w / 2. + offset, pos[1] - h / 2.]
       if orientation == 'W':
         dPos = [pos[0] - w + offset, pos[1] - h / 2.]
       elif orientation == 'E':
         dPos = [pos[0] + offset, pos[1] - h / 2.]
       else:
-        dPos = [pos[0] - w / 2 + offset, pos[1] - h / 2.]
+        dPos = [pos[0] - w / 2. + offset, pos[1] - h / 2.]
```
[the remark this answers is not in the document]
IchiruTake: Thank you for your review. In this change, I have felt that this is trivial but to increase consistency, I am happy with this change.

IchiruTake:
Thank you for your review. I have changed as requested.

greglandrum (approved):
LGTM

PR #4893, opened 2022-01-18
Clean-up Python #4815 - P1.4: Chem\EState

IchiruTake:

-----------------------------------
Updated Changes:
- [New]: Update function `EStateIndicies` in Estate.py

### DIFF
--- a/rdkit/Chem/EState/AtomTypes.py
+++ b/rdkit/Chem/EState/AtomTypes.py
@@ -12,6 +12,7 @@
 
  defined in: Hall and Kier JCICS _35_ 1039-1045 (1995)  Table 1
 """
+import sys
 from rdkit import Chem
 
 _rawD = [
@@ -136,9 +137,10 @@ def TypeAtoms(mol):
         res[idx] = [name]
       elif name not in res[idx]:
         res[idx].append(name)
+  
   for i, v in enumerate(res):
     if v is not None:
       res[i] = tuple(v)
     else:
-      res[i] = ()
+      res[i] = tuple()
   return res
--- a/rdkit/Chem/EState/EState.py
+++ b/rdkit/Chem/EState/EState.py
@@ -44,7 +44,7 @@ def EStateIndices(mol, force=True):
 
   tbl = Chem.GetPeriodicTable()
   nAtoms = mol.GetNumAtoms()
-  Is = [0.0] * nAtoms
+  Is = numpy.zeros(nAtoms, dtype=numpy.float64)
   for i in range(nAtoms):
     at = mol.GetAtomWithIdx(i)
     d = at.GetDegree()
@@ -55,7 +55,7 @@ def EStateIndices(mol, force=True):
       Is[i] = (4. / (N * N) * dv + 1) / d
   dists = Chem.GetDistanceMatrix(mol, useBO=0, useAtomWts=0) + 1
   
-  accum = [0.0] * nAtoms
+  accum = numpy.zeros(nAtoms, dtype=numpy.float64)
   for i in range(nAtoms):
     for j in range(i + 1, nAtoms):
       p = dists[i, j]
@@ -63,7 +63,8 @@ def EStateIndices(mol, force=True):
         tmp = (Is[i] - Is[j]) / (p * p)
         accum[i] += tmp
         accum[j] -= tmp
-  res = numpy.add(accum, Is, dtype='float')
+  
+  res = accum + Is
   mol._eStateIndices = res
   return res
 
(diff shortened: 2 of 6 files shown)


rdkit/Chem/EState/AtomTypes.py
```diff
@@ -12,6 +12,7 @@
 
  defined in: Hall and Kier JCICS _35_ 1039-1045 (1995)  Table 1
 """
+import sys
```
gedeck, line 15: I don't think this is required.

rdkit/Chem/EState/EState.py
```diff
@@ -45,28 +44,26 @@ def EStateIndices(mol, force=True):
...
   for i in range(nAtoms):
     for j in range(i + 1, nAtoms):
       p = dists[i, j]
       if p < 1e6:
         tmp = (Is[i] - Is[j]) / (p * p)
         accum[i] += tmp
         accum[j] -= tmp
-
-  res = accum + Is
+  res = numpy.add(accum, Is, dtype='float')
```
gedeck, line 66: This will convert the `accum` and `Is` lists into numpy arrays. It therefore is better to leave the original definition of these arrays using `numpy.zeros` . You can then leave the old statement `res = accum + Is`.

rdkit/Chem/EState/EState_VSA.py
```diff
@@ -41,7 +41,7 @@ def VSA_EState_(mol, bins=None, force=1):
   propContribs = EStateIndices_(mol, force=force)
   volContribs = VSAContribs_(mol)
 
-  ans = numpy.zeros(len(bins) + 1, numpy.float)
+  ans = numpy.zeros(len(bins) + 1, dtype='float')
```
gedeck, line 44:
```
ans = numpy.zeros(len(bins) + 1, dtype=numpy.float)
```
No need to add the burden of interpreting the string.

rdkit/Chem/EState/Fingerprinter.py
```diff
@@ -35,15 +35,14 @@ def FingerprintMol(mol):
...
-
+  counts = [0] * nPatts
+  sums = [0.0] * nPatts
   for i, (_, pattern) in enumerate(AtomTypes.esPatterns):
     matches = mol.GetSubstructMatches(pattern, uniquify=1)
     counts[i] = len(matches)
     for match in matches:
       sums[i] += esIndices[match[0]]
-  return counts, sums
+  return numpy.array(counts, dtype='int'), numpy.array(sums, dtype='float')
```
gedeck, line 45: Same argument about using `numpy.zeros` as above.

gedeck (changes requested):

rdkit/Chem/EState/EState_VSA.py
IchiruTake: I don't agree with it. I changed it intentionally and got accepted in PR 4815. 
In particularly, this is a fix of `DeprecationWarning` in numpy 1.20 
Reference: https://numpy.org/devdocs/release/1.20.0-notes.html

rdkit/Chem/EState/Fingerprinter.py
IchiruTake: > Same argument about using `numpy.zeros` as above.
As mentioned above: 
I don't agree with it. I changed it intentionally and got accepted in PR 4815.
In particularly, this is a fix of DeprecationWarning in numpy 1.20
Reference: https://numpy.org/devdocs/release/1.20.0-notes.html

rdkit/Chem/EState/EState_VSA.py
gedeck: Ok, What about this?
```
ans = numpy.zeros(len(bins) + 1, dtype=float)
```

rdkit/Chem/EState/EState.py
IchiruTake: I don't agree with it. 
Yes, the main reason I make it here due to the element-wise access to `np.array` which is not recommended in numpy.

gedeck: If np.add is more efficient and recommended, then use it. However, do you think that using numpy arrays for `accum` and `Is` in the code above is more efficient?

IchiruTake: I am so sorry about that, but to your question, at least it is. At current, I did not make any strong optimization, 
If you want, I would check it in the near future.

rdkit/Chem/EState/EState_VSA.py
IchiruTake: > Ok, What about this?
> 
> ```
> ans = numpy.zeros(len(bins) + 1, dtype=float)
> ```

I am aware of that change but at the very least, I would prefer numpy.float64. 
According to PR 4815, he prefers a general solution to handle it.

rdkit/Chem/EState/Fingerprinter.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thanks you for your review.

rdkit/Chem/EState/EState.py
IchiruTake: I believe a greater option is to make heavy optimization.
I can merge `accum` and `Is` at once. 
1) Initialize `res` as a python list
2) Calculate
3) Switch back to numpy.
Does it sound good?
@gedeck

greglandrum (approved):

PR #4896, opened 2022-01-18
Clean-up Python #4815 - P1.6: Chem\Features

IchiruTake:

------------------------------
Updated Changes:
- [Q&A]: Documentation in FeatDirUtilsRD: Two docs are similar between `GetDonor3FeatVects()` and `GetAcceptor3FeatVect()`
- [Remark]: In `ShowFeats.py` we groups similar operation into one by its condition. And remove one unused variable named `nm` in line 178

### DIFF
--- a/rdkit/Chem/Features/FeatDirUtilsRD.py
+++ b/rdkit/Chem/Features/FeatDirUtilsRD.py
@@ -21,9 +21,10 @@ import math
 
 
 def cross(v1, v2):
-  res = numpy.array([v1[1] * v2[2] - v1[2] * v2[1], -v1[0] * v2[2] + v1[2] * v2[0],
-                     v1[0] * v2[1] - v1[1] * v2[0]], numpy.double)
-  return res
+  return numpy.array([v1[1] * v2[2] - v1[2] * v2[1], 
+                      -v1[0] * v2[2] + v1[2] * v2[0],
+                      v1[0] * v2[1] - v1[1] * v2[0]], 
+                     dtype=numpy.float64)
 
 
 def findNeighbors(atomId, adjMat):
@@ -34,16 +35,11 @@ def findNeighbors(atomId, adjMat):
   atomId - atom of interest
   adjMat - adjacency matrix for the compound
   """
-  nbrs = []
-  for i, nid in enumerate(adjMat[atomId]):
-    if nid >= 1:
-      nbrs.append(i)
-  return nbrs
+  return [i for i, nid in enumerate(adjMat[atomId]) if nid >= 1]
 
 
 def _findAvgVec(conf, center, nbrs):
-  # find the average of the normalized vectors going from the center atoms to the
-  # neighbors
+  # find the average of the normalized vectors going from the center atoms to the neighbors
   # the average vector is also normalized
   avgVec = 0
   for nbr in nbrs:
@@ -51,7 +47,7 @@ def _findAvgVec(conf, center, nbrs):
     pt = conf.GetAtomPosition(nid)
     pt -= center
     pt.Normalize()
-    if (avgVec == 0):
+    if avgVec == 0:
       avgVec = pt
     else:
       avgVec += pt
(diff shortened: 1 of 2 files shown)


rdkit/Chem/Features/FeatDirUtilsRD.py
```diff
@@ -320,6 +295,8 @@ def GetDonor2FeatVects(conf, featAtoms, scale=1.5):
       bvec += cpt
       res.append((cpt, bvec))
     return tuple(res), 'linear'
+  
+  return None
```
gedeck, line 299: This looks like something pylint is complaining about. Did you use pylint to check the code?

rdkit/Chem/Features/ShowFeats.py
```diff
@@ -194,33 +203,24 @@ def ShowMolFeats(mol, factory, viewer, radius=0.5, confId=-1, showOnly=True, nam
...
           featAtom = mol.GetAtomWithIdx(aids[0])
           hvyNbrs = [x for x in featAtom.GetNeighbors() if x.GetAtomicNum() != 1]
-          if len(hvyNbrs) == 1:
-            ps, fType = FeatDirUtils.GetDonor1FeatVects(mol.GetConformer(confId), aids, scale=1.0)
-          elif len(hvyNbrs) == 2:
-            ps, fType = FeatDirUtils.GetDonor2FeatVects(mol.GetConformer(confId), aids, scale=1.0)
-          elif len(hvyNbrs) == 3:
-            ps, fType = FeatDirUtils.GetDonor3FeatVects(mol.GetConformer(confId), aids, scale=1.0)
+          ps, fType = FeatVectsDictMethod[(family, len(hvyNbrs))](
+            mol.GetConformer(confId), aids, scale=1.0)
```
gedeck, line 215: I suggest to make this two statements to make it easier to read.

```
method = FeatVect...
ps, fType = method(mol.Get...)
```

rdkit/Chem/Features/ShowFeats.py
```diff
@@ -169,13 +169,22 @@ def ShowMolFeats(mol, factory, viewer, radius=0.5, confId=-1, showOnly=True, nam
...
-  for i, feat in enumerate(molFeats):
+  FeatVectsDictMethod = \
+    {('Aromatic', None): FeatDirUtils.GetAromaticFeatVects, 
+     ('Donor', 1): FeatDirUtils.GetDonor1FeatVects,
+     ('Donor', 2): FeatDirUtils.GetDonor2FeatVects,
+     ('Donor', 3): FeatDirUtils.GetDonor3FeatVects,
+     ('Acceptor', 1): FeatDirUtils.GetDonor1FeatVects,
+     ('Acceptor', 2): FeatDirUtils.GetDonor2FeatVects,
+     ('Acceptor', 3): FeatDirUtils.GetDonor3FeatVects,
+     } # TODO: add more feat types if needed. Error should not be thrown if a type is not found due to the lack of method.
```
gedeck, line 180: I can see why you do this and it reduces the code for the donors and acceptors. On the one hand that is nice, on the other hand however, we loose the clarity of which functions are called when the dictionary is used to select the method. A alternative would be to keep the original call for the aromatic feature and use two dictionaries one for the donor and one for the acceptor feature. These can be defined in the respective if block.

gedeck (commented):
A few comments for discussion.

rdkit/Chem/Features/FeatDirUtilsRD.py
IchiruTake: Unfortunately, I did not use pylint to check, but since I saw couple of empty return (`return `) and probably unclear result to handle. This is just to make thing clear.

rdkit/Chem/Features/ShowFeats.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thank you for your review.

rdkit/Chem/Features/ShowFeats.py
IchiruTake: Agree ! I would fix it in the next patch. Thank you for your review.

gedeck (approved):

PR #4898, opened 2022-01-18
Clean-up Python #4815 - P1.8: Chem\fmcs

IchiruTake:
- [Line 689]: I expect a better compatibility from Python between `dict` and `Counter`  or a better solution

### DIFF
--- a/rdkit/Chem/fmcs/fmcs.py
+++ b/rdkit/Chem/fmcs/fmcs.py
@@ -673,9 +673,6 @@ def _check_atom_classes(molno, num_atoms, atom_classes):
 # prune. But so far I don't have a test set which drives the need for
 # that.
 
-# UPDATED by Ichiru Take: Use collections.Counter to count the bond types, which is much faster.
-# See here: https://stackoverflow.com/questions/44012479/intersection-of-two-counters
-
 # Return a dictionary mapping iterator item to occurrence count
 def get_counts(it):
   return dict(Counter(it))
@@ -735,7 +732,8 @@ def remove_unknown_bondtypes(typed_mol, supported_canonical_bondtypes):
 
 
 def find_upper_fragment_size_limits(rdmol, atoms):
-  max_num_atoms, max_twice_num_bonds = 0, 0
+  max_num_atoms = 0
+  max_twice_num_bonds = 0
   for atom_indices in Chem.GetMolFrags(rdmol):
     max_num_atoms = max(max_num_atoms, len(atom_indices))
 
@@ -745,8 +743,8 @@ def find_upper_fragment_size_limits(rdmol, atoms):
     for atom_index in atom_indices:
       # XXX Why is there no 'atom.GetNumBonds()'?
       # Ichiru Take: len(atoms[atom_index].GetBonds()) would be more efficient but I don't know the input type.
-      twice_num_bonds += sum(1 for bond in atoms[atom_index].GetBonds()) 
-
+      twice_num_bonds += len(atoms[atom_index].GetBonds())
+       
     max_twice_num_bonds = max(max_twice_num_bonds, twice_num_bonds)
 
   return max_num_atoms, max_twice_num_bonds // 2
@@ -1164,7 +1162,7 @@ def canon(cangen_nodes):
 def get_closure_label(bond_smarts, closure):
   if closure < 10:
     return bond_smarts + str(closure)
-  return bond_smarts + "%%%02d" % closure
+  return bond_smarts + f"%{closure:02d}"
 
 # Precompute the initial closure heap. *Overall* performance went from 0.73 to 0.64 seconds!
 _available_closures = list(range(1, 101))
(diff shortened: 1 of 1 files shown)


rdkit/Chem/fmcs/fmcs.py
```diff
@@ -674,24 +673,23 @@ def _check_atom_classes(molno, num_atoms, atom_classes):
 # prune. But so far I don't have a test set which drives the need for
 # that.
 
+# UPDATED by Ichiru Take: Use collections.Counter to count the bond types, which is much faster.
+# See here: https://stackoverflow.com/questions/44012479/intersection-of-two-counters
```
gedeck, line 677: You don't need to add this comment. It doesn't add anything

rdkit/Chem/fmcs/fmcs.py
```diff
@@ -674,24 +673,23 @@ def _check_atom_classes(molno, num_atoms, atom_classes):
...
-      v = min(v1, counts2[k])
-      d[k] = v
-  return d
+  if not isinstance(counts1, Counter) and not isinstance(counts2, Counter):
+    return dict(Counter(counts1) & Counter(counts2))
+  if isinstance(counts1, Counter) and not isinstance(counts2, Counter):
+    return dict(counts1 & Counter(counts2))
+  if not isinstance(counts1, Counter) and isinstance(counts2, Counter):
+    return dict(Counter(counts1) & counts2)
+  return dict(counts1 & counts2)
```
gedeck, line 692: With all the if statements this is quite hard to read. This should be equivalent.
```
If not isinstance(Counter, counts1):
  Counts1 = Counter(counts1)
If not isinstance(Counter, counts2):
  Counts2 = Counter(counts2)
Return dict(counts1 & counts2)
```
The question is however, is it any faster if we have to add two conversions from dict to counter and one for counter to dict?

I looked at the implementation of the & operation and it is done in Python. I doubt that it is any faster than the implementation we currently have. I expect it to be slower

rdkit/Chem/fmcs/fmcs.py
```diff
@@ -743,20 +741,19 @@ def remove_unknown_bondtypes(typed_mol, supported_canonical_bondtypes):
 
 
 def find_upper_fragment_size_limits(rdmol, atoms):
-  max_num_atoms = max_twice_num_bonds = 0
+  max_num_atoms, max_twice_num_bonds = 0, 0
```
gedeck, line 744: No. I'm not a fan of multiple assignments, but converting this to tuple unpacking makes no sense.

rdkit/Chem/fmcs/fmcs.py
```diff
@@ -743,20 +741,19 @@ def remove_unknown_bondtypes(typed_mol, supported_canonical_bondtypes):
...
     # Every bond is connected to two atoms, so this is the
     # simplest way to count the number of bonds in the fragment.
     twice_num_bonds = 0
     for atom_index in atom_indices:
       # XXX Why is there no 'atom.GetNumBonds()'?
-      twice_num_bonds += sum(1 for bond in atoms[atom_index].GetBonds())
-    if twice_num_bonds > max_twice_num_bonds:
-      max_twice_num_bonds = twice_num_bonds
+      # Ichiru Take: len(atoms[atom_index].GetBonds()) would be more efficient but I don't know the input type.
+      twice_num_bonds += sum(1 for bond in atoms[atom_index].GetBonds())
```
gedeck, line 754: @greglandrum what do you think?

rdkit/Chem/fmcs/fmcs.py
```diff
@@ -1171,8 +1170,7 @@ def canon(cangen_nodes):
 def get_closure_label(bond_smarts, closure):
   if closure < 10:
     return bond_smarts + str(closure)
-  else:
-    return bond_smarts + "%%%02d" % closure
+  return bond_smarts + "%%%02d" % closure
```
gedeck, line 1173: F strings

gedeck (changes requested):

rdkit/Chem/fmcs/fmcs.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thank you for your review.

rdkit/Chem/fmcs/fmcs.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thank you for your review.

rdkit/Chem/fmcs/fmcs.py
gedeck: I really think this needs to be reversed to the original code. It will be slower.

rdkit/Chem/fmcs/fmcs.py
```diff
@@ -827,6 +815,9 @@ def get_typed_fragment(typed_mol, atom_indices):
       orig_bonds.append(orig_bond)
     elif count == 1:
       raise AssertionError("connected/disconnected atoms?")
+    elif count == 0:
+      # This means that the bond is not in the fragment or the source code above is the error.
+      raise AssertionError("Error Source Code?")
```
gedeck, line 820: Why add an exception if the original code doesn't contain one.

gedeck (changes requested):

rdkit/Chem/fmcs/fmcs.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thank you for your review.

rdkit/Chem/fmcs/fmcs.py
IchiruTake: I am so sorry, this is my mistake. I will revert it to the original

rdkit/Chem/fmcs/fmcs.py
IchiruTake: I have tested `len(atoms[atom_index].GetBonds())` and it worked

gedeck (approved):

PR #4911, opened 2022-01-19
Clean-up Python #4815 - P1.12: Chem\MolStandardize

IchiruTake:

----------------
Changes:
- [Remark]: In charge.py, the method in line 245 and 255 seems redundant. I has dropped it and no issue raised
- [Remark]: Method `normalize()` in `normalize.py`  is relatively weird due to the excessive use of list. This patch is to correct the method and free some unused memory. No speed-boost here.

### DIFF
--- a/rdkit/Chem/MolStandardize/charge.py
+++ b/rdkit/Chem/MolStandardize/charge.py
@@ -310,9 +310,7 @@ class Uncharger(object):
                     neg_surplus -= 1
                     log.info('Removed negative charge')
         else:
-            #
-            negativeAtomIdxs = [mol.GetAtomWithIdx(x) for x in n]
-            for atom in negativeAtomIdxs:
+            for atom in [mol.GetAtomWithIdx(x) for x in n]:
                 while atom.GetFormalCharge() < 0:
                     atom.SetNumExplicitHs(atom.GetNumExplicitHs() + 1)
                     atom.SetFormalCharge(atom.GetFormalCharge() + 1)
--- a/rdkit/Chem/MolStandardize/charge.py
+++ b/rdkit/Chem/MolStandardize/charge.py
@@ -50,7 +50,7 @@ class AcidBasePair(object):
         return Chem.MolFromSmarts(self.base_str)
 
     def __repr__(self):
-        return fr'AcidBasePair({self.name}, {self.acid_str}, {self.base_str})'
+        return 'AcidBasePair({!r}, {!r}, {!r})'.format(self.name, self.acid_str, self.base_str)
 
     def __str__(self):
         return self.name
@@ -116,7 +116,7 @@ class ChargeCorrection(object):
         return Chem.MolFromSmarts(self.smarts_str)
 
     def __repr__(self):
-        return fr'ChargeCorrection({self.name}, {self.smarts_str}, {self.charge})'
+        return 'ChargeCorrection({!r}, {!r}, {!r})'.format(self.name, self.smarts_str, self.charge)
 
     def __str__(self):
         return self.name
--- a/rdkit/Chem/MolStandardize/charge.py
+++ b/rdkit/Chem/MolStandardize/charge.py
@@ -34,23 +34,23 @@ class AcidBasePair(object):
         :param string acid: SMARTS pattern for the protonated acid.
         :param string base: SMARTS pattern for the conjugate ionized base.
         """
-        log.debug('Initializing AcidBasePair: %s', name)
+        log.debug(f'Initializing AcidBasePair: {name}')
         self.name = name
         self.acid_str = acid
         self.base_str = base
 
     @memoized_property
     def acid(self):
-        log.debug('Loading AcidBasePair acid: %s', self.name)
+        log.debug(f'Loading AcidBasePair acid: {self.name}')
         return Chem.MolFromSmarts(self.acid_str)
 
     @memoized_property
     def base(self):
-        log.debug('Loading AcidBasePair base: %s', self.name)
+        log.debug(f'Loading AcidBasePair base: {self.name}')
         return Chem.MolFromSmarts(self.base_str)
 
     def __repr__(self):
-        return 'AcidBasePair({!r}, {!r}, {!r})'.format(self.name, self.acid_str, self.base_str)
+        return fr'AcidBasePair({self.name}, {self.acid_str}, {self.base_str})'
 
     def __str__(self):
         return self.name
@@ -105,18 +105,18 @@ class ChargeCorrection(object):
         :param string smarts: SMARTS pattern to match. Charge is applied to the first atom.
         :param int charge: The charge to apply.
         """
-        log.debug('Initializing ChargeCorrection: %s', name)
+        log.debug(f'Initializing ChargeCorrection: {name}')
         self.name = name
         self.smarts_str = smarts
         self.charge = charge
 
     @memoized_property
     def smarts(self):
-        log.debug('Loading ChargeCorrection smarts: %s', self.name)
+        log.debug(f'Loading ChargeCorrection smarts: {self.name}')
         return Chem.MolFromSmarts(self.smarts_str)
 
     def __repr__(self):
-        return 'ChargeCorrection({!r}, {!r}, {!r})'.format(self.name, self.smarts_str, self.charge)
+        return fr'ChargeCorrection({self.name}, {self.smarts_str}, {self.charge})'
 
     def __str__(self):
         return self.name
@@ -186,8 +186,7 @@ class Reionizer(object):
                 ppos, poccur = self._strongest_protonated(mol)
                 if ppos is None:
                     break
-                log.info('Ionizing %s to balance previous charge corrections',
-                         self.acid_base_pairs[ppos].name)
+                log.info(f'Ionizing {self.acid_base_pairs[ppos].name} to balance previous charge corrections')
                 patom = mol.GetAtomWithIdx(poccur[-1])
                 patom.SetFormalCharge(patom.GetFormalCharge() - 1)
                 if patom.GetNumExplicitHs() > 0:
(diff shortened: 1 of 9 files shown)


rdkit/Chem/MolStandardize/charge.py
```diff
@@ -245,16 +243,10 @@ def reionize(self, mol):
         return mol
 
     def _strongest_protonated(self, mol):
-        for position, pair in enumerate(self.acid_base_pairs):
-            for occurrence in mol.GetSubstructMatches(pair.acid):
-                return position, occurrence
-        return None, None
+        return 0, mol.GetSubstructMatches(self.acid_base_pairs[0].acid)
```
gedeck, line 246: This looks very different!

rdkit/Chem/MolStandardize/charge.py
```diff
@@ -245,16 +243,10 @@ def reionize(self, mol):
...
 
     def _weakest_ionized(self, mol):
-        for position, pair in enumerate(reversed(self.acid_base_pairs)):
-            for occurrence in mol.GetSubstructMatches(pair.base):
-                return len(self.acid_base_pairs) - position - 1, occurrence
-        return None, None
+        return len(self.acid_base_pairs) - 1, mol.GetSubstructMatches(self.acid_base_pairs[-1].base)
```
gedeck, line 249: Same

rdkit/Chem/MolStandardize/normalize.py
```diff
@@ -137,20 +137,24 @@ def normalize(self, mol):
...
-            outmol = Chem.CombineMols(outmol, fragment)
+        # Update: Aggregated normalized fragments into a single molecule according to the previous code.
+        # Step 1: Normalize each fragment separately to get around quirky RunReactants behaviour
+        # Step 2: Join normalized fragments into a single molecule again (previous intention: bottom-up)
+        NORMALIZE_FRAGMENT = self._normalize_fragment
+
+        MolFrags = Chem.GetMolFrags(mol, asMols=True)
+        outmol = NORMALIZE_FRAGMENT(MolFrags[-1])
+        for i in range(len(MolFrags) - 2, -1, -1):
+            outmol = Chem.CombineMols(outmol, NORMALIZE_FRAGMENT(MolFrags[i])) # from bottom to top
```
gedeck, line 150: The original code is a lot easier to read. The naming is also inconsistent with the rest of the code. And why reverse the order of combining the fragments. The order shouldn't matter.

gedeck (changes requested):

rdkit/Chem/MolStandardize/charge.py
IchiruTake: The old code iterate only 1 time and only return at the beginning. If you think it may indicate some incomplete function, I will revert it.

rdkit/Chem/MolStandardize/charge.py
IchiruTake: Similar

rdkit/Chem/MolStandardize/normalize.py
IchiruTake: At the old code, it did exactly as I did. The order I made here is 100% the same as in the original version

IchiruTake: > The original code is a lot easier to read. The naming is also inconsistent with the rest of the code. And why reverse the order of combining the fragments. The order shouldn't matter.

I have already mentioned that.

rdkit/Chem/MolStandardize/charge.py
```diff
@@ -314,13 +313,16 @@ def uncharge(self, mol):
...
+            for atom in negativeAtomIdxs:
                 while atom.GetFormalCharge() < 0:
                     atom.SetNumExplicitHs(atom.GetNumExplicitHs() + 1)
                     atom.SetFormalCharge(atom.GetFormalCharge() + 1)
                     log.info('Removed negative charge')
+        
         # Neutralize positive charges
-        for atom in [mol.GetAtomWithIdx(x) for x in p]:
+        positiveAtomIdxs = [mol.GetAtomWithIdx(x) for x in p]
+        for atom in positiveAtomIdxs:
```
gedeck, line 325: The definition of the list is so short that I don't see a benefit of separating it from the for loop in both cases. It also gives the impression that it could be used later on in the code.

rdkit/Chem/MolStandardize/normalize.py
```diff
@@ -137,33 +137,35 @@ def normalize(self, mol):
...
-        # Join normalized fragments into a single molecule again
-        outmol = fragments.pop()
-        for fragment in fragments:
-            outmol = Chem.CombineMols(outmol, fragment)
+        NORMALIZE_FRAGMENT = self._normalize_fragment
+
+        Fragments = Chem.GetMolFrags(mol, asMols=True)
+        outmol = NORMALIZE_FRAGMENT(Fragments[-1])
+        for i in range(len(Fragments) - 2, -1, -1):
+            outmol = Chem.CombineMols(outmol, NORMALIZE_FRAGMENT(Fragments[i]))
```
gedeck, line 148: I find this a lot less readable than the original version. 

Considering the rest of the code base, `Fragments` looks like a class or a module name and not a variable name.

gedeck (changes requested):

rdkit/Chem/MolStandardize/charge.py
```diff
@@ -36,23 +34,23 @@ def __init__(self, name, acid, base):
...
 
     @memoized_property
     def base(self):
-        log.debug('Loading AcidBasePair base: %s', self.name)
+        log.debug(f'Loading AcidBasePair base: {self.name}')
         return Chem.MolFromSmarts(self.base_str)
 
     def __repr__(self):
-        return 'AcidBasePair({!r}, {!r}, {!r})'.format(self.name, self.acid_str, self.base_str)
+        return fr'AcidBasePair({self.name}, {self.acid_str}, {self.base_str})'
```
gedeck, line 53: This is not identical. Use this:
```suggestion
        return f'AcidBasePair({self.name!r}, {self.acid_str!r}, {self.base_str!r})'
```

rdkit/Chem/MolStandardize/charge.py
```diff
@@ -107,18 +105,18 @@ def __init__(self, name, smarts, charge):
...
 
     @memoized_property
     def smarts(self):
-        log.debug('Loading ChargeCorrection smarts: %s', self.name)
+        log.debug(f'Loading ChargeCorrection smarts: {self.name}')
         return Chem.MolFromSmarts(self.smarts_str)
 
     def __repr__(self):
-        return 'ChargeCorrection({!r}, {!r}, {!r})'.format(self.name, self.smarts_str, self.charge)
+        return fr'ChargeCorrection({self.name}, {self.smarts_str}, {self.charge})'
```
gedeck, line 119:
```suggestion
        return f'ChargeCorrection({self.name!r}, {self.smarts_str!r}, {self.charge!r})'
```

rdkit/Chem/MolStandardize/fragment.py
```diff
@@ -40,7 +40,7 @@ def smarts(self):
         return Chem.MolFromSmarts(self.smarts_str)
 
     def __repr__(self):
-        return 'FragmentPattern({!r}, {!r})'.format(self.name, self.smarts_str)
+        return fr'FragmentPattern({self.name}, {self.smarts_str})'
```
gedeck, line 43:
```suggestion
        return f'FragmentPattern({self.name!r}, {self.smarts_str!r})'
```

rdkit/Chem/MolStandardize/normalize.py
```diff
@@ -29,17 +29,17 @@ def __init__(self, name, transform):
...
 
     @memoized_property
     def transform(self):
-        log.debug('Loading Normalization transform: %s', self.name)
+        log.debug(f'Loading Normalization transform: {self.name}')
         return AllChem.ReactionFromSmarts(str(self.transform_str))
 
     def __repr__(self):
-        return 'Normalization({!r}, {!r})'.format(self.name, self.transform_str)
+        return fr'Normalization({self.name}, {self.transform_str})'
```
gedeck, line 42:
```suggestion
        return f'Normalization({self.name!r}, {self.transform_str!r})'
```

rdkit/Chem/MolStandardize/charge.py
gedeck:
```
>>> fr'AcidBasePair({self.name}, {self.acid_str}, {self.base_str})'
'AcidBasePair(abc, def, xyz)'
>>> f'AcidBasePair({self.name!r}, {self.acid_str!r}, {self.base_str!r})'
"AcidBasePair('abc', 'def', 'xyz')"
```

rdkit/Chem/MolStandardize/tautomer.py
```diff
@@ -58,7 +58,7 @@ def tautomer(self):
         return Chem.MolFromSmarts(self.tautomer_str)
 
     def __repr__(self):
-        return 'TautomerTransform({!r}, {!r}, {!r}, {!r})'.format(self.name, self.tautomer_str, self.bonds, self.charges)
+        return fr'TautomerTransform({self.name}, {self.tautomer_str}, {self.bonds}, {self.charges})'
```
gedeck, line 61:
```suggestion
        return f'TautomerTransform({self.name!r}, {self.tautomer_str!r}, {self.bonds!r}, {self.charges!r})'
```

gedeck (changes requested):
The use of f-strings in logging commands is controversial. The call to e.g. `log.debug` with a f-string will format the string even if the log level is not debug.

IchiruTake:
> The use of f-strings in logging commands is controversial. The call to e.g. `log.debug` with a f-string will format the string even if the log level is not debug.

Thank you for pointing out my mistake. However, I don't prefer this change as it cause some noisy when looking at it. I will revert all these six changes back to original

gedeck (approved):

greglandrum (approved):
LGTM

PR #4914, opened 2022-01-19
Clean-up Python #4815 - P1.15: Chem\Subshape

IchiruTake:

----------------------------------
Updated Changes:
- [Remark]: Optimize `__call__()` in `SubshapeAligner.py`
- [Remark]: Measure time by `perf_counter()` (Introduced in Python 3.3)
- [New]: Remove unused import in `testCombined.py`

### DIFF
--- a/rdkit/Chem/Subshape/SubshapeAligner.py
+++ b/rdkit/Chem/Subshape/SubshapeAligner.py
@@ -23,17 +23,25 @@ class SubshapeAlignment(object):
 
 
 def _getAllTriangles(pts, orderedTraversal=False):
-  for i in range(len(pts)):
+  nPts: int = len(pts)
+  for i in range(nPts):
     if orderedTraversal:
       jStart = i + 1
     else:
       jStart = 0
-    for j in range(jStart, len(pts)):
+    for j in range(jStart, nPts):
       if j == i:
         continue
-      for k in range(j + 1, len(pts)):
+      for k in range(j + 1, nPts):
         if k == i or k == j:
           continue
+        """
+        Dead code (Keep for potential bug)
+        if orderedTraversal:
+          kStart = j + 1
+        else:
+          kStart = 0
+        """
         yield (i, j, k)
 
 
@@ -48,8 +56,9 @@ def GetShapeShapeDistance(s1, s2, distMetric):
     # print(s1.grid.GetOccupancyVect().GetTotalVal(), s2.grid.GetOccupancyVect().GetTotalVal())
     if s1.grid.GetOccupancyVect().GetTotalVal() < s2.grid.GetOccupancyVect().GetTotalVal():
       return Geometry.ProtrudeDistance(s1.grid, s2.grid)
+    else:
+      return Geometry.ProtrudeDistance(s2.grid, s1.grid)
     
-    return Geometry.ProtrudeDistance(s2.grid, s1.grid)
   return Geometry.TanimotoDistance(s1.grid, s2.grid)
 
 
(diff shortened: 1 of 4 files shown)


rdkit/Chem/Subshape/SubshapeAligner.py
```diff
@@ -33,10 +31,6 @@ def _getAllTriangles(pts, orderedTraversal=False):
     for j in range(jStart, len(pts)):
       if j == i:
         continue
-      if orderedTraversal:
```
bp-kelley, line 36: This dead code might be indicative of a real bug.

rdkit/Chem/Subshape/SubshapeAligner.py
```diff
@@ -51,15 +45,12 @@ class SubshapeDistanceMetric(object):
 def GetShapeShapeDistance(s1, s2, distMetric):
   """ returns the distance between two shapes according to the provided metric """
   if distMetric == SubshapeDistanceMetric.PROTRUDE:
-    # print(s1.grid.GetOccupancyVect().GetTotalVal(),s2.grid.GetOccupancyVect().GetTotalVal())
```
bp-kelley, line 54: The else statements make the code paths a lot more clear here IMO

rdkit/Chem/Subshape/SubshapeAligner.py
```diff
@@ -282,42 +274,44 @@ def PruneMatchesUsingShape(self, targetMol, target, queryMol, query, builder, al
   def GetSubshapeAlignments(self, targetMol, target, queryMol, query, builder, tgtConf=-1,
                             queryConf=-1, pruneStats=None):
     import time
+
     if pruneStats is None:
       pruneStats = {}
+
     logger.info("Generating triangle matches")
-    t1 = time.time()
+    t1 = time.perf_counter()
```
bp-kelley, line 282: This is a general comment, but time.time is not a reliable measurement for profiling.  We may think about adjust this to perf_counter in other places, especially unit tests.

rdkit/Chem/Subshape/SubshapeAligner.py
```diff
@@ -282,42 +274,44 @@ def PruneMatchesUsingShape(self, targetMol, target, queryMol, query, builder, al
...
 
   def __call__(self, targetMol, target, queryMol, query, builder, tgtConf=-1, queryConf=-1,
                pruneStats=None):
     for alignment in self.GetTriangleMatches(target, query):
-      if (not self._checkMatchFeatures(target.skelPts, query.skelPts, alignment) and
-          builder.featFactory):
-        if pruneStats is not None:
-          pruneStats['features'] = pruneStats.get('features', 0) + 1
-        continue
+      if builder.featFactory:
```
bp-kelley, line 309: This code doesn't do the same thing as the original.

The original code always called _checkMatchFeatures before checking if featFactory existed.  This checks featFactory first.

bp-kelley (changes requested):

rdkit/Chem/Subshape/SubshapeAligner.py
IchiruTake: I know that the old method do `_checkMatchFeatures()` first at the time refactoring. However, as said, if the function `_checkMatchFeatures()` doing ahead of the design, it would be the damaging of function design rather than the naive optimization. I have discussed this type in PR #4815. 
As discussed, this type of PR do naive-blindful optimization, which means that given unknown input and processing (little datatype is known and guessed), how much optimization we can do ahead.

In the end, all unit tests passed and zero issues raised.
@gedeck What do you think?

gedeck: While the test coverage is reasonable on the Python side, I wouldn't assume that a piece of code is covered by tests. This means passing tests doesn't necessarily imply correctness of refactoring.

rdkit/Chem/Subshape/SubshapeAligner.py
IchiruTake: OK, I got your idea! I would fix it in the next patch. Thank you for your review. I will make a doc of this for further usage

rdkit/Chem/Subshape/SubshapeAligner.py
IchiruTake: > This code doesn't do the same thing as the original.
> 
> The original code always called _checkMatchFeatures before checking if featFactory existed. This checks featFactory first.

If you worrried about this change, I will revert back to the original

IchiruTake: OK, I got your idea! I would fix it in the next patch. Thank you for your review.

PR #4922, opened 2022-01-20
Clean-up Python #4815 - P2.3: Chem:E

IchiruTake:

### DIFF
--- a/rdkit/Chem/EnumerateStereoisomers.py
+++ b/rdkit/Chem/EnumerateStereoisomers.py
@@ -231,7 +231,7 @@ def EnumerateStereoisomers(m, options=StereoEnumerationOptions(), verbose=False)
     >>> isomers = tuple(EnumerateStereoisomers(m, options=opts))
     >>> len(isomers)
     10
-    >>> for smi in sorted(Chem.MolToSmiles(x,isomericSmiles=True) for x in isomers):
+    >>> for smi in sorted(Chem.MolToSmiles(x, isomericSmiles=True) for x in isomers):
     ...     print(smi)
     ...
     F[C@@H](Cl)/C=C/C=C/[C@@H](F)Cl
@@ -251,7 +251,7 @@ def EnumerateStereoisomers(m, options=StereoEnumerationOptions(), verbose=False)
     >>> isomers = tuple(EnumerateStereoisomers(m))
     >>> len(isomers)
     8
-    >>> for smi in sorted(Chem.MolToSmiles(x,isomericSmiles=True) for x in isomers):
+    >>> for smi in sorted(Chem.MolToSmiles(x, isomericSmiles=True) for x in isomers):
     ...     print(smi)
     ...
     F[C@@]12C[C@@]1(Cl)C[C@@H](/C=C/Br)O2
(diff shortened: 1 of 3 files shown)


rdkit/Chem/EnumerateStereoisomers.py
```diff
@@ -66,30 +66,36 @@ def __init__(self, group):
...
+    for atom, original_parity in self._original_parities:
+      try:
+        if flag:
+          chirality = original_parity 
+        else:
+          chirality = CHIRAL_DICT[original_parity]
+        atom.SetChiralTag(chirality)
+      except KeyError:
+        pass
+    del CHIRAL_DICT
```
gedeck, line 83: Deleting the dictionary shouldn't be necessary. However, I think the modified code is less efficient. 
- Constructing and deleting the dictionary costs time.
- If flag is True, the `for` loop has no additional condition (likely not critical)
- If flag is False, `atom.setChiralTag` is only called for atoms that are chiral. In the new code, it is called for all atoms.

rdkit/Chem/EnumerateStereoisomers.py
```diff
@@ -354,4 +359,4 @@ def EnumerateStereoisomers(m, options=StereoEnumerationOptions(), verbose=False)
       if options.maxIsomers != 0 and numIsomers >= options.maxIsomers:
         break
     elif verbose:
-      print("%s    failed to embed" % (Chem.MolToSmiles(isomer, isomericSmiles=True)))
+      print(f"{Chem.MolToSmiles(isomer, isomericSmiles=True)}    failed to embed")
```
gedeck, line 362: `isomericSmiles=True` is the default. On the other hand, adding it here makes it clear that we want to see chirality information.

rdkit/Chem/UnitTestEnumerateHeterocycles.py
```diff
@@ -1,29 +1,20 @@
...
+                                              GetHeterocycleReactions,
+                                              GetHeterocycleReactionSmarts)
+
 
 def has_radical(mol):
-    for atom in mol.GetAtoms():
-        if atom.GetNumRadicalElectrons():
-            return True
-    return False
+    return any(atom.GetNumRadicalElectrons() > 0 for atom in mol.GetAtoms())
```
gedeck, line 14: You probably don't need the `> 0`:
```suggestion
    return any(atom.GetNumRadicalElectrons() for atom in mol.GetAtoms())
```

rdkit/Chem/UnitTestEnumerateHeterocycles.py
```diff
@@ -58,8 +49,7 @@ def get_unique_products(rxn, mol):
             continue
         uniq_smiles.add(isosmiles)
         newmol = Chem.MolFromSmiles(isosmiles)
-        assert newmol is not None, '%s produced %s when applying' % (Chem.MolToSmiles(mol, isomericSmiles=True),
-                                                                     isosmiles)
+        assert newmol is not None, f'{Chem.MolToSmiles(mol, isomericSmiles=True)} produced {isosmiles} when applying'
```
gedeck, line 52: Same comment about `isomericSmiles`

rdkit/Chem/UnitTestEnumerateHeterocycles.py
```diff
@@ -108,7 +98,7 @@ def test_at_least_something_matches_every_negative_example(self):
                 if mol.HasSubstructMatch(substruct):
                     something_hit = True
                     break
-            assert something_hit, ('nothing matched %s' % Chem.MolToSmiles(mol, isomericSmiles=True))
+            assert something_hit, f'nothing matched {Chem.MolToSmiles(mol, isomericSmiles=True)}'
```
gedeck, line 101: and again.

rdkit/Chem/UnitTestEnumerateHeterocycles.py
```diff
@@ -149,20 +139,16 @@ def test_apply_every_rule_to_every_fragment(self):
...
-                        'SMILES'  : orig_can_smi,
-                        'MUTATED' : smiles,
-                        'REACTION' : src.SMARTS + '>>' + src.CONVERT_TO,
-                        'DESCRIPTION' : src.DESCRIPTION,
-                        }
-                    writer.writerow(row)
+                    desc = {'SMILES'  : orig_can_smi, 'MUTATED' : smiles,
+                           'REACTION' : src.SMARTS + '>>' + src.CONVERT_TO,
+                           'DESCRIPTION' : src.DESCRIPTION,
+                           }
```
gedeck, line 145: Restore the original version. Dictionaries should either be all on one line or all key: value pairs on individual lines.

rdkit/Chem/UnitTestEnumerateHeterocycles.py
```diff
@@ -261,29 +247,25 @@ def test_fuzz_atom_mutations(self):
...
-                    row = {
-                        'SMILES'  : orig_can_smi,
-                        'MUTATED' : smiles,
-                        'REACTION' : src.SMARTS + '>>' + src.CONVERT_TO,
-                        'DESCRIPTION' : src.DESCRIPTION,
-                        }
+                    row = {'SMILES': orig_can_smi, 'MUTATED': smiles,
+                           'REACTION': src.SMARTS + '>>' + src.CONVERT_TO, 
+                           'DESCRIPTION': src.DESCRIPTION, 
+                           }
```
gedeck, line 253: same comment as above

gedeck (changes requested):

rdkit/Chem/EnumerateStereoisomers.py
IchiruTake: Agree, I would reverse it back to originial

rdkit/Chem/EnumerateStereoisomers.py
IchiruTake: As mentioned in the last comment, also since the file is targeted on Stereo-Isomers, we want to ensure that even if the default argument in `MolToSmiles()` is switched to False, we will always get `isomericSmiles=True`. Otherwise, it is quite hard to fix them

PR #4923, opened 2022-01-20
Clean-up Python #4815 - P2.4: Chem: F & Graphs.py

IchiruTake:

-----
Updated Changes:
- [New]: Missing import in `FragmentCatalog.py`
- [Warning]: In `FilterCatalog.py`, errors is raised because Chem module is not imported. Please check the source


IchiruTake:
The error here indicated incorrect time measurment of `time.time()` in `UnitTestLogging.py`. Should we change it to `time.perf_counter()` for better time measurement @greglandrum

PR #4929, opened 2022-01-21
Clean-up Python #4815 - P2.6: Chem: R & S & T

IchiruTake:

A second patch later after review. If the warning below is correct. This could be a bug-fix, I guessed so.

------------
Note:
- [Warning]: Weird behavior at line 384 and 386 in `TorsionFingerprints.py`, is that: `if len(nb1) > 1 and len(nb2) > 1:`


IchiruTake:
@gedeck Do you have any issue related to the module `TorsionFingerprints.py` at line 384-386. Does the algorithm should be `if len(nb1) > 1 and len(nb2) > 1:` instead of `if len(nb1) > 1 and len(nb1) > 1:`

PR #4944, opened 2022-01-23
Clean-up Python #4815 - P7: utils modules

IchiruTake:

----------------------------------
Updated Changes:
- [Q&A]: Line 12 use `import win32clipboard` in `chemdraw
_qax.py`. I cannot find any particular usage here. Is it safe to delete?
- [Remark]: Fix a numerical issue in function `SplitComposition()` in `chemutils.py`.


IchiruTake:
Updated:
[Q&A]: Line 12 use import win32clipboard in chemdraw _qax.py. I cannot find any particular usage here. Is it safe to delete?
-> It worked

PR #4948, opened 2022-01-24
Refactor mol draw2 d

DavidACosgrove:
[WIP] - Re-factoring of the 2D drawing code - not yet ready for code review
This is a re-factoriing of the MolDraw2D classes to make them more maintainable in future.  There is a new class DrawMol which does nearly all the drawing in terms of a new set of classes derived from DrawShape.  The new classes are all private and friends of MolDraw2D to emphasise that they are implementation classes and not for general use.  When the re-factoring is finished, MolDraw2D will contain a set of functions for drawing primitives, and the top level functions for drawing molecules and reactions.  Everything else will be removed as it has mostly been copied into DrawMol. 

#### Any other comments?
There is still all the tidying up to do.  This is an early PR so that people can start looking at the new pictures and decide if any changes are required.
To help with the comparison, there's a very basic script $RDBASE/Code/GraphMol/MolDraw2D/side_by_side.py which will makes an HTML table of corresponding image files from 2 directories as the rows.  You will need to edit it to change the hard-coded directory names.  I will improve this in due course.


DavidACosgrove:
I think this is now ready for review.  The testMolDraw2D.py now works on Ubuntu 20.04.  It crashes for me on macOS, but then so does testSparseIntVect.py so it may not be relevant.
I haven't changed any of the file hash codes yet, I'll do that once everything is accepted.
The C++ code is compiling and working on Ubuntu 2.0.04 and macOS Big Sur 11.6.2.  I expect the other language wrappers will need some work, and I'm not able to try the Qt.

Code/GraphMol/MolDraw2D/DrawMol.h
```diff
@@ -0,0 +1,303 @@
...
+          bool isReactionMol = false);
+  // this one is for when there's no molecule to draw, but we still want
+  // a DrawMol in the MolDraw2D for scale, conversion of atom coords to
+  // draw coords etc.  And so DrawMol starts sounding like a poor name for
+  // the class.
+  DrawMol(int width, int height, const MolDrawOptions &drawOptions,
+          DrawText &textDrawer, double xmin, double xmax, double ymin,
+          double ymax, double scale, double fontscale);
+  DrawMol(const DrawMol &) = delete;
+  DrawMol(const DrawMol &&) = delete;
```
greglandrum, line 88: argument shouldn't be const

Code/GraphMol/MolDraw2D/DrawMol.h
```diff
@@ -0,0 +1,303 @@
...
+  DrawMol &operator=(const DrawMol &) = delete;
```
greglandrum, line 89: maybe forgot move-assignment

Code/GraphMol/MolDraw2D/DrawMolMCH.h
```diff
@@ -0,0 +1,57 @@
...
+ protected:
+  DrawMolMCH(const ROMol &mol, const std::string &legend, int width, int height,
+             MolDrawOptions &drawOptions, DrawText &textDrawer,
+             const std::map<int, std::vector<DrawColour>> &highlight_atom_map,
+             const std::map<int, std::vector<DrawColour>> &highlight_bond_map,
+             const std::map<int, double> &highlight_radii,
+             const std::map<int, int> &highlight_linewidth_multipliers,
+             int confId = -1);
+  DrawMolMCH(const DrawMol &) = delete;
+  DrawMolMCH(const DrawMol &&) = delete;
```
greglandrum, line 37: argument shouldn't be const

Code/GraphMol/MolDraw2D/DrawMolMCH.h
```diff
@@ -0,0 +1,57 @@
...
+  DrawMolMCH &operator=(const DrawMol &) = delete;
```
greglandrum, line 38: maybe forgot move assignment?

Code/GraphMol/MolDraw2D/DrawShape.h
```diff
@@ -0,0 +1,250 @@
...
+ public:
+  virtual ~DrawShape() = default;
+
+ protected:
+  DrawShape(const std::vector<Point2D> &points, int lineWidth = 2,
+            bool scaleLineWidth = false,
+            DrawColour lineColour = DrawColour(0, 0, 0), bool fill = false,
+            int atom1 = -1, int atom2 = -1, int bond = -1);
+  DrawShape(const DrawShape &ds) = delete;
+  DrawShape(const DrawShape &&ds) = delete;
```
greglandrum, line 49: argument shouldn't be const

Code/GraphMol/MolDraw2D/DrawShape.h
```diff
@@ -0,0 +1,250 @@
...
+  DrawShape &operator=(const DrawShape &ds) = delete;
+  DrawShape &operator=(const DrawShape &&ds) = delete;
```
greglandrum, line 51: argument shouldn't be const

Code/GraphMol/MolDraw2D/Qt/DrawTextQt.h
```diff
@@ -20,6 +20,10 @@ namespace RDKit {
 class DrawTextQt : public DrawText {
  public:
   DrawTextQt(double max_fnt_sz, double min_fnt_sz, QPainter *qp);
+  MolDraw2DQt(const MolDraw2DQt &rhs) = delete;
+  MolDraw2DQt(const MolDraw2DQt &&rhs) = delete;
```
greglandrum, line 24: argument shouldn't be const

Code/GraphMol/MolDraw2D/Qt/DrawTextQt.h
```diff
@@ -20,6 +20,10 @@ namespace RDKit {
...
+  MolDraw2DQt &operator=(const MolDraw2DQt &rhs) = delete;
+  MolDraw2DQt &operator=(const MolDraw2DQt &&rhs) = delete;
```
greglandrum, line 26: argument shouldn't be const

Code/GraphMol/MolDraw2D/Qt/DrawTextFTQt.h
```diff
@@ -22,10 +22,14 @@ namespace RDKit {
...
 // ****************************************************************************
 
-class DrawTextFTQt : public DrawTextFT {
- public:
+class DrawTextFTQt : protected DrawTextFT {
+ protected:
   DrawTextFTQt(double max_fnt_sz, double min_fnt_sz,
                const std::string &font_file, QPainter *qp);
+  DrawTextFTQt(const DrawTextFTQt &rhs) = delete;
+  DrawTextFTQt(const DrawTextFTQt &&rhs) = delete;
```
greglandrum, line 30: argument shouldn't be const

Code/GraphMol/MolDraw2D/Qt/DrawTextFTQt.h
```diff
@@ -22,10 +22,14 @@ namespace RDKit {
...
+  DrawTextFTQt &operator=(const DrawTextFTQt &rhs) = delete;
+  DrawTextFTQt &operator=(const DrawTextFTQt &&rhs) = delete;
```
greglandrum, line 32: argument shouldn't be const

greglandrum (changes requested):
I just reviewed the header files here.
I have a few more comments about the API which I'll make  offline since it's probably more efficient.

Code/GraphMol/MolDraw2D/MolDraw2D.h
```diff
@@ -707,283 +383,124 @@ class RDKIT_MOLDRAW2D_EXPORT MolDraw2D {
...
+
+  // split the reaction up into the reagents, products and agents, each as
+  // a separate entity with its own scale.
+  void getReactionDrawMols(
+      const ChemicalReaction &rxn, bool highlightByReactant,
+      const std::vector<DrawColour> *highlightColorsReactants,
+      const std::vector<int> *confIds,
+      std::vector<std::shared_ptr<MolDraw2D_detail::DrawMol>> &reagents,
+      std::vector<std::shared_ptr<MolDraw2D_detail::DrawMol>> &products,
+      std::vector<std::shared_ptr<MolDraw2D_detail::DrawMol>> &agents,
```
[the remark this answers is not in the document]
DavidACosgrove: I had these as unique_ptr initially.  However, MolDraw2D::getAtomCoords(int atnum) uses drawMols_[activeMolIdx_] to do the transformation.  Thus the reagents etc. are also stuffed into drawMols_ hence the change to shared_ptr.
The behaviour of getAtomCoords(int atnum) pretty much mirrors what the original code did but rather emphasises that that only ever really worked when there was 1 molecule in the drawing.  You can't say which molecule the atnum refers to.  But that's a whole different issue.

Code/GraphMol/MolDraw2D/MolDraw2D.h
```diff
@@ -409,26 +60,12 @@ class RDKIT_MOLDRAW2D_EXPORT MolDraw2D {
     \c drawMolecules() is called.
   */
   MolDraw2D(int width, int height, int panelWidth, int panelHeight);
+  MolDraw2D(const MolDraw2D &rhs) = delete;
+  MolDraw2D(const MolDraw2D &&rhs) = delete;
```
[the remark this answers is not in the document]
DavidACosgrove: A good point and a useful mnemonic.

Code/GraphMol/MolDraw2D/MolDraw2D.h
```diff
@@ -565,18 +279,27 @@ class RDKIT_MOLDRAW2D_EXPORT MolDraw2D {
...
   virtual Point2D getAtomCoords(const std::pair<int, int> &screen_cds) const;
   //! transform a point from drawing coordinates to the molecule coordinate
-  //! system
+  //! system. Prefers globalDrawTrans_ if it exists.
   virtual Point2D getAtomCoords(
       const std::pair<double, double> &screen_cds) const;
   //! returns the molecular coordinates of a particular atom
   virtual Point2D getAtomCoords(int at_num) const;
   //@}
+  //! returns the coordinates of the atoms of the current molecule in molecular
```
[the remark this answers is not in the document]
DavidACosgrove: Yep.  Scattered references to activeMolIdx_ as seemed appropriate.

Code/GraphMol/MolDraw2D/MolDraw2D.h
```diff
@@ -586,28 +309,14 @@ class RDKIT_MOLDRAW2D_EXPORT MolDraw2D {
   //! return the height of the drawing panels.
   virtual int panelHeight() const { return panel_height_; }
   virtual int drawHeight() const { return panel_height_ - legend_height_; }
+  // calculate the width to draw a line in draw coords.
```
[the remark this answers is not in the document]
DavidACosgrove: I suppose it returns a calculated value but the user doesn't need to know that.

DavidACosgrove:
I think I've resolved all the issues here, bar the biggy about whether the new classes should be protected.  That's pending an offline discussion with Greg.

Code/GraphMol/MolDraw2D/DrawText.h
```diff
@@ -186,6 +214,22 @@ RDKIT_MOLDRAW2D_EXPORT bool setStringDrawMode(const std::string &instring,
 std::vector<std::string> atomLabelToPieces(const std::string &label,
                                            OrientType orient);
 
+bool doesLineIntersect(const StringRect &rect, const Point2D &end1,
```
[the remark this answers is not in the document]
DavidACosgrove: It was somewhere else, but it looked like only DrawText was using it, so I moved it there.  I can move it back.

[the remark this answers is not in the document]
greglandrum: we can always move it later, but it does seem like it could be generally useful

Code/GraphMol/MolDraw2D/DrawMol.h
```diff
@@ -0,0 +1,301 @@
...
+
+  // Make the object, scaled to a given pixel size.
+  /*!
+    \param mol             : the molecule to draw
+    \param legend          : the legend (to be drawn under the molecule)
+    \param width           : width (in pixels) of the rendering
+    set this to -1 to have the canvas size set automatically
+    \param height          : height (in pixels) of the rendering
+    set this to -1 to have the canvas size set automatically
+    \param highlight_atoms : (optional) vector of atom ids to highlight
```
greglandrum, line 51: This and the other multi-word parameters are all mis-named. Should be highlightAtoms

Code/GraphMol/MolDraw2D/DrawMolMCH.h
```diff
@@ -0,0 +1,55 @@
...
+//
+//   @@ All Rights Reserved @@
+//  This file is part of the RDKit.
+//  The contents are covered by the terms of the BSD license
+//  which is included in the file license.txt, found at the root
+//  of the RDKit source tree.
+//
+// Original author: David Cosgrove (CozChemIx Limited)
+//
+// This class, derived from DrawMol, does multi-coloured highlights.
```
greglandrum, line 12: Maybe document this below with the class definition?

Code/GraphMol/MolDraw2D/DrawText.cpp
```diff
@@ -509,6 +502,7 @@ void DrawText::getStringRects(const std::string &text, OrientType orient,
                               std::vector<TextDrawType> &draw_modes,
                               std::vector<char> &draw_chars, bool dontSplit,
                               TextAlignType textAlign) const {
+  PRECONDITION(text.size() > 0, "empty string");
```
greglandrum, line 505: `PRECONDITION(!text.empty(),"empty string");`

Code/GraphMol/MolDraw2D/DrawText.cpp
```diff
@@ -538,11 +532,11 @@ void DrawText::getStringRects(const std::string &text, OrientType orient,
     alignString(TextAlignType::START, draw_modes, rects);
   } else {
     double running_y = 0;
-    for (size_t i = 0; i < text_bits.size(); ++i) {
+    for (auto tb : text_bits) {
```
greglandrum, line 535: can be `const auto &tb`, right? No need to copy those strings
same thing for line 528 above, but I can't make a comment on that line

greglandrum (changes requested):
Another partial review

Code/GraphMol/MolDraw2D/DrawAnnotation.cpp
```diff
@@ -0,0 +1,161 @@
...
+    ymax = std::max(tr.y, ymax);
+    r->trans_ = otrans;
+  }
+}
+
+// ****************************************************************************
+void DrawAnnotation::getDimensions(double &width, double &height) const {
+  double xMin, yMin, xMax, yMax;
+  xMin = yMin = std::numeric_limits<double>::max();
+  xMax = yMax = -xMin;
```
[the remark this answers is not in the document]
DavidACosgrove: min() is always positive and essentially zero for floating point numbers.  Apparently, lowest() is recommended for floats and doubles where it is -FLT_MAX etc.  I will make it more portable with lowest().

Code/GraphMol/MolDraw2D/DrawShape.h
```diff
@@ -0,0 +1,231 @@
...
+  void myDraw(MolDraw2D &drawer) const override;
+  void findExtremes(double &xmin, double &xmax, double &ymin,
+                    double &ymax) const override;
+  void move(const Point2D &trans) override;
+  bool doesRectClash(const StringRect &rect, double padding) const override;
+
+  double ang1_, ang2_;
+};
+
+std::vector<Point2D> calcScaledWedgePoints(const Point2D &point,
```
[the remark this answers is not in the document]
DavidACosgrove: Sorry.  Left over from a previous iteration.

Code/GraphMol/FileParsers/testPropertyLists.cpp
```diff
@@ -150,7 +150,7 @@ M  END
 >  <atom.iprop.NumHeavyNeighbors>  (1) 
 2 2 2
 
->  <atom.prop.AtomLabel>  (1) 
+>  <atom.prop.AtomSymbol>  (1)
```
greglandrum, line 153: Is this a result of automatic refactoring gone wrong?

Docs/Book/RDKit_Book.rst
```diff
@@ -1498,7 +1498,7 @@ Here's a sample block from an SDF that demonstrates all of the features, they ar
   >  <atom.iprop.NumHeavyNeighbors>  (1)
   2 2 2
 
-  >  <atom.prop.AtomLabel>  (1)
+  >  <atom.prop.AtomSymbol>  (1)
```
greglandrum, line 1501: Is this a result of automatic refactoring gone wrong?

Code/GraphMol/FileParsers/testPropertyLists.cpp
DavidACosgrove: I would imagine so.  I did call it AtomLabel to begin with, and then did a re-factor.  It didn't ever put anything in that file or the one below, though.

Docs/Book/RDKit_Book.rst
DavidACosgrove: Sorry, yes.  I will pay more attention to a re-factor in future.  I didn't realise CLion would try so hard, so far outside where I was working.

Code/GraphMol/MolDraw2D/DrawShape.cpp
```diff
@@ -0,0 +1,552 @@
...
+  drawer.setColour(ocolour);
+  drawer.setLineWidth(olw);
+  drawer.setFillPolys(ofill);
+  drawer.drawOptions().scaleBondWidth = lineScale;
+}
+
+// ****************************************************************************
+void DrawShape::findExtremes(double &xmin, double &xmax, double &ymin,
+                             double &ymax) const {
+  for (auto &p : points_) {
```
greglandrum, line 62:
```suggestion
  for (const auto &p : points_) {
```

Code/GraphMol/MolDraw2D/DrawShape.cpp
```diff
@@ -0,0 +1,552 @@
...
+  DrawShape::scale(scale_factor);
+  at1Cds_.x *= scale_factor.x;
+  at1Cds_.y *= scale_factor.y;
+
+  // We may want to adjust the number of lines, so set everything up like
+  // the original triangle was passed in and rebuild.
+  Point2D point{at1Cds_};
+  Point2D end1{points_[points_.size() - 1]};
+  Point2D end2{points_[points_.size() - 2]};
+  points_ = std::vector<Point2D>{point, end1, end2};
```
greglandrum, line 438: This could involve less copying:
```suggestion
  points_ = {at1Cds_, points_[points_.size() - 1], points_[points_.size() - 2]};
```

Code/GraphMol/MolDraw2D/DrawShape.cpp
```diff
@@ -0,0 +1,552 @@
...
+// ****************************************************************************
+DrawShapeArc::DrawShapeArc(const std::vector<Point2D> points, double ang1,
+                           double ang2, int lineWidth, bool scaleLineWidth,
+                           const DrawColour &col1, bool fill, int atom1)
+    : DrawShape(points, lineWidth, scaleLineWidth, col1, fill, atom1),
+      ang1_(ang1),
+      ang2_(ang2) {
+  PRECONDITION(points_.size() == 2, "arc wrong points");
+  PRECONDITION(ang1_ >= 0.0 && ang1_ <= 360.0, "ang1_ not 0-360")
+  PRECONDITION(ang2_ >= 0.0 && ang2_ <= 360.0, "ang2_ not 0-360")
```
greglandrum, line 505:
```suggestion
  RANGE_CHECK(0., ang1_, 360.);
  RANGE_CHECK(0., ang2_, 360.);
```

Code/GraphMol/MolDraw2D/DrawShape.cpp
```diff
@@ -0,0 +1,552 @@
...
+
+// ****************************************************************************
+void DrawShape::move(const Point2D &trans) {
+  for (auto &p : points_) {
+    p += trans;
+  }
+}
+
+// ****************************************************************************
+bool DrawShape::doesRectClash(const StringRect &rect, double padding) const {
```
greglandrum, line 86: This will remove a compiler warning:
```suggestion
bool DrawShape::doesRectClash(const StringRect &, double ) const {
```

greglandrum (changes requested):
I think this is really almost there at this point.
There are a bunch of suggestions here, but they are all small.
The overall design looks good, as does the output.

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -0,0 +1,2927 @@
...
+  };
+
+  size_t rad_num = 0;
+  for (auto atom : drawMol_->atoms()) {
+    int num_rade = atom->getNumRadicalElectrons();
+    if (!num_rade) {
+      continue;
+    }
+    auto rad_rect = get<0>(radicals_[rad_num]);
+    OrientType draw_or = get<1>(radicals_[rad_num]);
```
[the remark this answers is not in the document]
DavidACosgrove: Is the surprise that it deals with the missing std::?  Me too, in that case.  Or have I missed something more sinister?

Code/GraphMol/MolDraw2D/MolDraw2DCairo.cpp
```diff
@@ -44,14 +45,18 @@ void MolDraw2DCairo::initDrawing() {
...
+      static_cast<MolDraw2D_detail::DrawTextCairo *>(text_drawer_.get())
           ->setCairoContext(dp_cr);
+    } else {
+      MolDraw2D_detail::DrawTextFTCairo *dt =
+          static_cast<MolDraw2D_detail::DrawTextFTCairo *>(text_drawer_.get());
+      dt->setCairoContext(dp_cr);
     }
 #else
-    dynamic_cast<DrawTextCairo *>(text_drawer_.get())->setCairoContext(dp_cr);
+    static_cast<DrawTextCairo *>(text_drawer_.get())->setCairoContext(dp_cr);
```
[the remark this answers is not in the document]
DavidACosgrove: IIRC, this is from the old iteration when all of the DrawText classes had private inheritance including private constructors.  In that case, the dynamic_cast compiles ok but returns nullptr at runtime because the cast fails.  It took some hunting to discover this was the cause the sudden runtime failures in the test code.  I will try reverting to dynamic_cast with the current fully public interface.

greglandrum:
Yeah, I was surprised that the std:: wasn't necessary here. One of the
header files must have a `using std::get`.

Code/GraphMol/MolDraw2D/AtomSymbol.cpp
```diff
@@ -0,0 +1,141 @@
...
+// ****************************************************************************
+void AtomSymbol::adjustColons() {
+  if (symbol_.empty()) {
+    return; // but probable it's always got something in it.
+  }
+  size_t colonPos = symbol_.find(':');
+  if (colonPos == std::string::npos) {
+    return;
+  }
+  double leftHeight = colonPos ? rects_[colonPos - 1]->height_ : 0;
```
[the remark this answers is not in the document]
DavidACosgrove: It's ok for there not to be a colon in the string, in fact it will be the commonest case.  Most atom symbols won't have atom maps in them.
There's a different bug, though, exposed by <lit>[CH2;X2:4]</lit> which is that the markup doesn't have a StringRect so the indexing in the string doesn't match the indexing in the StringRects.  Not sure why that didn't show up in testing before.

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -0,0 +1,2927 @@
...
+    }
+  }
+  return 0;
+}
+
+// ****************************************************************************
+OrientType DrawMol::calcRadicalRect(const Atom *atom,
+                                    StringRect &rad_rect) const {
+  int num_rade = atom->getNumRadicalElectrons();
+  double spot_rad = 0.2 * drawOptions_.multipleBondOffset;
```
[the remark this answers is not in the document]
DavidACosgrove: It's a bit worse than that.  Everywhere in that function where I used spot_rad bar 1 place, I multiplied it by fontScale_.  Doh!

Code/GraphMol/MolDraw2D/AtomSymbol.cpp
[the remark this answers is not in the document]
greglandrum: > It's ok for there not to be a colon in the string, in fact it will be the commonest case. Most atom symbols won't have atom maps in them. 

I think you might have misunderstood where I was suggesting to put that test:
```
  if (colonPos == std::string::npos) {
    return;
  }
  CHECK_INVARIANT(colonPos<=rects_.size(),"bad rects_ size");
  double leftHeight = colonPos ? rects_[colonPos - 1]->height_ : 0;
```
it's not super important, but this would avoid the situation where the colon position is somehow beyond the end of your rectangles array.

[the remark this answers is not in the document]
DavidACosgrove: I had understood what you meant, but I was convinced that it could never happen.  And then it did.  I've fixed that occurrence, but I guess it will do no harm to check and exit a bit less gracelessly.  It will be indicative of a bug if it is fired.

greglandrum (approved):

PR #4956, opened 2022-01-27
Clean-up Python #4815 - P8.4: ML\DecTree

IchiruTake:

-------------------
Updated Changes:
- [Remark]: Function `MaxCount()` in `PruneTree.py` is now faster 2x and used less memory.
- [Warning]: Using `sum()` consumed less memory compared to `len()` but is slower time_len ~ 0.85 * time_sum
- [Remark]: Default variable of `nQuantBounds` in function `ChooseOptimalRoot()` in `CrossValidate.py()` is set to be None.
- [Remark]: Optimize function `ChooseOptimalRoot()` in `CrossValidate.py()`

### DIFF
--- a/rdkit/ML/DecTree/BuildQuantTree.py
+++ b/rdkit/ML/DecTree/BuildQuantTree.py
@@ -6,9 +6,9 @@
 """
 
 """
-
 import numpy
 import random
+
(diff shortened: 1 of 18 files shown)


IchiruTake:
This branch is a break, and I will delete it

PR #4989, opened 2022-02-06
Expose baseFontSize to Python and JSON drawOptions

ptosco:
With reference to discussion #4652, I thought it would be useful to expose both to Python and JSON `drawOptions` the `baseFontSize` setting that currently is exposed only to C++.

Side note: I find the naming `baseFontSize` a bit confusing since the settings seems to be a scale rather than an absolute size; @DavidACosgrove: should we consider changing the naming and deprecate the old one?

Also, the behavior of `setFontSize()` seems a bit odd:
```
drawer = rdMolDraw2D.MolDraw2DSVG(300, 300)
drawer.FontSize()
0.6
```
Again, this looks like a scale rather than a font size.
If I call `SetFontSize(0.6)` (same value as returned by the accessor) I would expect no change; instead, I get a warning and afterwards `fontSize` is set to a different value than it was before:
```
SetFontSize(0.6)
[22:41:12] The new font size 0.6 is below the current minimum (6).
drawer.FontSize()
6.0
```
Looking at the `DrawText::setFontSize()` code I can see why this happens: `setFontSize()` actually calls `setFontScale()`, which I also find a bit confusing as `setFontScale()` is itself a public function that I could have called directly (though not from Python, as it is not exposed).
Additionally, if I set `fontSize` to a quite different value, e.g.:
```
drawer.SetFontSize(30)
drawer.FontSize()
30
```
I see no difference in the size of the fonts in the molecule (while changing `baseFontSize` does change the lettering base size as one would expect).

### DIFF
--- a/Code/GraphMol/MolDraw2D/MolDraw2DUtils.cpp
+++ b/Code/GraphMol/MolDraw2D/MolDraw2DUtils.cpp
@@ -164,6 +164,7 @@ void updateDrawerParamsFromJSON(MolDraw2D &drawer, const std::string &json) {
   PT_OPT_GET(includeAtomTags);
   PT_OPT_GET(clearBackground);
   PT_OPT_GET(legendFontSize);
+  PT_OPT_GET(baseFontSize);
   PT_OPT_GET(maxFontSize);
   PT_OPT_GET(minFontSize);
   PT_OPT_GET(annotationFontScale);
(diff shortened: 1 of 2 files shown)


DavidACosgrove:
Hi @ptosco,
You make good points.  The whole issue of font sizes is a bit of a can of worms, and the various ways of changing them has grown organically without an overall plan, I think it's fair to say.  If we're going to deprecate baseFontSize then I wonder if it makes sense to expose it as such in the Python wrapper.  Maybe relativeFontSize would be a better description.  Tidying up the various setters accessors at the same time would make sense.  I can incorporate that into the currently open PR #4948.

ptosco:
@DavidACosgrove I think incorporating this in your currently open PR makes sense. I opened this PR mostly to test `baseFontSize` myself and raise the point; happy to close it then.

PR #5011, opened 2022-02-16
Fix warning when generating coordinates for ZOBs

d-b-w:
When generating coordinates for molecules with zero order bonds,
the coordgen adapter layer emitted a warning. This addresses that
warning.


greglandrum (approved):
LGTM

PR #5054, opened 2022-02-27
Expose MolzipParams::atomSymbols to python

bp-kelley:
Use a simpler method of exposing MolzipParams::atomSymbols to python


greglandrum (approved):
LGTM

PR #5057, opened 2022-03-02
To INCHI conversion leaks on kekulization failure

ricrogz:
I noticed this while writing tests for an external project that uses RDKit.

A kekulization exception would cause `MolToInchi()` to return without cleaning up `m`. The smart pointer should now take care of cleaning it up whenever we leave the function.


greglandrum (approved):
Nice catch, simple fix.
LGTM!

Issue #5061, opened 2022-03-04
Image generation: Issue with scaling bond width for reaction with no reactants


rachelnwalker:
**Describe the bug**
I am having trouble generating an image of a specific reaction when `scaleBondWidth = True`

**To Reproduce**
```
from rdkit import Chem
from rdkit.Chem.Draw import rdMolDraw2D
rxn = Chem.rdChemReactions.ReactionFromRxnBlock("""$RXN

  Mrv16425    091201171606

  0  1
$MOL

  Mrv1642509121716062D

  2  1  0  0  0  0            999 V2000
    3.5357    0.0000    0.0000 R#  0  0  0  0  0  0  0  0  0  0  0  0
    2.7107    0.0000    0.0000 R#  0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
M  RGP  2   1   1   2   2


M  END""")
drawer = rdMolDraw2D.MolDraw2DSVG(400,400)
draw_options = drawer.drawOptions()
draw_options.scaleBondWidth = True
drawer.DrawReaction(rxn)
drawer.FinishDrawing()
SVG(drawer.GetDrawingText())
```
will give
```
[12:13:32] 

****
Pre-condition Violation
bad active mol index
Violation occurred on line 613 in file /tmp/build/rdkit/Code/GraphMol/MolDraw2D/MolDraw2D.cpp
Failed Expression: activeMolIdx_ >= 0 && static_cast<size_t>(activeMolIdx_) <= drawMols_.size()
****
```

**Expected behavior**
We should be able to set this bond scaling option to true. This is the image when the option is not set:
<img width="408" alt="image" src="https://user-images.githubusercontent.com/39069546/156834381-6898b312-db9c-4976-9500-74923943fc27.png">

**Configuration (please complete the following information):**
 - RDKit version: master, commit 6418918
 - OS: mac
 - Are you using conda? no
 
**Additional context**
I'm guessing this was introduced at some point in the moldraw2d refactor, because I was able to produce an image of this reaction with the scaleBondWidth option set to True a couple months ago.

DavidACosgrove:
Thanks @rachelnwalker.  I’ll take a look over the weekend.

rachelnwalker:
Thank you for getting to this so quickly @DavidACosgrove!

PR #5063, opened 2022-03-05
Fix github 5061.

DavidACosgrove:
#### Reference Issue
Fixes https://github.com/rdkit/rdkit/issues/5061

When no reagents are present in the reaction, there's nothing to use to scale the line width when drawing the arrow.  Duck the problem by drawing the arrow last.

#### Any other comments?
On my macBook, I am getting what look spurious different hash codes messaged for the moldraw2DRxnTest1 SVGs.


greglandrum (commented):
LGTM

PR #5069, opened 2022-03-08
Add sanitize option to molzip

bp-kelley:
This code allows non sanitizable fragments to be combined.  This is useful when combining multiple fragments that may eventually become a sensible molecule.

### DIFF
--- a/Code/GraphMol/ChemTransforms/MolFragmenter.cpp
+++ b/Code/GraphMol/ChemTransforms/MolFragmenter.cpp
@@ -955,7 +955,7 @@ std::unique_ptr<ROMol> molzip(const ROMol &a, const ROMol &b,
       }
     }
   }
-  newmol->updatePropertyCache(params.sanitize);
+  newmol->updatePropertyCache(params.enforceValenceRules);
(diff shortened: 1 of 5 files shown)


greglandrum (commented):
I'm not convinced that `sanitize` is the best name for this option since it's not really sanitizing the molecule: you're just changing how `updatePropertyCache()` is called

Unfortunately I don't have a great alternative suggestion. Maybe `enforceValenceRules`?

bp-kelley:
Give me the weekend to think of a better name, if not let’s merge Monday.

----
Brian Kelley

greglandrum:
@bp-kelley : any thoughts on a new name? Should we go ahead and merge as-is?

bp-kelley:
I'm fine with enforceValences, it was the best of all options I thought of.  I'll make the change.

bp-kelley:
@greglandrum changed.  Just waiting on the checks.

greglandrum (approved):
LGTM

greglandrum:
> @greglandrum changed. Just waiting on the checks.

Thanks!

PR #5086, opened 2022-03-12
[ENH]: Function `FindBondEnvironmentOfRadiusN()`

IchiruTake:
**Why does this PR:** 
I did the small project in which the properties of a particular atom inside a molecule can be predicted within a defined subgraph. The core idea worked well with `Subsgraph::findAtomEnvironmentOfRadiusN()`. But I don't like I have to write extra code to match up the `bondPath` and `radius`, so PR #4970 is introduced.

My second project is aimed on the particular bond (its property can also be predictable within a defined radius). 
In the legacy version, I have to merge the result from both connected atoms. Let's say the `bondID=5, atomIdx=(4, 5)`; thus I have to perform `Subsgraph::findAtomEnvironmentOfRadiusN()` twice and uniquely aggregating the result to run `pathToSubmol()`. Although there are zero issues arised but the time spent is relatively large `O(2 * (N*d))`. 

**What needed to be implemented** 
- Generalize the function `findAtomEnvironmentOfRadiusN()` into two smaller components which are `prepareNeighborStack()` and `findEnvironmentOfRadiusN()` into a private namespace.

**Result**
- Reduce time complexity from `2 * (N**d)` to `N**d + N`

**Future Work**
None, but if successfully, I wished the function `bond.getNeighborBonds()` or `bond.getBonds()` (by RDKit notation) can be implemented using the idea of `findBondEnvironmentOfRadiusN()`

### DIFF
--- a/Code/GraphMol/Subgraphs/Subgraphs.cpp
+++ b/Code/GraphMol/Subgraphs/Subgraphs.cpp
@@ -559,11 +559,11 @@ namespace {
   unsigned int findEnvironmentOfRadiusN(
     const ROMol &mol, unsigned int radius, PATH_TYPE &path,
     std::list<std::pair<int, int>> &nbrStack, 
-    boost::dynamic_bitset<> &bondsIn, bool useHs, 
-    std::unordered_map<unsigned int, unsigned int> *atomMap) {
+    std::unordered_map<unsigned int, int> &bondsInMap,
+    bool useHs, std::unordered_map<unsigned int, unsigned int> *atomMap) {
       
-      unsigned int maxRolledRadius;
-      for (maxRolledRadius = 0; maxRolledRadius < radius; ++maxRolledRadius) {
+      unsigned int i;
+      for (i = 0; i < radius; ++i) {
         if (nbrStack.empty()) {
           break;
         }
@@ -573,24 +573,24 @@ namespace {
           int bondIdx, startAtom;
           boost::tie(startAtom, bondIdx) = nbrStack.front();
           nbrStack.pop_front();
-          if (!bondsIn.test(bondIdx)) {
-            bondsIn.set(bondIdx);
+          if (bondsInMap.find(bondIdx) == bondsInMap.end()) {
+            bondsInMap[bondIdx] = -1;
             path.push_back(bondIdx);
 
             // add the next set of neighbors:
             int oAtom = mol.getBondWithIdx(bondIdx)->getOtherAtomIdx(startAtom);
             if (atomMap) {
               if (atomMap->find(oAtom) == atomMap->end()) {
-                (*atomMap)[oAtom] = maxRolledRadius + 1;
+                (*atomMap)[oAtom] = i + 1;
               } else {
-                (*atomMap)[oAtom] = std::min(atomMap->at(oAtom), maxRolledRadius + 1);
+                (*atomMap)[oAtom] = std::min(atomMap->at(oAtom), i + 1);
               }
             }
             // if we're going to do another iteration, then push the neighbors from
             // this round onto the stack
-            if (maxRolledRadius < radius - 1) {
+            if (i < radius - 1) {
               for (const auto bond : mol.atomBonds(mol.getAtomWithIdx(oAtom))) {
-                if (!bondsIn.test(bond->getIdx())) {
+                if (bondsInMap.find(bond->getIdx()) == bondsInMap.end()) {
                   if (useHs || mol.getAtomWithIdx(bond->getOtherAtomIdx(oAtom))
                                        ->getAtomicNum() != 1) {
                     nextLayer.emplace_back(oAtom, bond->getIdx());
@@ -602,9 +602,10 @@ namespace {
         }
         nbrStack = std::move(nextLayer);
       }
-      return maxRolledRadius;
+      return i;
     }
-}
+    
+} // Private namespace
 
 PATH_TYPE findAtomEnvironmentOfRadiusN(
     const ROMol &mol, unsigned int radius, unsigned int rootedAtAtom,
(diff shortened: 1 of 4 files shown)


IchiruTake:
The following code snipset described the idea
```python
def prepareNeighborStack(mol: Mol, atomIdx: int, neighborStack: List[Tuple[int, int]], useHs: bool):
    coreAtom = mol.GetAtomWithIdx(atomIdx)
    for bond in coreAtom.GetBonds():
        if useHs or bond.GetOtherAtom(coreAtom).GetAtomicNum() != 1:
            neighborStack.append((atomIdx, bond.GetIdx()))
    return None


def findEnvironmentOfRadiusN(mol: Mol, radius: int, bondPath: List[int], neighborStack: List[Tuple[int, int]],
                             visited: Dict[int, int], useHs: bool, atomMap: Dict[int, int]):
    maxRolledRadius = 0
    for dist in range(0, radius):
        if len(neighborStack) == 0:
            break

        canContinueIteration: bool = dist < (radius - 1)
        nextLayer: List[Tuple[int, ...]] = []
        while neighborStack:
            atomIdx, bondIdx = neighborStack.pop()
            if bondIdx not in visited:
                visited[bondIdx] = -1
                bondPath.append(bondIdx)

                nextAtomIdx = mol.GetBondWithIdx(bondIdx).GetOtherAtomIdx(atomIdx)
                if nextAtomIdx not in atomMap:
                    atomMap[nextAtomIdx] = dist + 1
                else:
                    atomMap[nextAtomIdx] = min(atomMap[nextAtomIdx], dist + 1)

                if canContinueIteration:
                    nextAtom = mol.GetAtomWithIdx(nextAtomIdx)
                    for neighbor_bond in nextAtom.GetBonds():
                        neighbor_bondIdx = neighbor_bond.GetIdx()
                        if neighbor_bondIdx not in visited:
                            if useHs or mol.GetAtomWithIdx(neighbor_bond.GetOtherAtomIdx(nextAtomIdx)).GetAtomicNum() != 1:
                                nextLayer.append((nextAtomIdx, neighbor_bondIdx))
        neighborStack = nextLayer
        maxRolledRadius = dist + 1
    return maxRolledRadius


def PySearchBondEnvV3(mol: Mol, bondIdx: int, radius: int, useHs: bool,
                      atomMap: Dict[int, int], enforceSize: bool = False) -> List[int]:
    # [1]: Initialization
    coreBond = mol.GetBondWithIdx(bondIdx)
    beginAtomIdx = coreBond.GetBeginAtomIdx()
    endAtomIdx = coreBond.GetEndAtomIdx()
    atomMap[beginAtomIdx] = 0
    atomMap[endAtomIdx] = 0

    neighborStack = []
    prepareNeighborStack(mol, atomIdx=beginAtomIdx, neighborStack=neighborStack, useHs=useHs)
    prepareNeighborStack(mol, atomIdx=endAtomIdx, neighborStack=neighborStack, useHs=useHs)
    # neighborStack = [(aId, bId) for aId, bId in neighborStack if bId != bondIdx] # Optional

    bondPath = [bondIdx]
    visited: Dict[int, int] = {bondIdx: -1}
    maxRolledRadius: int = findEnvironmentOfRadiusN(mol, radius, bondPath, neighborStack, visited, useHs, atomMap)
    if maxRolledRadius != radius and enforceSize:
        bondPath.clear()
        atomMap.clear()
    return bondPath

```

IchiruTake:
@greglandrum If you have time, can you take a look at it? It would be great. Although I can produce relatively good on Python with result, I don't understand where the failure came from. It was good on FindAtom but not on FindBond. 
Thank you very much. 

I hope Azure can accept my request as relying on RDKit build of C++ is not a good idea. 

P/s: I cannot build the C++ version of RDKit. If there is an explicit way, please let me know. Thank you!

IchiruTake:
@greglandrum Thank you for your support. I wish there is a donation channel to donate a little bit for my excessive use of Azure for 2 C++ PR.

First I has switched it to std::unordered_map instead of boost::dynamic_bitset. It may cause a little bit of memory extra on small molecules, but on average with large molecule and  small radius, this gain is for 'map' instead.

Related to your question about bondMap usage, I believe we can extract all bonds within a smaller radius. 
Let's say I have a molecule and I need to extract into 2 or more submols at different radius, (4, 2) or (6, 4, 2). By that, I can run function at radius 6 and apply filtering in bondMap to get the result without re-run the algorithm.

Thank you once again. I am welcomed to see your response

IchiruTake:
In fact, if our molecular graph space is just one normal molecule (no fragment nor broken bond), we can set it to be particularly useful for any scenario that is not limited to hydrogen: for example, if the `atomicNum=0` and `bond=BondType::DOUBLE` we can skip the traversing, or even nitrogen or hydrogen. The function I attempted to applies here is particularly useful. When I write the function on Python, the algorithm is faster in C++ for molecule with less than 10 heavy atoms or even more if we can scale the rule. It may not be approraite to do so, but it is also worth trying

``` python
class MolUtility:

    @staticmethod
    def EvalAtomForNeighborLoop(atom, bond) -> bool:
        """
            This function is particularly useful as it can be applied in many function that need
            neighbor traversing.  Logical operation: not condition
            Example (1): in function _FindEnvWith..._ to prevent looping if condition is not fulfilling.
            If True, the atom may definitely have neighbors to perform looping (Most frequent atom first).
        """
        AtomicNum: int = atom.GetAtomicNum()
        if AtomicNum == 1:
            return False
        if AtomicNum == 6 or bond is None:
            return True
        if AtomicNum == 8:
            return bond.GetBondType() != 2  # rdBondType.DOUBLE
        if AtomicNum == 7:
            return bond.GetBondType() != 3 or atom.GetFormalCharge() != 0
        if AtomicNum in (9, 17, 35, 53):
            return False
        return True

    @staticmethod
    def IsNotHydrogen(atom) -> bool:
        return atom.GetAtomicNum() != 1
```

PR #5103, opened 2022-03-17
Fixes #5017

### LINKED ISSUE #5017: Mol images in DataFrames are drawn only once in Jupyter Lab
**Describe the bug**
With the latest RDKit version `2021.09.4`, the images of molecules in DataFrames are only drawn on the first display when using Jupyter Lab. On subsequent display attempts, the images are no longer drawn. 

I have tested it with Pandas versions `1.4.1`, `1.4.0`, `1.3.5`, `1.2.5` and `1.1.5` and the same problem occurs with all of them.

The error does not occur with RDKit version `2021.09.2` in combination with Pandas `1.4.1` or `1.1.5`, the DataFrames are all displayed correctly here. 

With RDKit version `2021.09.3`, the already fixed error mentioned in ticket #4864 occurs with Pandas `1.4.1` as well as `1.1.5`.

If you look specifically at the Jupyter notebook attached below in a text editor, you will see that on the first call, the molecule image is given as `<img data-content=\"rdkit ...`, while on the second call, the entry looks like `&lt;img data-content=\"rdkit ...`.

**To Reproduce**
Basically, all that needs to be done in a Jupyter notebook is to create any DataFrame with RDKit mol objects and then display it with e.g. `df.head()`. If the same cell is executed again, the molecules are no longer drawn.

This is an example notebook: [1.4.1_2021.09.4_bug.ipynb.txt](https://github.com/rdkit/rdkit/files/8087891/1.4.1_2021.09.4_bug.ipynb.txt)

**Expected behavior**
The molecule images should be shown correctly in every call.

**Screenshots**
How it looks like:
bug

How it should look:
works

 - RDKit version: **2021.09.4**
 - OS: **Windows 10 and Ubuntu 18.04**
 - Python version (if relevant): **3.9.10**
 - Are you using conda? **Yes**
 - If you are using conda, which channel did you install the rdkit from? **conda-forge**

ptosco:
#### Reference Issue
This PR fixes #5017

`patchPandasrepr()` patches `pd.io.formats.html.HTMLFormatter._write_cell()`, then it restores the original function before returning. However, while doing so, it also sets an `_rdkitpatched` attribute on `pandas.io.formats.html.HTMLFormatter`.
The next time patchPandasrepr() is called, the patch is not applied anymore due to the present of the `_rdkitpatched` attribute, even though the patch has actually been reverted.
This explains why the DataFrame displays molecules as PNG images only the first time.
The fix is removing the usage of `_rdkitpatched` attribute entirely as it is not necessary, since the patch is applied and then reverted. Instead, a `try..finally` block is used to make sure that the patch is reverted before returning even if an exception is raised.

After this PR, running the test described in #5017 yields the desired behavior:

```
from rdkit import Chem
import pandas as pd

import rdkit
rdkit.__version__
'2022.03.1pre'

df = pd.DataFrame(dict(a=[1], s=['CCN']))
PandasTools.AddMoleculeColumnToFrame(df, smilesCol='s', molCol='m')
df
```
image
```
df
```
image
```
PandasTools.RenderImagesInAllDataFrames(False)
df
```
image
```
df
```
image
```
PandasTools.RenderImagesInAllDataFrames(True)
df
```
image
```
df
```
image

### DIFF
--- a/rdkit/Chem/PandasTools.py
+++ b/rdkit/Chem/PandasTools.py
@@ -197,19 +197,19 @@ def patchPandasrepr(self, **kwargs):
   global defPandasGetAdjustment
 
   import pandas.io.formats.html
-  if not hasattr(pandas.io.formats.html.HTMLFormatter, "_rdkitpatched"):
-    defHTMLFormatter_write_cell = pd.io.formats.html.HTMLFormatter._write_cell
-    pd.io.formats.html.HTMLFormatter._write_cell = _patched_HTMLFormatter_write_cell
-    pandas.io.formats.html.HTMLFormatter._rdkitpatched = True
-  get_adjustment_attr = getAdjustmentAttr()
-  if get_adjustment_attr:
-    defPandasGetAdjustment = getattr(pd.io.formats.format, get_adjustment_attr)
-    setattr(pd.io.formats.format, get_adjustment_attr, _patched_get_adjustment)
-  res = defPandasRepr(self, **kwargs)
-  if get_adjustment_attr:
-    setattr(pd.io.formats.format, get_adjustment_attr, defPandasGetAdjustment)
-  pd.io.formats.html.HTMLFormatter._write_cell = defHTMLFormatter_write_cell
-  return res
+  defHTMLFormatter_write_cell = pd.io.formats.html.HTMLFormatter._write_cell
+  pd.io.formats.html.HTMLFormatter._write_cell = _patched_HTMLFormatter_write_cell
+  try:
+    get_adjustment_attr = getAdjustmentAttr()
+    if get_adjustment_attr:
+      defPandasGetAdjustment = getattr(pd.io.formats.format, get_adjustment_attr)
+      setattr(pd.io.formats.format, get_adjustment_attr, _patched_get_adjustment)
+    res = defPandasRepr(self, **kwargs)
+    if get_adjustment_attr:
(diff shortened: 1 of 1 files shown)


ptosco:
The `macOS_x64_java` failure was caused by Avalon failing to download from SourceForge, nothing to do with this PR.

greglandrum (approved):
LGTM

greglandrum:
@ptosco : I merged this without actually trying it, and it doesn't seem to actually work in jupyter:
image
I've tried it in both jupyter notebook and jupyterlab (v3.4.4) and get the same behavior in both of them 

Which version/OS did you test this on?

ptosco:
@greglandrum The fix works, the problem is that I inadvertently reverted it in PR #5132: 
https://github.com/rdkit/rdkit/pull/5132/files#diff-4d3e0e17d686a9de85fb42b78a37f8fe69e95258f4e6be45899873fa292a2090L200
because I did not merge #5103 into my `rdkit-structure-renderer` branch before submitting #5132.
Sorry for the inconvenience, I have just submitted #5381 to restore the #5103 fix.

greglandrum:
@ptosco : sorry, I should have caught that in the review... the PandasTools code is a real brain bender though.

It would be super shiny if we could figure out some way to actually automatically test this stuff, but I don't see how that would work.

PR #5128, opened 2022-03-24
Swig MolDraw2D cairo

jones-gareth:

Add SWIG bindings for `MolDraw2DCairo` so you can create PNG images in Java and C#

Added `toByteArray()` in Java and `getImage()` in C# to return image bytes.

There is a Java test and a C# test

#### Any other comments?

The C# test is just a standalone file.  I have plans to automate C# testing, which I think can be done using dot net core, but that is a larger project.  If it doesn't make sense to include the test file I can take it out of the PR.


greglandrum (approved):
Looks good to me.
Thanks for including tests for both sets of wrappers!

PR #5148, opened 2022-03-29
Modified the JS tests to comply with older nodejs versions

ptosco:
`String.prototype.matchAll()` is only available in `nodejs` >= 12.
This PR switches to using `String.prototype.match()` to avoid the problem.


greglandrum (approved):
yay! this removes the bit of javascript that I found particularly nauseating.
:-)

PR #5204, opened 2022-04-15
Make atom, bond iterators usable in STL algorithms

ricrogz:
This is something that had been bugging me for some time: I had tried a couple of times to use atom and/or bond iterators with the STL algorithms, but never got it to work. Now I finally looked it up, and it turned out to be a quite easy fix, it turns out we just had to define some tags on the iterators!

### DIFF
--- a/Code/GraphMol/ROMol.h
+++ b/Code/GraphMol/ROMol.h
@@ -19,6 +19,8 @@
 #define RD_ROMOL_H
 
 /// Std stuff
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
Oh, wow, that's so simple. Thanks for figuring out how to make that work.

There are now *even more* opportunities open for refactoring old code. :-)

bp-kelley:
Nice!

PR #5205, opened 2022-04-16
Fixes array overflow in FMCS code

ricrogz:
I've been playing with address sanitization on Python tests (fortunately, asan is way faster than valgrind when working around Python).

Some Python test revealed an array overflow in line 204 of Code/GraphMol/FMCS/MaximumCommonSubgraph.cpp. I'm not familiar with the FMCS code, but this looks like a bug to me.

Sadly, I don't know how to write a test that would expose this.

### DIFF
--- a/Code/GraphMol/FMCS/MaximumCommonSubgraph.cpp
+++ b/Code/GraphMol/FMCS/MaximumCommonSubgraph.cpp
@@ -114,8 +114,8 @@ void MaximumCommonSubgraph::init() {
   unsigned currentLabelValue = 1;
   std::vector<LabelDefinition> labels;
   nq = QueryMolecule->getNumAtoms();
-  QueryAtomLabels.resize(nq);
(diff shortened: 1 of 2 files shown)


Code/GraphMol/FMCS/MaximumCommonSubgraph.cpp
```diff
@@ -157,8 +157,8 @@ void MaximumCommonSubgraph::init() {
   labels.clear();
   currentLabelValue = 1;
   nq = QueryMolecule->getNumBonds();
-  QueryBondLabels.resize(nq);
-  for (size_t aj = 0; aj < nq; aj++) {
```
bp-kelley, line 161: I think using  "aj" as a bond iterator was the recipe for disaster here.

 nice catch.

greglandrum (approved):
LGTM

PR #5211, opened 2022-04-17
Fix use of not thread safe function localtime() 

ricrogz:
`localtime()` is not thread safe (https://linux.die.net/man/3/localtime): 

> 
> The four functions asctime(), ctime(), gmtime() and localtime() return a pointer to static data and hence are not thread-safe. 
> Thread-safe versions asctime_r(), ctime_r(), gmtime_r() and localtime_r() are specified by SUSv2, and available since libc 5.2.5. 
> 

RDkit uses `localtime()` in logging, and the thread sanitizer flags its usage as a data race. Fortunately, it's easy to fix.


greglandrum (approved):
We should probably switch this code to use C++ time functions, but that’s clearly a different PR

PR #5232, opened 2022-04-24
Fix github5222

### LINKED ISSUE #5222: Core with query atoms and no user definded attachment points may create poor decompostions
**Describe the bug**
The tie breaking algorithm used in RGD can result in poor decompositions when the core has query features.

When the core is this:

image

And we decompose these guys:
image

The decomposition of the Pyrimidine is wrong:

image
 
**To Reproduce**
```python
#%%

from rdkit import Chem
from rdkit.Chem.rdRGroupDecomposition import (RGroupDecomposition,
                                              RGroupDecompositionParameters,
                                              RGroupMatching, RGroupScore)

import pandas as pd
from rdkit.Chem import Draw, PandasTools

#%%

smi = ['COc1ccccc1']*10
smi.append('COc1ccncn1')
mols = [Chem.MolFromSmiles(smiles) for smiles in smi]
Draw.MolsToImage(mols)

#%%
block="""
  ChemDraw04112214222D

  6  6  3  0  0  0  0  0  0  0999 V2000
   -0.7145    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.7145   -0.4125    0.0000 L   0  0  0  0  0  0  0  0  0  0  0  0
    0.0000   -0.8250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.7145   -0.4125    0.0000 L   0  0  0  0  0  0  0  0  0  0  0  0
    0.7145    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.0000    0.8250    0.0000 L   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  2  0      
  2  3  1  0      
  3  4  2  0      
  4  5  1  0      
  5  6  2  0      
  6  1  1  0      
  2 F    2   6   7
  4 F    2   6   7
  6 F    2   6   7
M  ALS   2  2 F C   N   
M  ALS   4  2 F C   N   
M  ALS   6  2 F C   N   
M  END
"""
core = Chem.MolFromMolBlock(block)
core
#%%

def do_decomp(groups):
    options = RGroupDecompositionParameters()
    options.matchingStrategy = RGroupMatching.GreedyChunks
    group_decomp = RGroupDecomposition([core], options)
    for mol in groups:
        group_decomp.Add(mol) 
    group_decomp.Process()
    cols = group_decomp.GetRGroupsAsColumns()
    df = pd.DataFrame(cols)
    return df

#%%
df = do_decomp(mols)
HTML(df.to_html())

```

**Expected behavior**
I expect to see all methoxy groups in R1

RDKit master on Ubuntu 20.04

**Additional context**

This issue arises because of the tie-breaking code.  For unindexed (not user defined) R group positions the tie breaking code will push heavy groups onto the lowest index.  For the compounds with the benzene ring that is -6.  After adding a few compounds the greedy chunks matcher will call the tie breaker and put all the methoxy groups on -6. The problem is that the pyrimidine core that matches the final compound has only 5 attachment points and doesn't have a -6 index.

I have a fix for the tie breaking code and should have a PR soon.  The fix is easy, but unfortunately breaks a number of tests.

This issue is somewhat related to issue #4505.  At least if we could solve #4505 for unindexed and user rgroups this issue would not arise.  I'm just not sure how to fix #4505.

jones-gareth:
#### Reference Issue

Fixes #5222 

Fixes #5222 by tie breaking on the highest unindexed labels (where an index label is not a user defined label) rather than the lowest.

I've included a new unit text to confirm the fix.

#### Any other comments?

I had to edit failing tests where the labels in the expected decomposition were wrong (`github1705`, `testScorePermutations`, three tests in `catch_rgd.cpp` and `simple1.out1.json`  for `rgd_testing.ipynb`).  In each case I've confirmed that the new decomposition is correct.


Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
```diff
@@ -314,7 +314,8 @@ struct RGroupDecompData {
     }
 
     // Deal with non-user supplied labels
-    for (auto newLabel : indexLabels) {
+    for (auto it = indexLabels.begin(); it != indexLabels.end(); it++) {
```
bp-kelley, line 317: Why the change here?

And (I'm very guilty of this my self) ++it

bp-kelley: I think this can be reverted as you aren't changing the iterator order.

bp-kelley (changes requested):
One small reversion, but otherwise looks good to me.

Code/GraphMol/RGroupDecomposition/RGroupDecompData.h
jones-gareth: I had put in a reverse iterator as I thought it could reduce the number of test failures (it didn't), then forgot to revert.

it++ - too much C and Java!

I have reverted in the latest commit.

greglandrum (approved):
LGTM

PR #5237, opened 2022-04-26
Allow linking against Python 3.6

ptosco:
`_Py_IsFinalizing()` was introduced in Python 3.7, so RDKit will fail to build against Python 3.6, as I realized today.
I do realize that Python 3.6 is deprecated, but I think it is a bit too early to prevent building RDKit against it.
As I understand the workaround that I propose will only cause a minor memory leak.

### DIFF
--- a/Code/RDBoost/Wrap/RDBase.cpp
+++ b/Code/RDBoost/Wrap/RDBase.cpp
@@ -78,11 +78,11 @@ struct PyLogStream : std::ostream, std::streambuf {
   }
 
   ~PyLogStream() {
-    #if PY_VERSION_HEX > 0x03070000
+#if PY_VERSION_HEX > 0x03070000
     if (!_Py_IsFinalizing()) {
       Py_XDECREF(logfn);
     }
-    #endif
+#endif
   }
 
   int overflow(int c) override {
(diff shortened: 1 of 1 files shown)


Code/RDBoost/Wrap/RDBase.cpp
```diff
@@ -78,9 +78,11 @@ struct PyLogStream : std::ostream, std::streambuf {
   }
 
   ~PyLogStream() {
+    #if PY_VERSION_HEX > 0x03070000
```
bp-kelley, line 81: I think the rest of the code base has the # on the first character.

I’m sadly old enough when this was required.  :)

ptosco: Oh, I hadn’t noticed! I’ll fix it straightaway.

greglandrum (commented):
Given that Python 3.6 hasn't received any form of patching (security or otherwise) since Sept of 2021 (https://peps.python.org/pep-0494/), which was 6 months before the release which included this dependency, and is no longer supported by conda-forge, I am reluctant to add noise to the code in order to continue to support it.

@ptosco : if this is important to you, I'll go along with it anyway though

ptosco:
@greglandrum Yeah, I realized how old and unsupported Python 3.6 is _after_ submitting this PR, and I completely agree with you. I am closing this.

PR #5240, opened 2022-04-27
Fixes #5239

### LINKED ISSUE #5239: Precondition violation on chiral Atoms with zero order bonds
Since we merged #5201 to fix #5196, actual chiral centers which are attached to one of the bond types we don't consider for chirality raise a precondition violation when trying to resolve their chirality:

```python
mb = """
     RDKit          3D

M  V30 BEGIN CTAB
M  V30 COUNTS 5 4 0 0 0
M  V30 BEGIN ATOM
M  V30 1 C -0.446600 -0.713700 1.305600 0
M  V30 2 Fe -1.628200 -0.983200 -0.412000 0
M  V30 3 Cl -0.049300 -1.876700 2.613900 0
M  V30 4 C -1.544600 0.306200 1.588200 0
M  V30 5 F 0.673700 0.029200 0.993700 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 1 1 3
M  V30 2 1 1 4 CFG=1
M  V30 3 1 1 5
M  V30 4 0 2 1
M  V30 END BOND
M  V30 END CTAB
M  END
"""
# if we sanitize here, the issue happens right away, and import fails
m = Chem.MolFromMolBlock(mb, sanitize=False)
Chem.AssignStereochemistryFrom3D(m)
---------
[11:08:37] bond with order 0 found on line 18. This is not part of the MDL specification.
[11:08:37] 

****
Pre-condition Violation
size mismatch
Violation occurred on line 55 in file /tmp/rdkit_builder/rdkit/Code/RDGeneral/utils.h
Failed Expression: ref.size() == probe.size()
****

Input In [1], in <cell line: 25>()
     23 # if we sanitize here, the issue happens right away, and import fails
     24 m = Chem.MolFromMolBlock(mb, sanitize=False)
---> 25 Chem.AssignStereochemistryFrom3D(m)

	size mismatch
	Violation occurred on line 55 in file Code/RDGeneral/utils.h
	Failed Expression: ref.size() == probe.size()
	RDKIT: 2022.09.1pre
	BOOST: 1_76

```
 I think I have a fix, I'll post a PR soon.

ricrogz:
The problem was happening because, as a side effect of not considering zerp/coordinate/unspecified bonds when determining if an atom could be chiral, these bonds were also not added to the list of bonds used to calculate parity. But they still were included in the other list, the one we use as reference, which caused the precondition violation inside `Atom::getPerturbationOrder()`, which requires both lists to have the same size.

This fixes #5239  by also excluding zero/coordinate/unspecified bonds from the reference list.

### DIFF
--- a/Code/GraphMol/Atom.cpp
+++ b/Code/GraphMol/Atom.cpp
@@ -23,6 +23,19 @@
 
 namespace RDKit {
 
+unsigned int getAtomNonzeroDegree(const Atom *atom) {
+  PRECONDITION(atom, "bad pointer");
+  PRECONDITION(atom->hasOwningMol(), "no owning molecule");
+  unsigned int res = 0;
+  for (auto bond : atom->getOwningMol().atomBonds(atom)) {
+    if (!bondAffectsAtomChirality(bond, atom)) {
+      continue;
+    }
+    ++res;
+  }
+  return res;
+}
+
 bool isAromaticAtom(const Atom &atom) {
   if (atom.getIsAromatic()) {
     return true;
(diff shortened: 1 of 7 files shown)


ricrogz:
Hm. There's more to consider here. Like, we'll have to patch the CIP labeler. It will also try to use zero/coordinate/undefined bonds when calculating CIP labels... and fail or even segfault.

ricrogz:
Closing in favor of Greg's fix.

PR #5343, opened 2022-05-31
Update docs for ReplaceSubstructs

i-tub:
- Use O[CH3] instead of OC for the existing examples, to ensure that the
  the examples produce the output shown (assuming the query is parsed as
  a SMARTS).
- Add a note and an example showing how bonds are formed only to the
  atoms that were connected to the first atom in the substructure query,
  and suggest using ChemicalReaction for finer control.

#### Reference Issue

Documents behavior explained in #5329


greglandrum (approved):
LGTM

PR #5398, opened 2022-07-01
Fix GCC 12/Qt build after #5387

ricrogz:
After merging #5387, builds using GCC 12 and Qt fail with:

```
/tmp/rdkit_builder/rdkit/Code/GraphMol/MolDraw2D/Qt/MolDraw2DQt.cpp:136:18: error: aggregate ‘QPainterPath path’ has incomplete type and cannot be defined
  136 |     QPainterPath path;

```

Adding this include fixes the build.


greglandrum (approved):
LGTM

PR #5446, opened 2022-07-22
MolZip: add atom property labels for mol-zipping

bp-kelley:
Implements #5451 

This adds an option to allow unsigned integer (and integer) properties atom labels to indicate which atoms should be zipped together.

This will be primarily used in the cdxml parser.


greglandrum (approved):
LGTM

PR #5452, opened 2022-07-23
Fix poor join between wedge and double bond.

DavidACosgrove:
#### Reference Issue
No issue to reference

The new drawing code from #5425 did a very poor job joining a wedged bond to a double bond, as seen in acs1996_9.svg.  This fixes that.


greglandrum (approved):
That's a pretty subtle change (to my eye at least), but this does look better.

PR #5478, opened 2022-08-01
Fixes "DeprecationWarning: invalid escape sequence \C" when importing EnumerateStereoisomers

ricrogz:
Currently, importing EnumerateStereoisomers triggers this warning:
```
In [4]: from rdkit.Chem import EnumerateStereoisomers
/Users/rodrigue/Documents/code/rdkit-install/lib/python3.8/site-packages/rdkit/Chem/EnumerateStereoisomers.py:165: DeprecationWarning: invalid escape sequence \C
  """ returns a generator that yields possible stereoisomers for a molecule

```
The warning doesn't seem too bad, as anything still works as expected, but when I use `EnumerateStereoisomers` in my pytest unit tests, it causes them to fail due to the unexpected warning while tests are collected.

This patch fixes the issue but making the docstring literal, and removing the duplicate slashes so that the doctests still pass.

### DIFF
--- a/rdkit/Chem/EnumerateStereoisomers.py
+++ b/rdkit/Chem/EnumerateStereoisomers.py
@@ -162,7 +162,7 @@ def GetStereoisomerCount(m, options=StereoEnumerationOptions()):
 
 
 def EnumerateStereoisomers(m, options=StereoEnumerationOptions(), verbose=False):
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
Thanks for fixing this. I had been studiously ignoring the warning for a while

PR #5479, opened 2022-08-01
Fix double to float downcasting warning

ricrogz:
Not a big issue, but the Visual Studio compiler raises a C4305 warning when using this header. If anyone includes this header while building with the `/WX` flag (treat all compiler warnings as errors), the build will fail.

The cause of the issue is that we are using a literal double value to initialize a float value, which counts as downcasting. Making the value a literal float fixes the issue.


bp-kelley (approved):

greglandrum (approved):
LGTM

PR #5503, opened 2022-08-17
MolDraw2D::drawMolecules() should not crash on null molecules

ptosco:
This bug was originally reported by Richard Lewis.
`MolDraw2D::drawMolecules()` under certain conditions will attempt to access non-existent vector items and hence segfault.
This simple PR fixes the problem and adds a test.

### DIFF
--- a/Code/GraphMol/MolDraw2D/MolDraw2D.cpp
+++ b/Code/GraphMol/MolDraw2D/MolDraw2D.cpp
@@ -191,6 +191,7 @@ void MolDraw2D::drawMolecules(
   int nCols = width() / panelWidth();
   int nRows = height() / panelHeight();
 
+  size_t j = 0;
   for (size_t i = 0; i < mols.size(); ++i) {
     if (!mols[i]) {
       continue;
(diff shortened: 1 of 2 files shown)


Code/GraphMol/MolDraw2D/MolDraw2D.cpp
```diff
@@ -231,12 +232,13 @@ void MolDraw2D::drawMolecules(
     drawMols_.back()->setOffsets(col * panelWidth(), row * panelHeight());
     drawMols_.back()->createDrawObjects();
     if (drawMols_.back()->getScale() < drawMols_[minScaleMol]->getScale()) {
-      minScaleMol = i;
```
DavidACosgrove, line 234: Oops, sorry.
You could perhaps use drawMols_.size() here rather than carrying j around.  It might be a bit more immediately obvious what the intention was.

ptosco: @DavidACosgrove Thanks for your review, I think it is a good idea and I have made the change you suggested.

greglandrum (approved):
LGTM

PR #5507, opened 2022-08-19
Performance issue with setting aromaticity for very large ring systems

rachelnwalker:
Greg, I know you mentioned a couple of other ideas for how to approach this issue. This PR is mostly a draft to see what you think.

We noticed that aromaticity detection is taking an extremely long time for very dense ring systems. We have one example with around ~1000 rings in a single system that takes at least 8 hours to complete (I ran it overnight and it hadn’t finished yet).

It looks like this happens because`applyHuckelToFused` checks whether every subset of rings within a ring system is fused, which is a lot of combinations for large fused ring systems. I tried setting maxNumFusedRings to 2, however this still took ~2.5 seconds to finish for this particular structure and didn’t set aromaticity for all ring bonds.

The best solution I could think of was to enumerate subgraphs of the ring neighbor graph directly instead trying all combinations of rings then checking if they are fused. This PR is a simpler version of that — basically just enumerating all sub-paths of the ring system graph. When the number of fused rings is >= 4, this approach will definitely miss some some of the combinations. Because of this, I think it only makes sense to use this approach when the current approach would take an unreasonable amount of time to complete. For now, that is determined by the number of rings in a ring system (I though 500 would be a good place to start but if we want to merge this we should probably do a little more testing to determine what the right number is).

For reference, here are the times for a couple of structures with a single, very dense ring system:

Nanotube (2412 atoms, 1182 rings — doesn’t make a performance difference): 
	 Time elapsed (before): .27s
        Time elapsed (now): .26s
        Aromatic atom count: 2388
        Aromatic bond count: 3570

Polymer crosslink-out (2304 atoms, 889 rings): 
        Time elapsed (before): very long time, never finished for me
        Time elapsed (now): .76s
        Aromatic atom count: 2160
        Aromatic bond count: 3168

### DIFF
--- a/Code/GraphMol/Aromaticity.cpp
+++ b/Code/GraphMol/Aromaticity.cpp
@@ -159,6 +159,29 @@ typedef VECT_EDON_TYPE::const_iterator VECT_EDON_TYPE_CI;
 static bool applyHuckel(ROMol &mol, const INT_VECT &ring,
                         const VECT_EDON_TYPE &edon, unsigned int minRingSize);
 
+static void applyHuckelToSubFusedSystem(
+    ROMol &mol,                   // molecule of interest
+    const VECT_INT_VECT &srings,  // list of all ring as atom IDS
+    const VECT_INT_VECT &brings,  // list of all rings as bond ids
+    const INT_VECT &curRs,  // list of current rings in subs fused system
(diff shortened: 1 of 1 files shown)


rachelnwalker:
Closing this for now in favor of https://github.com/rdkit/rdkit/pull/5648

PR #5519, opened 2022-08-27
Fixed scale

DavidACosgrove:
#### Reference Issue
No issue.

When using ACS1996 mode on a drawer with fixed size (not -1, -1 for width and height) the drawing was being scaled as normal.  ACS1996 mode is supposed to always have bonds of length 14.4px.


greglandrum (approved):
LGTM

Issue #5569, opened 2022-09-15
RGD may yield poor depictions


jones-gareth:
**Describe the bug**

When decomposing structures and cores with coordinates the RGD algorithm will copy those coordinates to the output groups.

In some cases the attachment dummy atom is poorly positioned

**To Reproduce**

```python
# %%

from rdkit import Chem
from rdkit.Chem import AllChem
from rdkit.Chem.rdRGroupDecomposition import RGroupDecomposition, RGroupDecompositionParameters, \
    RGroupMatching, RGroupScore, RGroupLabels, RGroupCoreAlignment, RGroupLabelling
from rdkit.Chem.Draw import IPythonConsole

IPythonConsole.ipython_useSVG = True

from rdkit.Chem import PandasTools
from rdkit.Chem import Draw
from IPython.display import HTML

import pandas as pd

# %%

core_block = """
  ChemDraw09152209202D

  6  6  0  0  0  0  0  0  0  0999 V2000
    0.0001    0.8250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.7144    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.7144   -0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.0001   -0.8250    0.0000 N   0  0  0  0  0  0  0  0  0  0  0  0
    0.7144   -0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.7144    0.4125    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0      
  2  3  1  0      
  3  4  1  0      
  4  5  1  0      
  5  6  1  0      
  6  1  1  0      
M  END
"""
core = Chem.MolFromMolBlock(core_block)
core
# %%

test_block = """MolHeader
     RDKit          2D

 20 21  0  0  0  0  0  0  0  0999 V2000
   -2.6437    1.7625    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -3.3582    1.3500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -3.3582    0.5250    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -2.6438    0.1125    0.0000 N   0  0  0  0  0  0  0  0  0  0  0  0
[...]

PR #5570, opened 2022-09-16
Fixed R-group attachment positions in RGD

jones-gareth:
#### Reference Issue
Fixes https://github.com/rdkit/rdkit/issues/5569

See summary in next comment

### DIFF
--- a/Code/GraphMol/ChemTransforms/ChemTransforms.cpp
+++ b/Code/GraphMol/ChemTransforms/ChemTransforms.cpp
@@ -506,7 +506,6 @@ ROMol *replaceCore(const ROMol &mol, const ROMol &core,
             });
   std::vector<std::pair<int, Atom *>> dummies;
 
-
   std::list<Bond *> allNewBonds;
   for (const auto &match : matches) {
     const auto &mappingInfo = match.second;
@@ -637,7 +636,7 @@ ROMol *replaceCore(const ROMol &mol, const ROMol &core,
       for (auto &newBond : newBonds) {
         newBond->setProp(replaceCoreDummyBond, 1);
         newMol->addBond(newBond, true);
-     }
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


jones-gareth:
Restoring the order of adding atom positions (first set terminal coordinates, next apply guesses from the `dummyAtomMap`) gives a decent depiction:
image

I noticed that, in master, any coordinates set by `MolOps::setTerminalAtomCoords` would be discarded by the later call to `updateSubMolConfs`.

jones-gareth:
`dummyAtomMap` in `RDKit::replaceCore` maps neighbor indices to dummy atoms.  However, a single neighbor index be mapped to multiple dummy atoms.  I have reversed the map and everything looks good.

image

I'm not sure that we need the terminal coordinate calculation for the dummy atoms at all.

jones-gareth:
**A summary of the changes here**

**Original logic**

In `RDKit::replaceCore` dummy atom coordinates are set firstly using `MolOps::setTerminalAtomCoords` when new bonds are created and secondly (from neighbors in the input structure) using `dummyAtomMap` towards the end of the function.

**Issues**

- The dummy atom coordinates set by `MolOps::setTerminalAtomCoords` were not correct or not calculated as the core atoms were still part of the new molecule (so the neighbor atom had incorrect valence)
- Regardless of what was set by `MolOps::setTerminalAtomCoords` those coordinates were removed by `updateSubMolConfs`
- Because `dummyAtomMap` mapped neighbor index to new atom it was lossy as multiple new atoms can be mapped to a single neighbor index.

**Fixes**

- Reverse `dummyAtomMap` so it is no longer lossy
- Add dummy atom coordinates using `MolOps::setTerminalAtomCoords` after removing the core and calling `updateSubMolConfs`, but before setting coordinates from `dummyAtomMap`

**Remaining Issues**

- Not sure if these is ever a dummy atom that has coordinates set by `MolOps::setTerminalAtomCoords` but not by `dummyAtomMap`, so perhaps we can remove the code that calculates the coordinates using `MolOps::setTerminalAtomCoords`
- If a fragment is cyclic on the same core atom then both dummy atoms will have the same coordinates.  This seems correct to me, but may cause depiction issues.

The rgroup depiction now correctly reflects the input structure coordinates:

image

bp-kelley (approved):

greglandrum (approved):
LGTM

greglandrum:
Thanks @jones-gareth

PR #5587, opened 2022-09-23
Remove unnecessary compiler flags that would be ignored anyway

ptosco:
`clang` legitimately emits warnings about those flags which are meant for the linker, that the compiler, and which are anyway ignored while compiling. This PR removes those unnecessary flags, and hence the warnings.


greglandrum (approved):
LGTM

PR #5619, opened 2022-10-05
Avoid incurring into division by zero in normalizeDepiction

ptosco:
Small fix to avoid incurring into division by zero, e.g., when coordinates of the molecule are all zero.


greglandrum (approved):
LGTM

PR #5628, opened 2022-10-08
PandasTools and InteractiveRenderer improvements

ptosco:
1. Avoid that Jupyter Notebooks become painfully slow when using PandasTools with hundreds of molecules by using `pandas` formatters rather than patching `Chem.Mol.__str__`
2. Make sure that static PNG images in `pandas` `DataDrames` honor `PandasTools.molSize`
3. Make sure that `PandasTools` works on the whole range of `pandas` versions
4. Improved and more robust loading of `rdkit-structure-renderer` in Jupyter Notebooks
5. `MolsToGridImage` is now also interactive
6. Interactive rendering now displays highlights and query molecules correctly

Take a look at the gists linked below to see how the new `PandasTools` works correctly on a wide range of `pandas` versions and avoids incurring in the sluggishness that currently affects notebooks using `PandasTools` on hundreds of molecules.

This PR `PandasTools` on various versions of `pandas`:
[1.5.0](https://gist.github.com/ptosco/c53d3dad1275e59cd129b90ba52aea11)
[1.4.3](https://gist.github.com/ptosco/989e5995c39bf64769781d6379a1477f)
[1.3.5](https://gist.github.com/ptosco/a049c08e1603e0ff723b15b181ae8158)
[1.2.5](https://gist.github.com/ptosco/69fdc01b03de3fedb9c61812e6dfa331)
[1.1.5](https://gist.github.com/ptosco/ad2e7b4ea73256e8760fafa1823adab4)
[1.0.5](https://gist.github.com/ptosco/459e3cbb4356e48c9dd2df144b4733fd)
[0.25.3](https://gist.github.com/ptosco/d06a3c1354fff6f8bb44f38b0f3df3a2)

`master` `PandasTools` performance issues with many molecules in the notebook and other visualization issues:
[Jupyter painfully slow with many molecules](https://gist.github.com/ptosco/807215ae3f329d3d9a48403511d269ff)
[PandasTools visualization issues](https://gist.github.com/ptosco/85d24e3f3476d98987af99134fc80f9d)

### DIFF
--- a/rdkit/Chem/Draw/IPythonConsole.py
+++ b/rdkit/Chem/Draw/IPythonConsole.py
@@ -93,13 +93,16 @@ def _toJSON(mol):
 
 
 def _wrapHTMLIntoTable(html):
-  return InteractiveRenderer.injectHTMLHeaderBeforeTable(
+  return InteractiveRenderer.injectHTMLFooterAfterTable(
     f'<div><table><tbody><tr><td style="width: {molSize[0]}px; ' +
     f'height: {molSize[1]}px; text-align: center;">' + html.replace(" scoped", "") +
     '</td></tr></tbody></table></div>')
 
 
 def _toHTML(mol):
+  def escape_special_chars(s):
+    return str(s).replace('"', '&quot;').replace("'", "&apos;").replace("<", "&lt;").replace(">", "&gt;").replace("&", "&amp;")
+
   useInteractiveRenderer = InteractiveRenderer.isEnabled(mol)
   if _canUse3D and ipython_3d and mol.GetNumConformers() and mol.GetConformer().Is3D():
     return _toJSON(mol)
@@ -107,7 +110,7 @@ def _toHTML(mol):
   if not ipython_showProperties or not props:
     if useInteractiveRenderer:
       return _wrapHTMLIntoTable(
-        InteractiveRenderer.generateHTMLBody(ipython_useSVG, mol, molSize, drawOptions))
+        InteractiveRenderer.generateHTMLBody(mol, molSize, useSVG=ipython_useSVG))
     else:
       return _toSVG(mol)
   if mol.HasProp('_Name'):
@@ -117,29 +120,31 @@ def _toHTML(mol):
 
   res = []
   if useInteractiveRenderer:
-    content = InteractiveRenderer.generateHTMLBody(ipython_useSVG, mol, molSize)
+    content = InteractiveRenderer.generateHTMLBody(mol, molSize, legend=nm, useSVG=ipython_useSVG)
   else:
     if not ipython_useSVG:
-      png = Draw._moltoimg(mol, molSize, [], nm, returnPNG=True, drawOptions=drawOptions)
+      png = Draw._moltoimg(mol, molSize, [], nm, returnPNG=True,
+                           kekulize=kekulizeStructures, drawOptions=drawOptions)
       png = base64.b64encode(png)
       content = f'<image src="data:image/png;base64,{png.decode()}">'
     else:
       content = Draw._moltoSVG(mol, molSize, [], nm, kekulize=kekulizeStructures,
                                drawOptions=drawOptions)
-  res.append(f'<tr><td colspan=2 style="text-align:center">{content}</td></tr>')
+  res.append(f'<tr><td colspan="2" style="text-align: center;">{content}</td></tr>')
 
   for i, (pn, pv) in enumerate(props.items()):
     if ipython_maxProperties >= 0 and i >= ipython_maxProperties:
       res.append(
-        '<tr><td colspan=2 style="text-align:center">Property list truncated.<br />Increase IPythonConsole.ipython_maxProperties (or set it to -1) to see more properties.</td></tr>'
+        '<tr><td colspan="2" style="text-align: center">Property list truncated.<br />Increase IPythonConsole.ipython_maxProperties (or set it to -1) to see more properties.</td></tr>'
       )
       break
+    pv = escape_special_chars(pv)
     res.append(
-      f'<tr><th style="text-align:right">{pn}</th><td style="text-align:left">{pv}</td></tr>')
+      f'<tr><th style="text-align: right">{pn}</th><td style="text-align: left">{pv}</td></tr>')
   res = '\n'.join(res)
   res = f'<table>{res}</table>'
   if useInteractiveRenderer:
-    res = InteractiveRenderer.injectHTMLHeaderBeforeTable(res)
+    res = InteractiveRenderer.injectHTMLFooterAfterTable(res)
   return res
 
 
@@ -244,7 +249,9 @@ def ShowMols(mols, maxMols=50, **kwargs):
     kwargs['useSVG'] = ipython_useSVG
   if 'returnPNG' not in kwargs:
     kwargs['returnPNG'] = True
-  if _MolsToGridImageSaved is not None:
+  if InteractiveRenderer.isEnabled():
+    fn = InteractiveRenderer.MolsToHTMLTable
+  elif _MolsToGridImageSaved is not None:
     fn = _MolsToGridImageSaved
   else:
     fn = Draw.MolsToGridImage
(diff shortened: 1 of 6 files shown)


greglandrum:
The test failures here are because the build on master is currently broken due to a bad interaction between #5595  and #5576.
It's not related to this PR

rdkit/Chem/Draw/IPythonConsole.py
```diff
@@ -93,21 +93,24 @@ def _toJSON(mol):
...
 def _wrapHTMLIntoTable(html):
-  return InteractiveRenderer.injectHTMLHeaderBeforeTable(
+  return InteractiveRenderer.injectHTMLFooterAfterTable(
     f'<div><table><tbody><tr><td style="width: {molSize[0]}px; ' +
     f'height: {molSize[1]}px; text-align: center;">' + html.replace(" scoped", "") +
     '</td></tr></tbody></table></div>')
 
 
 def _toHTML(mol):
+  def escape_special_chars(s):
```
greglandrum, line 103: Why not use python's builtin `html.escape()` for this?

rdkit/Chem/Draw/InteractiveRenderer.py
```diff
@@ -76,45 +79,74 @@ def filterDefaultDrawOpts(molDrawOptions):
...
-  ({{ default: Renderer }}) =>
-    Renderer.init('{minimalLibJsUrl}').then(
-      () => jsLoader.innerHTML = 'Interactive molecule rendering is available in this Jupyter Notebook.'
-    ).catch(
-      e => setError(e)
-    )
-  ).catch(
-    e => setError(e)
+
+  def _wrapMsgIntoDiv(uuid, msg):
```
greglandrum, line 83: nice, this feels cleaner than what was there before

rdkit/Chem/PandasPatcher.py
```diff
@@ -0,0 +1,235 @@
+from io import StringIO
```
greglandrum, line 1: I'm not going to pretend that I read through this in any detail at all

rdkit/Chem/PandasTools.py
```diff
@@ -77,7 +108,7 @@
 The standard ForwardSDMolSupplier keywords are also available:
 
 >>> sdfFile = os.path.join(RDConfig.RDDataDir,'NCI/first_200.props.sdf')
->>> frame = PandasTools.LoadSDF(sdfFile,smilesName='SMILES',molColName='Molecule',
+>>> frame = PandasTools.LoadSDF(sdfFile, smilesName='SMILES', molColName='Molecule',
 ...            includeFingerprints=True, removeHs=False, strictParsing=True)
 
 Conversion to html is quite easy:
```
greglandrum, line 114: The bottom of the docstring still includes:
```
In order to support rendering the molecules as images in the HTML export of the dataframe,
the __str__ method is monkey-patched to return a base64 encoded PNG:

>>> molX = Chem.MolFromSmiles('Fc1cNc2ccccc12')
>>> print(molX) # doctest: +SKIP
<img src="data:image/png;base64,..." alt="Mol"/>
This can be reverted using the ChangeMoleculeRendering method
>>> ChangeMoleculeRendering(renderer='String')
>>> print(molX) # doctest: +SKIP
<rdkit.Chem.rdchem.Mol object at 0x10d179440>
>>> ChangeMoleculeRendering(renderer='PNG')
>>> print(molX) # doctest: +SKIP
<img src="data:image/png;base64,..." alt="Mol"/>
```
I believe that this is no longer correct.

rdkit/Chem/PandasPatcher.py
greglandrum: Still, thanks for pulling it out of PandasTools.py, which is now more approachable

rdkit/Chem/PandasTools.py
```diff
@@ -379,50 +188,48 @@ def _molge(x, y):
   else:
     return False
 
-def PrintAsBase64PNGString(x, renderer=None):
-  '''returns the molecules as base64 encoded PNG image
-    '''
+def PrintAsBase64PNGString(x):
```
greglandrum, line 191: Do you want to document this deprecation in the ReleaseNotes.md so that we can remove it in the next release?

rdkit/Chem/PandasTools.py
```diff
@@ -432,14 +239,140 @@ def RenderImagesInAllDataFrames(images=True):
...
-    pd.core.frame.DataFrame.to_html = defPandasRendering
-    if defPandasRepr is not None:
-      pd.core.frame.DataFrame._repr_html_ = defPandasRepr
+  try:
+    PandasPatcher.renderImagesInAllDataFrames(images)
+  except NameError:
+    log.warning("Failed to patch pandas - unable to change molecule rendering")
+
+
+try:
```
greglandrum, line 248: Why not just try the `import pandas as pd` at the very top of this file and let the `ImportError` come up to the user?
There's no point importing this if you don't have pandas installed.

rdkit/Chem/PandasTools.py
```diff
@@ -801,64 +638,55 @@ def _runDoctests(verbose=None):  # pragma: nocover
     sys.exit(failed)
 
 
+InstallPandasTools()
+
+
 if __name__ == '__main__':  # pragma: nocover
```
greglandrum, line 644: I don't believe that any of these tests are actually run. In order to make that happen, `PandasTools.py` would have to be added to the `test_list.py`

greglandrum (changes requested):
Small stuff and a couple of questions.

rdkit/Chem/PandasTools.py
ptosco: Right, I had missed that. I have changed the wording and removed most of that section.

rdkit/Chem/PandasTools.py
ptosco: I admit I did not touch that part - I can fix that after the beta is released.

rdkit/Chem/Draw/IPythonConsole.py
ptosco: Done. I actually also need to escape the ampersand after everything else has been escaped, so I still need `escape_special_chars()`, but I can let `html.escape()` take care of most of the work.

rdkit/Chem/PandasTools.py
ptosco: I would change the logic that I have in place at the moment since it is quite complex and tested to work across a range of `pandas` version. My reasoning was mostly to enable access to `PrintAsImageString()` (formerly `PrintAsBase64PNGString()`) which has been part of the `PandasTools` module for a long time and someone might be using even without importing `pandas` (unlikely, but not impossible).

rdkit/Chem/PandasTools.py
ptosco: Done.

rdkit/Chem/PandasPatcher.py
ptosco: Going through all versions of `pandas` from 0.25.3 onwards to make sure the patching works on all of them was not particularly entertaining, but hopefully we have something more robust in place now.

rdkit/Chem/PandasTools.py
greglandrum: We don't need to do it here, but I think it's worth making the change later.

rdkit/Chem/Draw/IPythonConsole.py
```diff
@@ -93,21 +94,24 @@ def _toJSON(mol):
...
+    return html.escape(str(s)).replace("&", "&amp;")
```
greglandrum, line 105: I don't think this is correct. 

Escaping the &s in the results from `html.escape` gives the following in the notebook:
image

On the other hand, if I just use:
```
return html.escape(str(s)) 
```
I get:
image

greglandrum (changes requested):
I think one of the fixes results in some "over quoting"... is this what you intended?

rdkit/Chem/Draw/IPythonConsole.py
ptosco: @greglandrum When I saw the final `.replace("&", "&amp;")` in the `escape_special_chars()` function I had the feeling it was wrong, but then the 'If I wrote it that way, that must have been a reason' feeling prevailed, while it looks like it was just wrong. I wrote that piece of code ~2 months ago and honestly I can't remember why I introduced that further level of escaping at the time. Thanks for noticing and testing that!

greglandrum (approved):
LGTM

PR #5630, opened 2022-10-09
Add updateMolDrawOptionsFromJSON()

ptosco:
Currently it is a bit cumbersome to set `IPythonConsole.drawOptions` to options read from a JSON file.
In fact, I have to instantiate a dummy `MolDraw2D` object that I do not actually need:
```
from rdkit.Chem.Draw import rdMolDraw2D, IPythonConsole

with open("JupyterDrawOptions.json") as hnd:
    jupyter_draw_options = hnd.read()
drawer = rdMolDraw2D.MolDraw2DSVG(300, 200)
rdMolDraw2D.UpdateDrawerParamsFromJSON(drawer, jupyter_draw_options)
IPythonConsole.drawOptions = jupyter_draw_options
```
This PR simplifies the above to
```
from rdkit.Chem.Draw import rdMolDraw2D, IPythonConsole

with open("JupyterDrawOptions.json") as hnd:
    rdMolDraw2D.UpdateMolDrawOptionsFromJSON(IPythonConsole.drawOptions, hnd.read())
```
Also, it seems more logical to have a function that takes directly `MolDrawOptions` as parameter. given that JSON settings are all about `MolDrawOptions` and not about `MolDraw2D`.

### DIFF
--- a/Code/GraphMol/MolDraw2D/MolDraw2DUtils.cpp
+++ b/Code/GraphMol/MolDraw2D/MolDraw2DUtils.cpp
@@ -104,10 +104,16 @@ void prepareAndDrawMolecule(MolDraw2D &drawer, const ROMol &mol,
   drawer.drawOptions().prepareMolsBeforeDrawing = old_prep_mol;
 }
 
-void updateDrawerParamsFromJSON(MolDraw2D &drawer, const char *json) {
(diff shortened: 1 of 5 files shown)


greglandrum (approved):
Agreed, this makes a lot of sense.

PR #5635, opened 2022-10-10
InteractiveRenderer.setEnabled() improvements

ptosco:
- `setEnabled()` now takes two optional parameters:
  1. `shouldEnable` (defaults to `True`)
  2. `quiet` (defaults to `False`)
- `setEnabled()` now works also if it has downstream commands in the Jupyter Notebook cell


greglandrum (approved):
LGTM
Thanks for adding these!

PR #5648, opened 2022-10-14
add stop condition for arom calc of large ringysystems

alexwahab:
Possible ring combinations expand combinatorially, so for large ring systems, just checking for ring fusions may be slow. For a reported example of 900 rings, the calculation never completes. 300 is about 2 million iterations, which takes approximately 1 minute.

###

From Hackathon RDkit UGM 2022.
Co-authors: @rachelnwalker @d-b-w


greglandrum (approved):
LGTM

thanks for this contribution @alexwahab, @d-b-w, and @rachelnwalker!

d-b-w:
Thanks, @alexwahab !

PR #5660, opened 2022-10-17
Fixes #5659

### LINKED ISSUE #5659: Checking whether double bond's controlling atoms are duplicate may cause an infinite loop.
I just noticed that #5560 contains a bug:

Hitting the line https://github.com/rdkit/rdkit/blob/7e901080813ecd49db20b8eed2950f2fa42680fe/Code/GraphMol/FindStereo.cpp#L423, which happens when both controlling atoms are part of the same ring with an odd number of atoms, will cause the code to enter an infinite loop where always `*it1 == *it2`, since we never change their value, and, of course, the ring will always have an odd size.

This bug is present in Release_2022_09_1b1 (it's where I noticed it). I'll post a fix soon.

ricrogz:
This fixes #5659. It is a trivial fix, and we should include it in the 2022_09 fix.

### DIFF
--- a/Code/GraphMol/FindStereo.cpp
+++ b/Code/GraphMol/FindStereo.cpp
@@ -418,10 +418,13 @@ bool areStereobondControllingAtomsDupes(
     }
 
     auto ring = ringInfo->atomRings().at(*it1);
+    ++it1;
+    ++it2;
+
     if (ring.size() % 2) {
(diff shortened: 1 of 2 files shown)


ricrogz:
Hm, I think this is not completely correct. There might be more than one ring... I'll update.

greglandrum (approved):
LGTM
Also: the original bug is a real 🤦

PR #5663, opened 2022-10-18
Speed up ring detection by reducing count() calls

d-b-w:
Counting the "on" bits in a bitset is a calculation - finding the size of a std::vector requires only a lookup! For my example system from https://github.com/rdkit/rdkit/pull/5654 , this reduces the duration by another minute or so.

Here's an example profile from before this change. The "big" `count()` calls are the two removed here. I think they're usually close to 30s each, or 25 with SSE4.2 enabled (on my computer):
ringfinding_profile

(I think that there's more clever stuff that could be done, but this seems like another trivial change with an inarguably positive effect.)


greglandrum (approved):
LGTM

PR #5696, opened 2022-10-26
Update warning message about aromaticity detection

d-b-w:
An internal user encountered the existing error message and became alarmed by the implication that no aromaticity detection was even attempted.

I think that the new error message more accurately reflects the logic - we halt the search eventually if the ring system is too large.


greglandrum (approved):
Agreed that this is clearer

PR #5708, opened 2022-10-30
Compile clang bytecode with -std=c++17 (#5685)

mbanck:
#### Reference Issue
Fixes: #5685

The clang Postgres bytecode needs to be compiled with `-std=c++17` or else fails with the build error mentioned in the issue.


greglandrum (approved):
LGTM

PR #5712, opened 2022-10-30
windows.h header should be lowercase

ptosco:
`windows.h` header should be lowercase to avoid issues when cross-compiling on Linux. In fact, `windows.h` is lowercase on Linux and since the filesystem is case sensitive it will fail if it is uppercase. On Windows it will make no difference since the filesystem is not case sensitive.


greglandrum (approved):
LGTM

PR #5716, opened 2022-11-01
Add stub files generation and installation

hadim:
See https://github.com/rdkit/rdkit/issues/3591 and https://github.com/rdkit/rdkit/issues/5588 for context.

The goal of this PR is to generate stub files for the RDKit Python library. The generation is done with the command called `stubgen` coming from the [`mypy`](https://mypy.readthedocs.io/en/stable/stubgen.html) library. This is the only dependency that must be added to the rdkit local env and CI env (`mamba install mypy`).

The generated `*.pyi` files can be then used by any text editor or checker tool to provide things like autocompletion, type checking, etc.

The default logic is to build the stub files, and it can be turned `OFF` with `-DRDK_GENERATE_PYTHON_STUBS=OFF`.

---

I tried hard to do it the "CMake way" using `add_custom_target`, `add_dependencies` and `add_custom_command` but I could not make it work. I think the complexity here comes from the need to call `stubgen` on an installed Python module.

That being said, the current logic is probably good enough: it executes the `stubgen` command after all the Python-related targets have been installed. The generation is performed directly into the installed Python rdkit module.

### DIFF
--- a/.azure-pipelines/linux_build.yml
+++ b/.azure-pipelines/linux_build.yml
@@ -10,7 +10,7 @@ steps:
         boost-cpp=$(boost_version) \
         py-boost=$(boost_version) \
         numpy pillow eigen pandas matplotlib-base \
-        qt=5.9.7 cairo
+        qt=5.9.7 cairo mypy
     conda activate rdkit_build
     conda install -c conda-forge nbval ipykernel>=6.0
   displayName: Setup build environment
@@ -69,7 +69,7 @@ steps:
 - bash: |
     source ${CONDA}/etc/profile.d/conda.sh
     conda activate rdkit_build
-    conda install ipython matplotlib-base sphinx=3.1.2 "jinja2<3.1" recommonmark 
+    conda install ipython matplotlib-base sphinx=3.1.2 "jinja2<3.1" recommonmark
     export RDBASE=`pwd`
     export PYTHONPATH=${RDBASE}:${PYTHONPATH}
     export LD_LIBRARY_PATH=${RDBASE}/lib:${LD_LIBRARY_PATH}
--- a/.azure-pipelines/linux_build_limitexternal.yml
+++ b/.azure-pipelines/linux_build_limitexternal.yml
@@ -10,7 +10,7 @@ steps:
         boost-cpp=$(boost_version) \
         py-boost=$(boost_version) \
         numpy pillow eigen pandas \
-        qt
+        qt mypy
     conda activate rdkit_build
     conda install -c rdkit nox cairo=1.14.6
   displayName: Setup build environment
--- a/.azure-pipelines/linux_build_py37.yml
+++ b/.azure-pipelines/linux_build_py37.yml
@@ -9,7 +9,7 @@ steps:
         boost-cpp=$(boost_version) boost=$(boost_version) \
         py-boost=$(boost_version) libboost=$(boost_version) \
         numpy matplotlib pillow eigen pandas \
-        sphinx=3.1.2 recommonmark jupyter
+        sphinx=3.1.2 recommonmark jupyter mypy
     conda activate rdkit_build
     conda install -c rdkit nox cairo=1.14.6
   displayName: Setup build environment
--- a/.azure-pipelines/mac_build.yml
+++ b/.azure-pipelines/mac_build.yml
@@ -22,7 +22,7 @@ steps:
         py-boost=$(boost_version) libboost=$(boost_version) \
         qt \
         numpy matplotlib cairo pillow eigen pandas  \
-        sphinx=3.1.2 recommonmark jupyter
+        sphinx=3.1.2 recommonmark jupyter mypy
     conda activate rdkit_build
     conda install -c conda-forge nbval
   displayName: Setup build environment
--- a/.azure-pipelines/mac_build_java.yml
+++ b/.azure-pipelines/mac_build_java.yml
@@ -20,7 +20,7 @@ steps:
     conda create --name rdkit_build  $(compiler)=$(compiler_version) libcxx=$(compiler_version) cmake=3.19.6 \
         boost-cpp=$(boost_version) boost=$(boost_version) \
         py-boost=$(boost_version) libboost=$(boost_version) \
-        cairo eigen swig=3
+        cairo eigen swig=3 mypy
   displayName: Setup build environment
 - bash: |
     source ${CONDA}/etc/profile.d/conda.sh
--- a/.azure-pipelines/vs_build.yml
+++ b/.azure-pipelines/vs_build.yml
@@ -8,7 +8,7 @@ steps:
         boost-cpp=$(boost_version) boost=$(boost_version) ^
         py-boost=$(boost_version) libboost=$(boost_version) ^
         numpy matplotlib cairo pillow eigen pandas ^
-        sphinx=3.1.2  "jinja2<3.1" recommonmark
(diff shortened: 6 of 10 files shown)


hadim:
@greglandrum a friendly ping to see whether you could be interested in that PR?

bzoracler:
What kind of stubs you're getting out of `stubgen -p rdkit --quiet --output ${PYTHON_INSTDIR}` as in the `CMakeLists.txt`?

https://github.com/rdkit/rdkit/blob/5bd2c5f5249aef2a672a590536cd7f70b44453a7/rdkit/CMakeLists.txt#L41-L45

 When I run this, I get (for example)

```python
# `rdkit.Chem.rdchem.pyi`

...

class MolBundle(Boost.Python.instance):
    __instance_size__: ClassVar[int] = ...
    @classmethod
    def __init__(cls, *args, **kwargs) -> None: ...
    @classmethod
    def AddMol(cls, RDKit, boost) -> Any: ...
    @classmethod
    def GetMol(cls, RDKit, unsignedlong) -> Any: ...
    @classmethod
    def GetSubstructMatch(cls, *args, **kwargs) -> Any: ...
    @classmethod
    def GetSubstructMatches(cls, *args, **kwargs) -> Any: ...
    @classmethod
    def HasSubstructMatch(cls, *args, **kwargs) -> Any: ...
    @classmethod
    def Size(cls, RDKit) -> Any: ...
    @classmethod
    def __getitem__(cls, RDKit, unsignedlong) -> Any: ...
    @classmethod
    def __len__(cls, RDKit) -> Any: ...
    @classmethod
    def __reduce__(cls) -> Any: ...

...
```

The correct type annotations for this example, based on the [`rdkit.Chem.rdchem.MolBundle` documentation](https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.MolBundle), is

```python
from typing import Any, ClassVar, overload

class MolBundle(Boost.Python.instance):
    __instance_size__: ClassVar[int] = ...
    def __init__(arg1, /) -> None: ...
    def AddMol(arg1, arg2: Mol, /) -> int: ...
    def GetMol(arg1, arg2: int, /) -> Mol: ...
    def Size(arg1, /) -> int: ...
    @overload
    def GetSubstructMatch(self, query: Mol, useChirality: bool = ..., useQueryQueryMatches: bool = ...) -> Any: ...
    @overload
    def GetSubstructMatch(self, query: MolBundle, useChirality: bool = ..., useQueryQueryMatches: bool = ...) -> Any: ...
    @overload
    def GetSubstructMatch(self, query: Mol, params: SubstructMatchParameters) -> Any: ...
    @overload
    def GetSubstructMatch(self, query: MolBundle, params: SubstructMatchParameters) -> Any: ...
    @overload
    def GetSubstructMatches(self, query: Mol, uniquify: bool = ..., useChirality: bool = ..., useQueryQueryMatches: bool = ..., maxMatches: int = ...) -> Any: ...
    @overload
    def GetSubstructMatches(self, query: MolBundle, uniquify: bool = ..., useChirality: bool = ..., useQueryQueryMatches: bool = ..., maxMatches: int = ...) -> Any: ...
    @overload
    def GetSubstructMatches(self, query: Mol, params: SubstructMatchParameters) -> Any: ...
    @overload
    def GetSubstructMatches(self, query: MolBundle, params: SubstructMatchParameters) -> Any: ...
    @overload
    def HasSubstructMatch(self, query: Mol, recursionPossible: bool = ..., useChirality: bool = ..., useQueryQueryMatches: bool = ...) -> bool: ...
    @overload
    def HasSubstructMatch(self, query: MolBundle, recursionPossible: bool = ..., useChirality: bool = ..., useQueryQueryMatches: bool = ...) -> bool: ...
    @overload
    def HasSubstructMatch(self, query: Mol, params: SubstructMatchParameters) -> bool: ...
    @overload
    def HasSubstructMatch(self, query: MolBundle, params: SubstructMatchParameters) -> bool: ...
    def __getitem__(arg1, arg2: int, /) -> Mol: ...
    def __len__(arg1, /) -> int: ...
    def __reduce__(arg1, /) -> Any: ...
```

Limitations of `stubgen` aside, the type signatures generated will cause an abundant number of false positives. Just from the following snippet,

```python
class MolBundle(Boost.Python.instance):
    @classmethod
    def __init__(cls, *args, **kwargs) -> None: ...
    @classmethod
    def AddMol(cls, RDKit, boost) -> Any: ...
    @classmethod
    def GetMol(cls, RDKit, unsignedlong) -> Any: ...
```

 - Boost Python makes no use of `@classmethod`, so there should be nothing decorated with it in any stubs;
 - `AddMol` and `GetMol` both have the wrong number of arguments, and the wrong names for these arguments; it seems as if stubgen tried to use the C++ signature
   ```
    C++ signature :

        unsigned long AddMol(RDKit::MolBundle {lvalue},boost::shared_ptr<RDKit::ROMol>)
                             ^^^^^                     ^^^^^

    ...

    C++ signature :

        boost::shared_ptr<RDKit::ROMol> GetMol(RDKit::MolBundle {lvalue},unsigned long)
                                               ^^^^^                     ^^^^^^^^^^^^^
   ```
   to infer the arguments, which will cause any non-`*args, **kwargs` signatures generated by `stubgen` for instance methods to have one more argument than than they're supposed to.
   
For these stubs to aid in basic code completion and method calling, it would be necessary to turn methods produced by `stubgen`  all into the permissive `(*args, **kwargs) -> Any` so that linters won't generate a copious number of false positives.

hadim:
Thanks @bzoracler, for looking into that. The generated stubs give you the correct method for a given method, but you are right the signatures are messed up.

Unfortunately, I don't really know how to fix that at the moment. It would be great to be able to generate the correct args without setting all of them to `(*args, **kwargs) -> Any`.

hadim:
xref with https://github.com/python/mypy/issues/14094 as it might be relevant.

bzoracler:
> Thanks @bzoracler, for looking into that. The generated stubs give you the correct method for a given method, but you are right the signatures are messed up.
> 
> Unfortunately, I don't really know how to fix that at the moment. It would be great to be able to generate the correct args without setting all of them to `(*args, **kwargs) -> Any`.

This is out of anyone's hands, including RDKit's. The first step is to suggest to `Boost.Python` to generate valid Python syntax in the documentation strings. For example,

> GetSubstructMatch((MolBundle)self, (Mol)query[, (bool)useChirality=False[, (bool)useQueryQueryMatches=False]]) → object

has the complete information to write a Python stub, but is invalid Python syntax. I wouldn't try to parse this e.g. using regex or similar, just because of situations like in https://www.rdkit.org/docs/source/rdkit.Chem.rdMMPA.html:

> `rdkit.Chem.rdMMPA.FragmentMol((Mol)mol[, (int)maxCuts=3[, (int)maxCutBonds=20[, (str)pattern='[#6+0;!$(*=, #[!#6])]!@!=!#[*]'[, (bool)resultsAsMols=True]]]]) → tuple :`

Once `Boost.Python` revises how it generates its Python signature, then it'll be easier to convince the developers of `stubgen` or other programs to make a parser for the `Boost.Python` docstrings. The necessary modifications to the docstrings should all be in 
https://github.com/boostorg/python/blob/develop/src/object/function_doc_signature.cpp and the associated header at https://github.com/boostorg/python/blob/develop/include/boost/python/object/function_doc_signature.hpp.

hadim:
Thanks for the details. It's too bad Boost Python can't handle that. That old ticket might be relevant as well: https://github.com/boostorg/python/issues/277

I will close here for now.

PR #5729, opened 2022-11-04
CFFI/MinimalLib fixes

ptosco:
- added missing `#include` that caused the emscripten build to break
- removed details from `get_maccs_fp` calls since there are no adjustable parameters
- exposed `get_maccs_fp` to JS
- added tests and adjusted existing ones since some deprecated functions were removed and do not need testing anymore


greglandrum (approved):
LGTM

PR #5750, opened 2022-11-11
Avoid passing a NULL pointer to CanSmiles()

ptosco:
Avoid passing a NULL pointer to CanSmiles (which will happen if there are not atoms in the molblock and hence a NULL SMILES) as undefined behavior will arise.
While I have already reported the problem to Bernd and submitted a patch for Avalon, this has not yet been incorporated upstream, so meanwhile it is better to make sure that RDKit is robust against this condition.


greglandrum (approved):
LGTM

PR #5772, opened 2022-11-17
Add a timeout protection for CIP calculations

tadhurst-cdd:
Authored-by: Tad Hurst <tad.hurst@collaborativedrug.com>

#### Reference Issue
CIP label calculation can run for seeming infinite times for compounds like dodecahedrane.

The CIP calculation (assignCIPLabels()) now has an optional parameter to limit the number of recursive calls it will use to try to make the label assignments.  If the max iterations value is exceeded, an error is thrown.

#### Any other comments?
A value of 1,250,000 is about 1 sec delay on my test machine.  Most structures use less than 5000, including most peptides and proteins.   Cubane takes more than 5,000,000 iterations


Code/GraphMol/CIPLabeler/CIPLabeler.h
```diff
@@ -20,6 +20,24 @@ class ROMol;
...
+
+/*
+  Some very symmetrical mols can cause pseudo infinite processing
+  (e.g. dodecahedrane)
+  To avoid this a maxinum number of iterations can be set by the caller as a
+  parameter to assignCIPLabels.
+  If that maximum value is exceeded, the following error is thrown
+*/
+
+class MaxIterationsExceeded : public std::runtime_error {
```
greglandrum, line 33: I believe that you want an export on this as well so that the exception is exposed across DLL boundaries
```suggestion
class RDKIT_CIPLABELER_EXPORT MaxIterationsExceeded : public std::runtime_error {
```

Code/GraphMol/CIPLabeler/CIPLabeler.cpp
```diff
@@ -163,21 +163,35 @@ void label(std::vector<std::unique_ptr<Configuration>> &configs) {
...
 }  // namespace
 
 void assignCIPLabels(ROMol &mol, const boost::dynamic_bitset<> &atoms,
-                     const boost::dynamic_bitset<> &bonds) {
+                     const boost::dynamic_bitset<> &bonds, unsigned int maxRecursiveIterations) {
+
+  if (maxRecursiveIterations != 0)
+    remainingCallCount  = maxRecursiveIterations;
+  else
+    remainingCallCount = UINT_MAX;  // really big - will never be hit
```
greglandrum, line 176: We always include curly brackets:
```suggestion
  if (maxRecursiveIterations != 0) {
    remainingCallCount  = maxRecursiveIterations;
  } else {
    remainingCallCount = UINT_MAX;  // really big - will never be hit
  }
```

Code/GraphMol/CIPLabeler/catch_tests.cpp
```diff
@@ -554,3 +556,128 @@ TEST_CASE("GitHub Issue #5142", "[bug][accurateCIP]") {
...
+  REQUIRE(mol);
+  bool caughtIt = false;
+  try {
+    CIPLabeler::assignCIPLabels(*mol, 100000);
+  }
+  catch(const CIPLabeler::MaxIterationsExceeded &e)
+  {
+    caughtIt = true;
+  } 
+  CHECK(caughtIt);
```
greglandrum, line 669: catch2 makes this pattern super easy:
```suggestion
  CHECK_THROWS_AS(CIPLabeler::assignCIPLabels(*mol, 100000), CIPLabeler::MaxIterationsExceeded);
```

Code/GraphMol/CIPLabeler/catch_tests.cpp
```diff
@@ -554,3 +556,128 @@ TEST_CASE("GitHub Issue #5142", "[bug][accurateCIP]") {
...
+  caughtIt = false;   
+  try {
+    CIPLabeler::assignCIPLabels(*mol, 100000);
+  }
+  catch(const CIPLabeler::MaxIterationsExceeded &e)
+  {
+    caughtIt = true;
+  }    
+
+  CHECK(caughtIt);
```
greglandrum, line 682:
```suggestion
  CHECK_THROWS_AS(CIPLabeler::assignCIPLabels(*mol, 100000), CIPLabeler::MaxIterationsExceeded);
```

Code/GraphMol/CIPLabeler/CIPLabeler.h
```diff
@@ -20,6 +20,24 @@ class ROMol;
 
 namespace CIPLabeler {
 
+RDKIT_CIPLABELER_EXPORT bool decrementRemainingCallCountAndCheck();
+
```
greglandrum, line 24: I'm not particularly fond of this being in the public API. 
It looks like it does need to be used across multiple files, so I'd suggest either putting it in a `detail` namespace to mark it as not-for-use in client code or moving it to a new header file, perhaps `CIPUtils.h`, which isn't installed on client systems (i.e. it isn't in the `rdkit_headers` section of `CMakeLists.txt`).

Code/GraphMol/CIPLabeler/CIPLabeler.cpp
```diff
@@ -163,21 +163,35 @@ void label(std::vector<std::unique_ptr<Configuration>> &configs) {
...
   boost::dynamic_bitset<> bonds(mol.getNumBonds());
   atoms.set();
   bonds.set();
-  assignCIPLabels(mol, atoms, bonds);
+  assignCIPLabels(mol, atoms, bonds, maxRecursiveIterations);
+}
+
+bool decrementRemainingCallCountAndCheck()
+{
+  return (--remainingCallCount) >0 ;
```
greglandrum, line 194: Code-formating consistency:
```suggestion
bool decrementRemainingCallCountAndCheck() {
  return (--remainingCallCount) >0 ;
```
if you run `clang-format` over the code pre-commit (or from your editor), it will fix stuff like this for you.

greglandrum (changes requested):
Some suggested changes.

greglandrum:
@tadhurst-cdd : thanks for the contribution and sorry it took me so long to do the review.

Code/GraphMol/CIPLabeler/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ rdkit_catch_test(testCIPLabeler catch_tests.cpp
                  CIPMol.cpp CIPLabeler.cpp
                  Mancude.cpp Digraph.cpp Node.cpp Edge.cpp Sort.cpp
                  ${configs_src} ${rules_src}
-                 LINK_LIBRARIES SmilesParse GraphMol)
+                 LINK_LIBRARIES SmilesParse GraphMol FileParsers)
```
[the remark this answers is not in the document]
tadhurst-cdd: the test case for the max iteractions in catch_tests.cpp calls MolBlockToMol which is in FileParsers.  The mol block is a 3D structure of dodecahedrane.

Code/GraphMol/CIPLabeler/CIPLabeler.cpp
tadhurst-cdd: I have been in places where the code style REQUIRES to not use unneeded braces.   Personally, I prefer to not put them in, but I don't care a lot.  I care more where they go (aligned braces), but I realize that the standards for this library is not going to change so ...

Code/GraphMol/CIPLabeler/CIPLabeler.h
tadhurst-cdd: I changed it to in a detail names as you suggested.

Code/GraphMol/CIPLabeler/CMakeLists.txt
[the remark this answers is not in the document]
greglandrum: In that case the dependency could be added to the tests; it doesn't need to be at the library level.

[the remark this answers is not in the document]
greglandrum: Sorry, just from looking at the diff I had mis-interpreted where this was.

greglandrum (approved):
LGTM

greglandrum:
@ricrogz : any objections to me merging this? I think it's a sensible addition for molecules where the iterations take a very long time.

Code/GraphMol/CIPLabeler/CIPLabeler.h
```diff
@@ -18,8 +18,29 @@ namespace RDKit {
...
+  (e.g. dodecahedrane)
+  To avoid this a maxinum number of iterations can be set by the caller as a
+  parameter to assignCIPLabels.
+  If that maximum value is exceeded, the following error is thrown
+*/
+
+class RDKIT_CIPLABELER_EXPORT MaxIterationsExceeded : public std::runtime_error {
+ public:
+  explicit MaxIterationsExceeded()
+      : std::runtime_error("MaxIterationsExceeded"){};
```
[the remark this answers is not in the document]
tadhurst-cdd: I took the suggested change, and also added a python error wrapper as suggested

ricrogz (approved):
Looks good, thank you!

greglandrum (commented):
LGTM

greglandrum:
@tadhurst-cdd : thanks for the contribution!

PR #5802, opened 2022-11-27
Update macOS target platform to 10.13

ptosco:
As discussed previously with @greglandrum, this PR updates the macOS target platform to 10.13 (2017) from the previous 10.9 (2013). 10.13 is also the minimum required OS version for KNIME.


bp-kelley (approved):

greglandrum (approved):
LGTM

PR #5835, opened 2022-12-02
Improvements to PandasTools.SaveXlsxFromFrame

ptosco:
A couple of improvements to `PandasTools.SaveXlsxFromFrame` to make it support a list of column names and not just a single column name, such that one can export, for example, a SAR table with R group decomposition.
I also added support for optional cell formatting, for example to enable multi-line cells containing newlines.


greglandrum (approved):
LGTM
We probably should do an example or some more documentation about the stuff that's in PandasTools in some future PR. And I do mean "we", not "you". :-)

PR #5836, opened 2022-12-04
Allow building PgSQL RPM and DEB packages

ptosco:
This issue was reported by @tdudgeon. The problem here is that CPack creates the RPM or DEB install tree in a local staging directory, and installs all relevant files there.
However, the `pgsql_install.sh` is configured to install the PostgreSQL cartridge and the other extension files in the target PostgreSQL directories rather than in the CPack local staging directory.
Therefore, at CPack packing time the packing either fails due to insufficient permissions (usually installing PostgreSQL files requires superuser privileges), or anyway the PostgreSQL files are not installed where they should.
This PR fixes the problem by using `DESTDIR` if defined (as it is at CPack packing time) to install in the proper path.
Intermediate directories are also created as needed.
I have verified that the CPack build now succeeds, does not require superuser privileges (as it shouldn't) and generates correctly populated `PgSQL.deb` and `PgSQL.rpm` packages.

### DIFF
--- a/Code/PgSQL/rdkit/CMakeLists.txt
+++ b/Code/PgSQL/rdkit/CMakeLists.txt
@@ -203,9 +203,10 @@ if(MSVC)
   set(installPgSQLName "${installPgSQLName}.bat")
   set(installPgSQLCopyCommand "copy /Y")
   set(installPgSQLSeparator "\\")
-  set(installPgSQLBody "")
   set(BAT_QUOTE "\"")
   set(SH_QUOTE "")
+  set(PG_CPACK_STAGING_DIR "")
+  set(installPgSQLBody "")
   string(REGEX REPLACE "/" ${REGEX_SEPARATOR} PG_CURRENT_SOURCE_DIR ${PG_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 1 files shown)


tdudgeon:
Thanks @ptosco . Does it make sense that this only started to be a problem after the 2021_03 releases (or maybe after I stopped using the postgres patch that you gave me some time ago)?

ptosco:
@tdudgeon I do not think CPack may have ever worked properly with PostgreSQL since we switched to CMake builds for PostgreSQL, because the install scripts never referenced `$DESTDIR`. It could be that my patch solved the problem; I remember we did something during a hackathon in Hamburg but I cannot recall the details.

greglandrum (approved):
LGTM

tdudgeon:
@ptosco @greglandrum Thanks for this. It works for me now on master branch.

tdudgeon:
@ptosco I'm having problems with this in the Release_2022_09_4 release tag. When I build on master the PgSQL deb looks OK, but on the branch the deb looks effectively empty (like it was before the problem was fixed).
It looks to my like the changes were merged from master into the branch, but seems like it's not working on for me for that release.
Could you take a look and see if you can see if anything is wrong?

tdudgeon:
All is now good on the 2022.09.5 branch. Thanks!

PR #5857, opened 2022-12-10
Set emsdk path for freetype in emscripten builds

ptosco:
Working on something completely different I realized what was wrong with `libfreetype`: `emscripten` compiles its own version of `libfreetype` and caches it within `emsdk`. Pointing `cmake` to the `emsdk` version of `libfreetype` solves all problems and it is possible to build again with the latest `emsdk`.
Since the latest `emsdk` is based on `clang16` the `-Wno-enum-constexpr-conversion` flag is needed to build `boost`.

- restore latest `emsdk`
- set `emsdk` path for `freetype`
- add compilation flag to enable building with latest `clang` compilers

### DIFF
--- a/Code/MinimalLib/docker/Dockerfile
+++ b/Code/MinimalLib/docker/Dockerfile
@@ -9,8 +9,7 @@ RUN apt-get update && apt-get upgrade -y && apt install -y \
   g++ \
   libeigen3-dev \
   git \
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM
Thanks for identifying and fixing the problem!

PR #5910, opened 2022-12-30
Add info about mergeHs to README.

DavidACosgrove:
#### Reference Issue
No issue

Adds a note to the PAINS README about explicit Hs.

#### Any other comments?
When I came to use the PAINS SMARTS recently, I got bad results because I hadn't realised that the SMARTS have explicit H atoms that need extra treatment.  This is so other people have less excuse for making the same error.


DavidACosgrove:
I'm guessing this isn't a genuine failure as all I've changed is a README.

greglandrum (approved):
LGTM

PR #5950, opened 2023-01-09
Fixes Github 5949

### LINKED ISSUE #5949: Arrow heads of dative bonds are different sizes
**Describe the bug**
The width of the arrow head of a dative bond is a function of the length of the line.  This means that sometimes in the same drawing they are different sizes, as below.  The separation between lines in a double bond is now a fixed number of pixels, and it should be the same for arrowheads as well, IMO.

**To Reproduce**

**Expected behavior**

**Screenshots**
image

 - RDKit version: master branch built from scratch
 - OS: macOS 12.6.2
 - Are you using conda?

**Additional context**
Fix on its way

DavidACosgrove:
#### Reference Issue
Fixes #5949

Several issues with the dative bonds are fixed:
The arrow head is now a fixed size, based on doubleBondOffset.
The length of the arrow is reduced a bit when drawing a filled one, because the mitre algorithm for the filled shape makes the point extend beyond the end coordinates of the line as supplied.
I have moved the code that calculates the triangle corners for the head, as the same calculation was being done in 2 places.
If the base of the arrow is a variable bond, the * symbol for the dummy atom is removed.  I have changed the order of the feature extraction because it was being done after extractBonds() which shortened the dative bond to accommodate the * which was later removed.
In catch_tests.cpp, clang spotted that some of the tests have arguments in the wrong order.  We got away with it because they were the same value.  It didn't seem worth a separate PR.

#### Any other comments?
Some of the tests will likely fail on merge to due 2 previous PRs.


greglandrum (approved):
LGTM

PR #5951, opened 2023-01-09
Fix typo in regex

ptosco:
Due to a typo, the regex check on `reaccsio.,c` always fails, so any build following the first will fail because `reaccsio.c` is patched multiple times. This does not show up in CI builds but does show up locally and needs to be fixed.


greglandrum (approved):
I won't say "LGTM" since it's a regex, but I think it's fine to merge. ;-)

PR #5978, opened 2023-01-14
Github5963

DavidACosgrove:
#### Reference Issue
Fixes 5963

The wrong points were being picked up in the wedged bond.

#### Any other comments?
We count from 0!!


greglandrum (approved):
LGTM
This is one of those :facepalm: moments. :-)

PR #5987, opened 2023-01-17
Exception in RegistrationHash for molecules with bad bond directions

d-b-w:
The specific exception is:

```    
        ..../site-packages/rdkit/Chem/EnumerateStereoisomers.py:329: in EnumerateStereoisomers
            Chem.SetDoubleBondNeighborDirections(isomer)
        E   RuntimeError: Pre-condition Violation
        E           bad dir
        E           Violation occurred on line 396 in file Code/GraphMol/Chirality.cpp
        E           Failed Expression: dir == Bond::ENDUPRIGHT || dir == Bond::ENDDOWNRIGHT
        E           RDKIT: 2022.09.3
        E           BOOST: 1_76
        ----------------------------
```

This happens if we are enumerating stereoisomers and there is a
center with a bond direction between an enumerated center and a
double bond. This is a bad way to draw a bond wedge/dash, but
in Registration, we often see weird depictions straight from
some drug database.
    
Encountered at some customer site in Schrödinger software.

### DIFF
--- a/rdkit/Chem/RegistrationHash.py
+++ b/rdkit/Chem/RegistrationHash.py
@@ -458,6 +458,13 @@ def _CanonicalizeStereoGroups(mol):
   opts.onlyStereoGroups = True
   opts.unique = False
 
+  # Before enumerating stereoisomers, we strip the bond directions to prevent
+  # an exception when enumerating centers near double bonds.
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #5995, opened 2023-01-20
Enable using STL algorithms on ROMol atoms and bonds

ptosco:
Small PR that enables using STL algorithms on `ROMol` atoms and bonds which were previously failing to compile due to missing `==` operator on the `CXXAtomIter` and `CXXBondIter` classes.


greglandrum (approved):
Nice!
I clearly hadn't thought of using STL algorithsm with those iterators, but being able to is certainly cool.

PR #6003, opened 2023-01-24
Remove `and` from C++ headers

d-b-w:
`and` is a C++ keyword, but it isn't supported by default on some compilers (msvc). I'd like to disable it on _all_ platforms in Schrödinger builds so that developers don't accidentally post code that fails on a single platform.

However, that means external headers need to use `&&` and `||` instead of `and` and `or`, even in platform-specific code.


bp-kelley:
I used to do this when going from python to c++ as well, Gcc should not accept this in my opinion!

greglandrum (approved):
LGTM

d-b-w:
yeah, I'm gonna turn on `-fno-operator-names` for our gcc and clang builds

PR #6037, opened 2023-01-31
Allow SMARTS of zero order bonds to match zero order bonds

d-b-w:
Before this commit, if you run MolToSmarts on a molecule with a zero order bond (ZOB), the SMARTS won't match the ZOB! It's a bit weird for a molecule not to be its own substructure...

This came up because in Schödinger we see ZOBs somewhat frequently in organometallics.

### DIFF
--- a/Code/GraphMol/SmilesParse/SmartsWrite.cpp
+++ b/Code/GraphMol/SmilesParse/SmartsWrite.cpp
@@ -365,6 +365,9 @@ std::string getBasicBondRepr(Bond::BondType typ, Bond::BondDir dir,
         res = "->";
       }
       break;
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

d-b-w:
This admittedly feels a bit weird, because ZOBs are not in the Daylight SMARTS specification, and `~` does have a specific meaning in SMARTS ("any bond").

I would love to move from ZOBs to dative bonds for internal use within Schrödinger, but I think that there are a couple of places that this may not be feasible. Another option we discussed was to store the ZOB information in the CXSMARTS portion.

I think using the ~ symbol is a sane fallback.

PR #6046, opened 2023-02-02
Replace the broken, commented out test for legacy stereo with a test that works

ptosco:
Small PR to replace a test that does not actually work (same stereochemistry is perceived regardless of whether legacy is used or not, same happens from Python), and hence is commented out, with a test that actually yields different results depending on the legacy stereo setting, and hence actually tests the underlying code.


greglandrum (approved):
LGTM

PR #6059, opened 2023-02-06
WIP - Disable alpha channel support on legacy browsers

ptosco:
Small PR to disable alpha channel when generating SVG and `canvas` images since it is not supported by legacy browsers.


ptosco:
Alpha channel can be conveniently handled JS-side - no need for a PR, thus closing.

PR #6066, opened 2023-02-08
Correct docstring for minFontSize.

DavidACosgrove:
#### Reference Issue
No issue.

Corrects the default value for minFontSize in the python doctoring.

#### Any other comments?
And that took longer and much more typing than the fix itself!


greglandrum (approved):
After a long and arduous review, I agree that this seems safe to merge

PR #6102, opened 2023-02-17
Set the useComplexQueryAtomSymbols default value to true

ptosco:
As previously discussed with @greglandrum and @DavidACosgrove, this PR changes the `useComplexQueryAtomSymbols` default value from `false` to `true`.


greglandrum (approved):
LGTM.
Will merge after the CI builds finish

PR #6110, opened 2023-02-19
A few SWIG tweaks

ptosco:
- wrap types that are otherwise not accessible from SWIG wrappers
- remove redundant `Point2D_Vect` definition (already defined in `Code/JavaWrappers/*/GraphMol*.`i)
- blacklist `Conformer::getPositions()` whose return type is not accessible from SWIG (and is anyway not necessary since `Conformer::getAtomPos()` and `Conformer::setAtomPos()` serve the same purpose just fine)

### DIFF
--- a/Code/JavaWrappers/Conformer.i
+++ b/Code/JavaWrappers/Conformer.i
@@ -37,6 +37,11 @@
 #include <Geometry/point.h>
 %}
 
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #6133, opened 2023-02-26
SWIG pickling improvements (and other cleanup)

ptosco:
- avoid leaking memory after instantiating `UChar_Vect` in methods added with `javacode` and `cscode`
- make helper methods private
- fix some indentation
- make it easier to read/write pickled molecules as native byte arrays from Java and C# as it happens from C++ and Python
- add tests


ptosco:
@greglandrum I have solved the conflicts.

PR #6196, opened 2023-03-14
Better name for areBondsLinear.

DavidACosgrove:
#### Reference Issue
No Issue

Function areBondsLinear is renamed areBondsParallel which is a better description of what it does.

#### Any other comments?
As the world's most overrated scribbler put it "If it were done when 'tis done, then 'twere well It were done quickly"


greglandrum (approved):
LGTM
Thanks for following up on this @DavidACosgrove

PR #6218, opened 2023-03-20
Fixes #6217

### LINKED ISSUE #6217: Atoms may get flagged with non-tetrahedral stereo even when it is not allowed
A minimal reproducible:
```python
from rdkit import Chem

Chem.SetAllowNontetrahedralChirality(False)

m = Chem.MolFromSmiles('CS(=O)(=O)O')

sinfos = Chem.FindPotentialStereo(m)
print(sinfos[0].type)
# ------------------------------
rdkit.Chem.rdchem.StereoType.Atom_SquarePlanar

# because of the non-tetrahedral stereo, the atom gets flagged
# as "potential stereo" even when it shouldn't be:
at = m.GetAtomWithIdx(1)
print(at.GetPropsAsDict())
# ------------------------------
{'_ChiralityPossible': 1}
```

The atom shouldn't be allowed to have that type of stereo, since it was explicitly disallowed.

I'll post a fix in a minute.

ricrogz:
Fixes #6217.

A very straightforward fix: just add a check to see if non-tetrahedral stereo is allowed before checking and assigning it.

### DIFF
--- a/Code/GraphMol/FindStereo.cpp
+++ b/Code/GraphMol/FindStereo.cpp
@@ -275,7 +275,8 @@ StereoInfo getStereoInfo(const Atom *atom) {
         default:
           UNDER_CONSTRUCTION("unrecognized chiral flag");
       }
-    } else if (isAtomPotentialNontetrahedralCenter(atom)) {
+    } else if (getAllowNontetrahedralChirality() &&
+               isAtomPotentialNontetrahedralCenter(atom)) {
       if (stereo == Atom::CHI_UNSPECIFIED) {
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
Thanks for catching and fixing this.

We may want to actually explore some additional special cases for S and P, but this fix should be made in any case.

ricrogz:
> Thanks for catching and fixing this.

No problem!

> We may want to actually explore some additional special cases for S and P, but this fix should be made in any case.

Yeah, I had that feeling when looking at some downstream code that we hit due to this bug, and I had a similar thought.

PR #6266, opened 2023-04-01
Fix oxidation numbers calculation

DavidACosgrove:
#### Reference Issue
No issue

This is a tweak to the experimental oxidation state calculator.  It ignores dative bonds when it does the calculation.

#### Any other comments?
There's a test that in the Pd complex, it gets the same numbers for the haptic-bond and dative-bond representations.


greglandrum (approved):
LGTM

PR #6270, opened 2023-04-03
Update assignSubstructureFilters.py

OleinikovasV:
importable package from another directory

#### Reference Issue
Static file path for inhousefilter makes it useable only when in the same directory.

This might make this contribution more readily usable, eg. in scripts
```
import sys
from rdkit.Chem import RDConfig
sys.path.append(RDConfig.RDContribDir)
from NIBRSubstructureFilters import assignSubstructureFilters

# run filter on Pandas DataFrame with column 'smiles'
assignSubstructureFilters.assignFilters(df, nameSmilesColumn='smiles')
```


greglandrum (commented):
LGTM
Thanks @OleinikovasV !

PR #6297, opened 2023-04-17
has_coords() now reports whether coords are 2D or 3D if present

ptosco:
With this PR `has_coords()` will return `2` or `3` depending whether coordinates are 2D or 3D, if present, rather than just returning `1` (CFFI) or `true` (JS).
Note that the perception of 2D/3D is only based on the return value of `Conformer::is3D()` - no attempt is made to perceive the actual 3D character. This is still very useful in many circumstances, and this change should not break existing code relying on `has_coords()`.


greglandrum (approved):
LGTM

PR #6299, opened 2023-04-18
Fix UnitTestPandasTools for running without pandas installed.

roccomoretti:
There's some existing code for making the UnitTestPandasTools tests robust to running in an environment where pandas is not present. (e.g. like the new conda environment one would make according to Docs/Book/Install.md). However, those precautions don't quite work with the current codebase.

Update the skip-test logic to check for the current situation. Also, defer using functionality which is only loaded in the presence of pandas until test running time. (As `__init__()` is called even when the test is being skipped.)

As a bonus, I added more test-generated files to the .gitignore file. (The files are not related to UnitTestPandasTools, but I figured it was close enough to not require a separate PR. )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,18 +88,33 @@ __pycache__/
 
 
 /Code/GraphMol/FileParsers/test_data/blah.sdf
+/Code/GraphMol/FileParsers/test_data/blah_molsupplier.sdf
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM
Thanks @roccomoretti !

PR #6308, opened 2023-04-21
Add Non-contiguous Atom Molecular Structural Similarity of Teixeira & Falcao

roccomoretti:
I've been looking for a way to better map the atoms of one molecule to another, one which is not as limited as MCS and which is not dependent on a particular 3D conformation. (And one that is accessible at the C++ level.) It looks like NAMS (Non-contiguous Atom Molecular Structural Similarity) method would work.
>AL Teixeira, AO Falcao. 2013. A non-contiguous atom matching structural similarity function. J. Chem. Inf. Model. DOI: 10.1021/ci400324u. http://dx.doi.org/10.1021/ci400324u

The general concept of NAMS is to come up with an atom similarity score, incorporating various components of the atom environment. Once an all-against-all similarity matrix is calculated, the Kuhn–Munkres algorithm (Hungarian method) is used to find the atom pairing which maximizes the score. By also calculating the self-similarity, the Jaccard (Tanimoto) coefficient can be used to compute a molecular similarity on a standard 0-1 scale.

The NAMS code is available under an MIT license from https://github.com/aofalcao/NAMS and https://github.com/aofalcao/nams-docker/  The preprocessing code needed to be converted from Python and OpenBabel usage to C++ and RDKit-usage. Most of the algorithm code is then usable with a few tweaks, though this means vendorizing (copy-paste) the code. (As much as possible, copyright notices and function sources have been noted.)  The Hungarian solver (libhungarian by Cyrill Stachniss) has been copied without modifications.

I did my best to attempt to match RDKit coding conventions for new or substantially written code. However, I'm a bit uncertain about formatting, documentation standards, code location and wrapping conventions. (e.g. which directories and namespaces to put things in.) Given the "primary" usage of NAMS as a similarity metric, I attempted to match the locations of the other Fingerprinting code. -- But please let me know if there's a better location for it, or if there's other alterations I should perform. I'm more than happy to make any necessary fixes.

A final note on performance. Best I can tell, the runtime for this implementation is about the same as that of the reference code (if not slightly faster), but is about 200 times slower than the standard RDKit fingerprint comparison. (About 400-500 comparisons per second in my hands.)


bp-kelley (changes requested):

Code/GraphMol/Fingerprints/NAMS.cpp
```diff
@@ -0,0 +1,661 @@
...
+  result.mapping1to2.clear();
+  result.mapping1to2.resize(mi1.natoms(), -1);
+  result.atom_scores.clear();
+  result.atom_scores.resize(mi1.natoms(), 0.0);
+  for (unsigned int row = 0; row < mi1.natoms(); row++) {
+    unsigned int col=p.my_assig[row];
+    if(col<mi2.natoms() && row < mi1.natoms()) {
+      v=m[row][col];
+      result.mapping1to2[row] = col;
+      result.atom_scores[row] = v/10000.0d;
```
bp-kelley, line 604: Drop the d.

bp-kelley (changes requested):

bp-kelley:
These changes appear to allow compilation on windows and osx btw.

bp-kelley:
After some digging, it appears to be a C99 feature:  https://www.open-std.org/jtc1/sc22/open/n4060.pdf

Code/GraphMol/Fingerprints/NAMS.h
```diff
@@ -0,0 +1,213 @@
...
+#include <string>
+
+namespace RDKit {
+class ROMol;
+namespace NAMS {
+
+struct NAMSParameters {
+
+  NAMSParameters() = default;
+  ~NAMSParameters();
```
bp-kelley, line 28: You've defined a constructor but haven't implemented it.  It's also a little odd to have a struct with private members.  Perhaps this should be a class.

bp-kelley (changes requested):

PR #6339, opened 2023-05-04
remove the dependency from python distutils in the top CMakeLists.txt file

rvianello:
This PR refactors a few steps in the cmake configuration of the python build, to remove the dependency from the deprecated distutils module.

#### Any other comments?
distutils is planned to be removed in python 3.12


greglandrum (approved):
LGTM

PR #6341, opened 2023-05-05
take drawOptions into account when exporting structure to xlsx format

ptosco:
This is a really tiny one. Currently `IPythonConsole.drawOptions` are being ignored when exporting a `pandas` dataframe to `xlsx` format with `SaveXlsxFromFrame()`. This PR corrects that.


greglandrum (approved):
LGTM

PR #6361, opened 2023-05-11
adjustQueryProperties cleanup

ptosco:
A small cleanup PR.
* In `adjustQueryProperties`, it is not necessary to dynamically allocate `QueryAtom` and `QueryBond` instances to be used with `RWMol::replaceAtom()` and `RWMol::replaceBond()`, since both functions will make a copy of the query rather than taking ownership. Therefore, it is better to use static allocation to avoid potential memory leaks in case an exception is thrown between allocation and deletion. Note that elsewhere in the same file static allocation is already used for `QueryAtom` and `QueryBond`; this PR only makes usage of static allocation consistent across the whole file
* switched to `auto` and `constexpr` where possible
* `ADJUST_IGNOREMAPPED` was not exposed to Python - this PR exposes it

### DIFF
--- a/Code/GraphMol/AdjustQuery.cpp
+++ b/Code/GraphMol/AdjustQuery.cpp
@@ -348,7 +348,7 @@ void adjustQueryProperties(RWMol &mol, const AdjustQueryParameters *inParams) {
   if (inParams) {
     params = *inParams;
   }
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #6363, opened 2023-05-11
postgres cartridge: cleanup a few obsolete build options

rvianello:
This PR removes a few conditional build settings that were required when building the cartridge with PostgreSQL 9.1-9.5 and are for this reason no longer useful.


greglandrum (approved):
Thank you for doing this!
I've been intending to clean that stuff up for a while and now I don't have to. :tada:

PR #6477, opened 2023-06-21
Fix RWMol::addAtom docstring

d-b-w:
The docstring claimed that it returned the new number of atoms. It actually returns the index of the new atom.

I think it makes more sense to change the documentation than the behavior, because callers are definitely out there using the current behavior.


greglandrum (approved):
LGTM

PR #6486, opened 2023-06-22
Fixed tests that weren't being run in testDepictor.py

rachelnwalker:
These tests weren't being ran due to an indentation issue. I think that when I wrote these, my formatting was incorrect, and the re-formatting done before the last release accidentally moved all of these tests into another function and skipped them.


greglandrum (approved):
LGTM
Thanks for noticing and fixing this.

PR #6487, opened 2023-06-22
Get tests to work when building without exception support (i.e., legacy pure JS library)

ptosco:
Small fix as I realized some JS tests were failing when run with MinimalLib built without exception support, as is the case for the legacy pure JS library.


greglandrum (approved):
LGTM

PR #6488, opened 2023-06-22
Replace a try..catch block with an if clause

ptosco:
Small refactoring to replace a `try..catch` block with an `if` clause. Exceptions are expensive, and not supported by the legacy JS MinimalLib builds, so it is good to avoid them when they can be replaced without pain.


greglandrum (approved):
LGTM

PR #6490, opened 2023-06-23
Fixes rdkit-js/issues/347

ptosco:
This PR allows building MinimalLib using the latest `emsdk` (see https://github.com/rdkit/rdkit-js/issues/347).
The issue described in https://github.com/rdkit/rdkit-js/issues/347 is not actually caused by anything changed in RDKit, but rather by an `emsdk` update.

- Add appropriate compiler flags such that MinimalLib builds also when using `clang` version >=15
- Use the `nodejs` version bundled with `emscripten` to run tests since the one in Ubuntu is ancient and does not support recent `wasm` features

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,8 +222,11 @@ if(RDK_BUILD_XYZ2MOL_SUPPORT)
     include_directories(${RDKit_ExternalDir})
 endif()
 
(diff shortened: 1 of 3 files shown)


greglandrum (approved):
LGTM

PR #6511, opened 2023-06-30
Drop the s_m_color_rgb property from MaeWriter

ricrogz:
The `s_m_color_rgb` property is not one of the required atom properties by the Maestro format, but it was added to MaeWriter in #6069. This PR removes this property. It also fixes a couple of bugs that I found during testing.


greglandrum (approved):
LGTM

PR #6513, opened 2023-07-03
update avalontools version to incorporate bug fixes

ptosco:
This PR fixes a crash in Avalon occurring when an atom idx in the bond table is out of bounds.
I also decided to fork Avalon as previously discussed with @greglandrum. This allows to patch it more conveniently and avoid using horrible regular expressions in CMake files.
Finally, I added a couple of unit tests RDKit-side to test the fixes.


greglandrum (approved):
Unsurprisingly I agree with using a fork for the avalontools code.
Do you think it would make sense to have the fork live in the RDKit organization?

ptosco:
@greglandrum absolutely. I did not want to do that without asking while you were on holiday, but since you are now proposing this I will go ahead and move it under https://github.com/rdkit.

ptosco:
@greglandrum Done.

greglandrum (approved):
LGTM

PR #6519, opened 2023-07-05
Exposed partial sanitization options to MinimalLib (both JS and CFFI)

ptosco:
This PR exposes partial sanitization options to MinimalLib (both JS and CFFI), namely:

- `setAromaticity` (defaults to true)
- `fastFindRings` (defaults to true; only used when sanitize=false)
- `assignStereo` (defaults to true)

These options avoid doing unnecessary work when the molecule is only used for specific purposes (e.g., computing FPs or doing substructure searches), as nicely described [in this blog post](https://rdkit.blogspot.com/2016/09/avoiding-unnecessary-work-and.html).

I also added `try..catch` blocks to the CFFI library FP functions to avoid that they throw exceptions that cannot be caught from C, but rather return `nullptr`, as other functions already do.

### DIFF
--- a/Code/MinimalLib/cffi_test.c
+++ b/Code/MinimalLib/cffi_test.c
@@ -2124,6 +2124,46 @@ M  END\n",
   free(tpkl);
 }
 
+void test_partial_sanitization() {
+  printf("--------------------------\n");
+  printf("  test_partial_sanitization\n");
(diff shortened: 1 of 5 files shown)


greglandrum (approved):
LGTM

PR #6522, opened 2023-07-06
Fix memory leak by making RDProps destructor virtual.

Efim-Shats:
Fixed memory leak for `RWMol` and `ROMol` objects by making `RDProps` destructor virtual, and explicitly calling `clear()` in `RWMol` destructor.

#### Any other comments?
Using RDKit SWIG Java wrapper, we noticed memory leaking by simple massive loading molecule from SMILES. These changes fixed the memory leak problem.


bp-kelley:
Rdprops has no virtual functions so it shouldn’t need a virtual destructor.

also, valgrind hasn’t complained so I’m suspicious that rdprops might be a red herring, I expect this might be best fixed in swig.

PR #6580, opened 2023-07-27
allow building the cartridge against PostgreSQL 16

ptosco:
Small PR to add the `varatt.h` header and define the `Abs` macro as it was defined on PostgreSQL <16.


greglandrum (approved):
LGTM

ptosco:
I did, even though only a beta was available when I did.

PR #6619, opened 2023-08-10
Fix bug in StereoGroup wrapper

ricrogz:
Ooops! This one slipped through!


greglandrum (changes requested):
Could you please add a test for this function?
If we'd done that originally, this would have been noticed immediately.

ricrogz:
> Could you please add a test for this function? If we'd done that originally, this would have been noticed immediately.

Done. I definitely have a memory of having added a test... but I just noticed it didn't go into the PR. I guess it got lost at some point.

greglandrum (approved):
LGTM

ricrogz:
> @ricrogz will merge once the last of the CI tests have finished. The win64_dll failure isn't real

Sounds good!

PR #6674, opened 2023-08-30
Expose RascalMCES singleLargestFrag option to Python.

DavidACosgrove:
#### Reference Issue
No issue.

Exposes the singleLargestFrag option of RascalMCES::RascalOptions to the Python wrapper.

#### Any other comments?
Writing documentation is always valuable!


greglandrum (changes requested):
I think it would be useful to have something in the documentation explaining what exactly this means and how/why it's different from what you get by calling `largestFragmentOnly()`

DavidACosgrove:
It might be worth leaving it until I finish writing the documentation and adding all the other bits that turn out to be missing from the Python wrapper.  I'll let you know...

DavidACosgrove:
Ok, I think that is now good for review again.  It's remarkable how much wasn't in the Python wrapper :-(.

greglandrum (approved):
LGTM

PR #6746, opened 2023-09-25
Enable chemist-friendly depiction of R-groups

ptosco:
- Added two `MolDrawOptions` (`mappedDummiesAreRGroups`, `isotopeDummiesAreRGroups`) to enable conveniently depicting R group labels in more chemist-friendly fashion (i.e., as `R#` rather than `#*` or `*:#`)
- Added relevant C++, Python and JS tests
- Added a `noFreeType` JSON parameter to enable generating non-FreeType SVGs in MinimalLib, mostly to make testing easier

# Current status
RDKit stores R group labels as either `atomMapNum` or isotope number. This results in a representation which is not very pretty, unless some post-procesing is applied.
For example, an R group decomposition currently looks like this in `PandasTools`:
```
from rdkit import Chem
from rdkit.Chem.Draw import IPythonConsole, rdDepictor
from rdkit.Chem import rdRGroupDecomposition

mols = [Chem.MolFromSmiles(smi) for smi in 'c1c(F)cccn1 c1c(Cl)c(C)ccn1 c1c(F)cncn1 c1c(F)c(C)ccn1'.split()]
sma_scaffold = Chem.MolFromSmarts('c:1:c(-[*:1]):c(-[*:2]):*:c:n:1')
groups, unmatched = rdRGroupDecomposition.RGroupDecompose([sma_scaffold], mols, asSmiles=False, asRows=False)
df = PandasTools.RGroupDecompositionToFrame(groups, mols, include_core=True)
df
```
image

This becomes even less chemist-friendly when R group decomposition results are generated from SMILES:
```
groups, unmatched = rdRGroupDecomposition.RGroupDecompose([sma_scaffold], mols, asSmiles=True, asRows=False)
df = PandasTools.RGroupDecompositionToFrame({k: [Chem.MolFromSmiles(smi) for smi in v] for k, v in groups.items()}, mols, include_core=True)
df
```
image

# After this PR
This PR adds two `MolDrawOptions` to enable a more chemistry-friendly depiction with a single setting, e.g.:
```
IPythonConsole.drawOptions.mappedDummiesAreRGroups = True

groups, unmatched = rdRGroupDecomposition.RGroupDecompose([sma_scaffold], mols, asSmiles=False, asRows=False)
df = PandasTools.RGroupDecompositionToFrame(groups, mols, include_core=True)
df
```
image

```
groups, unmatched = rdRGroupDecomposition.RGroupDecompose([sma_scaffold], mols, asSmiles=True, asRows=False)
df = PandasTools.RGroupDecompositionToFrame({k: [Chem.MolFromSmiles(smi) for smi in v] for k, v in groups.items()}, mols, include_core=True)
df
```
image

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawMol.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawMol.cpp
@@ -1289,6 +1289,8 @@ std::string DrawMol::getAtomSymbol(const Atom &atom,
   // as MolDrawing.py. My first thought was for a LaTeX-like system,
   // obviously...
   std::string symbol;
+  std::string mapNumAsString;
+  std::string isoAsString;
   bool literal_symbol = true;
   unsigned int iso = atom.getIsotope();
   if (drawOptions_.atomLabels.find(atom.getIdx()) !=
@@ -1341,9 +1343,13 @@ std::string DrawMol::getAtomSymbol(const Atom &atom,
 
     // first thing after the symbol is the atom map
     if (atom.hasProp("molAtomMapNumber")) {
-      std::string map_num = "";
-      atom.getProp("molAtomMapNumber", map_num);
-      postText.push_back(std::string(":") + map_num);
+      std::string mapNumSep;
+      atom.getProp("molAtomMapNumber", mapNumAsString);
+      if (!drawOptions_.mappedDummiesAreRGroups ||
+          (drawOptions_.isotopeDummiesAreRGroups && iso)) {
+        mapNumSep = ":";
+      }
+      postText.push_back(mapNumSep + mapNumAsString);
     }
 
     if (0 != atom.getFormalCharge()) {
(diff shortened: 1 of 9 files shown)


greglandrum:
@ptosco given that this can already be done relatively easily in code, and that it's pretty specific to the RGD code, I'm reluctant to add yet another drawing option for it.
What about adding a free function in the RGD library that makes the appropriate changes to the molecules add adds atom labels?

ptosco:
@greglandrum I respectfully disagree :-). I know that can be done (relatively) easily in code, but
1. That does not apply to MInimalLib, where currently this would require regex-based hackery on CXSMILES
2. That requires making another copy of the molecule just for visualization, and one normally already makes one with `prepareMoleculeForDrawing`
3. The free function would be even less discoverable than a visualization option, and would need to be exposed on JS side, where currently R group decomposition support is optional and not part of the standard build
4. In a Jupyter Notebook or in a web app that visualizes SAR tables it is way easier to set a global drawing option once and for all and always get the desired visualization than having to remember to tweak the molecule - there would easily be the odd place where the molecule pops up with the original `*:#` labels

greglandrum:
After our conversation last week, looking more at what this PR does, and spending some time investigating the options which are already available, I'm even more convinced that the PR itself is not necessary.

Users/developers who want R group labels in the output of the RGD code can ask for that using the `rgroupLabelling` parameter:
```
ps = rdRGroupDecomposition.RGroupDecompositionParameters()
ps.rgroupLabelling = rdRGroupDecomposition.RGroupLabelling.MDLRGroup
rg = RGroupDecomposition(core,ps)
```
The results of this RGroup decomposition have atomLabels on the dummy atoms; that leads to the RGroups being drawn as follows without needing any special code:
image

If you generate CXSmiles for that RGD output, then you end up with the atomLabel properties in the CXSMILES, so the values can even be stored:
image
and, fortunately, reading those SMILES back in preserves the labels:
image

What does this miss?

ptosco:
@greglandrum It does not miss anything. I already knew about CXSMILES, and to be honest we had already mentioned it in our discussion. Still, given [`rdRGroupDecomposition.RGroupDecompose` does not support CXSMILES output](https://rdkit.org/docs/source/rdkit.Chem.rdRGroupDecomposition.html?highlight=rgroupdecompose#rdkit.Chem.rdRGroupDecomposition.RGroupDecompose), I think the additional visualization option wouldn't have hurt. Anyway, I'll close this.

PR #6792, opened 2023-10-08
_moltoimg() should honor drawOptions.prepareMolsBeforeDrawing

ptosco:
`_moltoimg()` should honor `drawOptions.prepareMolsBeforeDrawing` and not run `PrepareMolForDrawing` if requested not to.
This is annoying when trying to visualize a molecule with original molblock wedging as, for example, chiral Hs were still being added even if one does not want to.


greglandrum (approved):
LGTM

PR #6848, opened 2023-10-28
Fix code formatting in docstrings for a few functions

bertiewooster:
Format code in docstrings as code blocks by adding `  >>> ` before lines of code

#### Any other comments?
I noticed this issue in the documentation for the [MolsMatrixToGridImage](https://www.rdkit.org/docs/source/rdkit.Chem.Draw.html#rdkit.Chem.Draw.MolsMatrixToGridImage) code I submitted (scroll down to EXAMPLES), then discovered that other docstrings had similar issues, so I addressed the couple that I found.

### DIFF
--- a/rdkit/Chem/AtomPairs/Pairs.py
+++ b/rdkit/Chem/AtomPairs/Pairs.py
@@ -74,32 +74,32 @@ def pyScorePair(at1, at2, dist, atomCodes=None, includeChirality=False):
 
 def ExplainPairScore(score, includeChirality=False):
   """
-  >>> from rdkit import Chem
-  >>> m = Chem.MolFromSmiles('C=CC')
-  >>> score = pyScorePair(m.GetAtomWithIdx(0), m.GetAtomWithIdx(1), 1)
-  >>> ExplainPairScore(score)
+    >>> from rdkit import Chem
+    >>> m = Chem.MolFromSmiles('C=CC')
+    >>> score = pyScorePair(m.GetAtomWithIdx(0), m.GetAtomWithIdx(1), 1)
+    >>> ExplainPairScore(score)
   (('C', 1, 1), 1, ('C', 2, 1))
-  >>> score = pyScorePair(m.GetAtomWithIdx(0), m.GetAtomWithIdx(2), 2)
-  >>> ExplainPairScore(score)
+    >>> score = pyScorePair(m.GetAtomWithIdx(0), m.GetAtomWithIdx(2), 2)
+    >>> ExplainPairScore(score)
   (('C', 1, 0), 2, ('C', 1, 1))
-  >>> score = pyScorePair(m.GetAtomWithIdx(1), m.GetAtomWithIdx(2), 1)
-  >>> ExplainPairScore(score)
+    >>> score = pyScorePair(m.GetAtomWithIdx(1), m.GetAtomWithIdx(2), 1)
+    >>> ExplainPairScore(score)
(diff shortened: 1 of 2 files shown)


greglandrum (commented):
I think the changes to your docs make them considerably less useful. 
As is, you can copy examples from the doc string and paste them directly jnto code.
After this change you'd have to edit out the prompts. 

I believe that the prompts are almost only useful when they are present as part of doctests: showing both input and output

bertiewooster:
In their current state, the code blocks are not presented as code blocks, but as [run-on paragraphs](https://www.rdkit.org/docs/source/rdkit.Chem.Draw.html#rdkit.Chem.Draw.MolsMatrixToGridImage):

Unformatted code in MolsMatrixToGridImage documentation

I proposed using `>>>` because the code blocks displayed as such, for example for [GetAtomPairFingerprintAsBitVect](https://www.rdkit.org/docs/source/rdkit.Chem.AtomPairs.Pairs.html#rdkit.Chem.AtomPairs.Pairs.GetAtomPairFingerprintAsBitVect):

Formatted code in GetAtomPairFingerprintAsBitVect documentation selected

are [formatted with `>>>`](https://github.com/rdkit/rdkit/blob/81e9cb23868c7d8e6fe391b7809acfedae306661/rdkit/Chem/AtomPairs/Pairs.py#L135) :

```
  >>> from rdkit import Chem
  >>> m = Chem.MolFromSmiles('CCC')
  >>> v = [pyScorePair(m.GetAtomWithIdx(0), m.GetAtomWithIdx(1), 1),
  ...      pyScorePair(m.GetAtomWithIdx(0), m.GetAtomWithIdx(2), 2),
  ...     ]
  >>> v.sort()
  >>> fp = GetAtomPairFingerprintAsBitVect(m)
  >>> list(fp.GetOnBits()) == v
  True
```

And as shown in the documentation screenshot above, when you select the code with the mouse the `>>>` are excluded (they are not within the blue highlighting). I verified that the `>>>` are excluded when you then copy and paste the selected code.

If the preference is to format code blocks that are not doctest-like (showing both input and output) without using `>>>`, perhaps triple backticks ```` ``` ```` around code blocks would work? I searched for triple backticks ```` ``` ```` and didn't find any in the RDKit codebase that produce documentation pages, so I can't cite any example of that working, but it's the usual way to format code blocks.

P.S. In the code block above, GitHub adds a Copy button Copy code button on GitHub which we might also want to add to the RDKit documentation to facilitate copying blocks of code. I believe the [sphinx-copybutton](https://github.com/executablebooks/sphinx-copybutton) extension achieves that. I can take that on if it's something we want to add.

greglandrum:
> In their current state, the code blocks in that documentation page are not presented as code blocks, but as [run-on paragraphs](https://www.rdkit.org/docs/source/rdkit.Chem.Draw.html#rdkit.Chem.Draw.MolsMatrixToGridImage):
> 

The solution to this is apparently to end the line above with two colons: https://stackoverflow.com/a/64452858

I tested this with a couple of doc strings and that seems to work.

bertiewooster:
Thanks for researching that. I'll try using two colons, ideally after Documentation preview builds on CI #6849 is in place so I can check that I formatted the code blocks correctly--an easy mistake to make is to format a first code block correctly but not a second code block, for example in [ExplainPairScore](https://www.rdkit.org/docs/source/rdkit.Chem.AtomPairs.Pairs.html#rdkit.Chem.AtomPairs.Pairs.ExplainPairScore).
ExplainPairScore second code block formatted incorrectly

PR #6857, opened 2023-10-31
RGD: Fix doEnumeration true for cores that are not bundles

jones-gareth:

If `doEnumeration` is set in RGD and a input core does not enumerate into a bundle that core is not added. This PR fixes this   by adding the original input core if it does not enumerate into a bundle.


greglandrum (approved):
LGTM

PR #6867, opened 2023-11-03
Fix build error when serialization is off.

vslashg:
Unlike other files that use the `RDK_USE_BOOST_SERIALIZATION` flag, EvenSamplePairs.h does not wrap the `BOOST_CLASS_VERSION` boilerplate in an #ifdef.

This happens to work anyway when built against older versions of boost, where the required header is being made visible by other boost includes, but this causes a build breakage when compiled against Boost 1.82 or later and with serialization off.


bp-kelley (approved):

greglandrum (approved):
LGTM
Thanks for the fix @vslashg !

PR #6913, opened 2023-11-15
Remove boost::regex support #6817

whosayn:
Made the following changes
        * removed RDK_USE_BOOST_REGEX cmake option and its references
        * removed usage of RDKIT_USE_BOOST_REGEX macro
        * removed install instructions related to boost regex

#### Reference Issue
Implements #6817 

removes boost::regex in favor of std::regex

#### Any other comments?
I remember @d-b-w mentioning std::regex having significant runtime costs in some internal sdgr workflows, so it's worth noting that a decrease in performance for some MolFileParser-related workflows could be a cost of this change.


greglandrum:
> I remember @d-b-w mentioning std::regex having significant runtime costs in some internal sdgr workflows, so it's worth noting that a decrease in performance for some MolFileParser-related workflows could be a cost of this change.

If this turns out to be true, I'd be really interested in hearing about it.

d-b-w:
It may be less true now. The first couple of compilers that supported std::regex were _terrible_. I think it was all platforms. 

Can either of you think of places where regex parsing or matching is a major bottleneck?

If it _truly_ matters, maybe we should swap to CTRE (https://github.com/hanickadot/compile-time-regular-expressions)?

bp-kelley:
I think it is still a problem in many cases.

There is a proposal to move CTRE into the standard in C++23

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1433r0.pdf

[image: image.png]
Now, we probably don't have complicated regex, but this is still pretty eye
opening and worth comparison.

bp-kelley:
image

d-b-w:
I wonder what version of clang _that_ was. comical that the paper doesn't say

whosayn:
It looks like the potential performance hit will only affect parseEnhancedStereo in MolFileParser.cpp https://github.com/rdkit/rdkit/blob/2957ab4576ddd422cba5d511e9e275ad69b8c628/Code/GraphMol/FileParsers/MolFileParser.cpp#L200, so this change shouldn't be terribly inefficient. std::regex also appears to be in use by some rdkit modules https://github.com/search?q=repo%3Ardkit%2Frdkit+regex+language%3AC%2B%2B&type=code&l=C%2B%2B in similar use cases e.g. SLNParse.cpp

greglandrum:
I really don't think we're using regexes anywhere that is even remotely performance critical. The majority is actually in the tests.

Given that, it seems like there's value in removing the possible dependency on a compiled boost library that we really don't need.

greglandrum (approved):
LGTM
Thanks @whosayn !

PR #6940, opened 2023-11-24
Segfault in JSONToMols when "commonchem" is an int #6890

### LINKED ISSUE #6890: Segfault in JSONToMols when "commonchem" is an int
Trying to use the example given at https://github.com/CommonChem/CommonChem/blob/8f74f8fde14e548e0bf3581e448ce3a1596e3692/README.md ,
I get a segfault!

```
from rdkit import Chem

s = '''{
  "commonchem": 10,
  "molecules": [
    {
      "name": "ethane",
      "atoms": [
        {"z": 6, "impHs": 3},
        {"z": 6, "impHs": 3}
      ],
      "bonds": [
        {"type": 1, "atoms": [0, 1]}
      ]
    }
  ]
}'''

print(Chem.JSONToMols(s)
```

I noticed that the spec given in the CommonChem documentation has the type of "commonchem" be an int holding the version number, while RDKit actually uses an object holding a key named "version":

```{
  "commonchem": {
    "version": 10
  },
  ...
```

I don't know which one is "right" given that this is a beta format, but still, getting an invalid input shouldn't trigger a segfault.

**Expected behavior**

Not crash. Maybe even accept the version number as documented.

I should note that after I change the representation of the version number to what RDKit expects, the conversion still fails, but not with a segfault: RDKit complains about "bad stereo value for atom". It would be nice if "unspecified" stereo were assumed by default, but that's another issue, and is not nearly as bad as a segfault. :-) (Finally, I also noticed that the spec has "type" for bonds, while RDKit uses "bo", but I'm getting off-topic.)
 
**Configuration:**
 - RDKit version: 2023.09.1-sch
 - OS: CentOS 7
 - Python version: 3.8.10
 - Are you using conda? No
 - If you are not using conda: how did you install the RDKit? Schrodinger

whosayn:
#### Reference Issue
this fixes #6890 

the DocToMols procedure assumes that the "commonchem" and "rdkitjson" fields are nested json docs, but that is not true. This change updates the procedure to stop relying on that wrong assumption and adds a test for validation.

### DIFF
--- a/Code/GraphMol/MolInterchange/Parser.cpp
+++ b/Code/GraphMol/MolInterchange/Parser.cpp
@@ -921,15 +921,18 @@ std::vector<boost::shared_ptr<ROMol>> DocToMols(
   if (!doc.IsObject()) {
     throw FileParseException("Bad Format: JSON should be an object");
   }
+
   if (doc.HasMember("commonchem")) {
-    if (!doc["commonchem"].HasMember("version")) {
+    if (!doc["commonchem"].IsObject() ||
+        !doc["commonchem"].HasMember("version")) {
       throw FileParseException("Bad Format: missing version in JSON");
     }
     if (doc["commonchem"]["version"].GetInt() != currentMolJSONVersion) {
       throw FileParseException("Bad Format: bad version in JSON");
     }
   } else if (doc.HasMember("rdkitjson")) {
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM
Thanks for fixing the rdkitjson one at the same time.

PR #6948, opened 2023-11-27
Fixes #6945

### LINKED ISSUE #6945: FindMolChiralCenters should honor RDK_USE_LEGACY_STEREO_PERCEPTION
The behavior of `FindMolChiralCenters` changes when the new stereo algorithm is enabled. Specifically, it won't return any chiral centers unless CIP labels have been assigned first (because the default stereo detection no longer assigns `_CIPCode` properties).

`FindMolChiralCenters` has a `useLegacyImplementation` argument (which defaults to `False`), but using it requires changes to code, which implies that one knows what's happening inside your program :)

Maybe we should make the default value of `useLegacyImplementation` dependent on `RDK_USE_LEGACY_STEREO_PERCEPTION` ?

ricrogz:
This might fix #6945.

It just makes `useLegacyImplementation` depend on the function we use in the C++ code to see if we should use the legacy or the new stereo algorithm. In this way, unless we manually override the argument, we should always use the same algorithm as specified via environment variable.

### DIFF
--- a/Code/GraphMol/Wrap/rough_test.py
+++ b/Code/GraphMol/Wrap/rough_test.py
@@ -6560,8 +6560,21 @@ M  END
     self.assertEqual(len(ctrs), 2)
     self.assertEqual(ctrs, [(1, 'S'), (5, '?')])
 
-  @unittest.skipUnless(hasattr(Chem,'MolFromPNGFile'),
-                     "RDKit not built with iostreams support")
+  def testGithub6945(self):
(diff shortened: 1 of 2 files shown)


greglandrum (changes requested):
I don't think this covers all the cases correctly
I submitted a PR to your branch with a proposed fix, including some tests.

ricrogz:
Thanks for your suggestion, it looks good to me.

greglandrum (approved):
LGTM

PR #6966, opened 2023-12-06
Drop unrequired zlib include which may break the windows build

ricrogz:
This `zlib.h` include is not required. After the changes in https://github.com/rdkit/rdkit/pull/6814, zlib is only required on windows if building with static libraries, so if the header is not in one of the include directories we have already added (e.g. system or conda directories), and needs to be specified separately, it will break the build.


greglandrum (approved):
LGTM
Thanks for cleaning this up.

PR #6980, opened 2023-12-12
Enable in-tree builds and improve overloaded constructor docstrings

ptosco:
This PR enables building stubs in both in-tree and out-of-tree builds with `cmake --build . --target stubs` (`make stubs` will also work on *NIX).
This PR also features improvements to the patching script to do a better assignment of overloaded constructor parameters, which results in a number of fixes to docstring parameters which were previously incorrectly assigned or in some cases even re-assigned.


greglandrum (approved):
I tested this locally and verified that it works for an in-tree build. I didn't spend a lot of time reviewing the cmake changes

PR #7002, opened 2023-12-20
Fixes #7001

### LINKED ISSUE #7001: HierarchicalClusterPicker::pick() randomly fails with Invariant Violation
One of our CI runs has been randomly, but frequently, failing due to this Invariant Violation since our last RDKit update:

```
E   RuntimeError: Invariant Violation
--
E
E   	Violation occurred on line 113 in file Code/SimDivPickers/HierarchicalClusterPicker.cpp
E   	Failed Expression: clusters.size() == pickSize
E   	RDKIT: 2023.09.3-sch
E   	BOOST: 1_81
```

A RDkit-only sample that reproduces the error is:
```python
from rdkit import Chem
from rdkit import SimDivFilters
from rdkit.Chem import rdMolDescriptors
from rdkit.Chem.Fingerprints.ClusterMols import GetDistanceMatrix
from rdkit.DataStructs import DiceSimilarity


def test_picker():
    smis = ['c1ccccc1', 'CCc1ccccc1', 'CCCc1ccccc1', 'CN']
    mols = [Chem.MolFromSmiles(smi) for smi in smis]

    fpgen = rdMolDescriptors.GetHashedAtomPairFingerprintAsBitVect
    fps = [(0, fpgen(mol)) for mol in mols]
    print(fps)

    distance_matrix = GetDistanceMatrix(fps, metric=DiceSimilarity)

    picker = SimDivFilters.HierarchicalClusterPicker(
        SimDivFilters.rdSimDivPickers.ClusterMethod.WARD)
    picks = list(picker.Pick(distance_matrix, len(fps), 2))
    assert picks == [2, 3]
    subset = [mols[pick] for pick in picks]
    assert [Chem.MolToSmiles(m) for m in subset] == ['CCCc1ccccc1', 'CN']


if __name__ == '__main__':
    test_picker()
```

Still, this code would hit the issue very rarely. We only noticed the error because some recent change (We don't know what it is) made our CI runs hit the Invariant Violation in about 1 time in each 4 or 5 runs.

ricrogz:
Fixes #7001

This fixes the cause of the Invariant Violation: accessing an array index out of bounds.

The explanation of the problem is simple: when the last proper removal happens, `j` is incremented one final time, so that `j == removed.size()`, which causes the following iterations to access`removed[j]`, which is out of bounds, and will likely contain some random value. If, by chance, this value happens to be equal to `i`, then an additional and unexpected "removal" will happen, returning a result that will not contain the number of clusters we expect, and will trigger the invariant violation in line 113, right after the call to `cluster()`: https://github.com/rdkit/rdkit/blob/31de6b09fba6cb62f51c6c1253b6744c8a42a2dd/Code/SimDivPickers/HierarchicalClusterPicker.cpp#L112-L113

I wonder if this is the cause of the random SimDivPickers test failure that we randomly see...

### DIFF
--- a/Code/SimDivPickers/HierarchicalClusterPicker.cpp
+++ b/Code/SimDivPickers/HierarchicalClusterPicker.cpp
@@ -96,7 +96,7 @@ RDKit::VECT_INT_VECT HierarchicalClusterPicker::cluster(
   RDKit::VECT_INT_VECT res;
   unsigned int j = 0;
   for (unsigned int i = 0; i < poolSize; i++) {
-    if (static_cast<int>(i) == removed[j]) {
+    if (j < removed.size() && static_cast<int>(i) == removed[j]) {
       j++;
       continue;
     }


ricrogz:
I'm intentionally not including a test: error is very difficult to trigger, given that it depends on very specific values in an out of bounds read.

bp-kelley (approved):
Good catch!

greglandrum (approved):
LGTM

PR #7012, opened 2023-12-30
change the handling of autocommit flag to support sqlite3 connections

### LINKED ISSUE #7009: rdkit.Dbase doesn't work correctly with Python 3.12
**Describe the bug**
When running the unit tests with Python 3.12, `pythonSourceTests` fails on `Dbase/UnitTestDbUtils.py` with `test1Txt`, `test3Txt`, and `test_DatabaseToDatabase` producing an output similar to the following:

    =================================== FAILURES ===================================
    ______________________________ TestCase.test1Txt _______________________________
    
    self = <rdkit.Dbase.UnitTestDbUtils.TestCase testMethod=test1Txt>
    
        def test1Txt(self):
          """ test reading from a text file """
          with open(os.path.join(self.baseDir, 'dbtest.csv'), 'r') as inF:
            tblName = 'fromtext'
            f = StringIO()
            with redirect_stdout(f):
              DbUtils.TextFileToDatabase(self.tempDbName, tblName, inF)
    >       self._confirm(tblName, dbName=self.tempDbName)
    
    Dbase/UnitTestDbUtils.py:77: 
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
    
    self = <rdkit.Dbase.UnitTestDbUtils.TestCase testMethod=test1Txt>
    tblName = 'fromtext', dbName = '/tmp/tmpi9sfh6oysqlt'
    colHeads = ('int_col', 'floatCol', 'strCol')
    colTypes = ('integer', 'float', 'string')
    
        def _confirm(self, tblName, dbName=None, colHeads=None, colTypes=None):
          dbName = dbName or self.dbName
          colHeads = colHeads or self.colHeads
          colTypes = colTypes or self.colTypes
        
          conn = DbConnect(dbName, tblName)
          res = conn.GetColumnNamesAndTypes()
    >     assert len(res) == len(colHeads), 'bad number of columns'
    E     AssertionError: bad number of columns
    E     assert 0 == 3
    E      +  where 0 = len([])
    E      +  and   3 = len(('int_col', 'floatCol', 'strCol'))

**To Reproduce**
Build the current master branch with Python 3.12 and run

    $ PYTHONPATH=$RDBASE LD_LIBRARY_PATH=$RDBASE/lib ctest --output-on-failure -R pythonSourceTests

**Expected behavior**
The unit tests should pass.

 - RDKit version: [cc36a86](https://github.com/rdkit/rdkit/commit/cc36a86ba22bfb62145596a30ac109dea35391cb) 
 - OS: Fedora 39
 - Python version (if relevant): 3.12
 - Are you using conda? no
 - If you are not using conda: how did you install the RDKit? from source code

**Additional context**
The issue I think is due to [Python 3.12 adding an `autocommit` attribute to the sqlite connection](https://docs.python.org/3/whatsnew/3.12.html#sqlite3) with a default value of `LEGACY_TRANSACTION_CONTROL`. This value is meant to preserve the pre-Python 3.12 behavior and will be replaced by `False` in future versions of Python, but because it's truthy, it confuses the logic at `DbUtils.py:358-363` and prevents the transaction from being committed.

rvianello:
#### Reference Issue
Fixes #7009

In Python 3.12 sqlite connection objects have an `autocommit` attribute with possible values `False`, `True` and `LEGACY_TRANSACTION_CONTROL`. This PR introduces a small change in `rdkit/Dbase/DbUtils.py` so that an explicit commit is executed when the sqlite connection's `autocommit` "flag" has value `LEGACY_TRANSACTION_CONTROL` (the default value in Python 3.12).

#### Any other comments?
Maybe there are better ways to address this issue, I don't find this one particularly pretty, but it's simple and I think should preserve the original behavior.

### DIFF
--- a/rdkit/Dbase/DbUtils.py
+++ b/rdkit/Dbase/DbUtils.py
@@ -355,12 +355,16 @@ def _AddDataToDb(dBase, table, user, password, colDefs, colTypes, data, nullMark
     block.append(tuple(entries))
     if len(block) >= blockSize:
       nDone += _insertBlock(cn, sqlStr, block)
-      if not hasattr(cn, 'autocommit') or not cn.autocommit:
+      # note: in Python 3.12 `cn.autocommit != True`
+      # is different from `not cn.autocommit` (GH #7009)
+      if not hasattr(cn, 'autocommit') or cn.autocommit != True:
         cn.commit()
       block = []
   if len(block):
     nDone += _insertBlock(cn, sqlStr, block)
-  if not hasattr(cn, 'autocommit') or not cn.autocommit:
+  # note: in Python 3.12 `cn.autocommit != True`
+  # is different from `not cn.autocommit` (GH #7009)
+  if not hasattr(cn, 'autocommit') or cn.autocommit != True:
     cn.commit()
 
 


bp-kelley (approved):

rvianello:
Well, they didn't actually change it, because the `autocommit` flag was missing for sqlite connections, but I think they also couldn't introduce it as a bool flag (what the DB API specifies) without breaking a lot of users' code that currently depends on the isolation mode settings for transaction control.  The current implementation is "surprising", also when taking the "practicality beats purity" clause into account, but as far as I understood it's meant to be temporary (the connection's `isolation_level` property should be deprecated in Python 3.14, and in Python 3.16 the default behavior should change to to PEP 249-compliant transaction control).

PR #7035, opened 2024-01-11
fix check of python version when updating Filters.cpp

rvianello:
I noticed that the update of FilterCatalog's `Filters.cpp` at cmake configuration time is no longer working. I think this is because line

    if(DEFINED ${PYTHON_VERSION_STRING} AND ${PYTHON_VERSION_STRING} VERSION_GREATER "2.6" )

should have been

    if(DEFINED PYTHON_VERSION_STRING AND ${PYTHON_VERSION_STRING} VERSION_GREATER "2.6" )

#### Any other comments?
Would it be preferable to just remove the if clause?

### DIFF
--- a/Code/GraphMol/FilterCatalog/CMakeLists.txt
+++ b/Code/GraphMol/FilterCatalog/CMakeLists.txt
@@ -4,7 +4,7 @@ endif()
 
 find_package(PythonInterp)
 if (PYTHONINTERP_FOUND)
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM
(sorry I'm so slow, I thought I had approved and merged this after I first looked at it)

rvianello:
Thanks Greg and no worries, considering that it's a small fix I thought it could be just deferred to the next patch release.

PR #7063, opened 2024-01-18
fix several warnings originating from the swig wrappers

rvianello:
This PR should fix a subset of the warnings produced during the build of the SWIG-based wrappers.


greglandrum (approved):
LGTM

PR #7103, opened 2024-01-28
chore: fix blog link to new blog location

pechersky:

The blog link on the README points to the old blog. Instead, point to the new blog (same link the old blog refers to).


greglandrum (approved):
LGTM
Thanks for fixing this @pechersky

PR #7124, opened 2024-02-01
Fixes #7123

### LINKED ISSUE #7123: AddHs sets "no implicit Hs" on the atoms were Hs are added
Setting this flag might cause any further transformations to result in the wrong number of hydrogens:

```python3
In [2]: # initially, the atoms do not have no implicit set
   ...: m = Chem.MolFromSmiles('CCC')
   ...: for at in m.GetAtoms():
   ...:     if at.GetAtomicNum() == 6:
   ...:         print(at.GetNoImplicit())
   ...: 
False
False
False

In [3]: # but after adding Hs, the C atoms have it set
   ...: m2 = Chem.AddHs(m)
   ...: for at in m2.GetAtoms():
   ...:     if at.GetAtomicNum() == 6:
   ...:         print(at.GetNoImplicit())
   ...: 
True
True
True

In [4]: # The flag stays after removing Hs again
   ...: m3 = Chem.RemoveHs(m2)
   ...: for at in m3.GetAtoms():
   ...:     if at.GetAtomicNum() == 6:
   ...:         print(at.GetNoImplicit())
   ...: 
True
True
True
```

ricrogz:
Fixes #7123

The fix is very simple: just remove the code that sets the flag. I'm also removing the property we were setting on the atoms, `origNoImplicit`, as it isn't used anywhere else.

After making the change, just one test needs to be updated (the H count flips from explicit Hs to implicit ones).

This is a fairly simple bug, so instead of adding a new test, I'm adding an additional check to one of the tests for `addHs`.

### DIFF
--- a/Code/GraphMol/AddHs.cpp
+++ b/Code/GraphMol/AddHs.cpp
@@ -566,11 +566,6 @@ void addHs(RWMol &mol, bool explicitOnly, bool addCoords,
           ++isoH;
         }
       }
-      // be very clear about implicits not being allowed in this
-      // representation
-      newAt->setProp(common_properties::origNoImplicit, newAt->getNoImplicit(),
-                     true);
-      newAt->setNoImplicit(true);
     }
     // update the atom's derived properties (valence count, etc.)
     // no sense in being strict here (was github #2782)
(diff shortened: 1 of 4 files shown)


greglandrum (approved):
LGTM

PR #7129, opened 2024-02-02
Fixes #7128

### LINKED ISSUE #7128: ReplaceBond may cause valence issues in specific edge cases
The edge cases in which this may happen is when the replaced bond is a wedged/dashed bond around a tricoordinate chiral center, and the incoming bond is either a double or triple bond. This is due to the explicit H we set on the chiral atom in `assignChiralTypesFromBondDirs()`. Since `replaceBond()` doesn't adjust the explicit H, and we are increasing the bond order, the chiral atom will end up with an invalid valence.

An example:
```python3
In [1]: mb = """
   ...:      RDKit          2D
   ...: 
   ...: M  V30 BEGIN CTAB
   ...: M  V30 COUNTS 4 3 0 0 0
   ...: M  V30 BEGIN ATOM
   ...: M  V30 1 C 1.787879 0.909091 0.000000 0
   ...: M  V30 2 C 3.287879 0.909091 0.000000 0
   ...: M  V30 3 N 1.037879 2.208129 0.000000 0
   ...: M  V30 4 O 1.037879 -0.389947 0.000000 0
   ...: M  V30 END ATOM
   ...: M  V30 BEGIN BOND
   ...: M  V30 1 1 1 2 CFG=3
   ...: M  V30 2 1 1 3
   ...: M  V30 3 1 1 4
   ...: M  V30 END BOND
   ...: M  V30 END CTAB
   ...: M  END
   ...: $$$$"""

In [2]: # Get a RWMol
   ...: m = Chem.RWMol(Chem.MolFromMolBlock(mb))
   ...: 

In [3]: b = m.GetBondWithIdx(0)

In [4]: # bond was a wedge bond (the wedge was cleaned up when it was converted into a chiral tag)
   ...: b.GetPropsAsDict()
   ...: 
Out[4]: {'_MolFileBondType': 1, '_MolFileBondCfg': 3}

In [5]: # The chiral atom has one explicit H set
   ...: at = b.GetBeginAtom()
   ...: at.GetNumExplicitHs()
   ...: 
Out[5]: 1

In [6]: # Now, replace the bond with a double
   ...: m2 = Chem.MolFromSmiles('C=C')
   ...: m.ReplaceBond(0, m2.GetBondWithIdx(0))
   ...: m.UpdatePropertyCache(strict=False)
   ...: 

In [7]: at.GetNumExplicitHs(), at.GetTotalValence()
Out[7]: (1, 5)

```

ricrogz:
Fixes #7128

This fixes the issue by updating the Explicit Hs according to the change in bond orders when we replace a bond. We only care about the case in which the bond order increases, as for a decrease, implicit Hs should be able to compensate the difference.

### DIFF
--- a/Code/GraphMol/RWMol.cpp
+++ b/Code/GraphMol/RWMol.cpp
@@ -184,6 +184,19 @@ void RWMol::replaceBond(unsigned int idx, Bond *bond_pin, bool preserveProps,
   bond_p->setIdx(idx);
   bond_p->setBeginAtomIdx(obond->getBeginAtomIdx());
   bond_p->setEndAtomIdx(obond->getEndAtomIdx());
+
+  // Update explicit Hs, if set, on both ends. This was github #7128
+  auto orderDifference =
+      bond_p->getBondTypeAsDouble() - obond->getBondTypeAsDouble();
+  if (orderDifference > 0) {
+    for (auto atom : {bond_p->getBeginAtom(), bond_p->getEndAtom()}) {
+      if (auto explicit_hs = atom->getNumExplicitHs(); explicit_hs > 0) {
+        auto new_hs = static_cast<int>(explicit_hs - orderDifference);
+        atom->setNumExplicitHs(std::max(new_hs, 0));
+      }
+    }
+  }
+
   if (preserveProps) {
     const bool replaceExistingData = false;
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #7137, opened 2024-02-05
Fixes #7044

### LINKED ISSUE #7044: Tautomer canonicalizer invariant violation
**Bug description:**
Canonicalizer has a runtime error on a specific structure `O=[N+](`O-])/C=C/c1ccccn1`. Error cannot be caught by try except. Error seems to be related to the stereochemistry as it does not occur when deleted from SMILES string. Error is related to latest versions of rdkit as it does not happen in rdkit==2023.3.3. Initial error happened on Google VM with python 3.11.5 and conda environment within installed rdkit==2023.9.4 but it may not be relevant as error reproduces in Google Colab using pip install.

**Code to reproduce:**
```
from rdkit import Chem

try:
  mol = Chem.MolFromSmiles('O=[N+]([O-])/C=C/c1ccccn1')
  enumerator.Canonicalize(mol)
except Exception as err:
  print(err)
```

Also look at [Google colab here](https://colab.research.google.com/drive/1MHAuIRJ5J5X6CH3bLIzFH9RlgmFQN82-?usp=sharing). 

**Expected behavior:**
Should not crash as molecule is valid and ideally should be able to be caught by try except to avoid interrupting workflow.

**Configuration:**
 - RDKit version: 2023.9.4
 - OS: [e.g. Ubuntu 20.04]
 - Python version (likely not relevant): 3.11.5
 - Are you using conda? Yes - miniconda

pechersky:
#### Reference Issue
Fixes #7044

During canonicalization of double bonds, neighbor bonds have to be found to correctly label the stereochem. During #6643, the allene case was not supported. This case comes up during a Tautomer Canonicalization of conjugated nitro or oximes. So I added tests around those (ones that came up in the issues and real-life use cases).

#### Any other comments?
I tried to make changes in the actual Tautomer Canonicalization to not copy stereo atoms in allene situations, but it seemed more direct to avoid a nullptr situation and _hope_ that that is doing the right thing. The new tests should shore up that the hope is true, and the rest of tests passing means that it is a targeted fix.


pechersky:
Thanks @greglandrum! I appreciate the direct fix. I had kept my fix localized to allenes because I was afraid of a similar issue popping up in other molecules which would have helped expose some other issue. If you're feeling good about exiting early in any such situation, I'm good with that too.

pechersky:
I've been trying to test locally, but have been hitting some cmake troubles (since the last time I had it working):
```
[...]
-- Found RingDecomposerLib source in /home/yakov/orion/rdkit/External/RingFamilies/RingDecomposerLib/src/RingDecomposerLib
CMake Error at Code/GraphMol/SmilesParse/CMakeLists.txt:51 (CONFIGURE_FILE):
  Operation not permitted


CMake Error at Code/GraphMol/SmilesParse/CMakeLists.txt:53 (CONFIGURE_FILE):
  Operation not permitted


== Updating Filters.cpp from pains file
[...]
```
So, excuse my clumsy CI-testing in the meantime please =)

pechersky:
@greglandrum Would be happy to have a re-review. The implementation comes from your PR.

greglandrum (approved):
LGTM
Sorry I forgot to come back to this one for so long

PR #7139, opened 2024-02-06
require SWIG 4.1+ at cmake config time

rvianello:
This PR just introduces an explicit requirement for SWIG version 4.1 or later in the CMake configuration for the Java and C# bindings. This requirement was introduced with #6767.


greglandrum (approved):
LGTM

PR #7178, opened 2024-02-21
Fix leaking Bonds on unmatched ring closures

ricrogz:
This cleans up the bonds that leak when you try to parse (incorrect) SMILES or SMARTS that contain unmatched ring closures. 

Background story:
For ring closures, the parsers create "half bonds" that are attached to a begin atom, but not an end one. These cannot be added to the mol because of the lack of an end atom. When we run `CloseMolRings` it tries to close the rings by adding the end atom with the matching bookmark to these half bonds. For unmatched closures, there's no such atom, so `CloseMolRings` it triggers a parsing failure.

The error:
When `CloseMolRings` triggers the failure, it raises an exception that we catch to print the error message and clean up. But the clean up is faulty: we reset `rest`, which deletes the problematic mol, without deleting the unmatched ring closure bonds first. Since they haven't been add to the mol, its destructor doesn't clean them up, and we need to do it by calling `CleanupAfterParseError`, which we do in the loop right after the catch. But, since we have reset `res` and zeroed the corresponding pointer in `molVect`, we never clean them up.

The fix:
Instead of `res.reset()`, we can use `res.release()`, which resets the `unique_ptr`, but without destroying the contained object. Then, we need to remove the code that resets the first pointer in `molVect`, because that would directly leak the mol.

### DIFF
--- a/Code/GraphMol/SmilesParse/SmilesParse.cpp
+++ b/Code/GraphMol/SmilesParse/SmilesParse.cpp
@@ -256,10 +256,11 @@ std::unique_ptr<RWMol> toMol(const std::string &inp,
     }
     BOOST_LOG(rdErrorLog) << nm << " Parse Error: " << e.what()
                           << " for input: '" << origInp << "'" << std::endl;
-    res.reset();
-    if (!molVect.empty()) {
-      molVect[0] = nullptr;
-    }
+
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #7211, opened 2024-03-04
Fix Uncharger applying to already neutralized perhalic groups

rvianello:
This PR fixes the neutralization of groups containing halogens in higher oxidation states. The neutralization of the negatively charged oxigen atoms sorrounding the halogen was supposed to happen at most once, but it was actually applied also to the already neutralized group.

#### Any other comments?
The problem addressed by this PR is probably a regression introduced with #7088.


greglandrum (approved):
LGTM

rvianello:
Thanks @greglandrum

PR #7240, opened 2024-03-12
feat(minilib): expose the options parameter in get_inchi

BenoitClaveau:
#### Reference Issue
Fixes: #449

expose options parameter in get_inchi like in the python api.

### DIFF
--- a/Code/MinimalLib/cffiwrapper.cpp
+++ b/Code/MinimalLib/cffiwrapper.cpp
@@ -135,6 +135,7 @@ SmilesWriteParams getParamsFromJSON(const char *details_json) {
   }
   return params;
 }
+
 std::string smiles_helper(const char *pkl, size_t pkl_sz,
                           const char *details_json) {
   if (!pkl || !pkl_sz) {
@@ -235,20 +236,8 @@ extern "C" char *get_inchi(const char *pkl, size_t pkl_sz,
   }
   auto mol = mol_from_pkl(pkl, pkl_sz);
   ExtraInchiReturnValues rv;
-  std::string options;
-  if (details_json && strlen(details_json)) {
-    boost::property_tree::ptree pt;
-    std::istringstream ss;
-    ss.str(details_json);
-    boost::property_tree::read_json(ss, pt);
-    PT_OPT_GET(options);
-  }
-
-  const char *opts = nullptr;
-  if (!options.empty()) {
-    opts = options.c_str();
-  }
-  return str_to_c(MolToInchi(mol, rv, opts));
+  auto options = MinimalLib::parse_inchi_options(details_json);
+  return str_to_c(MolToInchi(mol, rv, !options.empty() ? options.c_str() : nullptr));
 }
 
 extern "C" char *get_inchi_for_molblock(const char *ctab,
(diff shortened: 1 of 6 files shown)


Code/MinimalLib/minilib.cpp
```diff
@@ -110,10 +110,10 @@ std::string JSMol::get_svg_with_highlights(const std::string &details) const {
   return MinimalLib::mol_to_svg(*d_mol, w, h, details);
 }
 
-std::string JSMol::get_inchi() const {
+std::string JSMol::get_inchi(const std::string &options) const {
   assert(d_mol);
   ExtraInchiReturnValues rv;
-  return MolToInchi(*d_mol, rv);
+  return MolToInchi(*d_mol, rv, options.c_str());
```
ptosco, line 116:
```suggestion
  return MolToInchi(*d_mol, rv, !options.empty() ? options.c_str() : nullptr);
```

ptosco (changes requested):
@BenoitClaveau Thank you for volunteering to implement this enhancement.
For your contribution to be merged, some further work is needed:
1. A unit test should be added to `Code/MinimalLib/tests/tests.js` to test that passing some options actually leads to the desired non-standard InChI.
2. `options` should also be supported in the CFFI version of MinimalLib (see `Code/MinimalLib/cffiwrapper.cpp` and `Code/MinimalLib/cffiwrapper.h`), and likewise a test should be added to `Code/MinimalLib/cffi_test.c`.

Code/MinimalLib/minilib.cpp
BenoitClaveau: ``cffiwrapper`` already supports options but in json format.
 
```javascript
  inchi = get_inchi(pkl, pkl_size, "{\"options\":\"/FixedH\"}");
```
MolToInchi seems to accept options as string ``"-FixedH"`` https://github.com/rdkit/rdkit/blob/6f3275f9274eb958492679cb9c9e4af7cd3cbba3/External/INCHI-API/inchi.h#L105

```javascript
mol.get_inchi("-FixedH")
```

ptosco:
@ben-ikt Would you be willing to also address point (2) in my comment above?

BenoitClaveau:
> @ben-ikt Would you be willing to also address point (2) in my comment above?

I put a comment here, https://github.com/rdkit/rdkit/pull/7240#discussion_r1522783011, cffiwrapper already supports options but in json format via 
```javascript
inchi = get_inchi(pkl, pkl_size, "{\"options\":\"/FixedH\"}");
```

Code/MinimalLib/minilib.cpp
ptosco: Ah, sorry, I had missed this comment. Then I would suggest the following:
Extract the JSON options parsing snippet into a function:
```C++
std::string parse_inchi_options(const char *details_json) {
  std::string options;
  if (details_json && strlen(details_json)) {
    boost::property_tree::ptree pt;
    std::istringstream ss;
    ss.str(details_json);
    boost::property_tree::read_json(ss, pt);
    LPT_OPT_GET(options);
  }
  return options;
}
```
and put the function inside the `MinimalLib` namespace in `common.h`, thus it can be called from CFFI `get_inchi()`, `get_inchi_for_molblock()` (and we get rid the current code duplication in `get_inchi_for_molblock()`, which is a bit ugly), and also from JS MinimalLib `get_inchi()`.
This avoids code duplication and yields a consistent API across JS and CFFI MinimalLib.

Code/MinimalLib/cffiwrapper.cpp
```diff
@@ -235,42 +236,18 @@ extern "C" char *get_inchi(const char *pkl, size_t pkl_sz,
...
-  }
-
   ExtraInchiReturnValues rv;
-  const char *opts = nullptr;
-  if (!options.empty()) {
-    opts = options.c_str();
-  }
-  return str_to_c(MolBlockToInchi(ctab, rv, opts));
+  auto options = MinimalLib::parse_inchi_options(details_json);
+  return str_to_c(MolBlockToInchi(mol, rv, !options.empty() ? options.c_str() : nullptr));
```
ptosco, line 250:
```suggestion
  return str_to_c(MolBlockToInchi(ctab, rv, !options.empty() ? options.c_str() : nullptr));
```

ptosco: This should fix the build failures.

ptosco (approved):
LGTM. Thank you for your contributions and for keeping up with my requests!

greglandrum (approved):
LGTM

PR #7245, opened 2024-03-13
Avoid rebuilding FreeSASA at every build for no good reason

ptosco:
The patching code in the FreeSASA `CMakeLists.txt` was causing FreeSASA to be rebuilt at each RDKit build because `freesasa.h` was getting updated. This is unnecessary since the patching is actually a one-off operation.
Additionally, the patching code was prepending an `#include` line to `freesasa.h` at _every_ build.
My local copy had _many_ of those `#include` lines :facepalm:

- avoid rebuilding FreeSASA at every build for no good reason
- avoid prepending an `#include` to `freesasa.h` at every build for no good reason


greglandrum (approved):
LGTM

PR #7257, opened 2024-03-15
Fixes #7256

### LINKED ISSUE #7256: RDKit fails to parse "M RAD" lines were radical is 0
According to the Spec, 0 is a valid value (no radical), but RDKit only supports 1,2,3, and fails parsing when a 0 is found:

image

```python3
In [1]: mb = """
   ...:      RDKit          2D
   ...: 
   ...:   1  0  0  0  0  0  0  0  0  0999 V2000
   ...:     0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   ...: M  RAD  1   1   0
   ...: M  END
   ...: """

In [2]: m = Chem.MolFromMolBlock(mb)
[13:57:28] Unrecognized radical value 0 for atom 0 on line 6


In [3]: assert m is not None
AssertionError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 assert m is not None

AssertionError: 

```

ricrogz:
Fixes #7256

A simple fix to prevent the parser from throwing when it find M RAD 0. We shouldn't need to `setNumRadicalElectrons(0)`, since there's nothing that would change it in the atom or bond block, but just in case.

### DIFF
--- a/Code/GraphMol/FileParsers/MolFileParser.cpp
+++ b/Code/GraphMol/FileParsers/MolFileParser.cpp
@@ -422,6 +422,10 @@ void ParseRadicalLine(RWMol *mol, const std::string &text, bool firstCall,
       spos += 4;
 
       switch (rad) {
+        case 0:
+          // This shouldn't be required, but let's make sure.
+          mol->getAtomWithIdx(aid - 1)->setNumRadicalElectrons(0);
+          break;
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #7258, opened 2024-03-15
Make _NonExplicit3DChirality a public property

ricrogz:
`_NonExplicit3DChirality` was introduced in the atropoisomers patch (#6903), but it is not exposed through any header nor Python wrapped. This moves it to `common_properties` so that it is exported through the public header and wrapped.


greglandrum (approved):
LGTM

PR #7274, opened 2024-03-19
Update ring fusion cache when needed

ptosco:
Prompted by @sroughley reporting a failure in his KNIME testing pipeline involving `RingInfo::isFusedRings()`, I reviewed that code and, though I did not manage to reproduce a segfault (`RingInfo::isFusedRings()` uses `std::vector::at()` to access vector elements, so it should never segfault for going out of bounds), I assembled a couple of test cases where the current code would certain yield wrong results.
Here's a PR to fix that.


sroughley:
Thanks @ptosco - I will keep collecting error logs as and when they occur and keep on sending them over. It seems very intermittent!

ptosco:
The macOS Java failure is not related to this PR.

greglandrum (approved):
LGTM

PR #7287, opened 2024-03-21
restrict the application of 1,3- 1,5- conjugated cation normalization

rvianello:
This PR updates the reaction patterns defining the normalization of 1,3- and 1,5-conjugated cations, so that they won't apply to compounds where the positive charge is placed on a nitrogen bound to a negatively charged N, P, O, or S. The intention is to prevent the normalization from applying to some functional groups like e.g., diazo groups and azides.


greglandrum (approved):
LGTM

PR #7321, opened 2024-04-02
Wedge bond from atrop error

tadhurst-cdd:
#### Reference Issue
Lost atropiomser wedging when an neighboring bond already has a direction related to a double bond on it.

Now that bond is simply not considered for the wedge to define the atropisomer, and another bond is chosen.  Previously, the search for a bond to wedge was terminated


Code/GraphMol/FileParsers/file_parsers_catch.cpp
```diff
@@ -6010,10 +6010,10 @@ TEST_CASE("MaeWriter basic testing", "[mae][MaeWriter][writer]") {
     auto bondBlockStart = mae.find("m_bond[6]");
     REQUIRE(bondBlockStart != std::string::npos);
 
-    std::string_view ctBlock(&mae[ctBlockStart], atomBlockStart - ctBlockStart);
-    std::string_view atomBlock(&mae[atomBlockStart],
-                               bondBlockStart - atomBlockStart);
-    std::string_view bondBlock(&mae[bondBlockStart]);
+    std::string ctBlock(&mae[ctBlockStart], atomBlockStart - ctBlockStart);
```
greglandrum, line 6013: why are these `string_view` -> `string` changes here?

greglandrum (changes requested):
The changes to `file_parsers_catch.cpp` aren't connected to this change and I don't think I agree with them anyway, but otherwise this looks fine.

Please explain or revert those changes

Code/GraphMol/FileParsers/file_parsers_catch.cpp
tadhurst-cdd: You are right  - these changes have nothing to do with the bug or the fix.   I needed to do this to get the system to build.

When I remove the changes I get:
[build] /usr/bin/ld: CMakeFiles/fileParsersCatchTest.dir/file_parsers_catch.cpp.o: in function `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > Catch::Detail::stringify<std::basic_string_view<char, std::char_traits<char> > >(std::basic_string_view<char, std::char_traits<char> > const&)':
[build] /usr/local/include/catch2/catch_tostring.hpp:150: undefined reference to `Catch::StringMaker<std::basic_string_view<char, std::char_traits<char> >, void>::convert[abi:cxx11](std::basic_string_view<char, std::char_traits<char> >)'

I am using g++ (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
and clang version 10.0.0-4ubuntu1 .

Anyway, I have removed those changed and checked it back in.

Any suggestions on the build?


Suggestion

greglandrum: > Any suggestions on the build?

This could be because you are still using an older version of cmake (or a version installed the way the RDKit used to do it before #6898).

You might try looking to see if there's a `catch` directory in `$RDBASE/External/catch` and/or `build/External/catch` and, if so, removing those directories, then re-running cmake. This will complain about not finding catch2 and then grab a new copy and build it using the proper build flags.

greglandrum (approved):
LGTM

tadhurst-cdd:
I updated cmake, and I also removed catch2 libs and includes.  the cmake
command did not re-load catch2, but generated an error saying it could not
find catch2.
With the catch2 back in place, I still get the error about string_view.
 Any other ideas?

How do YOU load Catch2?

thanks, tad

PR #7329, opened 2024-04-04
cleanup RDKit::MolOps::detectBondStereochemistry

rvianello:
This PR deletes some code from `RDKit::MolOps::detectBondStereochemistry` that depends on a temporary "_isStereoAny" bond property that is never assigned.


greglandrum (approved):
> LGTM

PR #7344, opened 2024-04-10
Add option for RASCAL to restrict atom matching to atoms of same degree

DavidACosgrove:
#### Reference Issue
No issue

Consider the pair of molecules
image
The default RASCAL options give rise to 2 distinct MCESs, with the 2 different amide moieties matching, as in:
image
They don't both match at the same time, obviously.
Sometimes, for example when looking for pairs of molecules that differ only by a linking group, that's a bit too permissive.  This option only allows atoms to match if they have the same degree (i.e. same number of explicit connections).  In this case the SMARTS returned is
```
[#6&a&D2]1:[#6&a&D2]:[#6&a&D2]:[#6&a&D2]:[#6&a&D2]:[#6&a&D3]:1.[#6&A&D3](=[#8&A&D1])-[#6&A&D3](-[#7&A&D1])-[#7&A&D1]
```
which gives the single match:
image

### DIFF
--- a/Code/GraphMol/RascalMCES/RascalMCES.cpp
+++ b/Code/GraphMol/RascalMCES/RascalMCES.cpp
@@ -19,7 +19,6 @@
 #include <iostream>
 #include <map>
 #include <regex>
-#include <set>
 #include <stdexcept>
 #include <unordered_set>
 #include <vector>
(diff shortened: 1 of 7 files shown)


greglandrum (approved):
LGTM

PR #7359, opened 2024-04-13
Rascal exactConnectionsMatch bug

DavidACosgrove:
#### Reference Issue
No issue.

Further testing of the new exactConnectionsMatch option for RASCAL revealed that the MCES sometimes gave incomplete results with different atom orders of the same molecule.

#### Any other comments?
I have also updated the docstring for the C++ MolToSmiles function as it was incomplete.

### DIFF
--- a/Code/GraphMol/RascalMCES/RascalMCES.cpp
+++ b/Code/GraphMol/RascalMCES/RascalMCES.cpp
@@ -255,14 +255,26 @@ void getBondLabels(const ROMol &mol, const RascalOptions &opts,
   getAtomLabels(mol, opts, atomLabels);
   bondLabels = std::vector<std::string>(mol.getNumBonds());
   for (const auto &b : mol.bonds()) {
-    if (b->getBeginAtom()->getAtomicNum() < b->getEndAtom()->getAtomicNum()) {
-      bondLabels[b->getIdx()] = atomLabels[b->getBeginAtomIdx()] +
-                                std::to_string(b->getBondType()) +
-                                atomLabels[b->getEndAtomIdx()];
+    if (b->getBeginAtom()->getAtomicNum() == b->getEndAtom()->getAtomicNum()) {
+      if (b->getEndAtom()->getDegree() < b->getBeginAtom()->getDegree()) {
+        bondLabels[b->getIdx()] = atomLabels[b->getEndAtomIdx()] +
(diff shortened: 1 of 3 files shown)


Code/GraphMol/SmilesParse/SmilesWrite.h
```diff
@@ -103,7 +103,7 @@ RDKIT_SMILESPARSE_EXPORT std::string MolToSmiles(
 RDKIT_SMILESPARSE_EXPORT std::string MolToSmiles(
     const ROMol &mol, const SmilesWriteParams &params);
 
-//! \brief returns canonical SMILES for a molecule
+//! \brief returns SMILES for a molecule, canonical by default
```
bp-kelley, line 106: This might have snuck in accidentally, not that more docs are bad!

DavidACosgrove: I came across it whilst fixing this bug, and it didn't seem worth a separate PR.

Code/GraphMol/RascalMCES/RascalMCES.cpp
```diff
@@ -255,14 +255,26 @@ void getBondLabels(const ROMol &mol, const RascalOptions &opts,
   getAtomLabels(mol, opts, atomLabels);
   bondLabels = std::vector<std::string>(mol.getNumBonds());
   for (const auto &b : mol.bonds()) {
-    if (b->getBeginAtom()->getAtomicNum() < b->getEndAtom()->getAtomicNum()) {
-      bondLabels[b->getIdx()] = atomLabels[b->getBeginAtomIdx()] +
-                                std::to_string(b->getBondType()) +
-                                atomLabels[b->getEndAtomIdx()];
+    if (b->getBeginAtom()->getAtomicNum() == b->getEndAtom()->getAtomicNum()) {
```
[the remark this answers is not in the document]
DavidACosgrove: Rather selfishly, perhaps, I went with just what I needed to get the job done.  Element and bond type are already encoded, so I don't think getTotalDegree would be any different from getDegree.  Charge might be useful, maybe, and you could make a case for options for things like treating all halogens the same.  Given that enhancements now go in in the bug fixing releases, I think it would be ok to work on an on-demand basis.

greglandrum (approved):
LGTM
Sorry I missed this one @DavidACosgrove

PR #7377, opened 2024-04-21
Restore mono build

jones-gareth:

Restore mono library build on Linux when creating csharp wrapper.

#### Any other comments?

Removing the mono library build broke existing workflows for users and the new dot net build does not work on all systems.


ptosco (approved):
I tested this on Linux and I could get a functional C# build with `mono`.
Thank you @jones-gareth!

ptosco (changes requested):
@jones-gareth Apologies, I was too fast with my approval.
If I understand correctly, currently what happens is:

**Windows, macOS**
* If `RDK_BUILD_DOTNET_CSHARP_TESTS=ON` the `RDKit2DotNet.dll` DLL will be built with `dotnet`, which is correct
* If `RDK_BUILD_DOTNET_CSHARP_TESTS=OFF`, the `RDKit2DotNet.dll` DLL will *not* be built , which is incorrect

**Linux**
* If `RDK_BUILD_DOTNET_CSHARP_TESTS=ON` the `RDKit2DotNet.dll` DLL will be built with `dotnet`, which is correct. However, if it succeeds, it will be built again with `mono`, which is incorrect.
* If `RDK_BUILD_DOTNET_CSHARP_TESTS=OFF` the `RDKit2DotNet.dll` DLL will be built with `mono` build, which is correct.

I think what we should rather aim for is:

* On all platforms, when `RDK_BUILD_DOTNET_CSHARP_TESTS=OFF`, the `RDKit2DotNet.dll` DLL is built with `mono`. I think this is reasonable stopgap until we are confident that we can get a functional C# build on all platforms./architectures
* On all platforms, when `RDK_BUILD_DOTNET_CSHARP_TESTS=ON`, the `RDKit2DotNet.dll` DLL is built with `dotnet`

ptosco:
@jones-gareth I think the logic is still incorrect.
As it is now, I believe the DLL will not be built if `MSVC` is defined and `RDK_BUILD_DOTNET_CSHARP_TESTS=OFF`, while we want the DLL to be always built on all platforms regardless of `RDK_BUILD_DOTNET_CSHARP_TESTS`.
In fact, the `dotnet` build is not done because `RDK_BUILD_DOTNET_CSHARP_TESTS=OFF`, and the `mono` build is not done either because `MSVC` is defined.
I think you should do the following changes:
* line 84: `if (RDK_BUILD_DOTNET_CSHARP_TESTS)` to `if (RDK_BUILD_DOTNET_CSHARP_TESTS OR MSVC)`: this will trigger a `dotnet` build if `MSVC` is defined, or if `RDK_BUILD_DOTNET_CSHARP_TESTS=ON` (by which we assume that the user knows that requesting tests to be built implies the availability of `dotnet`, even in the absence of MSVC)
* line 114: `if(NOT MSVC AND NOT RDK_BUILD_DOTNET_CSHARP_TESTS)` to `else(RDK_BUILD_DOTNET_CSHARP_TESTS OR MSVC)`: this will fall back to a `mono` build if `MSVC` is not defined and `RDK_BUILD_DOTNET_CSHARP_TESTS=OFF`
* line 134: `endif(NOT MSVC)` to `endif(RDK_BUILD_DOTNET_CSHARP_TESTS OR MSVC)`

jones-gareth:
I think the new logic is fine, but premature as it may break the csharp pipeline if dotnet 8 is not present in the Azure container. I will give it a go.

jones-gareth:
Great, this now runs the dotnet tests in the csharp azure container.

ptosco (approved):
I think this should be good for now, thanks @jones-gareth

greglandrum (approved):
Deferring to @ptosco's detailed review

PR #7395, opened 2024-04-27
Add option for non-isomeric SMILES creation in the PostgreSQL cartridge

### LINKED ISSUE #7307: Add option for non-isomeric SMILES creation in the PostgreSQL cartridge
I need to create non-isomeric SMILES inside the database cartridge. Therefore it would be helpful to have an additional argument on the `mol_to_smiles` function in the cartridge that would allow this, 

The same should be possible for cxsmiles. Also while not having an explicit use-case yet, having the option for cxsmiles flags in the cartridge also seems to make a lot of sense.

An additional argument to say if the returned SMILES should be isomeric or not

Doing it in python code but this can be more cumbersome and much slower for say "data migration" tasks when upgrading rdkit version. Doing it directly in the database would be preferable.

rvianello:
#### Reference Issue
Fixes #7307 

This PR should extend mol_to_smiles and mol_to_cxsmiles to support optional non-isomeric output.


greglandrum (approved):
LGTM

PR #7403, opened 2024-04-30
Do not apply the normalization of conjugated cations to the oxime oxygen

rvianello:
This PR further constraints the normalization of 1,3- and 1,5- aliphatic conjugated cations. The motivating case was about preventing the transformation of oxime groups into nitroso compounds.


greglandrum (approved):
LGTM

PR #7440, opened 2024-05-14
Update test_rfrag.py to pytest

watson-clara:

### DIFF
--- a/Contrib/mmpa/test_rfrag.py
+++ b/Contrib/mmpa/test_rfrag.py
@@ -5,25 +5,19 @@
 #
 import os
 import subprocess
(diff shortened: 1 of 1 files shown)


greglandrum (changes requested):
Hi @watson-clara, thanks for the PR and sorry for the very long time it took me to get to reviewing this.

The rest of the RDKit's testing is set up to use unittest for the tests (like what was being done here before your changes) and then use pytest to discover and run all of the tests.

I think that it makes sense to remain consistent with this for the contents of the Contrib folder too, even though these tests are not run as part of the standard RDKit testing.

PR #7465, opened 2024-05-27
Remove unnecessary wrapped pointers

bp-kelley:
This removes a collection of unnecessary wrapped pointers that may cause memory leaks.


greglandrum (approved):
LGTM

PR #7504, opened 2024-06-06
Avoid inconsistency between V2K and V3K MDL output wrt isotopic labelling of R groups

ptosco:
This was originally reported by @MariaDolotova.
While the V2000 MDL writer has logic to prevent writing the isotopic label associated to an R-group as a separate `M  ISO` record, the V3000 MDL writer lacks such logic, and hence the R-group atom has a `MASS` record. This is undesired and inconsistent with the V2000 behavior.
This PR corrects that.
Additionally, the decision whether the isotopic label should be written to the MDL output for an atom with atomic number 0 is made based on the presence of the `common_properties::_MolFileRLabel` rather than on the presence of an atom query as previously happened for V2000, which seems a preferrable choice to me as it is less restrictive.

### DIFF
--- a/Code/GraphMol/FileParsers/MolFileWriter.cpp
+++ b/Code/GraphMol/FileParsers/MolFileWriter.cpp
@@ -139,6 +139,11 @@ int getQueryBondSymbol(const Bond *bond) {
   }
   return res;
 }
+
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #7532, opened 2024-06-17
Multiple templates

ZontaNicola:
The templating code will now search for alternatives when a template is found that would put substituents inside templated rings. This allows to have different templates for the same smiles, in order of preference, and the code will try to find one that does not cause substituents inside rings, or at least to have as few as possible
Screenshot 2024-06-17 at 11 09 18


Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -346,14 +346,151 @@ void EmbeddedFrag::setupAttachmentPoints() {
...
+        for (auto atom : vec) {
+          average += ringSystemAtomsCoords.at(atom);
+        }
+        average /= vec.size();
+        return average;
+      };
+      auto averageNeighborCoords = averageCoords(neighborsInsideTemplate);
+      auto substitutionV = (averageNeighborCoords - coords);
+      substitutionV.normalize();
+      substitutionV *= -0.4;
```
rachelnwalker, line 475: How did you come up with `-0.4`? Does this signify where the non-templated neighbor atom would be placed?

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -346,14 +346,151 @@ void EmbeddedFrag::setupAttachmentPoints() {
...
+
+bool EmbeddedFrag::matchToTemplate(const RDKit::VECT_INT_VECT &fusedRings,
+                                   const RDKit::INT_VECT &ringSystemAtoms,
                                    unsigned int ring_count) {
   CoordinateTemplates &coordinate_templates =
       CoordinateTemplates::getRingSystemTemplates();
 
-  // only look for an exact match to the ring system because our method of
-  // completing rings from a template isn't reliably better than not using
-  // a template at all
```
rachelnwalker, line 356: Isn't this partially true in the scope of this function? I think this would make sense to be removed in your other review

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -346,14 +346,151 @@ void EmbeddedFrag::setupAttachmentPoints() {
...
+        }
+        average /= vec.size();
+        return average;
+      };
+      auto averageNeighborCoords = averageCoords(neighborsInsideTemplate);
+      auto substitutionV = (averageNeighborCoords - coords);
+      substitutionV.normalize();
+      substitutionV *= -0.4;
+      for (auto ring : fusedRings) {
+        if (pointInsideRing(coords + substitutionV, ringSystemAtomsCoords,
```
rachelnwalker, line 477: Would it be worth only checking if a point is inside a ring if the templated atoms are in that ring?

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -396,11 +534,27 @@ bool EmbeddedFrag::matchToTemplate(const RDKit::INT_VECT &ringSystemAtoms,
     auto matches = RDKit::SubstructMatch(RDKit::ROMol(rs_mol), *mol, params);
```
rachelnwalker, line 534: Don't you need to set `maxMatches` higher?

rachelnwalker (commented):
I trust your math here, it sounds like you know what you're talking about :) 

I made a few comments, but the diff is a little complicated to read because I think its based off your other review and has slightly diverged. I think it would be helpful to add a test and show a few more examples if possible, and probably run the benchmark Greg suggested here https://github.com/rdkit/rdkit/pull/7468#issuecomment-2196153262 just to track performance.

cdvonbargen:
I think that this should also allow us to address https://github.com/rdkit/rdkit/issues/6753#issuecomment-2423487215

PR #7544, opened 2024-06-20
Restore mol writing functions in the RDKFuncs module

ptosco:
As part of #7077 mol writers were moved to `Code/GraphMol/FileParsers/FileWriters.h`, but since `Code/JavaWrappers/MolWriters.i` was not updated accordingly, symbols which used to be available in the `RDKFuncs` module are now missing.
This went unnoticed as there were no unit tests in RDKit for those `RDKFuncs` free functions, since the same functions are available as `ROMol` methods.
However, as there may be Java or C# code using those free functions, I think it would be good to restore them; for example, this was noticed by @manuelschwarze while running KNIME tests on 2024.03.3.
This PR restores the missing symbols and adds a couple of tests to avoid similar regressions in the future.

### DIFF
--- a/Code/JavaWrappers/MolWriters.i
+++ b/Code/JavaWrappers/MolWriters.i
@@ -32,9 +32,11 @@
 */
 
 %{
+#include <GraphMol/FileParsers/FileWriters.h>
 #include <GraphMol/FileParsers/MolWriters.h>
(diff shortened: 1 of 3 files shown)


greglandrum (approved):
LGTM

PR #7546, opened 2024-06-20
Avoid duplicate enhanced stereo labels in drawing

ptosco:
This is the simple fix for #7543.
However this fix does not address the following point:

> However, users might still want to call `Chirality::addStereoAnnotations` to display custom stereochemistry labels with their own preferred format, potentially including enhanced stereo labels, and hence a way to prevent default enhanced stereo annotations from being added.
A possibility could be to add a `hideEnhancedStereoAnnotations` flag defaulting to false.
We have briefly discussed this with @greglandrum and he would like to think more carefully about this.


greglandrum (approved):
LGTM

Issue #7545, opened 2024-06-20
Maximal common substructure smaller than expected


maabuu:
**Describe the bug**
Maximal common substructures are smaller than expected when taking stereochemistry into account.

**To Reproduce**
```
from rdkit.Chem.rdmolfiles import MolFromSmiles, MolFromSmarts
from rdkit.Chem.rdFMCS import FindMCS, RingCompare, AtomCompare, BondCompare

# VRI: https://www.rcsb.org/ligand/VRI
smiles = "CC(=O)[C@H]1CC[C@@H]2[C@@]1(CC[C@H]3[C@H]2CC[C@@H]4[C@@]3(CC[C@H](C4)OS(=O)(=O)N)C)C"
mol = MolFromSmiles(smiles) 
```

Using default settings, the MCS finds the entire molecule:
```
mcs = FindMCS([mol, mol])
mol.GetSubstructMatch(MolFromSmarts(mcs.smartsString))
mol
```
![image](https://github.com/rdkit/rdkit/assets/42468905/4246223a-2a22-4c10-9051-e2f26da9b6dc)

Using additional options, it does not:
```
mcs = FindMCS([mol, mol], matchChiralTag=True)
mol.GetSubstructMatch(MolFromSmarts(mcs.smartsString))
mol
```
![image](https://github.com/rdkit/rdkit/assets/42468905/ca79464e-f9f9-4380-bc84-c37815b7529f)

```
mcs = FindMCS([mol, mol], matchChiralTag=True, ringMatchesRingOnly=True)
mol.GetSubstructMatch(MolFromSmarts(mcs.smartsString))
mol
```
![image](https://github.com/rdkit/rdkit/assets/42468905/09cec18b-ad82-4cde-85b4-2b2f385a5d7b)

```
mcs = FindMCS([mol, mol], matchChiralTag=True, ringMatchesRingOnly=True, ringCompare=RingCompare.StrictRingFusion)
mol.GetSubstructMatch(MolFromSmarts(mcs.smartsString))
mol
```
![image](https://github.com/rdkit/rdkit/assets/42468905/fad8d889-fd9c-408c-8c30-afa78fad5f51)

**Expected behaviour**
The MCS should be the entire molecule for all cases. 

**Configuration (please complete the following information):**
 - RDKit version: 2024.03.3
 - OS: Fedora release 38 (Thirty Eight)
[...]

PR #7559, opened 2024-06-24
Add MCS unit tests for stereochemistry

maabuu:
#### Reference Issue
Adds unit tests for #7545

Adds stereochemistry sensitive unit tests for the maximal common substructure module. This is needed to test the expected behaviour of the option `matchChiralTag=True`.


bertiewooster:
Nice job implementing tests. Perhaps include each parameter in FindMCS (say, Boolean and Compare ones) to catch any problems with other parameters in [FindMCS](https://www.rdkit.org/docs/source/rdkit.Chem.rdFMCS.html#rdkit.Chem.rdFMCS.FindMCS)?

PR #7561, opened 2024-06-24
fix one case of undesired 1-3 charge recombination

rvianello:
This PR fixes one of the "1-3 charge recombination" normalization transformations to prevent the undesired application on two adjacent N-oxide groups.

#### Any other comments?
A similar constraint I think is already in place for the aliphatic case, and this is basically the same, but applying to aromatic rings.


greglandrum (approved):
LGTM

PR #7575, opened 2024-06-27
Add missing GetRow method and fix Python parameter names

ptosco:
Small PR to add a missing `PeriodicTable` method in Python and also to fix a few misnamed Python parameters.


greglandrum (approved):
LGTM

PR #7591, opened 2024-07-09
Update CMake modules for finding flex and bison(#7590)

### LINKED ISSUE #7590: Update FindFLEX.cmake and FindBISON.cmake modules.
The current modules have been around for about 15 years, and currently limits us from using newer flex features. i.e. there's currently not an easy way to use the include directory of the found executable for C++ lexers.

We should use the up-to-date modules from https://github.com/Kitware/CMake/tree/master/Modules

**Additional context**

whosayn:
#### Reference Issue
Fixes #7590

This will improve the executable finding procedures and allow us to require minimum release versions. The current modules have been around for about 15 years, and currently limits us from using newer flex features. I just updated the modules with the current versions from https://github.com/Kitware/CMake/tree/master/Modules


whosayn:
> Given that these are now provided by cmake (I think maybe they weren't when we first started using cmake?), how about we remove these files entirely?

Good point!

greglandrum (approved):
LGTM

PR #7647, opened 2024-07-22
MinimalLib: fix misformatted molblock and make sure unit test is actually executed

ptosco:
Small PR as I have just realized that one of the unit test added as part of #7540 1. was not actually being run 2. featured a misformatted molblock which fail to yield a molecule.


greglandrum (approved):
LGTM

PR #7672, opened 2024-07-31
Highlighted hetero atoms disappear

### LINKED ISSUE #7654: Highlighted hetero atoms disappear
image

1 picture > 1000 words, apparently.

**To Reproduce**

**Expected behavior**
It would be nice to know that the highlighted group is a carboxylic acid.

**Screenshots**

 - RDKit version: Current master
 - OS: [e.g. Ubuntu 20.04] macOS, but not OS specific
 - Are you using conda?
 - If you are not using conda: how did you install the RDKit? Built from source

**Additional context**
Working on a fix that makes the atoms black in these circumstances.

DavidACosgrove:
#### Reference Issue
Fixes #7654

If a hetero atom is highlighted by a background colour it is drawn in black rather than its normal element colour, so it stands out better.
Formerly:
image
Now:
image


Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -3543,6 +3543,13 @@ DrawColour DrawMol::getColour(int atom_idx) const {
         retval = *highCol;
       }
     }
+  } else {
+    // There's going to be a colour behind the atom, so if the
+    // atom has a symbol, it should be black and this should only
+    // be called if there is an atom symbol.
+    if (highlightedAtom) {
+      retval = DrawColour(0.0, 0.0, 0.0);
```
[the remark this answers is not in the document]
DavidACosgrove: Interestingly, DarkMode gives this at present:
image
What colour should the background be?  I'm guessing not white.

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -3545,10 +3545,17 @@ DrawColour DrawMol::getColour(int atom_idx) const {
     }
   } else {
     // There's going to be a colour behind the atom, so if the
-    // atom has a symbol, it should be black and this should only
-    // be called if there is an atom symbol.
+    // atom has a symbol, it should be the same colour as carbon.  This
+    // function should only be called if there is an atom symbol.
     if (highlightedAtom) {
-      retval = DrawColour(0.0, 0.0, 0.0);
+      if (drawOptions_.atomColourPalette.find(6) !=
```
greglandrum, line 3551: I think you mean `-1` here.

greglandrum (changes requested):
There's an inconsistency in what you search for as a default color

DavidACosgrove:
I deliberately decided to make the atom colour the same as the carbon atom, rather than the default.  The only place it makes a difference, at least with the system colour palettes, is with dark mode, where the default colour is a slightly darker grey.  I think it's more consistent this way, but can change it if you prefer.  The behaviour is commented in the code.

greglandrum:
> I deliberately decided to make the atom colour the same as the carbon atom, rather than the default. The only place it makes a difference, at least with the system colour palettes, is with dark mode, where the default colour is a slightly darker grey. I think it's more consistent this way, but can change it if you prefer. The behaviour is commented in the code.

It's just a little bit strange because:
1. none of the palettes the RDKit provides actually have an explicit entry for 6 (the default is generally used for C)
2. you search to see if 6 is there, but then you use -1

greglandrum:
Also: there is unfortunately now a conflict

DavidACosgrove:
> > I deliberately decided to make the atom colour the same as the carbon atom, rather than the default. The only place it makes a difference, at least with the system colour palettes, is with dark mode, where the default colour is a slightly darker grey. I think it's more consistent this way, but can change it if you prefer. The behaviour is commented in the code.
> 
> It's just a little bit strange because:
> 
> 1. none of the palettes the RDKit provides actually have an explicit entry for 6 (the default is generally used for C)
All the palettes except the BW one set colours for 6 to be the same as 1, e.g.
https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/MolDraw2D/MolDraw2DHelpers.h#L71

> 2. you search to see if 6 is there, but then you use -1
Clearly I was having a bad day!  It now uses the answer correctly, I think, and falls back to the default value.

DavidACosgrove:
I think I've fixed the conflicts correctly, though maybe the order of the tests has changed a bit.  The normally awesome CLion merge tool made a bit of a hash of it and I ended up doing it by hand.

greglandrum:
> > 1. none of the palettes the RDKit provides actually have an explicit entry for 6 (the default is generally used for C)
> >    All the palettes except the BW one set colours for 6 to be the same as 1, e.g.
> >    https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/MolDraw2D/MolDraw2DHelpers.h#L71

Hah! right you are. This is what I get for just scanning the code and not actually reading it.

> 
> > 2. you search to see if 6 is there, but then you use -1
> >    Clearly I was having a bad day!  It now uses the answer correctly, I think, and falls back to the default value.

👍

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -3562,6 +3562,19 @@ DrawColour DrawMol::getColour(int atom_idx) const {
...
+    // There's going to be a colour behind the atom, so if the
+    // atom has a symbol, it should be the same colour as carbon.  This
+    // function should only be called if there is an atom symbol.
+    if (highlightedAtom) {
+      if (drawOptions_.atomColourPalette.find(6) !=
+          drawOptions_.atomColourPalette.end()) {
+        retval = drawOptions_.atomColourPalette.find(6)->second;
+      } else {
+        // Use the default if no carbon.
+        retval = drawOptions_.atomColourPalette.find(-1)->second;
```
greglandrum, line 3575: This would be cleaner (and safer) as:
```suggestion
      if (drawOptions_.atomColourPalette.find(6) !=
          drawOptions_.atomColourPalette.end()) {
        retval = drawOptions_.atomColourPalette.at(6);
      } else {
        // Use the default if no carbon.
        retval = drawOptions_.atomColourPalette.at(-1);
```

greglandrum (approved):
LGTM

PR #7695, opened 2024-08-06
fix RDKitEquals strategy of gmol_consistent and greaction_consistent

### LINKED ISSUE #7493: Index anomaly with mol @= mol operator on PostgreSQL. Exact match is extremely slow!

rvianello:
#### Reference Issue
Fixes #7493 

This PR includes @ergo70's fix for the RDKitEquals strategy of gmol_consistent and applies the same changes to greaction_consistent.

#### Any other comments?
@ergo70: sorry I couldn't find your branch `ergo70-issue_7493` in github (was it deleted?) and I was for this reason unable to include your original commit objects.

### DIFF
--- a/Code/PgSQL/rdkit/rdkit_gist.c
+++ b/Code/PgSQL/rdkit/rdkit_gist.c
@@ -517,7 +517,7 @@ Datum gmol_consistent(PG_FUNCTION_ARGS) {
       if (!ISALLTRUE(key)) {
         /*
         ** verify the necessary condition that key should contain the query
-        ** (on leaf nodes, couldn't it also verify that query contains key?)
+        ** (on leaf nodes, check that key and query match)
         */
         if (siglen != SIGLEN(query)) {
           elog(ERROR, "All fingerprints should be the same length");
@@ -526,7 +526,11 @@ Datum gmol_consistent(PG_FUNCTION_ARGS) {
         uint8 *k = (uint8 *)VARDATA(key);
         uint8 *q = (uint8 *)VARDATA(query);
 
-        res = bitstringContains(siglen, k, q);
+        if (GIST_LEAF(entry)) {
+          res = (memcmp(k, q, siglen) == 0);
+        } else {
+          res = bitstringContains(siglen, k, q);
+        }
       }
       break;
     default:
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

But that's not because I understand this code. :-)

PR #7706, opened 2024-08-11
Only test InChI functionality if it was built into JS MinimalLib

ptosco:
#7698 enables building MinimalLib without support for InChI, so we need to make sure that JS InChI tests run only if InChI support was built in.


greglandrum (commented):
LGTM

greglandrum (approved):

PR #7715, opened 2024-08-14
Reduce code duplication in ff

nmaeder:
Move `DistanceConstraint`, `TorsionConstraint` and `PositionConstraint` to the `ForceField` level. The implementations of all three of them are the same in `MMFF` and `UFF`, so this is removing some code duplication. The respective header files are kept, to keep the restraints in their respective namespaces for backwards compatibility.

#### Any other comments?
Removing some includes that are not used.

In `DistanceConstraint.cpp`, the definition of `dist` is moved inside the `if(relative)` as we have done it for the angle too.


greglandrum (approved):
LGTM

PR #7743, opened 2024-08-22
Expose two additional boolean flags for mol creation to MinimalLib

ptosco:
@greglandrum As discussed, I extracted into this PR the two additional mol creation flags for MinimalLib which were originally part of #7322.


greglandrum (approved):
LGTM

PR #7748, opened 2024-08-23
Exposure to minilib: molzip

StLeonidas:
molzip feature is exposed to minilib for fragments linking:
- feature itself
- test at JS
- parseing the details


StLeonidas:
@ptosco FYI

PR #7749, opened 2024-08-23
Fix vectToString such that it outputs valid JSON

ptosco:
This small PR fixes `vectToString` such that it outputs valid JSON (i.e., without a trailing comma) such that it can be directly parsed as a JSON string.
@greglandrum As discussed, I have extracted this change out of #7322.


greglandrum (approved):
LGTM

PR #7755, opened 2024-08-23
Dev/remove iter from vector wrappers

bp-kelley:
@greglandrum this was my attempt to fix the iteration problem at a deeper level.

Python has two iterator protocols, the ancient one uses __len__ and __getitem__ the newer one uses __iter__.  This remove the __iter__ wrapping from the boost iterator when registering vectors.

It all works great... except for when you send one of these wrapped vectors into a function that uses pyObjectToVector.  We only test this for rdAbbreviations, but I think users might do this a lot more often potentially?  Anyway, it's here for your thoughts and review.


Code/RDBoost/rdkit_indexing_suite.hpp
```diff
@@ -0,0 +1,295 @@
...
+            // Hook into the class_ generic visitation .def function
+            proxy_handler::register_container_element();
+
+            cl
+                .def("__len__", base_size)
+                .def("__setitem__", &base_set_item)
+                .def("__delitem__", &base_delete_item)
+                .def("__getitem__", &base_get_item)
+                .def("__contains__", &base_contains)
+                 // .def("__iter__", def_iterator())
```
bp-kelley, line 185: All this code was copied so I could comment this single line out...

PR #7778, opened 2024-08-30
Fix issue in #7720

ricrogz:
It seems that the newest versions of CMake may have problems with the `Python3_LIBRARY` variable (we've seen some on Mac with CMake 3.27.1) that was introduced in #7720 (before we used `Python3_LIBRARIES`).

The CMake documentation for `FindPython3` mentions `Python3_LIBRARY` as an "Artifact" (I'm not sure what that is), but it seems that the correct "result" variable is `Python3_LIBRARIES`.


greglandrum (approved):
LGTM

PR #7786, opened 2024-09-03
HasPropWithValueQueryBase needs a virtual destructor

ricrogz:
Since #7692, including `GraphMol/QueryOps.h` in a .cpp file results in a build error:

```

In file included from /tmp/rdkit-install/include/rdkit/GraphMol/QueryAtom.h:17,
                 from /tmp/test/rdkit.cpp:15:
/tmp/rdkit-install/include/rdkit/GraphMol/QueryOps.h:866:7: error: ‘class RDKit::HasPropWithValueQueryBase’ has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
  866 | class HasPropWithValueQueryBase {
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors

```

This PR gives `HasPropWithValueQueryBase()` a default constructor and a virtual destructor, which fixes the build issue.

### DIFF
--- a/Code/GraphMol/QueryOps.h
+++ b/Code/GraphMol/QueryOps.h
@@ -865,6 +865,8 @@ Queries::EqualityQuery<int, const Target *, true> *makeHasPropQuery(
 // ! Query whether an atom has a property with a value
 class HasPropWithValueQueryBase {
   public:
+  HasPropWithValueQueryBase() = default;
+  virtual ~HasPropWithValueQueryBase() = default;
   virtual Dict::Pair getPair() const = 0;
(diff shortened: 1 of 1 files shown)


bp-kelley:
Thanks.  This error is a new one for me since the virtual functions are abstract and need to be overridden, but it's way better to be safe!

bp-kelley (approved):

bp-kelley:
I'm not sure we need the default constructor though.

greglandrum (approved):
LGTM

PR #7801, opened 2024-09-10
Fix typo (Fingeprint -> Fingerprint)

ptosco:
Given that those classes cannot be instantiated from Python and currently
```
grep -l FingeprintGenerator -r .
```
in the `rdkit` source tree only yields hits in `./Docs/Book` and `./rdkit-stubs` I would not bother deprecating the old name with the typo and just fix it.


greglandrum (approved):
LGTM

PR #7804, opened 2024-09-11
Drop mol_hash for legacy databases to enable upgrade to recent versions

mwojcikowski:
Discussion #6711

We lost track of `mol_hash` function removed in #3047. No clue how to test this properly. Should be harmless, as it contains the `IF EXISTS` clause.


mwojcikowski:
I do not think the failures are real. Cannot rerun...

greglandrum (commented):
LGTM

greglandrum (approved):

PR #7805, opened 2024-09-11
A couple of fixes to the InteractiveRenderer Jupyter integration

ptosco:
This small PR fixes two issues that I noticed when trying to set a custom size on a specific image through the `data-width` and `data-height` `div` attributes, namely:

- Allow overriding the default image size through individual image options in the InteractiveRenderer
- `float` and `int` attributes should be converted to `str` before calling `setAttribute`


greglandrum (approved):
@ptosco we don't (and can't) have any automated tests for this. I assume you tested it yourself?

ptosco:
@greglandrum yes. I realised that there was a problem while preparing the UGM demo.

Issue #7781, opened 2024-09-01
Embed error with srETKDGv3 with extreme examples


phonglam3103:
**Describe the bug**
The function rdDistGeom.EmbedMultipleConfs using srETKDGv3 returns an error on extreme cases with cyclopentane scaffold. Interestingly, upon trying with cyclopentene, and trying using different random seeds, rdkit was able to handle it. 

In this example, random seed 1231, 12331 works (in 2024.3.4). As for 2024.3.5, so far I have not found any random seed from 1 to 25000 that works.

**To Reproduce**
"""
from rdkit import Chem
from rdkit.Chem import Draw, rdDistGeom
smiles = 'C[C@H]1C[C@H](Cl)CC1'
mol = Chem.AddHs(Chem.MolFromSmiles(smiles))
params = rdDistGeom.srETKDGv3()
params.numThreads = 0  
params.pruneRmsThresh = 0.35  
params.randomSeed = 41 
params.useRandomCoords = True
rdDistGeom.EmbedMultipleConfs(mol, numConfs=300, params=params)
mol.GetNumConformers()
"""

Output:
"""
RuntimeError: Invariant Violation
	bad direction in linearSearch
	Violation occurred on line 240 in file Code/Numerics/Optimizer/BFGSOpt.h
	Failed Expression: status >= 0
	RDKIT: 2024.03.4
	BOOST: 1_85

RuntimeError: Invariant Violation
	bad direction in linearSearch
	Violation occurred on line 240 in file Code/Numerics/Optimizer/BFGSOpt.h
	Failed Expression: status >= 0
	RDKIT: 2024.03.5
	BOOST: 1_85
"""
**Expected behavior**
I understand that the embedding could be dependent on the initial coordinates of the atoms, but should be at least not returning any error and should produce at least one conformation. I believe this is the same problem as https://github.com/rdkit/rdkit/issues/7726.
**Screenshots**

**Configuration (please complete the following information):**
 - RDKit version: 2024.3.4 and 2024.3.5
 - OS: Ubuntu 20.04
 - Python version (if relevant): 3.9
 - Are you using conda? yes
 - If you are using conda, which channel did you install the rdkit from? 
 - If you are not using conda: how did you install the RDKit? pip

phonglam3103:
Hi @greglandrum, just want to make sure that you are informed of this bug. Thank you.

rachelnwalker:
[...]

Issue #7726, opened 2024-08-18
Embed Molecule numerical instability edge case?


fredtony:
**Describe the bug**
When I try to create conformers of this specific molecule `N[C@@H]1CCC[C@@H]1O` using `ETKDGv3` params and `useSmallRingTorsions`, it seems to fail with this error:
```
RuntimeError: Invariant Violation
	bad direction in linearSearch
	Violation occurred on line 240 in file Code/Numerics/Optimizer/BFGSOpt.h
	Failed Expression: status >= 0
	RDKIT: 2024.03.5
	BOOST: 1_85
```

Some rdkit versions have slight differences in behavior. For example, in version `2024.03.5`, this happens with any random seed, but in version `2023.09.6`, the only seed that seems to fail is `5031` but changing to `5032` (or other seeds) no longer fails.

**To Reproduce**
```
from rdkit import Chem
from rdkit.Chem import rdDistGeom

smiles = "N[C@@H]1CCC[C@@H]1O"
mol = Chem.AddHs(Chem.MolFromSmiles(smiles))
conformer_params = rdDistGeom.ETKDGv3()
conformer_params.useSmallRingTorsions = True
conformer_params.randomSeed = 5031
rdDistGeom.EmbedMultipleConfs(mol, 5, conformer_params)
```

**Expected behavior**
I would expect typical `rdDistGeom.EmbedMultipleConfs` behavior where it does not error out even when no conformers are found.

**Screenshots**
N/A

**Configuration (please complete the following information):**
 - RDKit version: 2024.03.5
 - OS: [e.g. Ubuntu 20.04]: Ubuntu 22.04.4
 - Python version (if relevant): 3.10
 - Are you using conda? Yes, but I'm installing rdkit with pip
 - If you are using conda, which channel did you install the rdkit from? N/A
 - If you are not using conda: how did you install the RDKit? `pip install rdkit==2024.03.5` and `pip install rdkit==2023.09.6`

**Additional context**
N/A

rachelnwalker:
This looks like the same error in the minimizer as #7671 

I looked into that issue a little bit and traced it back to some torsion energy calculations are being evaluated to `nan`.

rachelnwalker:
Also #7781 -- @ricrogz is looking into this.

j-adamczyk:
[...]

PR #7814, opened 2024-09-13
Torsion energy calculations are sometimes nan

### LINKED ISSUE #7671: constrained minimization
**Describe the bug**
I want to do a constrained minimization. but sometimes, it failed. After trouble shooting, I found the forcefild energy and grad were nan. How to solve it？ the following are my code and sdf, atoms 0-37 were constrained.

```
suppl_ref = Chem.SDMolSupplier('test.sdf')

mol = suppl_ref[0]

molRmHs = Chem.RemoveHs(mol)

mp = AllChem.MMFFGetMoleculeProperties(molRmHs)

ff = AllChem.MMFFGetMoleculeForceField(molRmHs, mp)

constrained_indices = [atomid for atomid in range(38)]

for ii in constrained_indices:

    ff.MMFFAddPositionConstraint(ii, 0, 1.e4)

ff.Minimize(maxIts=1000)
```

**Screenshots**
image

image

image

rachelnwalker:
#### Reference Issue
Fixes #7671

`calcTorsionCosPhi` was returning `nan` when two adjacent bonds in a torsion are parallel. For example, atoms 37, 38, 39, 42 here:
image

This change makes the return 0.0 instead so that the minimization doesn't fail. I'm not sure if that is the right thing to do or not.

#### Any other comments?
I also spent some RDKit UGM hackathon time looking into #7781 and #7726 because they had the same error message. I didn't have time to actually fix them, but I believe they are due to the entire gradient being set to 0 causing the linearSearch to return early.

### DIFF
--- a/Code/ForceField/MMFF/TorsionAngle.cpp
+++ b/Code/ForceField/MMFF/TorsionAngle.cpp
@@ -26,7 +26,12 @@ double calcTorsionCosPhi(const RDGeom::Point3D &iPoint,
   RDGeom::Point3D r4 = lPoint - kPoint;
   RDGeom::Point3D t1 = r1.crossProduct(r2);
   RDGeom::Point3D t2 = r3.crossProduct(r4);
-  double cosPhi = t1.dotProduct(t2) / (t1.length() * t2.length());
+  auto t1_len = t1.length();
+  auto t2_len = t2.length();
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

PR #7817, opened 2024-09-14
Avoid pubchem-align3d build failure on g++ >=13

ptosco:
Currently `pubchem-align3d` fails to build on g++ >=13 because of a missing `#include`.
I have already submitted a PR to the `pubchem-align3d` repo to fix the root issue:
https://github.com/ncbi/pubchem-align3d/pull/1

but here's a patch to fix the issue on the RDKit side until the problem is fixed upstream.


greglandrum (approved):
regexps are bad enough, but the way they work in cmake is really ugly

PR #7834, opened 2024-09-19
Additional compatibility updates for commutator/negator settings for …

esiaero:
…operator @=

#### Reference Issue

Follow-up for additional fixes for https://github.com/rdkit/rdkit/issues/7459. 
Also see conversation at the bottom of https://github.com/rdkit/rdkit/pull/7596.

Allows upgrades from postgres versions <= 16 with existing installs to 17 and bumps cartridge version to 4.6.1.

### DIFF
--- a/Code/PgSQL/rdkit/CMakeLists.txt
+++ b/Code/PgSQL/rdkit/CMakeLists.txt
@@ -261,8 +261,6 @@ else()
 endif()
 
 if (${PostgreSQL_VERSION_STRING} VERSION_GREATER_EQUAL 17.0)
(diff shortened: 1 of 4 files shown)


esiaero:
@rvianello Bumping this topic, since PostgreSQL 17.0 was released yesterday - hope you can find some time to take a look. Thanks

rvianello:
@esiaero sorry for the slow feedback, I don't have the time to test the PR, and especially do so in a few different scenarios, but (also following the discussion we had on #7596) your approach definitely makes sense to me and it would be a clear improvement compared to my original proposal.

greglandrum (approved):
LGTM based on @rvianello's feedback

PR #7853, opened 2024-09-26
Expose preset color palettes to MinimalLib through JSON

ptosco:
Small PR to expose preset color palettes to MinimalLib via JSON.
Custom palette could already be loaded via a JSON object as follows:
```
{
    "atomColourPalette": {
        "7", [1.0, 0.0, 0.0]
    }
}
```
Now in addition to the JSON object mapping atomic numbers to RGB(A) tuples, it is also possible to associate the `atomColourPalette` key to a string value chosen among `default, avalon, cdk, darkmode, bw` (all case-insensitive), e.g.:
```
{
    "atomColourPalette": "bw"
}
```


greglandrum (approved):
LGTM

PR #7861, opened 2024-09-27
Gitignore fix

PatrickPenner:

- There seems to be some inconsistency between the .gitignore entries and the files written during test execution
- A number test files appear to be written to different paths than in the .gitignore
- Add an external lib to the .gitignore


greglandrum (approved):
LGTM

PR #7862, opened 2024-09-27
Improve Morgan Fingerprint Performance

scal444:
Fingerprinting is improved via the following changes:

* Neighborhoods are stored in a set rather than a vector, as lookup is the only operation performed
* Removed an extraneous neighborhood lookup that was already covered elsewhere in the code
* Pull vectors out of loops and preallocate where possible
* Use swap on  current and next level invariants.

The aggregate speedup on a random sample of Chembl molecules was about ~60% on radius = 2, and 100% on radius = 3

### DIFF
--- a/Code/GraphMol/Fingerprints/MorganGenerator.cpp
+++ b/Code/GraphMol/Fingerprints/MorganGenerator.cpp
@@ -1,5 +1,5 @@
 //
-//  Copyright (C) 2018-2022 Boran Adas and other RDKit contributors
+//  Copyright (C) 2018-2024 Boran Adas and other RDKit contributors
 //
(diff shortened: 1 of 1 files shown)


scal444:
On my machine, here's a scan of measured speedups per radius on a random sample of chembl molecules, single threaded.
Screenshot from 2024-08-22 14-34-18

I've attached a zip file with a small cmake project and build instructions to run the benchmarks using nanobench. 
[benchmark.zip](https://github.com/user-attachments/files/17171100/benchmark.zip)

greglandrum (approved):
LGTM

greglandrum:
Thanks @scal444 !

PR #7864, opened 2024-09-28
Fixes #7685

### LINKED ISSUE #7685: rdMolAlign.GetBestRMS gets stuck
Hello, while trying to get the best RMS value between 2 conformers, the code runs infinitely without returning any results. I have attached an example SDF file, and below is the reproduction code:
```python
    from rdkit.Chem import rdMolAlign


    path = '/home/jack/Auto3D_pkg/src/debug.sdf'
    supp = Chem.SDMolSupplier(path, removeHs=True)
    mol1 = supp[0]
    mol2 = supp[1]
    rmsd = rdMolAlign.GetBestRMS(mol1, mol2)
    print(rmsd)
```
[debug.sdf.txt](https://github.com/user-attachments/files/16465581/debug.sdf.txt)

**Configuration:**

- RDKit Version: 2023.9.5
- Operating system: Ubuntu 20.04.6 LTS
- Python version (if relevant): 3.9.19
- Are you using conda? Yes
- If you are using conda, which channel did you install the rdkit from? pypi
- If you are not using conda: how did you install the RDKit? pypi

**Description:**

When my code executes the line containing `rmsd = rdMolAlign.GetBestRMS(mol1, mol2)`, it gets stuck there forever. Does anyone know the cause for this? Thanks a lot!

ricrogz:
The cause of the issue seems to be the bad performance of `std::list` in some C++ implementations (probably older ones). For large number of matches, like in the example issue description, this difference may become very large, making the call intractable.

This fixes the issue by replacing `std::list` with `std::vector`. I have benchmarked the change on 3 different platforms using the example script in the issue (maxMatches=1000000).

```
2019 Intel Mac on macOS 14.6.1 / xcode 16: 0.06 minutes
Intel i7-10870H / Ubuntu 22.04 / gcc 12.3.0: 0.35 minutes
Intel w5-2465X / Rocky8 container / gcc 11.3.0: ... never finishes running (I interrupted the run after several minutes).
```

After this patch:

```
2019 Intel Mac on macOS 14.6.1 / xcode 16: 0.03 minutes
Intel i7-10870H / Ubuntu 22.04 / gcc 12.3.0: 0.28 minutes
Intel w5-2465X / Rocky8 container / gcc 11.3.0: 0.20 minutes
```

### DIFF
--- a/Code/GraphMol/Substruct/SubstructMatch.cpp
+++ b/Code/GraphMol/Substruct/SubstructMatch.cpp
@@ -175,7 +175,7 @@ void ResSubstructMatchHelper_(const ResSubstructMatchHelperArgs_ &args,
                               std::set<MatchVectType> *matches, unsigned int bi,
                               unsigned int ei);
 
-typedef std::list<
+typedef std::vector<
     std::pair<MolGraph::vertex_descriptor, MolGraph::vertex_descriptor>>
     ssPairType;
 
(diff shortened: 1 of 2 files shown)


bp-kelley (approved):
Nice!

greglandrum (approved):
LGTM

PR #7883, opened 2024-10-02
Use smart pointer for buffer.

DavidACosgrove:
#### Reference Issue
No issue.

Since we're fixing possible memory leaks, I just stumbled upon this one.  If the 2nd exception is thrown, buff will leak, I think.


greglandrum (approved):
LGTM

PR #7884, opened 2024-10-02
Modernise ExplicitBitVect.

DavidACosgrove:
#### Reference Issue
No issue

Holds the bitset in a unique_ptr rather than a raw pointer.
Adds move c'tor and move operator=.
Some minor tidying.

#### Any other comments?
Technically, this is a breaking change because someone could be uncouth enough to be using dp_bits directly, since it's public.  Maybe it should be made private?


greglandrum (approved):
LGTM

PR #7898, opened 2024-10-04
Restore "oops, exponential is a pain" code snippet in the RGD algorithm

ptosco:
Restores a RGD code snippet that was intentionally removed in #7186 because 1. it is unclear 2. it is not tested 3. It introduced an undocumented change in the documented `GreedyChunks` behavior based on an arbitrary threshold. Discussed in #7896.

### DIFF
--- a/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
+++ b/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
@@ -388,11 +388,14 @@ int RGroupDecomposition::add(const ROMol &inmol) {
   // Exhaustive - saves all matches and optimizes later exhaustive
   //               May never finish due to combinatorial complexity
   // Greedy - matches to *FIRST* available match
-  // GreedyChunks - default - process every N chunks
+  // GreedyChunks - default - process every N chunks, unless
+  // MAX_PERMUTATIONS is exceeded, in which case it falls back to
+  // Greedy
 
   //  Should probably scan all mols first to find match with
   //  smallest number of matches...
   std::vector<RGroupMatch> potentialMatches;
+  constexpr size_t MAX_PERMUTATIONS = 100000;
 
   std::unique_ptr<ROMol> tMol;
   for (const auto &tmatche : tmatches) {
@@ -559,8 +562,10 @@ int RGroupDecomposition::add(const ROMol &inmol) {
       size_t sz = matche->size();
       N *= sz;
     }
-    // oops, exponential is a pain
-    if (N * potentialMatches.size() > 100000) {
+    // Highly symmetric cores can lead to a very large number of
+    // permutations to test. Fall back to Greedy when the number
+    // is too high.
+    if (N * potentialMatches.size() > MAX_PERMUTATIONS) {
       data->process(data->prunePermutations);
     }
   }
--- a/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
+++ b/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
@@ -552,11 +552,18 @@ int RGroupDecomposition::add(const ROMol &inmol) {
     return -2;
   }
 
-  // in case the value ends up being changed in a future version of the code:
-  if (data->prunePermutations) {
-    data->permutationProduct = 1;
+  if (data->params.matchingStrategy != GA) {
+    size_t N = 1;
+    for (auto matche = data->matches.begin() + data->previousMatchSize;
+         matche != data->matches.end(); ++matche) {
+      size_t sz = matche->size();
+      N *= sz;
+    }
+    // oops, exponential is a pain
+    if (N * potentialMatches.size() > 100000) {
+      data->process(data->prunePermutations);
+    }
   }
-
   data->matches.push_back(std::move(potentialMatches));
 
   if (!data->matches.empty()) {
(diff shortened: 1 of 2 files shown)


bp-kelley:
I will point out that the original code was absolutely tested, just undocumented.  The removal was not tested adequately during the review which I apologize for.----Brian KelleyOn Oct 4, 2024, at 18:01, Paolo Tosco ***@***.***> wrote:﻿Restores a RGD code snippet that was intentionally removed in #7186 because 1. it is unclear 2. it is not tested 3. It introduced an undocumented change in the documented GreedyChunks behavior based on an arbitrary threshold.



You can view, comment on, or merge this pull request online at:
  https://github.com/rdkit/rdkit/pull/7898

Commit Summary

  ae2ecb0  restore "oops, exponential is a pain" code snippet in the RGD algorithm


File Changes  (1 file)

  
    M
    Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
    (15)
  


Patch Links:

  https://github.com/rdkit/rdkit/pull/7898.patch
  https://github.com/rdkit/rdkit/pull/7898.diff

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
```diff
@@ -552,11 +552,18 @@ int RGroupDecomposition::add(const ROMol &inmol) {
...
+    size_t N = 1;
+    for (auto matche = data->matches.begin() + data->previousMatchSize;
+         matche != data->matches.end(); ++matche) {
+      size_t sz = matche->size();
+      N *= sz;
+    }
+    // oops, exponential is a pain
+    if (N * potentialMatches.size() > 100000) {
+      data->process(data->prunePermutations);
+    }
```
greglandrum, line 565: I don't think you want to remove the `if (data->prunePermutations)` block, right? You just intend to add this piece back?

bp-kelley:
I did some digging from my wayback machine.  This heuristic was included in the first merge with the rdkit sources, but the commit with the description was squashed.

It was added due to some initial tests on highly symmetric ring systems (i.e. benzene which is where the slow down was noted by Joann Prescott-Roy back at Novartis).  The squashed commit from my original repo read:

```
Highly symmetric cores can explode the number of perumtations to test.  Fall back to Greedy when the number is too high.
```

We don't need to keep the typo, but I'm happy to add the comment to the code as opposed to the current tongue in cheek one.  We could also expose this number as a parameter like "maxNumberOfPermutationsForGreedyChunks", but I believe there is a better solution as I mentioned before.

I do recall that the processes of going from exhaustive to greedy to greedy chunks consisted of us pouring over large decompositions and just stopping when it got "good enough" and balanced time with quality.

And as an aside, I use the GA version for the FreeWilson because it has much better run time characteristics for larger datasets and I like the idea of minimizing fingerprint variance in that case.

ptosco:
> I did some digging from my wayback machine. This heuristic was included in the first merge with the rdkit sources, but the commit with the description was squashed.

Thanks Brian!

> It was added due to some initial tests on highly symmetric ring systems (i.e. benzene which is where the slow down was noted by Joann Prescott-Roy back at Novartis). The squashed commit from my original repo read:
> 
> ```
> Highly symmetric cores can explode the number of perumtations to test.  Fall back to 
> Greedy when the number is too high.
> ```
> 
> We don't need to keep the typo, but I'm happy to add the comment to the code as opposed to the current tongue in cheek one. 

I’ll add the comment as soon as I get home. I had figured out correctly what the code was doing, but the comment will help. I will also document that in the `GreedyChunks` Python docstring.

> We could also expose this number as a parameter like "maxNumberOfPermutationsForGreedyChunks", but I believe there is a better solution as I mentioned before.

I had also thought of exposing the parameter, even just to be able to add a quick test covering that bit of code without having to waste minutes until the 10^5 threshold is reached.
I think until we have a better solution it would be good to expose that parameter.

> I do recall that the processes of going from exhaustive to greedy to greedy chunks consisted of us pouring over large decompositions and just stopping when it got "good enough" and balanced time with quality.
> 
> And as an aside, I use the GA version for the FreeWilson because it has much better run time characteristics for larger datasets and I like the idea of minimizing fingerprint variance in that case.

bp-kelley:
The N=10_000 parameter can be reached absurdly quickly which really surprised me back in the day, I can't remember exactly what my time threshold was for setting it I kind of remember it was around 5-10 seconds per chunk size 5 on pathological cases on 2017 hardware.

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
ptosco: @greglandrum That was actually intentional, because `data->permutationProduct` is unused. I also removed it from `RGroupDecompData` for the same reason. I have added comments about what was discussed. Also based on Brian’s advice, I have refrained from exposing `MAX_PERMUTATIONS` as a user-adjustable parameter.

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
```diff
@@ -388,11 +388,14 @@ int RGroupDecomposition::add(const ROMol &inmol) {
   // Exhaustive - saves all matches and optimizes later exhaustive
   //               May never finish due to combinatorial complexity
   // Greedy - matches to *FIRST* available match
-  // GreedyChunks - default - process every N chunks
+  // GreedyChunks - default - process every N chunks, unless
+  // MAX_PERMUTATIONS is exceeded, in which case it falls back to
+  // Greedy
```
bp-kelley, line 393: perhaps

```
// MAX_PERMUTATIONS is exceeded, in which case it falls back to
// Greedy for the current chunk
```

bp-kelley (approved):
One small comment, I can review this in the morning with my tests cases.

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
ptosco: Done.

bp-kelley:
Thanks @ptosco, I'll continue discussions at #7896 for more thoughts regarding algorithm changes.  There are still a lot of potential reductions in complexity that can be done IMO

PR #7902, opened 2024-10-08
Switch from awful macros to Better Enums for enum reflection

ptosco:
## **** WIP - need to fix Windows builds ****

@greglandrum I was just about to add the 3rd instance of awful macros for home-made `enum` reflection when I thought how much you would have suffered, so I thought I'd rather explore Better Enums after being disappointed by magic_enum.

Indeed the experience was totally positive this time, and the code looks much cleaner now.
Please don't be put off by the number of changed file: almost all of the changes are about `enum` member namespace scoping, which I believe is a good thing as it makes source code clearer. However, that introduces a backwards-incompatible change for developers using RDKit from C++, which is why I added a note to Release Notes and scheduled this for `2025_03_1`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,18 @@ if(NOT Catch2_FOUND)
   FetchContent_MakeAvailable(Catch2)
 endif()
 
+#include better-enums
+find_package(better_enums 0 QUIET)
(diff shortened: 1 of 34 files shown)


ptosco:
Closed as superseded by #7913.

PR #7924, opened 2024-10-16
... and more mem errors fixed

ricrogz:
This is round 3.

- The change in `Code/RDBoost/python_streambuf.h` that takes ownership of the pointer allows fixing the leaking `streambuf`s in `getTDTWriter()`, `getSmilesWriter()`, `getSDWriter()` and `getPDBWriter()`.

- The changes in `Code/GraphMol/Fingerprints/Wrap/FingerprintGeneratorWrapper.cpp` are due to the `std::vector<std::uint32_t> *`s leaking in `getFingerprint()`.  `customAtomInvariants` and `customBondInvariants` are never cleaned up, but `fromAtoms` and `ignoreAtoms` would also leak if something goes wrong in `getFingerprint()` and an exception is thrown.

- In `Code/GraphMol/Wrap/MolOps.cpp`, there's two fixes: `sb` and `istr` leaking if `parseQueryDefFile()` throws, and `atomCountsV` in `wrapLayeredFingerprint()` also leaking if we either hit the `throw_value_error()` or `LayeredFingerprintMol()` throws.

- Finally, in `Code/ML/Cluster/Murtagh/Clustering.cpp`, `PyArray_SimpleNewFromData()` does not take ownership of the `ia`, `ib` or `crit` arrays it is called on. So, to allow Python to manage the lifetime of these arrays, we use `PyCapsule_New()` to encapsulate them and associate a destructor function to them, and then attach the capsules to the array, so that the lifetime of the encapsulated C array is dependent on the PyArray.

### DIFF
--- a/Code/GraphMol/CIPLabeler/catch_tests.cpp
+++ b/Code/GraphMol/CIPLabeler/catch_tests.cpp
@@ -944,22 +944,20 @@ TEST_CASE("atropisomers", "[basic]") {
   }
 }
 
-std::string cipLabels(std::string molBlock) {
-  RDKit::ROMol *mol = nullptr;
+std::string cipLabels(const std::string &molBlock) {
+  std::unique_ptr<RDKit::ROMol> mol;
   std::string result = "";
   try {
     // try parsing the mol block with sanitize ob
     try {
-      mol = MolBlockToMol(molBlock, true, false);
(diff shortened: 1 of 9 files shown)


greglandrum (approved):
LGTM

PR #7939, opened 2024-10-19
Add to Cookbook "Include a Bond Index" recipe and image

bertiewooster:
Adds an example to the Cookbook for numbering bonds by their index on a molecular structure.
RDKitCB_40_im0

#### Any other comments?
To show the image in a Jupyter notebook, I ended the code block with

`display(SVG(svg))`

which required adding as the `testoutput` text `<IPython.core.display.SVG object>`, so that text appears on the page.

I put this in section "Drawing Molecules (Jupyter)" after "Include an Atom Index". It could alternatively go in section "Bonds and Bonding".


greglandrum:
Any particular reason that you didn't just use the `addBondIndices` draw option?

bertiewooster:
Just that I wasn't aware of `addBondIndices`. I'll play with that and check whether the function in this PR has additional value.

In either case, it may be worth adding a Cookbook article "Include a Bond Index," what do you think?

bertiewooster:
I switched to `addBondIndices` and followed the template of @iwatobipen's "Include an Atom Index" recipe. Screenshot for reference:
Includ a Bond Index recipe

greglandrum (approved):
LGTM

greglandrum:
Thanks @bertiewooster

bertiewooster:
Great, glad this recipe will be in the Cookbook to help others. Thanks for your guidance.

PR #7946, opened 2024-10-21
Update RDKit_Book.rst

johnmay:
Update the section on missing ligands following #6777. 

Documentation

#### Any other comments?

It should also be noted the existing diagram was not correct and actually depicted ``O[Mn@OH10](Cl)(C)(N)(F)`` and *not* ``O[Mn@OH1](Cl)(C)(N)(F)``, if it were ``@OH1`` then when they were at the end ``O[Mn@OH1](Cl)(C)(N)(F)*`` the Cl < C < N < F would go CCW looking from the oxygen.

image

Might be worth double checking the depiction code.

### DIFF
--- a/Docs/Book/RDKit_Book.rst
+++ b/Docs/Book/RDKit_Book.rst
@@ -970,8 +970,8 @@ This also works with multiple implicit Hs: ``C[Pt@SP1H2]Cl`` and ``C[Pt@SP1]([H]
 Missing ligands
 ^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


greglandrum:
not stale, I still need to make time for this one

johnmay:
Just bumping this as the doc doesn't currently match the toolkit behaviour.

Might also be useful to explain it with a diagram like this:

Screenshot 2025-03-07 at 16 08 12

johnmay:
Not stale - book is out of sync with code base.

johnmay:
I give up

greglandrum (approved):
LGTM
Sorry this took so absurdly long

PR #7952, opened 2024-10-22
Fix misnamed test

ricrogz:
While working on https://github.com/rdkit/rdkit/pull/7943 I noticed that `pyMolDraw2D` was mentioned twice in the `ctest` tests list. I thought the test had been become duplicated somehow, but it's just that the Python Rascal MCE test is misnamed.


greglandrum (approved):
LGTM

PR #7960, opened 2024-10-25
Expose propertyFlags to CFFI and MinimalLib

ptosco:
- added property support to CFFI library
- added `propertyFlags` JSON parser
- added support for setting `propertyFlags` to MinimalLib


greglandrum (approved):
LGTM

PR #7962, opened 2024-10-25
Make sure that loggers can be enabled, disabled, captured and tee'd from MinimalLib

ptosco:
I noticed that capturing/tee'ing loggers from MinimalLib was not as reliable as it should be, so I made it more robust and added a good amount of tests on both CFFI and MinimalLib sides.


bp-kelley:
@ptosco I'm going to take a bit with this one if you don't mind.  If you don't hear from me by Monday Nov 11 ping me again

Code/MinimalLib/cffi_test.c
```diff
@@ -2170,6 +2176,160 @@ void test_partial_sanitization() {
...
+}
+
+CapturedStreams *capture_streams(unsigned int buf_size) {
+  CapturedStreams *res;
+  res = (CapturedStreams *)malloc(sizeof(CapturedStreams));
+  if (!res) {
+    return NULL;
+  }
+  memset(res, 0, sizeof(CapturedStreams));
+  res->stdout_pipes[0] = -1;
```
bp-kelley, line 2236: suggest using
```
for(int i=0;i<PIPE_FD_SIZE; ++i) {
   res->stdout_pipes[i] = -1;
   res->stderr_pipes[i] = -1;
}
```

Code/MinimalLib/common.h
```diff
@@ -1090,26 +1094,159 @@ std::string parse_inchi_options(const char *details_json) {
 }
 #endif
 
-struct LogHandle {
+namespace {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+std::mutex &logger_mutex_get() {
+  // create on demand
```
bp-kelley, line 1100: If this is C++11 or above you only need the logger_mutex_get, the static creation of _mutex will be thread-safe initialized and not require std::call_once

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/n4594.pdf

From the section 6.7:

such a variable is initialized the first time control passes through its declaration; such a variable is considered initialized upon the completion of its initialization. [...] If control enters the declaration concurrently while the variable is being initialized, the concurrent execution shall wait for completion of the initialization.

Then there's a footnote:

The implementation must not introduce any deadlock around execution of the initializer.

Code/MinimalLib/cffi_test.c
ptosco: Done.

Code/MinimalLib/common.h
ptosco: We have the same code elsewhere in RDKit (e.g., in `Code/GraphMol/DistGeomHelpers/Embedder.cpp` and `Code/GraphMol/MolPickler.cpp`); I would rather leave it as is here and then submit a separate PR to remove `std::call_once` if we believe that is unnecessary.

bp-kelley: Back then it was necessary, now it is not.  For new code we should use the current spec IMO

bp-kelley: As it makes code a lot simpler

bp-kelley: But the code is correct in both cases so I won’t fight this one

ptosco: @bp-kelley Hi Brian, I have applied your suggestion and removed `std::call_once` from this PR.
Are you OK to merge this PR now or do you need more time?

ptosco: @bp-kelley Gentle poke: I have applied your suggestion and removed std::call_once from this PR.
Are you OK to merge now or do you need more time?

bp-kelley: I admit to wishing there was an easier way to do this, but it finally makes sense to me.

LGTM

bp-kelley (approved):

greglandrum (approved):
LGTM

PR #7964, opened 2024-10-26
Need to reserve space before assignment

bp-kelley:
@greglandrum  a newer Xcode compiler revealed this issue, we need to reserve space before assignment.  It's unclear to my why this didn't show up in the tests.


greglandrum (approved):
Yeah, it's very weird that the UBSAN builds didn't catch this.

ricrogz:
Yeah, ASAN didn't catch it either.

ricrogz:
> Yeah, it's very weird that the UBSAN builds didn't catch this.

Oh, interesting: it seems that UBSAN won't fail a test when it finds issues... it will just print a report an continue executing, unless you tell it to fail (which we don't do):

> For most checks ([checks](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#ubsan-checks)), the instrumented program prints a verbose error report and continues execution upon a failed check. You can use the following options to change the error reporting behavior

Seen here:  https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#id5

PR #7965, opened 2024-10-28
Upstream patches from rdkit-feedstock

skearnes:
Applies patches from https://github.com/conda-forge/rdkit-feedstock/tree/main/recipe so we can remove them from the conda build.


ptosco (approved):
LGTM. I have not built the PostgreSQL cartridge for macOS for a long time, but I’ll trust you that the `.dylib` is now needed (AFAICR when I originally wrote the CMake recipe, contrary to expectations, it worked with the `.so` extension).

Code/PgSQL/rdkit/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 if(APPLE)
   set (EXTENSION_PREFIX "")
-  set (EXTENSION_SUFFIX ".so")
-  set (EXTENSION_DEST_SUFFIX ".so")
+  set (EXTENSION_SUFFIX ".dylib")
+  set (EXTENSION_DEST_SUFFIX ".dylib")
```
ptosco, line 4: I suggest to revert this change as #7869 does the same, but maintaining backwards compatibility with PostgreSQL <16.
```suggestion
  set (EXTENSION_SUFFIX ".so")
  set (EXTENSION_DEST_SUFFIX ".so")
```

ptosco (changes requested):

PR #7985, opened 2024-11-04
Make sure that FreeSASA can be built as DLL on Windows

ptosco:
I have recently noticed that building support for FreeSASA fails when requesting a Windows MinGW DLL build.
This fixes the problem and also simplifies the FreeSASA build by only using a `.def` file rather than a mixture of `.def` file  and a source code patch.


bp-kelley (approved):

greglandrum (approved):
LGTM

PR #7996, opened 2024-11-10
Remove unused code from GraphMol/SmilesParse

whosayn:
These files don't appear to be used anywhere. `SmilesParse/primes.h` is particularly odd since it declares some duplicate symbols in `RDGeneral/utils.h`  but doesn't define them


greglandrum (approved):
LGTM

PR #8011, opened 2024-11-16
Avoid division by zero in DrawMol

ricrogz:
Another easy fix: if X is too small, then just look at Y to decide the orientation.


greglandrum (approved):
LGTM

PR #8012, opened 2024-11-16
remove no-op macros and dead code (pt 1)

whosayn:
This removes some no-op conditionals and some dead code from the repo.


greglandrum (approved):
LGTM

Thanks for finding these and cleaning them up @whosayn

PR #8013, opened 2024-11-16
Avoid division by zero in AlignPoints

ricrogz:
Only check if we are below the tolerance if `diagNorm` is big enough, else, continue the calculation.


greglandrum (approved):
LGTM

PR #8020, opened 2024-11-19
Fix #8019

### LINKED ISSUE #8019: DistanceConstraintContrib is not exposed to SWIG wrappers anymore
I noticed that in `2024_09_2` release and in current `master` `DistanceConstraintContrib`  is not exposed to SWIG wrappers anymore.
Upon investigation, I found out that this regression was introduced in #7715 by consolidating the constraint headers between UFF and MMFF, which was a very good thing to do.
However, this needs some small changes in the SWIG wrappers such that `DistanceConstraintContrib` can be exposed as it used to.

ptosco:
Small PR to revert the regression introduced in #7715 and described in #8019.


greglandrum (approved):
LGTM

PR #8005, opened 2024-11-14
Mol to smiles docs

DavidACosgrove:
#### Reference Issue
Addresses #8003

#### What does this implement/fix? Explain your changes.
Just a doc change.

PR #8042, opened 2024-11-25
Fix #4570 Segfault on Property Getters/Setters

### LINKED ISSUE #4570: Segmentation fault using None as property name in SetProp
```
>>> from rdkit import Chem
>>> m = Chem.MolFromSmiles("CCC")
>>> m.SetProp(None, "foo")
Segmentation fault: 11
```

This happens at least on OSX.

bp-kelley:
Fixes #4570 

It turns out that any Boost python function that has a signature

```
T func(const char * value)
```

will segfault when called with

```
func(None)
```

Switching to:
```
T func(const std::string &value)
```

is the easy solution here as boost will not permit the type cast.  We should probably have a once over for char * in exposed functions.

### DIFF
--- a/Code/GraphMol/Wrap/Atom.cpp
+++ b/Code/GraphMol/Wrap/Atom.cpp
@@ -104,7 +104,11 @@ std::string AtomGetSmarts(const Atom *atom, bool doKekule, bool allHsExplicit,
 }
 
 void SetAtomMonomerInfo(Atom *atom, const AtomMonomerInfo *info) {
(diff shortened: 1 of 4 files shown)


greglandrum (changes requested):
This PR somehow overwrites the changes added to the documentation in #8005 
please undo that

bp-kelley:
Good catch, that’s really odd.

bp-kelley:
@greglandrum should be fixed now

greglandrum (approved):
LGTM

PR #8046, opened 2024-11-26
Remove explicit default constructor from Compute2DCoordParameters

d-b-w:
The explicit default constructor actually blocks the other defaults. It also means this is no longer a "plain struct", and so designated initializers aren't available.

I wanna be able to do:

    Compute2DCoordParameters params{.useRingTemplates = true};

Or even:

    RDDepict::compute2DCoords(*mol, {.useRingTemplates = true};

#### Any other comments?
I know that we haven't moved the RDKit code to C++20 - but I'm using it from C++20.


greglandrum (approved):
LGTM
will merge after the CI builds finish

PR #8085, opened 2024-12-11
Avoid that map::at() throws an exception in case atomColourPalette includes neither 6 nor -1

ptosco:
This addresses an issue noticed by @sroughley as a unit test was failing in KNIME.
Currently the code uses `map::at(-1)` to get the default color for atom symbols under a highlight, and will therefore throw an exception if the `-1` key is not present.
This PR adds a fallback to black as in `getColourByAtomicNum`.
It also avoids to call `map::find` twice in the above function.


greglandrum (approved):
LGTM

PR #8086, opened 2024-12-12
Synthon search fp bug

DavidACosgrove:
#### Reference Issue
No Issue

When doing a fingerprint search in synthon space, the connector patterns for the query fragments and synthons need to match, but the former weren't being compared with the right thing.

#### Any other comments?
This should make the search slightly quicker as it does a better screen out of things that don't need a full check.


greglandrum (approved):
LGTM

PR #8097, opened 2024-12-14
Rascal synthon search

DavidACosgrove:
#### Reference Issue
No issue

Adds search of synthon space by Rascal similarity.

#### Any other comments?
It's rather slower than fingerprint similarity (an order of magnitude, maybe, on average) but gives usefully different results, I think.


DavidACosgrove:
I don't think the failures are real.

DavidACosgrove:
@greglandrum Can you poke this again, please?  I don't think the failures were real.

DavidACosgrove:
Yeah.  I've paused this one pending the multi-threading PR getting passed.  It will need it, I think.  I will flag it as WIP.

PR #8105, opened 2024-12-16
Use endian-aware read/write for length of string.

DavidACosgrove:
#### Reference Issue
No issue

The std::string specialisations of streamRead and streamWrite did a raw read/write of the string length, rather than use the endian-aware versions.

#### Any other comments?
It's been this way since at least 2017 so probably not a big deal.  There's no test because of that and also because it is tested indirectly all over the place.
There's been a re-formatting as well, which wasn't intentional.  I run Clang-format on saving.  The changes are round about line 268.

### DIFF
--- a/Code/RDGeneral/StreamOps.h
+++ b/Code/RDGeneral/StreamOps.h
@@ -41,7 +41,7 @@ enum EEndian {
 
 // this function swap the bytes of values given it's size as a template
 // parameter (could sizeof be used?).
-template <class T, unsigned int size>
+template<class T, unsigned int size>
 inline T SwapBytes(T value) {
   if (size < 2) {
     return value;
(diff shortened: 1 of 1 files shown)


bp-kelley:
Yikes.  It looks like we should try and have some pickled streams written in both big and little endian in the test set to ensure we can read them...

DavidACosgrove:
> The reformatting that you did here is not consistent with the rest of the code formatting. Are you using clang-format and the `.clang-format` file that's at the top of the code tree? If so, which version of clang-format are you using?

I did it on my MacBook rather than my Mac mini which are conceivably running different versions of things.  The formatting is all done via a CLion rule but I’ve never looked to see how it does what it does.  I will have another go.

DavidACosgrove:
I have re-run clang-format, and it has inserted a load of newlines between functions but most of the other formatting has been reverted.

DavidACosgrove:
CLion is supposed to use clang-format when it finds a .clang-format file in the root of the project.  I upgraded my laptop to a newer version of CLion than on the Mac mini and it didn't do so, preferring its own formatter.

greglandrum (approved):
LGTM

PR #8112, opened 2024-12-16
Minor MinimalLib cleanup

ptosco:
- `str_to_c()` should initialize len to 0 even if memory allocation fails
- `str_to_c()` should check the ptr returned by malloc for non-nullness before using it
- change `has_coords()` `mol_pkl` parameter to `const`
- use `assert.equal` in JS tests where possible


greglandrum (approved):
LGTM

PR #8128, opened 2024-12-20
Add functionality for rotating rotatable bonds (#6710)

AdamSulek:
This change introduces functionality to allow the rotation of a single bond between two groups, provided that each side of the bond contains a group (not a single atom like H, Cl, or F).

Key details:

Rotation criteria:
Requires `atom_1_index` and `atom_2_index` to specify the bond.
If the bond does not exist or either side is a single atom, the rotation is not applied.
Rotation degree:
A third parameter specifies the degree of rotation in the degree scale.
This ensures precise control over bond rotation while maintaining structural validity.

### DIFF
--- a/rdkit/Chem/AllChem.py
+++ b/rdkit/Chem/AllChem.py
@@ -476,3 +476,234 @@ def AssignBondOrdersFromTemplate(refmol, mol):
     else:
       raise ValueError("No matching found")
   return mol2
+
+
+def RotateGroup(mol, start_idx, end_idx, angle_deg):
+  """ Rotates a group of atoms around a bond in an RDKit molecule.
+
+    Arguments
+      - mol: RDKit molecule object.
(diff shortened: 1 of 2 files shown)


greglandrum:
@AdamSulek thanks for the contribution.
I think that much (if not all) of what you want to do is already possible using the function `rdMolTransforms.SetDihedralDeg()`

As an example, I start with a conformer of this molecule:
image
here's the conformer:
image

I start by rotating about the C6-C7 bond by setting the dihedral between atoms 4,6,7,8:
image

And then rotate about the C4-O5 bond:
image

If I've misunderstood something, please let me know.

Here's a gist with the notebook itself: https://gist.github.com/greglandrum/8c486a90f84f63d61e2d7b037fc58569

greglandrum:
Closing because this functionality is provided by `rdMolTransforms.SetDihedralDeg()`

PR #8148, opened 2025-01-01
relax two tolerances for ARM64 builds

tadhurst-cdd:
#### Reference Issue
when building on ARM64 systems, the DistGeomHelper test (catch_test.cpp) generates two errors where the returned value is just slightly outside the expected range.

The expected values were increased slightly.


greglandrum (approved):
LGTM

PR #8196, opened 2025-01-22
Removes minimum font size from reaction rendering

### LINKED ISSUE #8195: Reaction Rendering Looks odd at small scales
There appears to be a minimum font size in reaction rendering that creates odd looking drawings at small scales.

Example:

Image

The reactants and products should be rendered as they are normally.

bp-kelley:
Fixes #8195 

Before:

image

After:

image

It's still not perfect as the text can now get cutoff, but it does look a tad better.

### DIFF
--- a/Code/GraphMol/MolDraw2D/MolDraw2D.cpp
+++ b/Code/GraphMol/MolDraw2D/MolDraw2D.cpp
@@ -287,9 +287,6 @@ void MolDraw2D::drawReaction(
 
   // Copy the reaction because processing for drawing alters it.
   ChemicalReaction nrxn(rxn);
-  // I think a larger minimum font size than the default works better for this
-  int mfs = drawOptions().minFontSize > 12 ? drawOptions().minFontSize : 12;
-  text_drawer_->setMinFontSize(mfs);
   int plusWidth;
   getReactionDrawMols(nrxn, highlightByReactant, highlightColorsReactants,
                       confIds, reagents, products, agents, plusWidth);
--- a/Code/GraphMol/MolDraw2D/rxn_test1.cpp
+++ b/Code/GraphMol/MolDraw2D/rxn_test1.cpp
@@ -45,28 +45,28 @@ namespace {
 // if the generated SVG hashes to the value we're expecting, delete
 // the file.  That way, only the files that need inspection will be
 // left at the end of the run.
-static const bool DELETE_WITH_GOOD_HASH = true;
+static const bool DELETE_WITH_GOOD_HASH = false;
 #ifdef RDK_BUILD_FREETYPE_SUPPORT
 static const std::map<std::string, std::hash_result_t> SVG_HASHES = {
-    {"rxn_test1_1.svg", 2645695747U},   {"rxn_test1_2.svg", 2640646320U},
+    {"rxn_test1_1.svg", 4280849065U},   {"rxn_test1_2.svg", 2640646320U},
     {"rxn_test1_3.svg", 2781928588U},   {"rxn_test1_4.svg", 437863065U},
-    {"rxn_test1_5.svg", 2762972460U},   {"rxn_test1_6.svg", 2941728011U},
-    {"rxn_test1_7.svg", 1389845691U},   {"rxn_test2_1.svg", 3791815067U},
+    {"rxn_test1_5.svg", 3590497359U},   {"rxn_test1_6.svg", 907723784U},
+    {"rxn_test1_7.svg", 1389845691U},   {"rxn_test2_1.svg", 2288835686U},
     {"rxn_test2_2_1.svg", 2068242269U}, {"rxn_test2_2_2.svg", 269990608U},
-    {"rxn_test2_2_3.svg", 1544919807U}, {"rxn_test2_2_4.svg", 2015871327U},
-    {"rxn_test3_1.svg", 1359061920U},   {"rxn_test4_1.svg", 1727161552U},
-    {"rxn_test4_2.svg", 2032588691U},
+    {"rxn_test2_2_3.svg", 4111685373U}, {"rxn_test2_2_4.svg", 2015871327U},
+    {"rxn_test3_1.svg", 3211858221U},   {"rxn_test4_1.svg", 3018376157U},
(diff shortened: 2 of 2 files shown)


bp-kelley:
Here is an extreme case:

image

bp-kelley:
Test Mol block

```
data = """$RXN

      Mrv2019  070820240955

  2  1
$MOL

  Mrv2019 07082409552D          

  9  9  0  0  0  0            999 V2000
   -7.2098    0.4625    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -7.9243    0.0500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -7.9243   -0.7750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -7.2098   -1.1875    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -6.4954   -0.7750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -6.4954    0.0500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -5.7809    0.4625    0.0000 C   0  0  0  0  0  0  0  0  0  1  0  0
   -5.7809    1.2875    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
   -5.0664    0.0500    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  2  3  2  0  0  0  0
  3  4  1  0  0  0  0
  4  5  2  0  0  0  0
  5  6  1  0  0  0  0
  1  6  2  0  0  0  0
  6  7  1  0  0  0  0
  7  8  2  0  0  0  0
  7  9  1  0  0  0  0
M  END
$MOL

  Mrv2019 07082409552D          

  4  3  2  0  0  0            999 V2000
   -2.1041    0.7645    0.0000 H   0  0  0  0  0  0  0  0  0  7  0  0
   -2.5166    0.0500    0.0000 N   0  0  0  0  0  0  0  0  0  4  0  0
   -2.1041   -0.6644    0.0000 L   0  0  0  0  0  0  0  0  0  5  0  0
   -3.3416    0.0500    0.0000 L   0  0  0  0  0  0  0  0  0  6  0  0
  1  2  1  0  0  0  0
  2  3  1  0  0  0  0
  2  4  1  0  0  0  0
  3 F    2   1   6
  4 F    2   1   6
M  ALS   3  2 F H   C   
M  ALS   4  2 F H   C   
M  END
$MOL

  Mrv2019 07082409552D          

 11 11  2  0  0  0            999 V2000
    3.3036    0.0500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    4.0180    0.4625    0.0000 C   0  0  0  0  0  0  0  0  0  1  0  0
    4.7325    0.0500    0.0000 N   0  0  0  0  0  0  0  0  0  4  0  0
    4.0180    1.2875    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
    2.5891    0.4625    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.8746    0.0500    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.8746   -0.7750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.5891   -1.1875    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    3.3035   -0.7750    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    5.4470    0.4625    0.0000 L   0  0  0  0  0  0  0  0  0  5  0  0
    4.7325   -0.7750    0.0000 L   0  0  0  0  0  0  0  0  0  6  0  0
  1  2  1  0  0  0  0
  2  3  1  0  0  0  0
  2  4  2  0  0  0  0
  5  6  2  0  0  0  0
  6  7  1  0  0  0  0
  7  8  2  0  0  0  0
  8  9  1  0  0  0  0
  1  5  1  0  0  0  0
  1  9  2  0  0  0  0
  3 10  1  0  0  0  0
  3 11  1  0  0  0  0
 10 F    2   1   6
 11 F    2   1   6
M  ALS  10  2 F H   C   
M  ALS  11  2 F H   C   
M  END
"""
```

greglandrum (changes requested):
Please add a test for this

bp-kelley:
@greglandrum @DavidACosgrove it turns out the reaction draw tests didn't actually fail when the depictions were different.  I've updated the results with the new renderings (after visual inspection) and added png hashes when FREETYPE support is not turned on, they were missing before.

So the tests: 
   CAIRO = on/off  not that exciting, turns off png
   FREETYPE on/off - affects png and sag apparently

DavidACosgrove:
@bp-kelley These are old tests, we rely on a visual inspection.  There's a script $RDBASE/Code/GraphMol/MolDraw2D/side_by_side_images.py that reads the same image files from 2 different directories and puts them in an HTML table.  I generally use that at the end.
The hash codes don't work for PNGs - it only takes a single tiny difference between 2 floats and you get a different hash code and that's pretty much always the case on different systems.  They are more reliable with SVGs because, being ASCII, we control the number of decimal places in the coords and the inter-system differences are much fewer.  That's why they aren't put in for the PNG tests.  Mostly we inspect the innards of the SVGs to check whether a change has been successful and assume the PNG will look the same since they use the same underlying drawing code.

bp-kelley:
In either case it was computing the hash but not actually testing it, so anything would be better!----Brian KelleyOn Jan 25, 2025, at 02:50, David Cosgrove ***@***.***> wrote:﻿
@bp-kelley These are old tests, we rely on a visual inspection.  There's a script $RDBASE/Code/GraphMol/MolDraw2D/side_by_side_images.py that reads the same image files from 2 different directories and puts them in an HTML table.  I generally use that at the end.
The hash codes don't work for PNGs - it only takes a single tiny difference between 2 floats and you get a different hash code and that's pretty much always the case on different systems.  They are more reliable with SVGs because, being ASCII, we control the number of decimal places in the coords and the inter-system differences are much fewer.  That's why they aren't put in for the PNG tests.  Mostly we inspect the innards of the SVGs to check whether a change has been successful and assume the PNG will look the same since they use the same underlying drawing code.

greglandrum:
Superseded by #8211

PR #8211, opened 2025-01-25
Reaction rendered badly with small canvas size

### LINKED ISSUE #8195: Reaction Rendering Looks odd at small scales

DavidACosgrove:
#### Reference Issue
Fixes #8195

This removes the over-riding of the minFontSize for reactions.

#### Any other comments?
It supersedes @bp-kelley's #8196.  It has a test in our currently preferred style.


greglandrum (approved):
LGTM

PR #8257, opened 2025-02-10
Fixes #8256

### LINKED ISSUE #8256: PR #8192 breaks AdjustQueryProperties / aromatizeIfPossible for some molecules.
Before #8192, the following resulted in a mol with an aromatic ring:
```python
from rdkit import Chem
mol = Chem.MolFromSmarts(r'[#6]1=[#6]-[#6]=[#6]-[#6]=[#6]-1/[#7](=[#7]\[#8])-*')
params = Chem.AdjustQueryParameters.NoAdjustments()
params.aromatizeIfPossible = True
mol2 = Chem.AdjustQueryProperties(mol, params)

# No aromatic bonds
mol2.Debug()
# Atoms:
# 	0 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb:  query: AtomAtomicNum
# 	1 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb:  query: AtomAtomicNum
# 	2 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb:  query: AtomAtomicNum
# 	3 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb:  query: AtomAtomicNum
# 	4 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb:  query: AtomAtomicNum
# 	5 6 C chg: 0  deg: 3 exp: 3 imp: 0 hyb:  query: AtomAtomicNum
# 	6 7 N chg: 0  deg: 3 exp: 3 imp: 0 hyb:  query: AtomAtomicNum
# 	7 7 N chg: 0  deg: 2 exp: 2 imp: 0 hyb:  query: AtomAtomicNum
# 	8 8 O chg: 0  deg: 1 exp: 0 imp: 0 hyb:  query: AtomAtomicNum
# 	9 0 * chg: 0  deg: 1 exp: 1 imp: 0 hyb:   query: AtomNull
# Bonds:
# 	0 0->1 order: 2
# 	1 1->2 order: 1
# 	2 2->3 order: 2
# 	3 3->4 order: 1
# 	4 4->5 order: 2
# 	5 5->6 order: 1 dir: /
# 	6 6->7 order: 2 stereo: CIS ats: (5 8)
# 	7 7->8 order: 1 dir: \
# 	8 6->9 order: 1
# 	9 5->0 order: 1
```

ricrogz:
Fixes #8256.

The issue is in `countAtomElec()`, which now returns a different electron count for the ring C atom connected to the N atom, than before #8256 was merged. 

The cause of this is the "single or aromatic" query that is added since #8256. This query is not caught by `isBondOrderQuery()` (a "single or aromatic" query is definitely a bond order query, even if a complex one), and causes `QueryBond::getValenceContrib()` to return 0, as query bonds don't contribute to valence. This means that `countAtomElec()` discounts the bond from the degree, which prevents the aromatization of the ring.

This patch updates `isBondOrderQuery()` to first check the bond order: if the bond is "real" (i.e. it has a non-zero order), and if that's not the case, then check if it's a bond order query bond (either a simple or complex query). In either of those cases, the bond will be considered to contribute to the atom's degree. Otherwise, it won't be counted.

The reason for the `static_cast<Bond *>(bond)` is that we don't want to use the `QueryBond` overload, as it will duplicate what we do in `isBondOrderQuery()`, but it will return "0" in case of a complex bond query, which we don't want.

### DIFF
--- a/Code/GraphMol/Aromaticity.cpp
+++ b/Code/GraphMol/Aromaticity.cpp
@@ -492,7 +492,7 @@ void applyHuckelToFused(
       std::copy(curRs.begin(), curRs.end(),
                 std::inserter(aromRings, aromRings.begin()));
     }  // end check huckel rule
-  }  // end while(1)
+  }    // end while(1)
   narom += rdcast<int>(aromRings.size());
 }
 
@@ -696,10 +696,11 @@ ElectronDonorType getAtomDonorTypeArom(
 namespace RDKit {
 namespace MolOps {
 bool isBondOrderQuery(const Bond *bond) {
-  if (bond->getBondType() == Bond::BondType::UNSPECIFIED && bond->hasQuery()) {
-    auto label =
-        dynamic_cast<const QueryBond *>(bond)->getQuery()->getTypeLabel();
-    if (label == "BondOrder") {
+  if (bond->hasQuery()) {
+    auto q = dynamic_cast<const QueryBond *>(bond)->getQuery();
+    // complex bond type queries are also bond order queries!
+    if (q->getTypeLabel() == "BondOrder" ||
+        QueryOps::hasComplexBondTypeQuery(*q)) {
       return true;
     }
   }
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

PR #8270, opened 2025-02-13
Revert Dockerfile to JavaScript-based exceptions (with an option for switching to native WASM)

ptosco:
This addresses https://github.com/rdkit/rdkit/pull/8267#issuecomment-2656454454


greglandrum (approved):
LGTM

PR #8277, opened 2025-02-17
Tweak mol fragmenter

DavidACosgrove:
#### Reference Issue
No issue

MolFragmenter rejects dummyLabels that are longer than bondIndices with a PRECONDITION.  It is not an error if there are more dummyLabels than bondIndices, as the extra are just ignored.  When fragmenting a molecule multiple times with different numbers of bonds, as in the SynthonSpaceSearch, it's very convenient to have one set of dummyLabels that's as large as will ever be needed that can be used for all fragmentations.  This PR relaxes the PRECONDITION so this is allowed.

#### Any other comments?
Whilst in the area, I have changed the use of boost::uint64_t to std::uint64_t and updated a couple of bond traversals to use the modern idiom.


greglandrum (approved):
LGTM

PR #8286, opened 2025-02-20
Synthon search memory optimisation

DavidACosgrove:
#### Reference Issue
No Issue

The new versions of FreedomSpace and REAL took hideous amounts of RAM (~500GiB for REAL, for example).  The datastructures and algorithm have been significantly re-factored to make it much more efficient.  The key changes are recognising that a lot of the synthons are re-used so only need to be stored once, and changing the order of the search from looping over all fragment sets and then reactions to looping over all reactions and then all fragment sets.  That way the reactions, which use most of the memory, are not being changed all the time.
I've also spent a lot of time with the profiler removing other inefficiencies.
There have been a lot of changes to the files, but the fundamental algorithm remains unchanged.

#### Any other comments?
On my Mac mini, REAL (72 billion compounds) now sits in ~10GiB of RAM and a simple substructure search takes ~1.5s, and FreedomSpace (142 billions compounds) uses ~3GiB of RAM and the same substructure search takes ~0.25s.  Fingerprint searches are slower, and can take up to 20s.  Reading the databases take ~20s and ~5s respectively. This is after the database has been converted to the binary format, which does a lot of processing of the synthons to make the search faster.  The conversion takes ~2 hours for REAL and 15 minutes for FreedomSpace but only needs to be done once per database per fingerprint type.  All still running on a single thread.  Multi-threading comes next.


Code/GraphMol/SynthonSpaceSearch/Synthon.h
```diff
@@ -22,35 +22,35 @@ class RWMol;
 
 namespace SynthonSpaceSearch {
 
+// These are the numbers of bits used in the internal fingerprints.
+// The user is not restricted to these numbers for the search.
+inline static constexpr unsigned int PATT_FP_NUM_BITS = 1024;
```
greglandrum, line 27: I think:
```suggestion
inline constexpr unsigned int PATT_FP_NUM_BITS = 1024;
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -99,54 +98,60 @@ struct RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpaceSearchParams {
...
    * @return std::int64_t
    */
-  std::int64_t getNumProducts() const;
+  std::uint64_t getNumProducts() const;
+  /*!
+   * Get the total number of products, formatted nicely in a string.
+   *
+   * @return std::string
+   */
+  std::string getFormattedNumProducts() const;
```
greglandrum, line 140: seems like a weird one to have in an API
I'd be happier to have this as a function at global scope

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -202,35 +206,44 @@ class RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpace {
...
   /*!
-  @param outFilename: name of the file to write
+   * Writes the enumerated library to file in SMILES format
+   * (1 compound per line, SMILES name)
+   *
+   * @param outFilename: name of the file to write
    */
   void writeEnumeratedFile(const std::string &outFilename) const;
 
+  // Below here are functions that have to be in the
```
greglandrum, line 242: Q: Would it make sense to make them protected and add the other classes that need them as friends?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceHitSet.h
```diff
@@ -0,0 +1,91 @@
...
+  std::string reactionId;
+  // The fragments that this hitset is derived from, useful for debugging.
+  std::vector<const ROMol *> frags;
+  std::vector<std::vector<std::pair<std::string, const ROMol *>>> synthonsToUse;
+  size_t numHits{0};
+};
+
+// This sub-class holds results from a SynthonSpaceFingerprintSearch.
+// It needs the synthon fingerprints as well.
+struct RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpaceFPHitSet
```
greglandrum, line 65: I think you need to/should disable the copy ctor and copy assignment operators here and then make it explicit that the dtor, move, and move assignment operators are default

greglandrum (changes requested):
This was a quick pass through the header files.
I still want to at least briefly look at the code itself too, but I will do that tomorrow

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
DavidACosgrove: I find it very useful for this specific purpose.  I guess a function for formatting large ints could be made more widely available, and then it could be used for things like SearchResults.getMaxNumProducts() as well.  Would a function still in the SynthonSpaceSearch namespace be ok, or can you see it being more generally useful?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.h
```diff
@@ -111,6 +119,34 @@ struct RDKIT_SYNTHONSPACESEARCH_EXPORT Stepper {
...
 
+// Return a molecule containing the portions of the molecule starting at
+// each dummy atom and going out up to 3 bonds.  There may be more than
+// 1 fragment if there are dummy atoms more than 3 bonds apart, and there
+// may be fragments with more than 1 dummy atom if their dummy atoms fall
+// within 3 bonds of each other.  E.g. the molecule [1*]CN(C[2*])Cc1ccccc1
+// will give [1*]CN(C)C[1*].  The 2 dummy atoms are 4 bonds apart, but the
+// fragments overlap.  All dummy atoms given isotope 1 whatever they had
+// before.
+RDKIT_SYNTHONSPACESEARCH_EXPORT std::unique_ptr<ROMol> getConnRegion(
```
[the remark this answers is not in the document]
greglandrum: Ah, my mistake. Sorry, I was reading this too quickly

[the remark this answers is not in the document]
DavidACosgrove: I have changed it to buildConnRegion() which is a more sensible name in any case, I think, and less confusing.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
DavidACosgrove: It looks differently inelegant, and probably a sign of bad design but yes, a bit tidier.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
DavidACosgrove: @greglandrum I've moved this but not pushed it yet pending any other review comments.

DavidACosgrove:
Closing this as it is now superseded.

PR #8302, opened 2025-02-26
Fix potential build error under MSVC

ricrogz:
This is an edge case that I've hit with MSVC VC v14.29.30133 compiler.

It seems there's an ambiguity of this `friend operator<<` declaration: if the compiler has seen a `::Atom` `class` or `struct` declaration (even a forward declaration) before including this header, it seems that the VC compiler will prefer that declaration to the `RDKit::Atom` one where this declaration is made. It will trigger a build error like this one:
```
my_code.cpp
..\rdkit-Release_2024_09_5\include\rdkit\GraphMol/Atom.h(79): error C2063: 'operator <<': not a function
```

This patch seems to resolve the ambiguity and fix the issue.


greglandrum (approved):
sometimes C++ is really weird...
LGTM

PR #8347, opened 2025-03-13
Ignore invalid chirality labels when reading MAE inputs

### LINKED ISSUE #8346: Ignore invalid chirality labels when reading MAE inputs
**Describe the bug**
For a host of reasons, inputs chirality labels can be invalid as we try to read an input MAE file or buffer. This can be annoying for situations when we have other means of calculating the stereochemistry information

**Expected behavior**
we should try to ignore bad labels and let the user deal with the "bad" stereochemistry.

whosayn:
#### Reference Issue

Fixes #8346

This updates the MAE reader to ignore invalid chirality labels


d-b-w:
Because of the way that stereochemistry/chirality is stored on .mae files (Molecule level property with explicit neighbor atom indices), there are certain workflows in which it's not hard to generate an invalid stereo label. However, those workflows usually _also add meaningful stereo labels_, so it is important not to ignore stereo information that is valid.

Code/GraphMol/FileParsers/testMolSupplier.cpp
```diff
@@ -2835,6 +2830,69 @@ void testGitHub3517() {
...
+    :::
+    1 1 1 2
+    2 2 1 3
+    3 2 1 4
+    4 2 1 5
+    :::
+  }
+}
+)DATA";
+  for (const auto &invalid_chirality_label : {
```
d-b-w, line 2872: nice descriptions of the test cases.

Code/GraphMol/FileParsers/testMolSupplier.cpp
```diff
@@ -2835,6 +2830,69 @@ void testGitHub3517() {
...
+           "12_R_1_3_4_5",     // missing chiral atom
+           "12_ANR_1_3_4_15",  // missing substituent atom
```
d-b-w, line 2874: This looks like it should use catch2 generators:
https://github.com/catchorg/Catch2/blob/devel/docs/generators.md

One disadvantage of loop based tests in C++ is that you can't usually see which test conditions caused a test failure in the test output. Using the test framework's native test generator tool helps with this (as I know you know from your work with boost::test :) ).

d-b-w (changes requested):
LGTM after updating to use a catch2 generator

Code/GraphMol/FileParsers/testMolSupplier.cpp
whosayn: thanks

Code/GraphMol/FileParsers/testMolSupplier.cpp
whosayn: I was not aware of this `catch2` functionality. thanks!

d-b-w (approved):
Looks great!

greglandrum (approved):
LGTM

PR #8371, opened 2025-03-21
Issue8340

tadhurst-cdd:
#### Reference Issue
Equivalent sidechains does not remove atropisomerism #8340

Fixes the molecule in issue 8340 when using legacy stereo (non-legacy stereo was right anyway)

#### Any other comments?
this new code BREAKS other atropisomers - it removes the atropisomer indications for true atropisomers.
using legacy stereo is flawed, and there will be errors.


tadhurst-cdd:
As per Greg L, we will NOT make the changed in this PR.  The reasons are documented in Issue https://github.com/rdkit/rdkit/issues/8340

tadhurst-cdd:
see https://github.com/rdkit/rdkit/issues/8340

this PR fixes one issue but creates others.   We will promote the new stereo instead

PR #8382, opened 2025-03-24
Open sources the Revvity ChemDraw document format including original source code

bp-kelley:
Thanks to Revvity for open sourcing the ChemDraw format and providing the original sources.

This releases the ChemDraw document format in an open source BSD license.  It includes the full internal specification of the document format and both the CDX and CDXML readers as used in ChemDraw.

Portions have been reimplemented where necessary.  Please see the README.md for details about the changes.  All changes have been documented in the code.

This package provides four new libraries to the rdkit.

1. expat (this is the c++ varian of the expat library as hosted by sourceforge.  Note that this is an old version that was used to develop CDXML.
2. ChemDraw this is the chemdraw document format.  CDXDocuments can be read independently of parsing molecules and gives access to all of the internal documents, graphics, stoichiometry tables and so on.
3. RDChemDrawLib this is the main library that reads and writes RDKit Molecules to and from CDX/CDXML
4. RDChemDrawReactionLib this does the same thing for reactions.

Note that the reactions have been split from the molecules only so we can replace the existing CDXML reader with the one provided here.

Revvity was kind enough to provide a test suite consisting of 6,000 CDXML files that were used to test the functionality of the ChemDraw parser.  Unfortunately, these files have not been cleared for release, but will be maintained by Glysade Inc for future testing with the hope that they will one day see the light of day.


bp-kelley:
@greglandrum enjoy!  There's a chance the windows tests will fail as the last commits weren't fully tested.  If so, I'll fix it in the AM.

There is also a documentation dump I won't be able to add until tuesday.

bp-kelley:
Ok, I think I merged off a bad branch, going to merge off the current master...

PR #8404, opened 2025-04-02
Bump MaeParser version to 1.3.2

ptosco:
Version 1.3.2 of `maeparser` can optionally be built without the `boost::iostreams` dependency.


greglandrum (approved):
LGTM

PR #8407, opened 2025-04-03
Do not free sslib until patternFpArray is still needed by the test

ptosco:
I noticed that `tests.js` occasionally fails on the
`assert.equal(patternFpArray.length, sslib.size())`
assertion, and I believe this might be causing by `sslib.delete()` being called too early, i.e. when the `patternFpArray` is still needed. This may cause objects in the array to be invalidated ahead of time.
If that is indeed the root cause, this should fix the problem.


greglandrum (approved):
LGTM

This would be easier if it were straightforward to tie the lifetime of one object to another in JS. I guess you can't do this?

PR #8448, opened 2025-04-16
add function to overwrite __setattr__ to only accept valid arguments

nmaeder:
Allows to overwrite __setattr__ for python classes from within c++.
This to raise errors in case a class attribute is misspelled.


greglandrum (approved):
LGTM

I think this is useful functionality for parameter classes like `EmbedParameters`, and think we should expand usage to our other parameter classes, but don't think it should be applied to our other classes like Mol, Atom, ExplicitBitVect, etc. where being able to take advantage of Python's dynamic nature is useful.

FWIW: there is a performance penalty associated with this. Here's master:
```
In [1]: from rdkit.Chem import rdDistGeom

In [2]: ps = rdDistGeom.EmbedParameters()

In [3]: %timeit ps.basinThresh=5.0
173 ns ± 2.25 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
with this change applied the timing info is:
```
1.08 µs ± 3.18 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
``` 
I don't think setting up a parameter object is ever a significant fraction of runtime, but it's good to be aware of this.

PR #8497, opened 2025-05-02
check if builds succeed with ccache

whosayn:


ricrogz:
Does it make sense to use ccache on the CI? The CI uses fresh machines on each run, so there's no saved cache to use.

whosayn:
> 

@ricrogz there are remote storage options for the cache (https://ccache.dev/manual/latest.html#_remote_storage_backends), which will be useful for the CIs. Was checking if the cmake version, platforms support etc allow ccache-enabled compilation so I haven't implemented that yet.

ricrogz:
> @ricrogz there are remote storage options for the cache (https://ccache.dev/manual/latest.html#_remote_storage_backends), which will be useful for the CIs. Was checking if the cmake version, platforms support etc allow ccache-enabled compilation so I haven't implemented that yet.

Which remote storage are you planning to use?

whosayn:
> > @ricrogz there are remote storage options for the cache (https://ccache.dev/manual/latest.html#_remote_storage_backends), which will be useful for the CIs. Was checking if the cmake version, platforms support etc allow ccache-enabled compilation so I haven't implemented that yet.
> 
> Which remote storage are you planning to use?

@ricrogz the plan is to use redis. current testing on my mac resulted in clean build times taking about 5mins with `make install -j10`.

@greglandrum for the purposes of testing this remote caching logic, would it be possible to provision test secrets for this repo? I have a test database that I'm working with, so I could provide you with the url and password to be set as secrets

ricrogz:
> > > @ricrogz there are remote storage options for the cache (https://ccache.dev/manual/latest.html#_remote_storage_backends), which will be useful for the CIs. Was checking if the cmake version, platforms support etc allow ccache-enabled compilation so I haven't implemented that yet.
> > 
> > 
> > Which remote storage are you planning to use?
> 
> @ricrogz the plan is to use redis. current testing on my mac resulted in clean build times taking about 5mins with `make install -j10`.

Oh, I meant where you plan to host the cache

PR #8504, opened 2025-05-05
Fix arrowhead8500

DavidACosgrove:
#### Reference Issue
Discussion #8500, 2nd bit, the solid arrow head positions.

Allows for the mitre on the arrowhead correctly rather than the ridiculous guess it was doing before.

@bertiewooster's test (slightly amended) now looks like this:

image

Both open and solid arrows are drawn on the same image to show that they are now the same.

At the same time I have fixed the Cairo/PNG polygon drawer so it uses mitred line joins, and also the SVG open polygon drawing.  So the reaction arrows look better, too.

The difference in the code for drawing the arrows is below.  Note that for some reason this uses the American spelling of colour, even though most of the code uses the British spelling.  I am still embarrassed about creating this confusion for post-colonial reasons.  Sorry.

```
# Draw arrows
p1 = Geometry.Point2D(0, 0)
for i in range(nRows + 1):
    for j in range(nCols + 1):
        p2 = Geometry.Point2D((j) * panelx, (i) * panely)
        d2d.DrawArrow(
            p1,
            p2,
            rawCoords=True,
            asPolygon=True,
            color=(0,1,0)
        )
for i in range(nRows + 1):
    for j in range(nCols + 1):
        p2 = Geometry.Point2D((j) * panelx, (i) * panely)
        d2d.DrawArrow(
            p1,
            p2,
            rawCoords=True,
            asPolygon=False,
            color=(0,0,1)
        )
```
#### Any other comments?
The hash codes for the files seem to be off again.  Maybe I'll summon the enthusiasm at some point to fix those in a separate PR.  I have compared all the drawings against the current master branch and they look ok.

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawShape.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawShape.cpp
@@ -99,13 +99,21 @@ DrawShapeArrow::DrawShapeArrow(const std::vector<Point2D> &points,
       frac_(frac),
       angle_(angle) {
   PRECONDITION(points_.size() == 2, "arrow bad points size");
-
-  // the two ends of the arrowhead are used for collision detection so we
-  // may as well store them.
+  // The two ends of the arrowhead are used for collision detection so we
+  // may as well store them.  They won't be used for drawing.  lineWidth_ is
+  // in pixels rather than drawing coords, so keep another set of coords that
+  // are the original ends and arrowhead ends, with a lineWidth_ of 0, used
+  // for findExtremes.
+  origPts_[0] = points_[0];
+  origPts_[1] = points_[1];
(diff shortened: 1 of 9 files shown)


DavidACosgrove:
> That was quick work, nice! The output image looks good to me.
> 
It didn't feel so quick - some "interesting" stuff came up.

> It might be nice for you to share the code you used to create that image, partly as an example of how to make the arrowhead color different than the line color.

That's not what it's doing.  It's drawing the solid arrow in green, then the open arrow in blue on top.

DavidACosgrove:
@greglandrum The failure is in pyDistGeom.py, which seems unlikely to be caused by this PR.

greglandrum (approved):
LGTM

PR #8540, opened 2025-05-17
Fix irritating warning when DrawMolMCHLasso compiled.

DavidACosgrove:
#### Reference Issue
No issue.

On my Ubuntu machine using gcc 14.2.0 (from the conda env) I was getting an irritating warning from the std::vector include file when compiling DrawMolMCHLasso.cpp.  This uses a different way of copying the vector that doesn't cause the warning.

#### Any other comments?
I've checked that the lasso images don't change.


ptosco (approved):
I suggested an alternative way to do the insertion which also suppresses the warning and should be more efficient.

greglandrum (approved):
LGTM

PR #8541, opened 2025-05-17
allow limiting the  number of stereo groups to process

tadhurst-cdd:
#### Reference Issue
If the number of OR groups or the number of AND groups is too large, the rigorous canonicalization of stereo groups effectively never returns.   

This allows the number of stereo ORs or stereo ANDs to be limited.   If the number is above that cutoff, the original molecule is not changed.    The default number allowed is 12, which makes the processing take something on the order of 3 seconds.


greglandrum (changes requested):
Please add the new argument to the python wrappers too

greglandrum (approved):
LGTM

PR #8555, opened 2025-05-23
Explicitly wrap raw pointer into emscripten::val object with allow_raw_pointers() policy

ptosco:
This change in `emscripten`

https://github.com/emscripten-core/emscripten/pull/24175

has broken our MinimalLib build, and this PR restores it by wrapping the raw pointer into an `emscripten::val` constructed with `allow_raw_pointers()` policy.


greglandrum (approved):
LGTM

Issue #4546, opened 2021-09-22
AllChem.CanonicalizeMol adds 3D coords


paconius:
**Describe the bug**
Calling CanonicalizeMol after Compute2DCoords adds unexpected 3D coords to some molecules

**To Reproduce**
```
import rdkit
from rdkit import Chem
from rdkit.Chem import AllChem

print(rdkit.__version__)
smiles = "CCOC1=CC(=O)CC1"
mol = Chem.MolFromSmiles(smiles)
AllChem.Compute2DCoords(mol)
print(Chem.MolToMolBlock(mol))
AllChem.CanonicalizeMol(mol)
print(Chem.MolToMolBlock(mol))
```

**Expected behavior**
For mols with only 2D coords, the computed centroid is expected to also be 2D (i.e. z-coord = 0.0). I therefore expect CanonicalizeMol **_not_** to transform the mol from 2D to 3D. Both printed molblocks are therefore expected to have only 2D coordinates (i.e. z-coords should all be 0.0).
This bug only happens for some molecules (I can't yet find a pattern). If you swap the smiles string to change the ethoxy group to a methoxy ("COC1=CC(=O)CC1"), it works as expected.

**Screenshots**

```
    RDKit          2D

  9  9  0  0  0  0  0  0  0  0999 V2000
   -4.1860   -0.1804    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -2.7677    0.3077    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
   -1.6357   -0.6765    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
   -0.2174   -0.1884    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.0112   -1.0489    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.2093   -0.1464    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    3.6437   -0.5851    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
    1.7212    1.2720    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.2214    1.2460    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0
  2  3  1  0
  3  4  1  0
  4  5  2  0
  5  6  1  0
  6  7  2  0
  6  8  1  0
  8  9  1  0
  9  4  1  0
[...]

PR #8573, opened 2025-06-06
Fix for 3dcoord in 2d mol file error

tadhurst-cdd:
#### Reference Issue
When a 2D mol file contains 3d coordinates, and has wedge bond information, the parse would convert the mol to a 3D molecule and ignore the wedges, If the 3d coords were all the same for the atoms related to a chiral center, the chirality of the center was not perceived.  Such a mol should not be converted to a 3d mol.

If the mol is 2D as specified in the header, but also has 3d coordinates and also has wedge bond informatoin, it is no longer converted to 3D.  This allows the wedge bond information to be used to determine the chirality.

#### Any other comments?
There already was treatment for similar things:  If mol was marked as 3d, but all the z-coords were zero AND there was wedge bond information present, the molecule was changed to 2D.   
Also, previously, is the mol was marked as 2D but had non-zero z coordinates, it was changed to 3D.   This second one is not NOT done (not converted to 3D) if there is wedge bond information.


greglandrum:
This one looks good to me. @ricrogz: you've done some work on the 2D vs 3D in mol blocks too, could you please also take a look?

tadhurst-cdd:
The original issue came from a CDD customer.  The 3D coords are clearly bogus.  I guess my general preference would be to treat mol files that say they are 2D as 2D, and treat those that say they are 3D as 3D.

The current code, however, tries to "fix" the 2D that have coords that are not zero to 3D.  

For the case  where the mol is marked 2D and has both wedging info AND 3D coordinates, and the chirality derived from the wedging and 3D coords does not agree,  I would prefer NOT to change it to 3D.

Why would we presume that the 3D coords that not supposed to be there are correct, and that the wedging is incorrect?

ricrogz:
> The original issue came from a CDD customer. The 3D coords are clearly bogus. I guess my general preference would be to treat mol files that say they are 2D as 2D, and treat those that say they are 3D as 3D.
> 
> The current code, however, tries to "fix" the 2D that have coords that are not zero to 3D.
> 
> For the case where the mol is marked 2D and has both wedging info AND 3D coordinates, and the chirality derived from the wedging and 3D coords does not agree, I would prefer NOT to change it to 3D.
> 
> Why would we presume that the 3D coords that not supposed to be there are correct, and that the wedging is incorrect?

That's why I have doubts... as there's conflicting information, nothing is certain in these cases. Except that customers always have crappy data :smile: 

My reason to prefer the 3D coordinates over the wedges is just that the wedges are an artifact to encode a feature of the 3D world (how neighbors are distributed around a central atom) into the 2D world. Given that this is calculated and cached, there's just more potential sources of errors (a wrong calculation, or just outdated information), I'd prefer the 3D coordinates as long as there isn't proof that it's wrong, because those coordinates will stick on the mol.

tadhurst-cdd:
Most wedging info probably does NOT come from 3D coords.  Users enter it.
In fact, they enter what they know about the stereo/chiral state of the
molecule.  A 3D structure will alway indicate the coords of a chiral center
even when it is not known.  I trust wedging in a 2D structure when it is
there.

Given that, we agree that customers have crappy data.

paconius:
> ...Given that, we agree that customers have crappy data.

I think there is room for a bit of humble introspection here, as until very recently, RDKit was a key source for garbage 3D coordinates being added to otherwise correct 2D molecules (see #4546).

bp-kelley:
@paconius I don't think anyone claimed that the RDKit is foolproof or perfect, crappy data has a lot of sources and I've always assumed that I'm one of them.  Registration systems are a bit different though in the assumptions that are made.  For instance, in the example you gave, we would reject it as you did, being  a 2D compound that has 3D coordinates.

It would be interesting to route rdkit test output through the sanity checker, this seems like a rather good idea to find issues like these.

The problem I have seen with 3D coordinates indicating atropisomers is that a conformer generator has to generate something.  It may look like an atropisomer, but actually might not be.  We flagged these for potential follow up and to indicate there may be a mixture for interesting compounds with iffy biological data.

If a chemist has taken the time to wedge an atropisomer, we assume that it has been backed up in a separation or has had another analysis been done.  Of course, all of these can be incorrect, so be prepared to reregister and fix issues.

greglandrum:
[This comment was updated and corrected on 21.08.2025]

Here's a summary of the current status (I tested using v2025.03.5)  of what the RDKit does:

|  flag |  coords  | wedging | result | notes |
|-------|----------|---------|--------|-------|
| 2D | 2D | no  | 2D | no chirality |
| 3D | 2D | no  | 3D | no chirality |
| 3D | 3D | no  | 3D | chirality from coords |
| 2D | 3D | no  | 3D | chirality from coords |
| 2D | 2D | yes | 2D | chirality from wedging |
| 3D | 2D | yes | 2D | chirality from wedging |
| 3D | 3D | yes | 3D | chirality from coords |
| 2D | 3D | yes | 3D | chirality from coords |

Here's what the Molfile specification from Biovia says:

> The “dimensional code” is maintained explicitly. Thus “3D” really means 3D, although “2D” will be interpreted as 3D if any non-zero Z-coordinates are found

We are consistent with this except for the case where the 3D flag is set for 2D coordinates and a wedge is present, in which case we ignore the 3D flag, mark the conformer as 2D, and set the stereochemistry based on the wedging

In cases where no 2D/3D flag is provided, the default value of the flag is 2D. 

In a 2D structure, wedging is interpreted as a signal to indicate that stereochemistry is present and to indicate what the stereochemistry is.

When 3D coordinates are provided, stereochemistry is perceived from the coordinates themselves. If wedging is also present it will be ignored; the 3D signal is "stronger" than the wedging signal. The exception to this rule is atropisomeric bonds, where the wedging simply indicates that atropisomerism should be perceived, but the direction of the wedge is ignored.

This PR would add a new behavior: if 2D is specified, 3D coordinates are provided, and there's a wedged bond then the coordinates will be interpreted as 2D and chirality will be perceived from the wedging, It relies upon the wedging being correct when all z coordinates are set to zero. The difference between this and the case where there's a 3D flag with 2D coordinates and a wedge is that we're ignoring even more information provided by the user: in addition to ignoring the 2D/3D flag, we're also discarding all of the z coordinates that they provide.

I recognize the problems with weird/conflicting user input that you're trying to solve here, but I think this is a step too far and is likely to lead to more problems than it resolves. I think specific business rules like this are better implemented as part of the checking/validation code in your software.

bp-kelley:
Perhaps a warning when the 3D stereo doesn't match the 2D would be appropriate.  Does one exist?

greglandrum:
> Perhaps a warning when the 3D stereo doesn't match the 2D would be appropriate. Does one exist?

if you mean a warning when there's a non-zero Z coordinate in a mol file marked as 2D, then "yes"

bp-kelley:
No, I mean if the 2d wedging doesn’t match the 3d

tadhurst-cdd:
The fourth item in your table of current behavior is not correct, I think.
It says that if the 3D flag is set, and there are NO non-zero coords, the mol is treated as 3D.  This is only true if there is NOT wedging information.   If the wedging is present, the mol is changed to 2D and the wedging is used.  See MolFileParser.cpp line 2497. 
` if (!nonzeroZ && marked3d == 1) {
`
This current behavior is NOT consistent with the Biovia Mol file spec, which says that if it says it is 3D, then it IS 3D.  The biovia spec makes no mention of the presence of wedging.

So the current code is already second-guessing the Biovia spec.

The change I suggested is consistent with this second-guessing of the Biovia spec.

Having said that, I don't really care.    Working to correct inconsistent input  is "iffy", but we are already doing it for case 3 of your table.  To be consistent with the Biovia spec, we should probably remove the code that changes 3D to 2D if there are no non-zero z coords and there is wedging.

Again, I don't really care.   But I think there is another potential serious error in that routine (calculate3dFlag).  

If none of the "fixing" logic is triggered, the last line returns "nonzeroZ".   This would screw up a legitimate 3D mol that happens to have all Z coords as zero (planar, in the XY plane - e.g. benzene).  I think it should return "marked3d" to be correct and consistent with the Biovia def.

greglandrum:
> No, I mean if the 2d wedging doesn’t match the 3d

Nope, no warning for that.

greglandrum:
> If none of the "fixing" logic is triggered, the last line returns "nonzeroZ". This would screw up a legitimate 3D mol that happens to have all Z coords as zero (planar, in the XY plane - e.g. benzene). I think it should return "marked3d" to be correct and consistent with the Biovia def.

The case of all z coordinates are zero but the molecule is marked as 3D is handled correctly by the return at line 2507, so I don't think this is a problem:
```
In [9]: mb = '''
   ...:      RDKit          3D
   ...: 
   ...: M  V30 BEGIN CTAB
   ...: M  V30 COUNTS 4 3 0 0 0
   ...: M  V30 BEGIN ATOM
   ...: M  V30 1 C 0.000000 0.000000 0.000000 0
   ...: M  V30 2 C 1.299038 0.750000 0.000000 0
   ...: M  V30 3 F 2.598076 -0.000000 0.000000 0
   ...: M  V30 4 Cl 1.299038 2.250000 0.000000 0
   ...: M  V30 END ATOM
   ...: M  V30 BEGIN BOND
   ...: M  V30 1 1 2 1
   ...: M  V30 2 1 2 3
   ...: M  V30 3 1 2 4
   ...: M  V30 END BOND
   ...: M  V30 END CTAB
   ...: M  END
   ...: '''

In [10]: m = Chem.MolFromMolBlock(mb)

In [11]: m.GetConformer().Is3D()
Out[11]: True
```

greglandrum:
> The fourth item in your table of current behavior is not correct, I think. It says that if the 3D flag is set, and there are NO non-zero coords, the mol is treated as 3D. This is only true if there is NOT wedging information. If the wedging is present, the mol is changed to 2D and the wedging is used. See MolFileParser.cpp line 2497. `if (!nonzeroZ && marked3d == 1) {` This current behavior is NOT consistent with the Biovia Mol file spec, which says that if it says it is 3D, then it IS 3D. The biovia spec makes no mention of the presence of wedging.
> 
> So the current code is already second-guessing the Biovia spec.
> 
Right you are. Sorry I missed that.

tadhurst-cdd:
> > If none of the "fixing" logic is triggered, the last line returns "nonzeroZ". This would screw up a legitimate 3D mol that happens to have all Z coords as zero (planar, in the XY plane - e.g. benzene). I think it should return "marked3d" to be correct and consistent with the Biovia def.
> 
> The case of all z coordinates are zero but the molecule is marked as 3D is handled correctly by the return at line 2507, so I don't think this is a problem:
> 
> ```
> In [9]: mb = '''
>    ...:      RDKit          3D
>    ...: 
>    ...:   0  0  0  0  0  0  0  0  0  0999 V3000
>    ...: M  V30 BEGIN CTAB
>    ...: M  V30 COUNTS 4 3 0 0 0
>    ...: M  V30 BEGIN ATOM
>    ...: M  V30 1 C 0.000000 0.000000 0.000000 0
>    ...: M  V30 2 C 1.299038 0.750000 0.000000 0
>    ...: M  V30 3 F 2.598076 -0.000000 0.000000 0
>    ...: M  V30 4 Cl 1.299038 2.250000 0.000000 0
>    ...: M  V30 END ATOM
>    ...: M  V30 BEGIN BOND
>    ...: M  V30 1 1 2 1
>    ...: M  V30 2 1 2 3
>    ...: M  V30 3 1 2 4
>    ...: M  V30 END BOND
>    ...: M  V30 END CTAB
>    ...: M  END
>    ...: '''
> 
> In [10]: m = Chem.MolFromMolBlock(mb)
> 
> In [11]: m.GetConformer().Is3D()
> Out[11]: True
> ```

You are right - I missed that return

tadhurst-cdd:
So  I can just close this PR, unless you think there is more to be done (like check to see if the wedging agrees with the 3D chirality).  I think we should just close this one.

greglandrum:
@tadhurst-cdd I have corrected and expanded my comment above - https://github.com/rdkit/rdkit/pull/8573#issuecomment-3112161228 
Unless you think that contains another factual error (I really hope not), I think it does make sense to close this PR and for you to add the logic to recognize and handle this particular case to your internal business rules.

Sorry it's taken me so long to make the time to dig into this.

tadhurst-cdd:
OK I am closing this PR - we at cdd with handle this case of bogus input on our side.  Or maybe we will - it is bugus input.

PR #8591, opened 2025-06-16
Replacing core GraphMol data structures with RDMol

ndickson-nvidia:
This pull request is a draft primarily due to plans to break up RDMol.h and RDMol.cpp to make them less monolithic, and because there are also still some undebugged test failures.

#### Reference Issue
**TODO:** Search for any issues related to performance or memory usage of ROMol, Atom, Bond, RDProps, accessing conformers, RingInfo, etc.

This is a very large change that replaces the implementations of several core data structures in GraphMol with data structures intended to use fewer memory allocations, less memory, and provide interfaces with less overhead on average.

#### Any other comments?
Wrappers are provided that try to maintain as much compatibility as is feasible, so that a vast majority of existing C++ code should function equivalently after recompiling, though some code not ported to use the new interfaces may be significantly slower than it was before. The hope is that most code ported to make effective use of the new interface will be about as fast as before or be significantly faster than before.

To keep the diff smaller, for now, large parts of GraphMol have not yet been ported to use the new interfaces, which in some notable cases, will prevent users from seeing any performance benefits.  For example, code corresponding with `sanitize` and `removeHs` options of `MolFromSmiles` haven't been ported, and the new function `SmilesToMol`, which uses the new interface and simplifies SMILES parsing, doesn't support those options yet either.  A separate pull request will be made with porting changes in GraphMol, unless having those changes in this pull request would be preferred.

### DIFF
--- a/Code/GraphMol/Atom.cpp
+++ b/Code/GraphMol/Atom.cpp
@@ -10,6 +10,7 @@
 #include <cmath>
 
 #include "ROMol.h"
+#include "RDMol.h"
 #include "Atom.h"
 #include "PeriodicTable.h"
 #include "SanitException.h"
@@ -20,9 +21,24 @@
 #include <RDGeneral/RDLog.h>
 #include <RDGeneral/types.h>
 #include <RDGeneral/Dict.h>
+#include <GraphMol/rdmol_throw.h>
 
 namespace RDKit {
 
+bool isAromaticAtom(ConstRDMolAtom atom) {
+  if (atom.data().getIsAromatic()) {
+    return true;
+  }
+  auto [begin, end] = atom.mol().getAtomBonds(atom.index());
+  for (; begin != end; ++begin) {
+    const auto &bond = atom.mol().getBond(*begin);
+    if (bond.getIsAromatic() ||
+        bond.getBondType() == Bond::BondType::AROMATIC) {
+      return true;
+    }
+  }
+  return false;
+}
 bool isAromaticAtom(const Atom &atom) {
   if (atom.getIsAromatic()) {
     return true;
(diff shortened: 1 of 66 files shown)


Code/GraphMol/Atom.cpp
```diff
@@ -20,9 +21,24 @@
...
 
+bool isAromaticAtom(ConstRDMolAtom atom) {
+  if (atom.data().getIsAromatic()) {
+    return true;
+  }
+  auto [begin, end] = atom.mol().getAtomBonds(atom.index());
+  for (; begin != end; ++begin) {
+    const auto &bond = atom.mol().getBond(*begin);
+    if (bond.getIsAromatic() ||
+        bond.getBondType() == Bond::BondType::AROMATIC) {
```
d-b-w: this would likely (bizarrely) be more efficient to search all bonds in the mol for bonds using `atom`, because each ROMol::getBond search for a bond is linear.

Code/GraphMol/Atom.cpp
```diff
@@ -228,426 +219,284 @@ Atom &Atom::operator=(const Atom &other) {
...
-  d_numRadicalElectrons = 0;
-  d_formalCharge = 0;
-  d_index = 0;
-  d_isotope = 0;
-  d_chiralTag = CHI_UNSPECIFIED;
-  d_hybrid = UNSPECIFIED;
-  dp_mol = nullptr;
-  dp_monomerInfo = nullptr;
+Atom::Atom(Atom &&other) {
+  dp_dataMol = other.dp_dataMol;
```
d-b-w, line 223: usually I see this as `*this = std::move(that)`.

Code/GraphMol/Bond.h
```diff
@@ -44,108 +49,107 @@ class Atom;
...
-class RDKIT_GRAPHMOL_EXPORT Bond : public RDProps {
-  friend class RWMol;
+class RDKIT_GRAPHMOL_EXPORT Bond {
   friend class ROMol;
+  friend class RWMol;
+  friend class RDMol;
+
+  // When a bond is created on its own, it owns dp_dataMol to store its data
+  // and dp_owningMol is nullptr. Calling setOwningMol will set dp_owningMol,
+  // but as before, this does not actually insert the Bond's data into the
```
d-b-w, line 58: comments shouldn't refer to the `way things used to be`.

Also, these seem like implementation comments, not comments for the consumers of the Bond class

Code/GraphMol/ROMol.h
```diff
@@ -125,13 +123,15 @@ struct CXXAtomIterator {
...
-    Atom *current;
+    value_type current;
 
     CXXAtomIter(Graph *graph, Iterator pos)
         : graph(graph), pos(pos), current(nullptr) {}
 
-    reference operator*() {
-      current = (*graph)[*pos];
+    const value_type &operator*() {
+      // The const_cast is to maintain the same behaviour as the previous
```
d-b-w, line 132: rdkit usually uses american spellings

Code/GraphMol/Atom.cpp
ndickson-nvidia: `ROMol::getBond` is no longer linear after this change.  It's constant-time.

ndickson-nvidia: Also, the call here is to `RDMol::getBond`, not `ROMol::getBond`, so it's just getting a reference to the `BondData` in the vector.  (Sorry for the double-comment.  I should have checked the code more closely before commenting.)

Code/GraphMol/Atom.cpp
ndickson-nvidia: If the other `Atom` owns an `RDMol`, `std::move(other)` won't normally clear the associated pointers in `other`, so when `other` is destructed, it will delete the `RDMol`, and then when this `Atom` is destructed (or accessed), it will try to delete (or access) the `RDMol` again and crash.  The pointers can't be a simple `std::unique_ptr`, because there are 3 states, only 2 of which have one of the pointers being owned by this `Atom`.

Code/GraphMol/ROMol.h
ndickson-nvidia: Thanks!  Sorry for missing this one.  I originally had a lot of the functions named with "neighbour", but had to go through and change it for consistency, but I didn't think to search for "behaviour".

Code/GraphMol/Atom.cpp
d-b-w: aha! good point. Nice!

Actually, I bet that O(1) lookup alone will be a largish perf improvement in many situations.

Code/GraphMol/Bond.h
ndickson-nvidia: They are implementation comments, not for consumers of the `Bond` class, which is why it's not a comment on the `Bond` class.  It's a comment on the data members inside the class, explaining how the data members behave.  We didn't change the comment on the `Bond` class, above.

Do you have a recommendation on how to avoid referring to the previous behaviour?  The unintuitive way these data members work is specifically to maintain compatibility with the previous behaviour of `setOwningMol`, so someone looking at the code without an explanation that it was to maintain compatibility might understandably wonder why they operate this way.  As we discovered, if the implementation is something simpler and more intuitive, it'll break some of the parsing code.

PR #8640, opened 2025-07-19
Mem fixes

ricrogz:
I ran asan again, and found a couple leaks in code, both of them leaking complete mols:

- in `rdEnumerateStereoisomers.cpp`, we release a `ROMol` to make a copy of it, which is then wrapped into a `shared_ptr`... but the released `ROMol`is never cleaned up. The fix is trivial: we don´t need the copy at all, since the released pointer is already a `ROMol`.

- The leak in `MultithreadedMolSupplier` is trickier. When we parse the text into mols, these are stored in the output queue as raw pointers, which are either extracted from the queue when they are written out, or cleaned up when the output queue is deleted. The problem is that when the `MultithreadedMolSupplier` shuts down, one of the first things we do in `MultithreadedMolSupplier::close()` is telling the output queue to stop accepting elements. Due to the `if(d_done) return;`added to `Code/RDGeneral/ConcurrentQueue.h`in #8556, this means that anything that we try to put into the output queue after that point is sliently discarded, leaking the mol pointers. This is easily fixed by removing the "return" line again and allowing mols to go into the output queue while the threads are being terminated. I also updated `MultithreadedMolSupplier::close()`  to make sure mols in the output queue are cleaned up also when `df_started == false`.

### DIFF
--- a/Code/GraphMol/EnumerateStereoisomers/Wrap/rdEnumerateStereoisomers.cpp
+++ b/Code/GraphMol/EnumerateStereoisomers/Wrap/rdEnumerateStereoisomers.cpp
@@ -42,7 +42,7 @@ class LocalStereoEnumerator {
     if (!iso) {
       return boost::shared_ptr<RDKit::ROMol>();
     }
-    return boost::shared_ptr<RDKit::ROMol>(new RDKit::ROMol(*iso.release()));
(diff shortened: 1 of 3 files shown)


greglandrum (approved):
LGTM

PR #8688, opened 2025-08-07
Add the useRingTemplates option to generateDepictionMatching2DStructure

ricrogz:
This adds a `useRingTemplates` option to the `ConstrainedDepictionParams` struct, so that the option can be passed to `compute2DCoords()` inside `generateDepictionMatching2DStructure()`.


greglandrum (approved):
LGTM

PR #8695, opened 2025-08-11
[WIP] Add osmordred support

bp-kelley:
This adds @thegodone 's osmordred port.

This initial pass is to make sure we can build it at scale, if it passes we'll add swig support and the remaining tests.


bp-kelley:
Moved to #8740

PR #8706, opened 2025-08-14
Fix drawing of very short arcs in lasso.

DavidACosgrove:
In lasso highlighting mode, the test molecule gave a very short arc that came out with only 2 points in it, which the polygon drawer objected to.

#### Any other comments?
DrawMol.cpp seemingly didn't go through clang-format last time someone put a PR in, but I have not re-formatted it here.


greglandrum (approved):
LGTM

PR #8772, opened 2025-09-12
Use ninja in the CI builds [WIP]

nmaeder:
Trying to build with ninja to save some time.

This also applies a small bug fix that  makes better enums part of the build if it's fetched.

### DIFF
--- a/.azure-pipelines/linux_build.yml
+++ b/.azure-pipelines/linux_build.yml
@@ -25,6 +25,7 @@ steps:
     export CXXFLAGS="${CXXFLAGS} -Wall -Werror"
     mkdir build && cd build && \
     cmake .. \
+    -G Ninja \
     -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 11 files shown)


.azure-pipelines/linux_build.yml
```diff
@@ -51,13 +51,14 @@ steps:
...
-    -DCMAKE_LIBRARY_PATH="${CONDA_PREFIX}/lib"
+    -DCMAKE_LIBRARY_PATH="${CONDA_PREFIX}/lib" \
+    -DCMAKE_GENERATOR=Ninja
   displayName: Configure build (Run CMake)
 - bash: |
     source ${CONDA}/etc/profile.d/conda.sh
     conda activate rdkit_build
     cd build
-    make -j $( $(number_of_cores) ) install
+    ninja -j $( $(number_of_cores) )
```
ricrogz: I think these `-j` options are redundant. iirc, Ninja's default is to build in parallel using (num cores + 2) processes.

PR #8774, opened 2025-09-12
Make InteractiveRenderer error reporting more robust and code cleanup

ptosco:
Currently, the `InteractiveRenderer.py` Python module has a bug:
if you set `InteractiveRenderer.rdkitStructureRendererJsUrl` to a non-existing and then run `InteractiveRenderer.setEnabled()`, It will report an error the first time, and then report that interactive rendering is available the second time, which is incorrect, because the JS module has not been loaded; see below:
```
from rdkit.Chem.Draw import InteractiveRenderer

InteractiveRenderer.rdkitStructureRendererJsUrl = "blah"

InteractiveRenderer.setEnabled()
Failed to load rdkit-structure-renderer.js:
SyntaxError: Unexpected token '<'
Interactive molecule rendering is not available in this Jupyter Notebook.

InteractiveRenderer.setEnabled()
Interactive molecule rendering is enabled in this Jupyter Notebook.
```
This PR fixes that bug:
```
from rdkit.Chem.Draw import InteractiveRenderer

InteractiveRenderer.rdkitStructureRendererJsUrl = "blah"

InteractiveRenderer.setEnabled()
Failed to load rdkit-structure-renderer.js:
SyntaxError: Unexpected token '<'
Interactive molecule rendering is not available in this Jupyter Notebook.

InteractiveRenderer.setEnabled()
Interactive molecule rendering is not available in this Jupyter Notebook.
```
Additionally, there is some code cleanup.

### DIFF
--- a/rdkit/Chem/Draw/InteractiveRenderer.py
+++ b/rdkit/Chem/Draw/InteractiveRenderer.py
@@ -31,6 +31,7 @@ parentNodeQuery = "div[class*=jp-NotebookPanel-notebook]"
 _enabled_div_uuid = None
 _camelCaseOptToTagRe = re.compile("[A-Z]")
 _opts = "__rnrOpts"
+_rdk_str_rnr_prefix = "rdk-str-rnr-"
 _disabled = "_disabled"
 _defaultDrawOptions = None
 _defaultDrawOptionsDict = None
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

PR #8798, opened 2025-09-22
update default ET version to 2 in getExperimentalTorsions

nmaeder:
Another default where I think it makes sense to move to ET version 2.


greglandrum (approved):
LGTM

greglandrum (changes requested):
Please add a note to the Backwards Incompatible changes section of ReleaseNotes.md for this

there’s a very low chance that anyone is actually using this somewhere, but we should still make the note

greglandrum (approved):
LGTM

PR #8811, opened 2025-09-24
Add extract mol fragment api

bp-kelley:
Fixes the root cause of #8649

This replaces the pr for #8742

### DIFF
--- a/Code/GraphMol/MolOps.cpp
+++ b/Code/GraphMol/MolOps.cpp
@@ -691,7 +691,7 @@ std::vector<std::unique_ptr<MolSanitizeException>> detectChemistryProblems(
 }
 
 namespace {
-  std::vector<std::unique_ptr<ROMol>> getTheFrags(
+std::vector<std::unique_ptr<ROMol>> getTheFrags(
     const ROMol &mol, bool sanitizeFrags, INT_VECT *frags,
     VECT_INT_VECT *fragsMolAtomMapping, bool copyConformers) {
   std::unique_ptr<INT_VECT> mappingStorage;
@@ -701,7 +701,7 @@ namespace {
   }
   int nFrags = getMolFrags(mol, *frags);
   std::vector<std::unique_ptr<RWMol>> res;
-  
+
   if (nFrags == 1) {
     res.emplace_back(new RWMol(mol));
     if (fragsMolAtomMapping) {
@@ -779,16 +779,14 @@ namespace {
         // empirical. This is mainly intended to catch situations like proteins
         // where you have a bunch of single-atom fragments (waters); the
         // standard approach below ends up being horribly inefficient there
-	SubsetOptions opts {
-	  .sanitize = sanitizeFrags,
-	  .clearComputedProps = true,
-	  .copyCoordinates = copyConformers,
-	  .method = SubsetMethod::BONDS_BETWEEN_ATOMS
-	};
-	std::vector<unsigned int> atoms{comp.begin(), comp.end()};
-	SubsetInfo info;
-	auto submol = copyMolSubset(mol, atoms, info, opts);
-	res.push_back(std::move(submol));
+        SubsetOptions opts{.sanitize = sanitizeFrags,
+                           .clearComputedProps = true,
+                           .copyCoordinates = copyConformers,
+                           .method = SubsetMethod::BONDS_BETWEEN_ATOMS};
+        std::vector<unsigned int> atoms{comp.begin(), comp.end()};
+        SubsetInfo info;
+        auto submol = copyMolSubset(mol, atoms, info, opts);
+        res.push_back(std::move(submol));
       } else {
         res.emplace_back(new RWMol(mol));
         auto &frag = res.back();
--- a/Code/GraphMol/Subset.cpp
+++ b/Code/GraphMol/Subset.cpp
@@ -18,16 +18,15 @@ namespace {
 
 inline void copyComputedProps(const ROMol &src, ROMol &dst) {
   dst.updateProps(src);
-  for(auto &v : dst.getPropList(true, false)) {
-    if(v != RDKit::detail::computedPropName)
-      dst.clearProp(v);
+  for (auto &v : dst.getPropList(true, false)) {
+    if (v != RDKit::detail::computedPropName) dst.clearProp(v);
   }
 }
 
 static void copySelectedAtomsAndBonds(RWMol &extracted_mol,
                                       const RDKit::ROMol &reference_mol,
                                       SubsetInfo &selection_info,
-				      const SubsetOptions &options) {
+                                      const SubsetOptions &options) {
   auto &[selectedAtoms, selectedBonds, atomMapping, bondMapping] =
       selection_info;
   for (const auto &ref_atom : reference_mol.atoms()) {
@@ -36,7 +35,7 @@ static void copySelectedAtomsAndBonds(RWMol &extracted_mol,
     }
 
     std::unique_ptr<Atom> extracted_atom{
-      options.copyAsQuery ? new QueryAtom(*ref_atom) : ref_atom->copy() };
+        options.copyAsQuery ? new QueryAtom(*ref_atom) : ref_atom->copy()};
 
     constexpr bool updateLabel = false;
     constexpr bool takeOwnership = true;
@@ -50,28 +49,28 @@ static void copySelectedAtomsAndBonds(RWMol &extracted_mol,
     }
 
     std::unique_ptr<Bond> extracted_bond{
-      options.copyAsQuery ? new QueryBond(*ref_bond) : ref_bond->copy() };
+        options.copyAsQuery ? new QueryBond(*ref_bond) : ref_bond->copy()};
 
     // Check the stereo atoms
     auto &atoms = extracted_bond->getStereoAtoms();
-    if(atoms.size() == 2) {
+    if (atoms.size() == 2) {
       auto map1 = atomMapping.find(atoms[0]);
       auto map2 = atomMapping.find(atoms[1]);
-      if(map1 != atomMapping.end() && map2 != atomMapping.end()) {
-	atoms[0] = map1->second;
-	atoms[1] = map2->second;
+      if (map1 != atomMapping.end() && map2 != atomMapping.end()) {
+        atoms[0] = map1->second;
+        atoms[1] = map2->second;
       } else {
-	atoms.clear(); // We couldn't map the stereo atoms
+        atoms.clear();  // We couldn't map the stereo atoms
       }
     }
-    
-    for(auto &atomidx : atoms) {
+
+    for (auto &atomidx : atoms) {
       auto map = atomMapping.find(atomidx);
-      if(map != atomMapping.end()) {
-	atomidx = map->second;
+      if (map != atomMapping.end()) {
+        atomidx = map->second;
       }
     }
-    
+
     extracted_bond->setBeginAtomIdx(atomMapping[ref_bond->getBeginAtomIdx()]);
     extracted_bond->setEndAtomIdx(atomMapping[ref_bond->getEndAtomIdx()]);
 
@@ -79,19 +78,16 @@ static void copySelectedAtomsAndBonds(RWMol &extracted_mol,
     auto num_bonds =
         extracted_mol.addBond(extracted_bond.release(), takeOwnership);
     bondMapping[ref_bond->getIdx()] = num_bonds - 1;
-
   }
   // we need to update rings now
 
-  if(selectedBonds.any() && reference_mol.getRingInfo()->isInitialized()) {
+  if (selectedBonds.any() && reference_mol.getRingInfo()->isInitialized()) {
     extracted_mol.getRingInfo()->reset();
   }
-
 }
 
-static bool isSelectedSGroup(
-    const SubstanceGroup &sgroup,
-    const SubsetInfo &selection_info) {
+static bool isSelectedSGroup(const SubstanceGroup &sgroup,
+                             const SubsetInfo &selection_info) {
   auto is_selected_component = [](auto &indices, auto &selection_test) {
     return indices.empty() ||
            std::all_of(indices.begin(), indices.end(), selection_test);
@@ -100,13 +96,14 @@ static bool isSelectedSGroup(
   auto atom_test = [&](int idx) { return selection_info.selectedAtoms[idx]; };
   auto bond_test = [&](int idx) { return selection_info.selectedBonds[idx]; };
   return is_selected_component(sgroup.getAtoms(), atom_test) &&
-    is_selected_component(sgroup.getBonds(), bond_test) &&
-    is_selected_component(sgroup.getParentAtoms(), atom_test);
+         is_selected_component(sgroup.getBonds(), bond_test) &&
+         is_selected_component(sgroup.getParentAtoms(), atom_test);
 }
 
-static void copySelectedSubstanceGroups(
-    RWMol &extracted_mol, const RDKit::ROMol &reference_mol,
-    const SubsetInfo &selection_info, const SubsetOptions &) {
+static void copySelectedSubstanceGroups(RWMol &extracted_mol,
+                                        const RDKit::ROMol &reference_mol,
+                                        const SubsetInfo &selection_info,
+                                        const SubsetOptions &) {
   auto update_indices = [](auto &sgroup, auto getter, auto setter,
                            auto &mapping) {
     auto indices = getter(sgroup);
@@ -125,15 +122,13 @@ static void copySelectedSubstanceGroups(
     SubstanceGroup extracted_sgroup(sgroup);
     extracted_sgroup.setOwningMol(&extracted_mol);
 
-    update_indices(
-        extracted_sgroup, std::mem_fn(&SubstanceGroup::getAtoms),
-        std::mem_fn(&SubstanceGroup::setAtoms), atomMapping);
-    update_indices(
-        extracted_sgroup, std::mem_fn(&SubstanceGroup::getParentAtoms),
-        std::mem_fn(&SubstanceGroup::setParentAtoms), atomMapping);
-    update_indices(
-        extracted_sgroup, std::mem_fn(&SubstanceGroup::getBonds),
-        std::mem_fn(&SubstanceGroup::setBonds), bondMapping);
+    update_indices(extracted_sgroup, std::mem_fn(&SubstanceGroup::getAtoms),
+                   std::mem_fn(&SubstanceGroup::setAtoms), atomMapping);
+    update_indices(extracted_sgroup,
+                   std::mem_fn(&SubstanceGroup::getParentAtoms),
+                   std::mem_fn(&SubstanceGroup::setParentAtoms), atomMapping);
+    update_indices(extracted_sgroup, std::mem_fn(&SubstanceGroup::getBonds),
+                   std::mem_fn(&SubstanceGroup::setBonds), bondMapping);
 
     addSubstanceGroup(extracted_mol, std::move(extracted_sgroup));
   }
@@ -177,16 +172,16 @@ static void copySelectedStereoGroups(RWMol &extracted_mol,
     std::vector<Atom *> atoms;
     for (const auto &atom : stereo_group.getAtoms()) {
       auto mapping = atomMapping.find(atom->getIdx());
-      if(mapping != atomMapping.end()) {
-	atoms.push_back(extracted_atoms[mapping->second]);
+      if (mapping != atomMapping.end()) {
+        atoms.push_back(extracted_atoms[mapping->second]);
       }
     }
 
     std::vector<Bond *> bonds;
     for (const auto &bond : stereo_group.getBonds()) {
       auto mapping = bondMapping.find(bond->getIdx());
-      if(mapping != bondMapping.end()) {
-	bonds.push_back(extracted_bonds[mapping->second]);
+      if (mapping != bondMapping.end()) {
+        bonds.push_back(extracted_bonds[mapping->second]);
       }
     }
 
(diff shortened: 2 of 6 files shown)


rachelnwalker:
Is it possible to add back some of the tests @whosayn had in the original PR?

Also, I would definitely find this useful in python :)

bp-kelley:
@rachelnwalker thanks for the note, I thought I had.   I'll readd them.

bp-kelley:
@rachelnwalker I did, there's a merge conflict though.  I'll fix it

bp-kelley:
I added back the tests with one change, in commitBatch we only remove atoms
from stereogroups, not the stereogroups themselves so I use this behavior.

And yes, python/swig all that good stuff should be implemented.

greglandrum:
@bp-kelley can you please push an empty commit here so that the CI builds run again?

Code/GraphMol/Subset.h
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * NOTE: Bookmarks are currently copied, StereoGroups that are not entirely
+ *        included in the subset are not copied.
+ *
+ */
+RDKIT_GRAPHMOL_EXPORT std::unique_ptr<RDKit::RWMol>
+copyMolSubset(const RDKit::ROMol& mol,
+	      const std::vector<unsigned int>& path,
+	      const SubsetOptions &options = SubsetOptions(),
+	      SubsetInfo *mappings = nullptr);
```
whosayn: from what I gather we only use the  atom mapping from this api. could we just have a `std::map<unsigned int, unsigned int> atom_mapping` variant instead? then we could make `SubsetInfo` private

Code/GraphMol/Subset.h
```diff
@@ -0,0 +1,123 @@
...
+  SubsetOptions(bool sanitize, bool copyAsQuery,
+		bool copyCoordinates, unsigned int conformerIdx) :
+    sanitize(sanitize), copyAsQuery(copyAsQuery), copyCoordinates(copyCoordinates),
+    conformerIdx(conformerIdx)
+  {
+  }
+};
+
+struct RDKIT_GRAPHMOL_EXPORT SubsetInfo {
+  std::vector<bool> selected_atoms;
```
whosayn: @greglandrum suggested making this a `boost::dynamic_bitset<>`, and I agree

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+  
+std::unique_ptr<RDKit::RWMol> copyMolSubset(
+    const RDKit::ROMol &mol, SubsetInfo & selection_info,
+    const SubsetOptions &options) {
+  auto extracted_mol = std::make_unique<::RDKit::RWMol>();
+  copySelectedAtomsAndBonds(*extracted_mol, mol, selection_info, options);
+  copySelectedSubstanceGroups(*extracted_mol, mol, selection_info, options);
+  copySelectedStereoGroups(*extracted_mol, mol, selection_info);
+  if(options.copyCoordinates)
```
whosayn: thanks!

Code/GraphMol/MolOps.cpp
```diff
@@ -751,35 +753,15 @@ std::vector<std::unique_ptr<ROMol>> getTheFrags(
...
-        }
+	SubsetOptions opts;
+	opts.copyCoordinates = copyConformers;
+	opts.sanitize = sanitizeFrags;
+	opts.clearComputedProps = true;
+	opts.method = SubsetMethod::BONDS_BETWEEN_ATOMS;
+	std::vector<unsigned int> atoms{comp.begin(), comp.end()};
+	SubsetInfo info;
+	auto submol = copyMolSubset(mol, atoms, opts, &info);
+	res.push_back(std::unique_ptr<RWMol>(submol.release()));
```
whosayn: `copyMolSubset` should already return a `std::unique_ptr<RWMol>`

Code/GraphMol/MolOps.cpp
```diff
@@ -751,35 +753,15 @@ std::vector<std::unique_ptr<ROMol>> getTheFrags(
...
-            conf->set3D((*cit)->is3D());
-            unsigned int cidx = 0;
-            for (auto ai : comp) {
-              conf->setAtomPos(cidx++, (*cit)->getAtomPos(ai));
-            }
-            frag->addConformer(conf);
-          }
-        }
+	SubsetOptions opts;
+	opts.copyCoordinates = copyConformers;
```
whosayn: nit: consider designated initializers like `SubsetOptions opts{.copyCoorsinates = copyConformers, ...};`

Code/GraphMol/MolOps.cpp
```diff
@@ -816,6 +798,50 @@ std::vector<std::unique_ptr<ROMol>> getTheFrags(
...
+  }
+  int nFrags = getMolFrags(mol, *frags);
+
+  std::vector<std::vector<int>> components;
+  components.resize(nFrags);
+  for(unsigned int atom_idx = 0; atom_idx < mol.getNumAtoms(); ++atom_idx) {
+    components[(*frags)[atom_idx]].push_back(atom_idx);
+  }
+  
+  std::vector<std::unique_ptr<ROMol>> res;
```
whosayn: same point about using designated initializers

Code/GraphMol/MolOps.cpp
```diff
@@ -816,6 +798,50 @@ std::vector<std::unique_ptr<ROMol>> getTheFrags(
   }
   return finalRes;
 }
+
+std::vector<std::unique_ptr<ROMol>> getTheFragsv2(
```
whosayn: cant find any use of this api or any tests.

Code/GraphMol/catch_molops.cpp
```diff
@@ -415,4 +416,239 @@ M  END)CTAB";
     CHECK(mol->getBondBetweenAtoms(10, 8)->getBondType() ==
           Bond::BondType::SINGLE);
   }
-}
\ No newline at end of file
+}
+
+// helper api to get test data for copyMolSubset
+struct SelectedComponents {
```
whosayn, line 422: could we move the tests into `catch_subset.cpp` or something of the sort?

Code/GraphMol/MolOps.cpp
bp-kelley: Yeah, I don't know why I got into this idiom, but I use it everywhere.

Code/GraphMol/MolOps.cpp
bp-kelley: This replaces the original getTheFrags with copyMolSubset and is how I noticed the issues with the multiple ways that getTheFrags actually works.  I'll toss it into a snippet and remove from thie PR, but it is worth noting that we have corner cases which are quite difficult to sort out

Code/GraphMol/Subset.h
bp-kelley: I think it would be useful to have a bond mapping API as well, I have certainly wanted one in the past.  It also seems odd to ask for a path by bonds and not get a bond mapping.

But if you want SubsetInfo private we could sort that out.

Code/GraphMol/Subset.h
bp-kelley: If it ends up being a dynamic_bitset I agree we shouldn't expose it to python.

Code/GraphMol/catch_molops.cpp
bp-kelley: They were initially there but I lost the source code when I first pushed.   I'll move them back.

bp-kelley:
@whosayn @greglandrum I'll tackle some of the review comments and I sorted the ubuntu error with the if statement.

I would like a mechanism to get mapped bonds as well as atoms.  For instance, just because the endpoints of a bond are mapped doesn't mean that the bond is.   This means to either keep SubsetInfo and probably make the following public as the other functions are helper functions around this.

```
copyMolSubset(RWMol &mol, SubsetInfo &info, const SubsetOptions &options)
```

Thoughts?

Once this is in place I'll do the python API.

Then I can show you how to start investigating the weird stuff.

greglandrum:
> @whosayn @greglandrum I'll tackle some of the review comments and I sorted the ubuntu error with the if statement.
> 
> I would like a mechanism to get mapped bonds as well as atoms. For instance, just because the endpoints of a bond are mapped doesn't mean that the bond is. This means to either keep SubsetInfo and probably make the following public as the other functions are helper functions around this.

I can definitely see that being able to track where bonds go would be useful.

Code/GraphMol/MolOps.cpp
bp-kelley: I'll make these changes when I add the python wrapper

Code/GraphMol/Subset.h
bp-kelley: It's now a dynamic bitset

bp-kelley:
@whosayn @rachelnwalker @greglandrum I think this is finally ready.  I think I have handled all critiques and made the Java and python wrappers.

Code/GraphMol/Subset.h
```diff
@@ -0,0 +1,133 @@
...
+//  This file is part of the RDKit.
+//  The contents are covered by the terms of the BSD license
+//  which is included in the file license.txt, found at the root
+//  of the RDKit source tree.
+//
+#ifndef RD_SUBSET_H
+#define RD_SUBSET_H
+
+#include <RDGeneral/export.h>
+#include <RDGeneral/BetterEnums.h>
```
greglandrum: you aren't using this here. Probably don't need to include it

Code/GraphMol/Subset.h
```diff
@@ -0,0 +1,133 @@
...
+ */ 
+enum class SubsetMethod {
+  BONDS_BETWEEN_ATOMS=0,
+  BONDS=1
+};  
+
+//! Subsetting Options for copyMolSubset
+/*
+ * These control what is copied over from the original molecule
+ *     \param sanitize - perform sanitization automatically on the subset
```
greglandrum: `\param` is for arguments to a function.
To document data members of a struct, we add the docs directly after declaring the member.
As an example, take a look at the docs of the options in FileParsers.h

Code/GraphMol/MolOps.cpp
```diff
@@ -751,35 +753,16 @@ std::vector<std::unique_ptr<ROMol>> getTheFrags(
...
-            conf->setId((*cit)->getId());
-            conf->set3D((*cit)->is3D());
-            unsigned int cidx = 0;
-            for (auto ai : comp) {
-              conf->setAtomPos(cidx++, (*cit)->getAtomPos(ai));
-            }
-            frag->addConformer(conf);
-          }
-        }
+	SubsetOptions opts {
```
greglandrum: Just noticing this here. I'm sure it applies elsewhere:
Please run clang-format over all of the code in this PR

Code/GraphMol/Wrap/test_subset.py
```diff
@@ -0,0 +1,96 @@
+#
+#  Copyright (C) 2003-2021  Greg Landrum and other RDKit contributors
```
greglandrum: might as well fix this

Code/GraphMol/Wrap/test_subset.py
```diff
@@ -0,0 +1,96 @@
...
+#         All Rights Reserved
+#
+""" This is a rough coverage test of the python wrapper
+
+it's intended to be shallow, but broad
+
+"""
+
```
greglandrum: might as well delete this; it's wrong and fixing it adds no value
```suggestion
```

Code/GraphMol/Wrap/test_subset.py
```diff
@@ -0,0 +1,96 @@
...
+import sys
+import tempfile
+import unittest
+from io import StringIO
+
+from rdkit import Chem
+
+def check(info, selection):
+  for i,v in enumerate(selection):
+    if v: assert i in info
```
greglandrum: we don't do python (or C++) one-liners
Running yapf over the file (which you should do anyway) *might* fix this

Code/GraphMol/Wrap/test_subset.py
```diff
@@ -0,0 +1,96 @@
...
+  if "RDTESTCASE" in os.environ:
+    suite = unittest.TestSuite()
+    testcases = os.environ["RDTESTCASE"]
+    for name in testcases.split(':'):
+      suite.addTest(TestCase(name))
+
+    runner = unittest.TextTestRunner()
+    runner.run(suite)
+  else:
+    unittest.main()
```
greglandrum: I don't think we need any of this boilerplate now that we're using pytest

Code/GraphMol/catch_molops.cpp
```diff
@@ -417,11 +418,247 @@ M  END)CTAB";
...
+
+  std::vector<unsigned int> extracted_atoms;
+  for (auto& atom : extracted_mol->atoms()) {
+    extracted_atoms.push_back(atom->template getProp<unsigned int>("orig_idx"));
+  }
+
+  CHECK(extracted_atoms == expected_atoms);
+}
+
+// This test makes sure we correctly extract atoms
```
greglandrum:
```suggestion
// This test makes sure we correctly extract bonds
```

Code/JavaWrappers/Subset.i
```diff
@@ -0,0 +1,73 @@
+//
+//  Copyright (C) 2025 and other RDKit contributors
```
greglandrum, line 2: need a name here before the "and"

Code/JavaWrappers/Subset.i
```diff
@@ -0,0 +1,73 @@
...
+//
+%{
+#include <GraphMol/Subset.h>
+%}
+
+%ignore RDKit::copyMolSubset;
+%include <GraphMol/Subset.h>
+
+%{
+RDKit::ROMol *copyMolSubsetHelper(const RDKit::ROMol& mol,
```
greglandrum, line 18: SWIG has some unique_ptr support: https://www.swig.org/Doc4.1/Library.html#Library_std_unique_ptr

Did you try this?

Code/JavaWrappers/gmwrapper/src-test/org/RDKit/ChemTransformsTests.java
```diff
@@ -70,6 +70,18 @@ public void test3BRICS() {
 
 	}
 
+        @Test
+	public void testSubset() {
+	    ROMol mol = RWMol.MolFromSmiles("c1ccccc1CCN");
+	    UInt_Vect vect = new UInt_Vect(mol.getNumAtoms());
+	    for(int i=0;i<6;++i)
+		vect.add(i);
+	    // atom copy
```
greglandrum: we don't have "coding standards" for the java tests, but this is just gross... curly brackets are free:

```suggestion
	    for(int i=0;i<6;++i) {
		  vect.add(i);
        }
	    // atom copy
```

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+#include <GraphMol/RDKitBase.h>
+#include <GraphMol/GraphMol.h>
+#include <GraphMol/QueryAtom.h>
+#include <GraphMol/QueryBond.h>
+#include "Subset.h"
+
+namespace RDKit {
+namespace {
+
+inline void copyComputedProps(const ROMol &src, ROMol &dst) {
```
greglandrum, line 19: Wouldn't this be more efficient if you just looped over the computed props and copied those?

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+    if (!selectedAtoms[ref_atom->getIdx()]) {
+      continue;
+    }
+
+    std::unique_ptr<::RDKit::Atom> extracted_atom{
+      options.copyAsQuery ? new QueryAtom(*ref_atom) : ref_atom->copy() };
+
+    static constinit bool updateLabel = false;
+    static constinit bool takeOwnership = true;
+    atomMapping[ref_atom->getIdx()] = extracted_mol.addAtom(
```
greglandrum: Curiosity: why are these constinit and not constexpr?

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+    static constinit bool takeOwnership = true;
+    auto num_bonds =
+        extracted_mol.addBond(extracted_bond.release(), takeOwnership);
+    bondMapping[ref_bond->getIdx()] = num_bonds - 1;
+
+    // we need to update rings now
+    if(reference_mol.getRingInfo()->isInitialized()) {
+      extracted_mol.getRingInfo()->reset();
+    }
+  }
```
greglandrum: Since we create the molecule just before the call to this function, we're never going to need to do this. Still, I guess it doesn't hurt to include it.

However, I don't think we need to do this for every bond that we add. Can't we move it outside of the loop and do:
```suggestion
    // we need to update rings now
    if(selectedBonds.any() && reference_mol.getRingInfo()->isInitialized()) {
      extracted_mol.getRingInfo()->reset();
    }
  }
```

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+[[nodiscard]] static bool is_selected_sgroup(
```
greglandrum: Why does this need to be `[[nodiscard]]`?

greglandrum: Also, it's not in the public API, so I won't insist, but we don't use underscores in function/variable names

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+    const ::RDKit::SubstanceGroup &sgroup,
+    const SubsetInfo &selection_info) {
+  auto is_selected_component = [](auto &indices, auto &selection_test) {
+    return indices.empty() ||
+           std::all_of(indices.begin(), indices.end(), selection_test);
+  };
+
+  // clang-format off
```
greglandrum: why?

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+inline void copyComputedProps(const ROMol &src, ROMol &dst) {
+  dst.updateProps(src);
+  for(auto &v : dst.getPropList(true, false)) {
+    if(v != RDKit::detail::computedPropName)
+      dst.clearProp(v);
+  }
+}
+
+static void copySelectedAtomsAndBonds(::RDKit::RWMol &extracted_mol,
```
greglandrum: Since we're inside the `RDKit` namespace, I don't think it needs to be used anywhere in this code.

greglandrum (changes requested):
I haven't finished with Subset.cpp. I'll finish it after the suggested changes have been made

Code/GraphMol/MolOps.cpp
bp-kelley: I'll run clang-format last to help with the PR.

Code/GraphMol/Subset.cpp
bp-kelley: Some of these (and the snake case) came from the original PR, I'll have to ask @whosayn, but constexpr or just const seem to be the right choice here.

Code/GraphMol/Subset.cpp
bp-kelley: I seem to remember needing this, but you are right about the location.

Code/GraphMol/Subset.cpp
bp-kelley: I think it's a safety thing, if the return value isn't used, it is a warning.

Code/GraphMol/Subset.cpp
bp-kelley: I think because the values are RDValues, this handles all the complexities for us.  I really think we should grow a updateProps(computed, private) actually.

Code/GraphMol/Subset.cpp
bp-kelley: ?

bp-kelley: Mainly because I didn't notice that.

Code/GraphMol/Subset.cpp
greglandrum: I don't understand your argument, but it's probably not killer for this PR

Code/GraphMol/Subset.cpp
greglandrum: I understand what `[[nodiscard]]` does, but I think it only makes sense to use it when there's a reason that you shouldn't ignore the return value of the function (like when it's passing back a pointer that you own).
I don't think that's the case here.

Code/GraphMol/Subset.cpp
greglandrum: You didn't move it. Was that intentional?

greglandrum:
@bp-kelley let me know when you think this is ready to be reviewed again

Code/GraphMol/Subset.cpp
DavidACosgrove: The compiler might helpfully optimise it out and you’re relying on a side effect of the function? That might not be the case here but it’s why I’ve used it on the past.

greglandrum: > The compiler might helpfully optimise it out and you’re relying on a side effect of the function? That might not be the case here but it’s why I’ve used it on the past.

Definitely could happen, but there are no side-effects here.

bp-kelley:
I’ll be able to make the changes this weekend focusing on the non testing code first.  I expect some of the stylistic issues were just not addressed in the original pr that I moved over.  I’ll do a style pass and then clang format

Code/GraphMol/Subset.h
bp-kelley: Fixed

Code/GraphMol/Subset.h
bp-kelley: Should be resolved

Code/GraphMol/Wrap/test_subset.py
bp-kelley: Fixed

Code/GraphMol/Subset.cpp
bp-kelley: resolved

Code/GraphMol/Subset.cpp
bp-kelley: Resolved

Code/GraphMol/Subset.cpp
bp-kelley: Resolved

Code/GraphMol/Subset.cpp
bp-kelley: Removed the qualifiers

bp-kelley:
@greglandrum should be up for a second look now

Code/GraphMol/Wrap/test_subset.py
greglandrum: just a reminder that there are still two comments in this file

Code/GraphMol/catch_molops.cpp
greglandrum: still here

Code/GraphMol/Subset.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+  
+std::unique_ptr<RDKit::RWMol> copyMolSubset(
+    const RDKit::ROMol &mol, SubsetInfo & selection_info,
+    const SubsetOptions &options) {
+  auto extracted_mol = std::make_unique<RWMol>();
+  copySelectedAtomsAndBonds(*extracted_mol, mol, selection_info, options);
+  copySelectedSubstanceGroups(*extracted_mol, mol, selection_info, options);
+  copySelectedStereoGroups(*extracted_mol, mol, selection_info);
+  if(options.copyCoordinates)
```
greglandrum: please remove all of the one liners

bp-kelley: fixed

bp-kelley: @greglandrum I believe these are all finished.

Code/GraphMol/Wrap/test_subset.py
bp-kelley: I missed the first comment, should be resolved now and the __main__ removed.

greglandrum (approved):
LGTM

PR #8813, opened 2025-09-24
Allow Multiple Core Hits in the Same Molecule in RGroupDecomposition

DavidACosgrove:
#### Reference Issue
No Issue.

If a core matches more than once in a molecule, the RGD is done for all instances.  Previously, it just did the match with the highest score.  For these purposes, "more than once" means a different set of atoms, so symmetric cases are handled as before.  

#### Any other comments?
The changes aren't as complicated as it would appear from the diff.  There's a new loop which pushed everything across.  Also, clang-format has made some spurious changes to RGroupDecompParams.h.

This work was funded by Scitegrity Limited (https://www.scitegrity.com/).


bp-kelley:
By same set, you mean the same set of atoms but in different matching orders I assume.  So if there are two phenyl groups and a phenyl query, one would get two cores.

Code/GraphMol/RGroupDecomposition/Wrap/rdRGroupComposition.cpp
```diff
@@ -288,7 +288,9 @@ struct rgroupdecomp_wrapper {
         "input structure\n"
         "    - doEnumeration: expand input cores into enumerated mol bundles\n"
         "    - allowMultipleRGroupsOnUnlabelled: permit more than one rgroup to "
-        "be attached to an unlabelled core atom";
+        "be attached to an unlabelled core atom\n"
+        "    - allowMultipleCoresInSameMol: permit a core to match more than "
```
bp-kelley: Perhaps: permit multiple cores to match in the molecule if the set of atoms matched are non-equal.

DavidACosgrove:
> By same set, you mean the same set of atoms but in different matching orders I assume. So if there are two phenyl groups and a phenyl query, one would get two cores.

Yes.  If the query matches 2 non-equal atom sets (which may be overlapping) you get 2 different cores.

DavidACosgrove:
@bp-kelley will you be able to finish the review for the next release?  It would be great if so.

bp-kelley:
I'll have time this week to finish the review, but it looks pretty straightforward right now

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
```diff
@@ -346,6 +346,36 @@ int RGroupDecomposition::getMatchingCoreInternal(
   return core_idx;
 }
 
+namespace {
+// Take the matches, all from the same molecule and split them so that
+// different atom sets are separated out.  So that if a core hits
+// more than once in the molecule, both sets of R Groups will be
+// returned.
+std::vector<std::vector<MatchVectType>> splitDifferentAtomMatches(
```
bp-kelley: We use the term unique in the substructure match, so perhaps 

```
splitNonUniqueMatches
```

obviously not a show stopper

Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
```diff
@@ -397,180 +428,192 @@ int RGroupDecomposition::add(const ROMol &inmol) {
...
-  for (const auto &tmatche : tmatches) {
-    const bool replaceDummies = false;
-    const bool labelByIndex = true;
-    const bool requireDummyMatch = false;
-    // TODO see if we need replaceCoreAtomsWithMolMatches or can just use
-    // rcore->core
-    auto coreCopy = rcore->replaceCoreAtomsWithMolMatches(*mol, tmatche);
-    tMol.reset(replaceCore(*mol, *coreCopy, tmatche, replaceDummies,
-                           labelByIndex, requireDummyMatch));
+  std::vector<std::vector<MatchVectType>> splitMatches;
```
bp-kelley: nonUniqueMatches would perhaps be more informative as above.

Code/GraphMol/RGroupDecomposition/RGroupDecompParams.h
```diff
@@ -75,6 +58,9 @@ struct RDKIT_RGROUPDECOMPOSITION_EXPORT RGroupDecompositionParameters {
   bool allowNonTerminalRGroups = false;
   //! unlabelled core atoms can have multiple rgroups
   bool allowMultipleRGroupsOnUnlabelled = false;
+  //! Cores can match more than once in the same molecule if the sets of
```
bp-kelley: `Allow cores to match...`

bp-kelley (approved):
Some minor verbiage tweaking, but it isn't critical.

DavidACosgrove:
> Some minor verbiage tweaking, but it isn't critical.

Thanks @bp-kelley. Those make sense. I’ll make the changes this afternoon.

Code/GraphMol/RGroupDecomposition/RGroupDecompParams.h
DavidACosgrove: I've gone with "Permit a core to match" which is the same as in the Python wrapper.

PR #8825, opened 2025-09-26
allow chiral borates

tadhurst-cdd:
#### Reference Issue
chiral borates were not allowed.  Example: 
image

changes the allowed valence of Boron to include 5, and fixes a bug in verifying valences of charged atoms


greglandrum:
I don't think this is necessary. If the structure above is drawn with a single bond between a positively charged N and a negatively charged B, the RDKit is happy with the valence (because B- is equivalent to C) and the chirality is preserved:
image

I found numerous other examples of chiral borate anions that can be drawn with a negative charge on the boron. These would all be handled perfectly happily by the RDKit.
For example:
https://pubs.rsc.org/en/content/articlelanding/2011/ce/c0ce00709a
https://onlinelibrary.wiley.com/doi/abs/10.1002/anie.202001637
https://pubs.rsc.org/en/content/articlelanding/2023/cc/d2cc05924j
https://pubs.rsc.org/en/content/articlelanding/2008/cc/b806529b

tadhurst-cdd:
We are checking with our customer to see if this works for them.   More in
a bit...

tad

tadhurst-cdd:
change not needed as described in the review

PR #8828, opened 2025-09-28
Fixes a leak in ChemDraw code

ricrogz:
This fixes some leaks in ChemDraw code, which can be detected in `External/ChemDraw/test.cpp`. The problem is happening in this block of code:
https://github.com/rdkit/rdkit/blob/2e1d946f6de4f7834106b1a5bf9a92d93ab8cd85/External/ChemDraw/utils.cpp#L176-L180

The problem is that `FragmentReplacement::replace()` runs a loop that creates some bonds, and then deletes some atoms. When this happens inside the batch edit that is initiated in the code I linked above, we can create bonds involving atoms that have already been marked for deletion. And that will cause the bond to leak, since the detection for bonds that need to be deleted together with the atom happens when `removeAtom()` is called, not when the batch edit is committed.

In other words, when we start a batch edit, then run `removeAtom()` and then use `addBond()` on one of the atoms that has been removed, the new Bond will leak, because it didn't exist when `removeAtom()` checked what needed to be deleted.

The fix I'm posting here consists in marking new bonds for deletion when they are created while in batch edit and the new bond involves a deleted atom. I think this is better than not creating the bond at all (even if we do more work by creating a bond that we're soon going to delete again) because there might be a case where the batch edit is rolled back, and in such case we'll probably want the new bond to be there. Although this is not the case in the context of this leak.

While investigating the bug, I also did some clang-formatting of the ChemDraw code, some minor refactoring in `External/ChemDraw/chemdraw.cpp` (in `MolsFromChemDrawDataStream()` and `MolsFromChemDrawBlock()` I deleted a couple of loops because the output of `molsFromCDXMLDataStream()` already has the same type of what we expect to return, so the loops are redundant), and fixed some typos (although the `return "Forumla";`might be considered a bug on its own :) ).

I highly recommend enabling "Ignore whitespace" to review this one.

### DIFF
--- a/Code/GraphMol/RWMol.cpp
+++ b/Code/GraphMol/RWMol.cpp
@@ -530,8 +530,12 @@ unsigned int RWMol::addBond(unsigned int atomIdx1, unsigned int atomIdx2,
   // we're in a batch edit, and at least one of the bond ends is scheduled
   // for deletion, so mark the new bond for deletion too:
   if (dp_delAtoms &&
-      (dp_delAtoms->test(atomIdx1) || dp_delAtoms->test(atomIdx2))) {
-    dp_delBonds->push_back(1);
+      ((atomIdx1 < dp_delAtoms->size() && dp_delAtoms->test(atomIdx1)) ||
+       (atomIdx2 < dp_delAtoms->size() && dp_delAtoms->test(atomIdx2)))) {
+    if (dp_delBonds->size() < numBonds) {
+      dp_delBonds->resize(numBonds);
+    }
+    dp_delBonds->set(numBonds - 1);
   }
 
   // if both atoms have a degree>1, reset our ring info structure,
--- a/Code/GraphMol/RWMol.cpp
+++ b/Code/GraphMol/RWMol.cpp
@@ -527,6 +527,13 @@ unsigned int RWMol::addBond(unsigned int atomIdx1, unsigned int atomIdx2,
   b->setBeginAtomIdx(atomIdx1);
   b->setEndAtomIdx(atomIdx2);
 
+  // we're in a batch edit, and at least one of the bond ends is scheduled
+  // for deletion, so mark the new bond for deletion too:
+  if (dp_delAtoms &&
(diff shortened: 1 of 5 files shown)


bp-kelley:
@ricrogz yikes.  Guess I just learned to run valgrind on all fixes.

Code/GraphMol/RWMol.cpp
```diff
@@ -527,6 +527,13 @@ unsigned int RWMol::addBond(unsigned int atomIdx1, unsigned int atomIdx2,
   b->setBeginAtomIdx(atomIdx1);
   b->setEndAtomIdx(atomIdx2);
 
+  // we're in a batch edit, and at least one of the bond ends is scheduled
```
bp-kelley, line 530: This appears to be the "Fix" for the leak, correct?

ricrogz: Yes. This will mark these bonds for deletion in the batch edit, so the bonds will be deleted if the batch edit is committed, or will stay if the batch commit is rolled back.

Code/GraphMol/RWMol.cpp
```diff
@@ -527,6 +527,13 @@ unsigned int RWMol::addBond(unsigned int atomIdx1, unsigned int atomIdx2,
...
+  // for deletion, so mark the new bond for deletion too:
+  if (dp_delAtoms &&
+      (dp_delAtoms->test(atomIdx1) || dp_delAtoms->test(atomIdx2))) {
+    dp_delBonds->push_back(1);
```
bp-kelley: Why is this 1 and not numBonds?

ricrogz: That's just to indicate we want the bit on. `push_back()` will add the bit at the end, since we just added the bond.

In any case, this approach was wrong, since we might have added other atoms and bonds that weren't reflected in the state of `dp_delAtoms` and `dp_delBonds` when we hit this. I already pushed a commit to fix it.

greglandrum (approved):
LGTM

PR #8829, opened 2025-09-28
fix

ricrogz:
`boost::python::object::ptr()` returns a pointer to the internal `PyObject` managed by the `boost::python::object`. Having a wrapper function return `PyObject *` is the same as returning a raw pointer; it won´t be managed by Python, and will leak. The right thing to do is return the `boost::python::object`, which will prevent the `PyObject` from leaking.

I split `translateMetadata()` in two separate functions returning `python::object`, which is a parent class to both `python::list`and `python::dict`, which both keeps the different implementations separate and makes it easier to handle the returned object type.


greglandrum (approved):
LGTM

PR #8844, opened 2025-10-03
Use std::string_view for property keys

ricrogz:
This is a WIP, but comments are welcome :)

I still need to do some more testing before this is ready to go.

In https://github.com/rdkit/rdkit/pull/8765 we mentioned using `std::string_view` in the properties interfaces, so that we could make turn `common_properties`  into `inline constexpr std::string_view` as Brian suggested. This works towards that.

This is a breaking API  change, but it shouldn't be that bad, because iirc , both `const char*` and `std::string` can be converted implicitly to `std::string_view` (but not the other way around!), so callers of the property functions should still be ok using `std::string` arguments. The complication comes with the Python wrappers: I had to write a converter that turns Python strings into `std::string_view`. Regarding memory safety, that's one of the things I still need to check, but this should be ok, given that we only use string `std::string_view` as property names, and using the Python `str` internal C-strings should be ok as long as the Python object isn't destroyed during the call, which shouldn't happen.


Code/RDBoost/Wrap/RDBase.cpp
```diff
@@ -235,6 +235,80 @@ struct python_ostream_wrapper {
...
+  /// Check PyObject is a str
+  static void *convertible(PyObject *object) {
+    if (object != nullptr && PyUnicode_Check(object)) {
+      return object;
+    }
+    return nullptr;
+  }
+
+  /// Construct a std::string_view from the internal string of the PyObject.
+  /// This shouldn´t fail, as we
```
bp-kelley: Missing some text here

bp-kelley: The string_view boost conversion looks okay, I would run valgrind though, the boost internals for conversion storage are a bit opaque to me.

ricrogz: Yup, I was interrupted while writing it, in mid sentence.

image

ricrogz: > The string_view boost conversion looks okay, I would run valgrind though, the boost internals for conversion storage are a bit opaque to me.

I have run asan and ubsan on a couple Python tests, and it looks good so far.

ricrogz:
Ok, this looks ready to review. I also gave it a quick run under asan and ubsan and it didn't seem to cause any issues.

greglandrum (approved):
LGTM

greglandrum:
The one downside to this is that, since it's a breaking API change, it needs to wait for the 2026.03 release.

greglandrum (changes requested):
Since this includes a breaking API change, please add something to the "Backwards incompatible changes" section of the release notes.

Which reminds me that I haven't done the PR to get things ready for the next release... I will do that now.

ricrogz:
> The one downside to this is that, since it's a breaking API change, it needs to wait for the 2026.03 release.

No problem, this can wait. It's just the first step to get @cdvonbargen's https://github.com/rdkit/rdkit/pull/8765 in. I already have some more changes on top of this prepared, but I decided to split the changes into 2 or 3 PRs to make them easier to review.

> Since this includes a breaking API change, please add something to the "Backwards incompatible changes" section of the release notes.
> 
> Which reminds me that I haven't done the PR to get things ready for the next release... I will do that now.

I see you already added the prep for the next release. I'll add something when I update this branch.

greglandrum:
@ricrogz this has a conflict since I merged the giant iostreams commit

ricrogz:
> @ricrogz this has a conflict since I merged the giant iostreams commit

Fixed! It was just a minor `#include` conflict due to the removal of `iostream` and the addition of `string_view`.

PR #8854, opened 2025-10-07
Homogenize maxAttempts / maxIterations documentation

nmaeder:
Related to #8807.

Copying the docstring of `maxIterations` in `EmbedParameters` to `maxAttempts` in `EmbedMolecule` and `EmbedMultipleConfs` since both of them accomplish exactly the same, but are unfortunately named different.


greglandrum (approved):
LGTM

PR #8855, opened 2025-10-08
Incremental synthon search

jgullingsrud:
The SynthonSpaceSearch routines currently accumulate all results as ROMols in the search results object.  For large searches this storage imposes significant memory overhead which impedes running searches on inexpensive, memory-constrained hardware (such as one might use in cloud compute platforms). 

This PR adds new interfaces which return ROMol results through a callback as the hits are generated.  The chunk size is configurable through the `max_chunk_size` setting.  The python wrappers expose the callback through an interface which accepts a callable object which will be passed python lists of Mols.

The new interface was implemented for substructure, fingerprint, and rascal searches.  In python, the incremental search has a different method name; this choice avoids with overloading the existing name (because of default keyword arguments the boost runtime can't always resolve which method was intended), and probably makes the most sense since the incremental search version doesn't return a search results object.

In practice, we use a `toTryChunkSize` of 4096, rather than the existing default of something like 2M.  It might be worth considering making the default smaller but no change was made here.

### DIFF
--- a/Code/GraphMol/SynthonSpaceSearch/fingerprint_search_catch_tests.cpp
+++ b/Code/GraphMol/SynthonSpaceSearch/fingerprint_search_catch_tests.cpp
@@ -113,7 +113,9 @@ TEST_CASE("FP Small tests") {
     // test with callback version
     std::set<std::string> cbSmis;
     auto cb = [&cbSmis](const std::vector<std::unique_ptr<ROMol>> &results) {
-      for (auto &r : results) cbSmis.insert(MolToSmiles(*r));
+      for (const auto &r : results) {
+        cbSmis.insert(MolToSmiles(*r));
+      }
       return false;
     };
     synthonspace.fingerprintSearch(*queryMol, *fpGen, cb, params);


greglandrum:
@jgullingsrud Thanks for the contribution!

@DavidACosgrove is the best person to review this and he's out through next week.  I may take a quick pass through just to look at the API, but he'll do the real review when he's back.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -222,6 +236,22 @@ class RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpace {
       const ROMol &query, const FingerprintGenerator<std::uint64_t> &fpGen,
       const SynthonSpaceSearchParams &params = SynthonSpaceSearchParams());
 
+  // Perform a RASCAL similarity search with the given query molecule
```
DavidACosgrove: This is a fingerprint search.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -194,6 +194,20 @@ class RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpace {
       const SubstructMatchParameters &matchParams = SubstructMatchParameters(),
       const SynthonSpaceSearchParams &params = SynthonSpaceSearchParams());
 
+  /*!
+   * Perform a substructure search with the given query molecule across
+   * the synthonspace library.  Duplicate SMILES strings produced by
+   * different reactions will be returned.
+   *
+   * @param query : query molecule
+   * @param callback: user-provided callback receiving chunks of ROMols.
```
DavidACosgrove, line 206: I think it would be very useful here if there was a brief explanation of the difference and flexibility that the callback offers.  @greglandrum may disagree, but I would be happy with a reference to an example in the test codes showing it stuffing a vector of SMILES strings.

Code/GraphMol/SynthonSpaceSearch/Wrap/rdSynthonSpaceSearch.cpp
```diff
@@ -98,6 +98,37 @@ SynthonSpaceSearch::SearchResults substructureSearch_helper2(
   return results;
 }
 
+struct CallbackAdapter {
```
DavidACosgrove, line 101: This is all witchcraft to me.  It makes superficial sense to me, but @greglandrum, I fear it's for you to pass judgement.

Code/GraphMol/SynthonSpaceSearch/SearchResults.h
```diff
@@ -11,19 +11,25 @@
...
 #define RDKIT_SYNTHONSPACE_SEARCHRESULTS_H
 
+#include <functional>
 #include <RDGeneral/export.h>
 #include <GraphMol/ROMol.h>
 
 namespace RDKit::SynthonSpaceSearch {
 
+// takes vector of search results; returns true if enough hits have been
+// returned, false if the search should continue.
```
DavidACosgrove, line 21: Is this correct?  I don't see the return value being used in the ::search() function.

DavidACosgrove (changes requested):
Some minor changes suggested.

I like the flexibility that this approach gives, but I think it could do with an explanation of the use somewhere.  I'm not sure quite where that would go, though.  Maybe in the https://www.rdkit.org/docs/GettingStartedInPython.html#searching-synthon-spaces documentation.  An example of how to use it in Python would be much appreciated, I imagine.  I would certainly find it helpful.  I'm tempted to say "that could be a separate PR" but we all know that documentation suggestions like that never happen!

In the PR notes, you mention a max_chunk_size parameter, but I don't find that anywhere.  Is that because toTryChunkSize does the job?  I note that I had failed to expose that to Python which you have now done.  Sorry.

I also wonder if a simpler solution for your specific case might be to give the option to accumulate and return the hits as SMILES strings or product names directly, similar to the asMols option of rdkit.Chem.rdmolops.GetMolFrags().  I wouldn't abandon this though, as I can see all manner of uses for it, including things like adding a "not" function for the hits, such as "don't return any hits that have a fingerprint similarity within a threshold to any of these fingerprints".  In fact, that might make a cool blogpost for @greglandrum who I'm sure is always seeking inspiration for his weekly posts.

Oh, and as @greglandrum said, many thanks for taking the time to do this and submit it.  Much appreciated.

greglandrum:
> In fact, that might make a cool blogpost for @greglandrum who I'm sure is always seeking inspiration for his weekly posts.

It is true that I am **always** happy for inspiration/contributions for the weekly blog posts!

Code/GraphMol/SynthonSpaceSearch/Wrap/rdSynthonSpaceSearch.cpp
```diff
@@ -98,6 +98,37 @@ SynthonSpaceSearch::SearchResults substructureSearch_helper2(
...
+  python::object py_callable;
+
+  bool operator()(std::vector<std::unique_ptr<ROMol>> &results) const {
+    python::list pyres;
+    for (auto &mol : results) {
+      pyres.append(boost::make_shared<ROMol>(*mol));
+    }
+    py_callable(pyres);
```
bp-kelley: I'm confused why this always returns false.    I would expect something like:

```
bool res = false;
try {
    python::object *callable_result = py_callable(pyres);
    res = PyObject_IsTrue(callable_result);
} catch (const python::error_already_set&) {
    res = false;
    throw;  // we we don't throw, at least we should return false
}

return res;
```

I'm unclear whether we should eat or throw the python error.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
jgullingsrud: I've extended the docstrings to say more about what the callback is doing, and I've added a section to the GettingStarted book on incremental search along with a usage example.

Code/GraphMol/SynthonSpaceSearch/SearchResults.h
jgullingsrud: You're quite right, I neglected to actually implement this logic in the initial set of commits.  It's now implemented and there are test cases in C++ and python.

Code/GraphMol/SynthonSpaceSearch/Wrap/rdSynthonSpaceSearch.cpp
jgullingsrud: Thanks to boost::python, exceptions thrown in the callback are propagated back to the invocation of the search function.  I've fixed the python wrapper to properly return the truthiness of the callback return value, but I didn't do any special exception handling.

jgullingsrud:
> Some minor changes suggested.
> 
> I like the flexibility that this approach gives, but I think it could do with an explanation of the use somewhere. I'm not sure quite where that would go, though. Maybe in the https://www.rdkit.org/docs/GettingStartedInPython.html#searching-synthon-spaces documentation. An example of how to use it in Python would be much appreciated, I imagine. I would certainly find it helpful. I'm tempted to say "that could be a separate PR" but we all know that documentation suggestions like that never happen!

Good idea, I added a section to that page, let me know if there's anything else that should be in there.

> 
> In the PR notes, you mention a max_chunk_size parameter, but I don't find that anywhere. Is that because toTryChunkSize does the job? I note that I had failed to expose that to Python which you have now done. Sorry.
> 

My bad, toTryChunkSize is indeed the relevant parameter.

> I also wonder if a simpler solution for your specific case might be to give the option to accumulate and return the hits as SMILES strings or product names directly, similar to the asMols option of rdkit.Chem.rdmolops.GetMolFrags(). I wouldn't abandon this though, as I can see all manner of uses for it, including things like adding a "not" function for the hits, such as "don't return any hits that have a fingerprint similarity within a threshold to any of these fingerprints". In fact, that might make a cool blogpost for @greglandrum who I'm sure is always seeking inspiration for his weekly posts.
> 

You mean something like having the SynthonSpaceSearcher class construct a list of smiles and titles, pass that list to to the SearchResults object, add a SearchResults.getSmiles() method, and have SearchResults.getHitMolecules() reconstruct the ROMols on the fly (as an iterator instead of returning a list)?  Seems possible, I don't know what the overhead is for the round trip to SMILES, or perhaps you had something else in mind...  

> Oh, and as @greglandrum said, many thanks for taking the time to do this and submit it. Much appreciated.

My pleasure, thanks for all the constructive feedback!

Code/GraphMol/SynthonSpaceSearch/Wrap/testSynthonSpaceSearch.py
```diff
@@ -53,14 +53,58 @@ def testSubstructSearch(self):
...
+    self.assertEqual(2, len(mols))
+
+    # Exceptions thrown in the callback propagate back here
+    mols = []
+    params.toTryChunkSize = 2
+    def callback_raises(chunk):
+        mols.extend(chunk)
+        raise StopIteration
+
+    try:
```
DavidACosgrove, line 94: This works as is and so this is a mild suggestion.  I think the unittest paradigm is normally
```
self.assertRaises(StopIteration, synthonspace.SubstructureSearchIncremental, query, callback_raises,
substructMatchParams=smParams, params=params)
```
Or something like that - I haven't got anything handy to check for the call.

DavidACosgrove (approved):
That all looks good to me, with one minor quibble.

DavidACosgrove:
> > Some minor changes suggested.
> > I like the flexibility that this approach gives, but I think it could do with an explanation of the use somewhere. I'm not sure quite where that would go, though. Maybe in the https://www.rdkit.org/docs/GettingStartedInPython.html#searching-synthon-spaces documentation. An example of how to use it in Python would be much appreciated, I imagine. I would certainly find it helpful. I'm tempted to say "that could be a separate PR" but we all know that documentation suggestions like that never happen!
> 
> Good idea, I added a section to that page, let me know if there's anything else that should be in there.
> 
> > In the PR notes, you mention a max_chunk_size parameter, but I don't find that anywhere. Is that because toTryChunkSize does the job? I note that I had failed to expose that to Python which you have now done. Sorry.
> 
> My bad, toTryChunkSize is indeed the relevant parameter.
> 
> > I also wonder if a simpler solution for your specific case might be to give the option to accumulate and return the hits as SMILES strings or product names directly, similar to the asMols option of rdkit.Chem.rdmolops.GetMolFrags(). I wouldn't abandon this though, as I can see all manner of uses for it, including things like adding a "not" function for the hits, such as "don't return any hits that have a fingerprint similarity within a threshold to any of these fingerprints". In fact, that might make a cool blogpost for @greglandrum who I'm sure is always seeking inspiration for his weekly posts.
> 
> You mean something like having the SynthonSpaceSearcher class construct a list of smiles and titles, pass that list to to the SearchResults object, add a SearchResults.getSmiles() method, and have SearchResults.getHitMolecules() reconstruct the ROMols on the fly (as an iterator instead of returning a list)? Seems possible, I don't know what the overhead is for the round trip to SMILES, or perhaps you had something else in mind...

Actually, I just meant have the option of returning a vector of SMILES or titles directly instead of a vector of mol objects.  I imagine most people just dump a SMILES string to file anyway.  So there'd be a params option "returnHitsAsSmiles" or something that would fill in an appropriate vector in the SearchResults, leaving the mols one empty, and then a method on SearchResults to return those SMILES strings.  I'll put a suggestion in the GitHub tracker so I remember to do this.

> 
> > Oh, and as @greglandrum said, many thanks for taking the time to do this and submit it. Much appreciated.
> 
> My pleasure, thanks for all the constructive feedback!

DavidACosgrove:
@greglandrum @bp-kelley This is good to go, I think, from my point of view.  I guess one of you needs to give it a once-over and press the button when you too are happy.

DavidACosgrove:
The CI failure is our old friend the Java diversity picker.

WentongZhou:
٩(•̤̀ᵕ•̤́๑)ᵒᵏᵎᵎᵎᵎ

获取Outlook for Android<https://aka.ms/AAb9ysg>

________________________________
发件人: Brian Kelley ***@***.***>
发送时间: 星期二, 十月 21, 2025 8:52:04 下午
收件人: rdkit/rdkit ***@***.***>
抄送: Subscribed ***@***.***>
主题: Re: [rdkit/rdkit] Incremental synthon search (PR #8855)

This Message Is From an External Sender
This message came from outside your organization.
Report Suspicious<https://us-phishalarm-ewt.proofpoint.com/EWT/v1/NcNAlj2UIVkjDA!6cAUYfWrHdmEEuf3uJcAFY4s9UwTVHejoXxfnKd72_JJky966fISIfrWLEV0WfYZhLiamyPPwkdUKTe2K8FQQ3w-vGxzxDfGmeWesfGX22k$>


@bp-kelley commented on this pull request.

________________________________

In Code/GraphMol/SynthonSpaceSearch/Wrap/rdSynthonSpaceSearch.cpp<https://urldefense.com/v3/__https://github.com/rdkit/rdkit/pull/8855*discussion_r2448112276__;Iw!!NcNAlj2UIVkjDA!97rmnkPlcUWPfw5K3Te8UVc3ozcNZPtJRdPwkj1q7jwziGoYnfRZtTgmlTSwju_i_kdOEB2XmOga0aCv7qdz6wTFlixp$>:

> @@ -98,6 +98,37 @@ SynthonSpaceSearch::SearchResults substructureSearch_helper2(
   return results;
 }

+struct CallbackAdapter {
+  python::object py_callable;
+
+  bool operator()(std::vector<std::unique_ptr<ROMol>> &results) const {
+    python::list pyres;
+    for (auto &mol : results) {
+      pyres.append(boost::make_shared<ROMol>(*mol));
+    }
+    py_callable(pyres);


I'm confused why this always returns false. I would expect something like:

bool res = false;
try {
    python::object *callable_result = py_callable(pyres);
    res = PyObject_IsTrue(callable_result);
} catch (const python::error_already_set&) {
    res = false;
    throw;  // we we don't throw, at least we should return false
}

return res;


I'm unclear whether we should eat or throw the python error.

jgullingsrud:
Hello, just checking back to see if there was anything left to be done here, to my understanding all the feedback has been addressed.

Code/GraphMol/SynthonSpaceSearch/Wrap/rdSynthonSpaceSearch.cpp
```diff
@@ -98,6 +98,36 @@ SynthonSpaceSearch::SearchResults substructureSearch_helper2(
   return results;
 }
 
+struct CallbackAdapter {
+  python::object py_callable;
+
+  bool operator()(std::vector<std::unique_ptr<ROMol>> &results) const {
+    python::list pyres;
+    for (auto &mol : results) {
+      pyres.append(boost::make_shared<ROMol>(*mol));
```
greglandrum, line 107: It's sad that we have to copy the molecules here, but I don't see any way around it with the way things are currently structured.

Code/GraphMol/SynthonSpaceSearch/fingerprint_search_catch_tests.cpp
```diff
@@ -110,6 +110,15 @@ TEST_CASE("FP Small tests") {
       resSmis.insert(MolToSmiles(*r));
     }
 
+    // test with callback version
+    std::set<std::string> cbSmis;
+    auto cb = [&cbSmis](std::vector<std::unique_ptr<ROMol>> &results) {
+      for (auto &r : results) cbSmis.insert(MolToSmiles(*r));
```
greglandrum: We don't do one liners; please put curly brackets around things:

```suggestion
      for (const auto &r : results) {
        cbSmis.insert(MolToSmiles(*r));
      }
```

Code/GraphMol/SynthonSpaceSearch/rascal_search_catch_tests.cpp
```diff
@@ -99,6 +99,16 @@ TEST_CASE("RASCAL Small tests") {
     for (const auto &r : results.getHitMolecules()) {
       resSmis.insert(MolToSmiles(*r));
     }
+
+    // test with callback version
+    std::set<std::string> cbSmis;
+    auto cb = [&cbSmis](std::vector<std::unique_ptr<ROMol>> &results) {
+      for (auto &r : results) cbSmis.insert(MolToSmiles(*r));
```
greglandrum:
```suggestion
      for (const auto &r : results) {
        cbSmis.insert(MolToSmiles(*r));
      }
```

Code/GraphMol/SynthonSpaceSearch/SearchResults.h
```diff
@@ -11,19 +11,25 @@
...
+using SearchResultCallback =
+    std::function<bool(std::vector<std::unique_ptr<ROMol>> &)>;
```
greglandrum: It seems like the callbacks shouldn't be modifying their arguments:
```suggestion
    std::function<bool(const std::vector<std::unique_ptr<ROMol>> &)>;
```
(this change requires other code changes)

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -50,6 +50,66 @@ SynthonSpaceSearcher::SynthonSpaceSearcher(
...
+        std::vector<std::unique_ptr<ROMol>> partResults;
+        processToTrySet(toTry, endTime, partResults);
+        hitCount += partResults.size();
+        stop = cb(partResults);
+        toTry.clear();
+        if (stop || (d_params.maxHits != -1 && hitCount >= d_params.maxHits)) break;
+      }
+      stepper.step();
+    }
+    if (stop || (d_params.maxHits != -1 && hitCount >= d_params.maxHits)) break;
```
greglandrum:
```suggestion
    if (stop || (d_params.maxHits != -1 && hitCount >= d_params.maxHits)) {
      break;
    }
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -50,6 +50,66 @@ SynthonSpaceSearcher::SynthonSpaceSearcher(
...
+    // process the synthons
+    while (stepper.d_currState[0] != numSynthons[0]) {
+      toTry.emplace_back(hitset.get(), stepper.d_currState);
+      if (toTry.size() == static_cast<size_t>(d_params.toTryChunkSize)) {
+        std::vector<std::unique_ptr<ROMol>> partResults;
+        processToTrySet(toTry, endTime, partResults);
+        hitCount += partResults.size();
+        stop = cb(partResults);
+        toTry.clear();
+        if (stop || (d_params.maxHits != -1 && hitCount >= d_params.maxHits)) break;
```
greglandrum:
```suggestion
        if (stop || (d_params.maxHits != -1 && hitCount >= d_params.maxHits)) {
         break;
       }
```

Code/GraphMol/SynthonSpaceSearch/Wrap/rdSynthonSpaceSearch.cpp
jgullingsrud: Yeah, the SearchResults version also makes a copy.  If `shared_ptr<ROMol>` were used here instead of `unique_ptr` this could be avoided, but that seems like a larger discussion.

jgullingsrud:
> I like the idea here and the implementation mostly looks good. I made some small suggestions.
> 
> One API question: wouldn't it be possible to make the callback part of `SynthonSpaceSearchParams` and avoid adding the extra argument/overload?

It would be possible, but then the `SearchResults` object returned by the original `search()` method wouldn't have any molecules to return, because the ROMol list get cleared after the callback gets invoked (to avoid consuming memory, which was the point of the callback approach).

DavidACosgrove:
Seeing this makes we wonder if my alternative suggestion of a SearchResults mode that keeps and returns SMILES strings instead of ROMols would be easier. But keep the callback method in for extra processing or filtering of hits prior to accepting them.   It would still save memory but wouldn’t have the requirement of passing in a vector to be filled by the callback. I can take a look at that if it sounds sensible.

greglandrum:
> > I like the idea here and the implementation mostly looks good. I made some small suggestions.
> > One API question: wouldn't it be possible to make the callback part of `SynthonSpaceSearchParams` and avoid adding the extra argument/overload?
> 
> It would be possible, but then the `SearchResults` object returned by the original `search()` method wouldn't have any molecules to return, because the ROMol list get cleared after the callback gets invoked (to avoid consuming memory, which was the point of the callback approach).

Ah, got it.
In that case having the separate API endpoint to make it clear that results won't be returned makes sense.
Thanks for that explanation.

greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @jgullingsrud !

PR #8868, opened 2025-10-14
[CONTRIB]: Freewilson now keeps the coordinates passed in

bp-kelley:
Keeps the original molecule rgroups as opposed to only keeping the smiles strings.

This allows coordinates to be kept for reconstruction purposes.

### DIFF
--- a/Contrib/FreeWilson/test/test_freewilson.py
+++ b/Contrib/FreeWilson/test/test_freewilson.py
@@ -15,13 +15,13 @@ import importlib
 fw = importlib.reload(fw)
 
 from rdkit import Chem, rdBase
-from rdkit.Chem import rdFMCS
+from rdkit.Chem import rdFMCS, rdMolAlign
 
 PATH = os.path.join(os.path.dirname(fw.__file__), 'data')
 assert os.path.exists(PATH), PATH
 
 
-def atest_chembl():
+def test_chembl():
   logging.getLogger().setLevel(logging.INFO)
   smilesfile = os.path.join(PATH, "CHEMBL2321810.smi")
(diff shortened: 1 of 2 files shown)


Contrib/FreeWilson/test/test_freewilson.py
```diff
@@ -68,3 +78,18 @@ def test_multicore():
   decomp = fw.FWDecompose(scaffolds, mols, [1, 2, 3, 4, 5, 6])
   s = io.StringIO()
   fw.predictions_to_csv(s, decomp, fw.FWBuild(decomp))
+
+def test_fep_benchmark_3D():
+  benchmark = os.path.join(PATH, "cmet_ligands.sdf")
+  mols = list(Chem.SDMolSupplier(benchmark))
```
badisa: Do you want to keep the hydrogens?

```suggestion
  mols = list(Chem.SDMolSupplier(benchmark, removeHs=False))
```

bp-kelley: For this particular test it doesn't matter, but we should test that we actually keep them in the right coords.  It's a good suggestion.

bp-kelley: I've updated the test to check for RMSD against the training set.  We can't directly compare coords since the RGroups come from different conformers.  I'll think about if we need to address this at some point.

bp-kelley:
I think at some point we should keep all the coordinates from all rgroups of the same type, but that will take more engineering and testing.

greglandrum (approved):
I just skimmed this since it's in contrib

PR #8939, opened 2025-11-10
Handle chiral atoms with 2 hydrogens in SCSR parser

tadhurst-cdd:
#### Reference Issue
When using the SCSR parser to parse an ACSR file and fully convert it into all-atom format, Some Sugar templates have leaving groups that are H atoms.  These centers are marked as chiral, but the leaving H-atom and one implicit H atom trigger a check in the chirality determination code that removes that chirality.    This removal is artificial, and the chirality should remain in the expanded molecule.

A new parameter has been added to disable the removal of chirality when there is an implicit H and an explicit H.  This parameter (allowTwoHs) defaults to False, so no current code needs to be changed.

#### Any other comments?
A test has been added, and this includes an SCSR file that has an H-atom leaving group in the Sugar template (DNASlurpErrorImport.mol).  In this particular case, then template was derived from deoxyribose, and could have an OH for that leaving group.  However, some custom "sugar" templates are not derived from an acutal sugar molecule, and could have an arbitrary leaving group (H-atom)

### DIFF
--- a/setup.bat
+++ b/setup.bat
@@ -1,4 +1,0 @@
-export RDBASE="/home/tad/cdd/cdd_rdkit/Allow2HChiralTemplates/rdkit"
-export PYTHONPATH="$RDBASE/:/opt/boost/lib/"
-export LD_LIBRARY_PATH="$RDBASE/lib:/opt/boost/lib/"
-


tadhurst-cdd:
That code change does not work very well in other cases.     The leaving
groups do not always "leave" when the SCSR mol is expanded.  If there is no
connection at that point, the atoms of the leaving group are added to the
expanded mol, and that includes bogus atoms with isotopes of 999.

For example, in the test molecule DNASlurpErrorImport.mol, the leaving
groups on the ends of the DNA strand are expanded to Hs with isotope of 999.

If the sugar connection to a base is not used in the SCSR mol, then the
chirality of the C that would normally be attached to a base is NOT
removed- it has an H with no isotope and an H with 999 isotope.    In this
case the chirality should be removed, and this in fact happens correctly
with my suggested code change  - that chirality is removed later inn the
code because the atom in the expanded mol has two identical H atoms on it.

You mentioned that your suggested change would also work for cases where
the duplicated atom is not an H atom.  This already works with the existing
code - the only issue is in the code at and around line 551
in atomChiralTypeFromBondDirPseudo3D (chirality.cpp) when it checks to see
if atom marked chiral has an implicit H and an explicit H.   It really only
is the Hs that have a problem.

I verified that when I add a chlorine atom the C that has the leaving group
to the base, and change the leaving group to a chlorine atom also, the
stereochemistry is NOT removed.  This is true with the original node before
my modifications.

greglandrum (commented):
This all made me wonder if we need to have the test for hSeen at all, so I just completely removed hSeen from `atomChiralTypeFromBondDirPseudo3D()` and everything seems to be fine: all the tests pass.

Based on that, I think a much simpler version of this fix would be to do that... no new global options required, no changes to the parsers required.

Could you please check whether or not your new tests still pass if you do that?

tadhurst-cdd:
Removing the check for hseen works fine for me and my test.  All tests pass.  This removal would allow atoms with wedged to be marked chiral even if they have two Hs.   These bad chiral markers are removed later if sanitation is done.

I am good with this as long was you are not concerned that someone has code that will be broken.  
I will go ahead and make the change.

greglandrum:
I think your argument about the way other atoms are handled is a good one; there's no need to treat Hs differently this way

setup.bat
greglandrum, line 1: This was, I guess, accidentally added. Please remove it

greglandrum (changes requested):
Looks good except for the extra file

tadhurst-cdd:
I made the changes to remove hSeen and it all tests OK.   The CI build in github shows a Java test error - I do not know how to fix or even inspect that.   Can you take a look, or give me hint how to check/run/test fix it?

greglandrum:
> I made the changes to remove hSeen and it all tests OK. The CI build in github shows a Java test error - I do not know how to fix or even inspect that. Can you take a look, or give me hint how to check/run/test fix it?

It's not a real error, don't worry about it.

My only remaining request is to remove the extra file (see the comment/review above)

tadhurst-cdd:
The setup.bat file has been removed.

greglandrum:
> The setup.bat file has been removed.

I'll merge after the CI builds go through.

Issue #8936, opened 2025-11-07
SVG exports of bicolor solid wedge bound not working with stroke-width


EliotRagueneau:
**Describe the bug**
When drawing and SVG of a molecule containing a solid bicolor wedge, the bicolor wedge is composed of 3 paths instead of 2, which creates issues when scaling the stroke-width up. An example and explanation is available at https://github.com/ebi-chebi/ChEBI/issues/4777#issuecomment-3377247271  . 
This is also problematic as it increase the weight and render time of the SVG I believe, as you are adding many little triangles instead of adding a point to a path to create a 4-sided polygon. 

**To Reproduce**
I'm sorry I'm a third-party user of a third-party user, it's a bit hard for me to be sure of how things are done exactly, but I think the code using RDKIT is the following from ChEMBL https://github.com/chembl/libRDChEBI/blob/main/libRDChEBI/depiction.py:
```python 
from typing import Optional, List, Tuple
from chembl_structure_pipeline.standardizer import parse_molblock
from rdkit.Chem.Draw import rdMolDraw2D
from rdkit import Chem

def depict(
    molfile: str,
    width: int = 300,
    height: int = 300,
    output_format: str = "svg",
    baseFontSize: float = -1,
    fixedFontSize: float = -1,
    minFontSize: float = -1,
    maxFontSize: float = -1,
    useCDKAtomPalette: bool = True,
    explicitMethyl: bool = True,
    scaleBondWidth: bool = False,
    addStereoAnnotation: bool = True,
    useMolBlockWedging: bool = True,
    atomLabelDeuteriumTritium: bool = True,
) -> Optional[bytes | str]:
    """Generate an SVG or PNG depiction of a molecule from a molfile.

    Args:
        molfile: A string containing the molecule data in molfile format
        width: Width of the output image in pixels
        height: Height of the output image in pixels
        output_format: Output format, either 'svg' or 'png'
        baseFontSize: Base font size for atom labels (-1 for auto)
        fixedFontSize: Fixed font size for all labels (-1 for variable)
        minFontSize: Minimum font size for atom labels (-1 for no limit)
        maxFontSize: Maximum font size for atom labels (-1 for no limit)
        useCDKAtomPalette: Use CDK atom colors if True
        explicitMethyl: Show explicit methyl groups if True
        scaleBondWidth: Scale bond widths with drawing size if True
        addStereoAnnotation: Add stereochemistry annotations if True
[...]

PR #8943, opened 2025-11-11
Better solid triangles

DavidACosgrove:
#### Reference Issue
Addresses #8936

Where appropriate, solid quadrilaterals are drawn rather than 2 solid triangles.  I don't know if it will address the underlying issue that Eliot was concerned about, but it is a bit tidier and hopefully doesn't do any harm.

#### Any other comments?
It was complicated by situations like
testHandDrawn-3
where the solid triangle is extended up to the 2 ring bonds.

I have looked at all the drawings produced by the 3 test programs and they all look the same.

The hash codes are stuffed in many cases.  I may update those on a separate PR.

There's no specific test for this, as they crop up all over so the change is implicitly tested in many places.


EliotRagueneau:
Awesome! Just out of curiosity, why do you prefer to flatten all points of different polygons into a single list of points? Wouldn't it be cleaner to have something like 

- polygons: Polygon[]
- Polygon: Point[]

(Not c++ syntax but understandable I hope)

PR #8948, opened 2025-11-14
Valence error when parsing SCSR

tadhurst-cdd:
#### Reference Issue
in SCSR template definitions, Nitro groups with 2 double bonds from O atoms to the N atom generate a valence error.  Also, 6 N atoms that are in aromatic rings and also bonded to a heavy metal give the same error.  The error message is: "Structure is invalid: Explicit valence for atom # 1 N, 4, is greater than permitted. Please check the format, try pasting it directly, or opening it in the structure editor."

The call to update the valence for the template ctabs was called with "strict" on.  This is not how that routine is called in parsing regular molecules, so that has been changed to be not strict.


greglandrum (approved):
LGTM

PR #8958, opened 2025-11-19
Add test to watch the state of canonicalization problems reported in #8775

ricrogz:
I was looking at #8957, and was wondering whether this would fix any of the canonicalization issues that we have open in #8775 (spoiler: no), and I thought it might be a good idea to have a test to track known canonicalization issues.

In this way, when we make changes, we can see if these fix any of these issues (or break some that had already been fixed...)

### DIFF
--- a/Code/GraphMol/catch_canon.cpp
+++ b/Code/GraphMol/catch_canon.cpp
@@ -1349,29 +1349,27 @@ TEST_CASE("Canonicalization issues watch (see GitHub Issue #8775)") {
        true},  // #8862
   };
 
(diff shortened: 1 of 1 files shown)


greglandrum (changes requested):
I really like the idea, and the implementation looks mostly ok, but please use the normal RDKit naming conventions (i.e. ditch the underscores).

ricrogz:
> I really like the idea, and the implementation looks mostly ok, but please use the normal RDKit naming conventions (i.e. ditch the underscores).

Will do. For some reason, I need to keep reminding myself to not user underscores...

PR #8962, opened 2025-11-21
Small fix in ring stereo/canonicalization

ricrogz:
I was a little bit bored, and looked into this item in the new canonicalization watch test: https://github.com/rdkit/rdkit/blob/3320d7b454f837b181d26ed8b35fd720c366e88c/Code/GraphMol/catch_canon.cpp#L1304

It brought me back to something I noticed in #8736, but didn't touch because we were in a hurry to fix a bug and I wasn't quite sure about it. But now that I was looking at a clear example where this is a problem, I confirmed my suspicions.

The patch is very simple (if you ignore whitespace in the diff): if something is detected as having a "dupe" in the neighbor ranks, it can't be chiral, not even in ring stereo. So we need to set `possibleRingStereoAtoms[aidx] = 0;`. The decrement is not enough because in `flagRingStereo()` we can flag the same atom pair for ring stereo multiple times. The code in the block starting at: https://github.com/rdkit/rdkit/blob/3320d7b454f837b181d26ed8b35fd720c366e88c/Code/GraphMol/FindStereo.cpp#L707-L709 can be triggered more than once if the same pair of potentially ring stereo atoms are part of more than one ring, as in the case of the mol I was looking at:

image

Here, atoms 4 and 6 are bonded and carry parities (DOWN bonds in the image), and are part of two rings (`(1, 2, 3, 4, 6)` and `(1, 2, 5, 4, 6)`), so that for each of them `possibleRingStereoAtoms[aidx] = 2`, and when atom 6 is found to have dupes, we only `--possibleRingStereoAtoms[aidx];` once, so that it was never completely discarded from having ring stereo, which meant that atom 4 was detected as a valid stereo atom because of the dependency, and ended up triggering the canonicalization issue.

### DIFF
--- a/Code/GraphMol/catch_canon.cpp
+++ b/Code/GraphMol/catch_canon.cpp
@@ -1172,8 +1172,7 @@ TEST_CASE("Canonicalization issues watch (see GitHub Issue #8775)") {
       {R"smi(C1=C\CCCCCC/C=C/C=C/1)smi", true, true},                   // #8759
       {R"smi(O=C=NC1=CC2C3=C(C=C1)C2=C(N=C=O)C=C3)smi", false, false},  // #8721
       {R"smi(O=C(c1ccccc1C(=O)N1C(=O)c2ccccc2C1=O)N1C(=O)c2ccccc2C1=O)smi",
-       false, false},                                                   // #8721
-      {R"smi(O=C=NC1=CC2C3=C(C=C1)C2=C(N=C=O)C=C3)smi", false, false},  // #8721
+       false, false},  // #8721
       {R"smi(O=[N+]([O-])c1cc/c2c(c1)=C(c1ccccc1)/N=c1\\ccc([N+](=O)[O-])cc1=C(c1ccccc1)/N=2)smi",
        false, true},  // #8721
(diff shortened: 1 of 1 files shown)


ricrogz:
btw, I also tried to keep track of the bonds that had been inspected in `flagRingStereo()`, and that also fixed this issue, but caused issues in chirality tests, as it left some potential ring stereo atoms unflagged.

Code/GraphMol/catch_canon.cpp
```diff
@@ -1220,11 +1219,6 @@ TEST_CASE("Canonicalization issues watch (see GitHub Issue #8775)") {
       {R"smi([H]/N=C(C=C)\C(/N=C\[O-])=N\[H])smi", false, true},  // #7759
       {R"smi([H]N=C(/N=C\[O-])/C(C=C)=N\[H])smi", true, true},    // #7759
       {R"smi([H]/N=C(/C=C)C(=N)/N=C\[O-])smi", true, true},       // #7759
-      {R"smi([H]/N=C(/C=C)C(=N)/N=C\[O-])smi", true, true},       // #7759
-      {R"smi([C@H]12[C@H]3[C@@H]4[C@H]5[C@@H]([C@H]1N24)N53)smi", false,
-       true},  // #7759
-      {R"smi([C@H]12[C@H]3[C@H]4[C@@H]5[C@@H]([C@H]1N25)N34)smi", false,
-       true},  // #7759
```
rachelnwalker, line 1227: Why did you remove these tests?

ricrogz: They are duplicates (see the commit's comment).

greglandrum (approved):
LGTM

PR #8968, opened 2025-11-25
Fix for issue #8965

### LINKED ISSUE #8965: Stereo bond inversion in SMILES Writer canonicalization
This is coming from the canonicalization example `C/C=C\C=C(/C=C\C)C(/C=C\C)=C/C`  in #8775.

Minimal reproducible:

```python

In [1]: smi2 = Chem.CanonSmiles(r'C/C=C\C=C(/C=C\C)C(/C=C\C)=C/C')

In [2]: print(smi2)
# C/C=C\C=C(/C=C\C)C(/C=C\C)=C\C
#                             ^ Last bond is inverted

In [3]: print(Chem.CanonSmiles(smi2))
# C/C=C\C=C(/C=C\C)C(/C=C\C)=C/C
#                             ^ Last bond is inverted again
```

The mol looks like this (numbers are atom indices):

Image

The reason the inversion is happening is because bonds 4-8 and 8-9 are assigned incompatible directions, and then the direction in 4-8 is removed, as the algorithm considers it redundant.

This happens because when the stereo bonds are canonicalized and the bond directions are assigned, it is done in the following order: first it's bond 1-2, then 3-4, 5-6, 9-10 and finally 8-12. So, when bond 3-4 is canonicalized, a direction is assigned to 4-8. Then, when we canonicalize 9-10, it has nothing in common with any of the other bonds, so the direction of 8-9 can be assigned arbitrarily, and it ends up in a direction that conflicts with 4-8.

Then, when 8-12 is canonicalized, we pick up 4-8 as the reference direction (because it has a lower atom index), and the direction of 12-13 is calculated. In this process, we do check that the directions of 4-8 and 8-9 are compatible, but the check seems to be flawed, and in any case, the check wouldn't trigger a fix, just raise an error.

I'm considering some options for a fix, but I need to do some more experimenting.

ricrogz:
This is my attempt to fix #8965.

It attempts to solve the issue by changing the way we canonicalize stereo bonds. Instead of traversing the `molStack` and assigning stereo bond directions to bonds as we discover them, it uses 2 queues to define the order in which stereo bonds are canonicalized.

The outer queue is a priority queue, in which stereo bonds are prioritized based on the number of other stereo bonds are constrained by assigning bond directions to the current bond, using the `molStack` order as secondary criterion. Once a bond is popped from this queue, we canonicalize it, and then we use the second queue to recurrently canonicalize the bonds that are affected by the newly assigned directions. In either queue, any bond that has already been canonicalized is skipped if seen again (so far, we cross our fingers that we never assign contradicting directions, but we should add more checks to `canonicalizeDoubleBond()` at some point).

~So far, this code fixes the canonicalization issue for the case I was looking at (`C/C=C\C=C(/C=C\C)C(/C=C\C)=C/C`), but triggers failures on other cases in `catch_canon.cpp` that did work before (note that I temporarily made the test stricter by making it check CIP codes instead of just counting the number of stereo features). I haven't looked at other tests yet, as I assume some of them fail just because the SMILES references we use have changed to something that is not necessarily wrong. I will investigate them once the issues in `catch_canon.cpp` are fixed.~

I tore out most of the old rules to decide the bond directions, since they were written on the assumption that we'd traverse the `molStack` in the forward direction, while the queues system described above means we can go back and forth, which badly broke the algorithm.

This probably doesn't fix all the double bond canonicalization issues, but seems to solve some of them, and it passes all tests (even though I had to update some of them).

( reposted from further down in the thread: )
I downloaded ~49 million SMILES from enamine real that contained stereo double bonds (the selection criterion was that the SMILES had to contain either a forward or a backward slash. I ran the test set twice: once with Release 2025.09.5, and another one applying this patch. The results were identical (seems that enamine doesn't include any cases affected by this).

### DIFF
--- a/Code/GraphMol/catch_canon.cpp
+++ b/Code/GraphMol/catch_canon.cpp
@@ -1213,7 +1213,7 @@ TEST_CASE("Canonicalization issues watch (see GitHub Issue #8775)") {
       {R"smi(CC1=C/[C@@H](C)C[C@]2(C)CC[C@H](O2)[C@]23CC[C@](CO)(C[C@H](O2)[C@@H]2O[C@@](C)(CC2=O)[C@@H](O)[C@H]2CC[C@@]4(CCC[C@@H](O4)[C@H](C)C(=O)O[C@H]4C[C@H]([C@]5(O)OCC[C@H](C)[C@@H]5O)O[C@H]4\C=C/1)O2)O3)smi",
        true, true},  // #8089
       {R"smi(COC(=O)C1=C/c2cc3c(cc2-c2c(cc(OC)c(OC)c2OC)\C=C/1C(=O)OC)OCO3)smi",
-       true, true},                                         // #8089
+       true, true},                                           // #8089
       {R"smi(N#C[P@@H]/C=C(/P=O)[P@@H]C#N)smi", true, true},  // #8089
       {R"smi(C1=C\C/C=C(\C)CC[C@H]2C(C)(C)[C@@H](\C=C/1)CC[C@]2(C)O)smi", true,
        true},                                                     // #8089
@@ -1340,8 +1340,8 @@ TEST_CASE("Canonicalization issues watch (see GitHub Issue #8775)") {
       {R"smi(O[C@H]1[C@H]2C[C@H](C2)[C@]12CC2)smi", false, true},  // #7759
       {R"smi(N[C@]1(C(=O)O)[C@@H]2C[C@H]3C[C@@H](C2)C[C@H]1C3)smi", false,
        true},  // #8862
-      {R"smi(C/C=C1/C(=C\Cl)/C(=C\F)/C(=C\N)/C/1=C\O)smi", true,
-       true},  // #8965
+      {R"smi(C/C=C1/C(=C\Cl)/C(=C\F)/C(=C\N)/C/1=C\O)smi", false,
+       false},  // #8965
   };
 
   auto count_features = [](RWMol m) {


Code/GraphMol/Canon.cpp
```diff
@@ -17,10 +17,68 @@
...
 #include <RDGeneral/utils.h>
+
+#include <ranges>
+#include <queue>
 #include <algorithm>
 
 namespace RDKit {
 namespace Canon {
+namespace {
+inline Bond::BondDir flipBondDir(Bond::BondDir bondDir) {
```
d-b-w: prefer `static` to `inline`. both of them are mere suggestion to the optimizer, but at least static is guaranteed to restrict the name availability

```suggestion
static constexpr Bond::BondDir flipBondDir(Bond::BondDir bondDir) {
```

(also, this _can_ be `constexpr`, so it probably should be)

Code/GraphMol/Canon.cpp
```diff
@@ -17,10 +17,68 @@
...
+  switch (bondDir) {
+    case Bond::ENDUPRIGHT:
+      return Bond::ENDDOWNRIGHT;
+    case Bond::ENDDOWNRIGHT:
+      return Bond::ENDUPRIGHT;
+    default:
+      throw std::logic_error("flipBondDir called on not double bond dir");
+  }
+
+  return (bondDir == Bond::ENDUPRIGHT) ? Bond::ENDDOWNRIGHT : Bond::ENDUPRIGHT;
```
d-b-w: this line is unreachable, nix it

Code/GraphMol/Canon.cpp
```diff
@@ -17,10 +17,68 @@
...
 namespace Canon {
+namespace {
+inline Bond::BondDir flipBondDir(Bond::BondDir bondDir) {
+  switch (bondDir) {
+    case Bond::ENDUPRIGHT:
+      return Bond::ENDDOWNRIGHT;
+    case Bond::ENDDOWNRIGHT:
+      return Bond::ENDUPRIGHT;
+    default:
+      throw std::logic_error("flipBondDir called on not double bond dir");
```
d-b-w: This didn't formerly throw, so we have a behavior change here. I think that's probably ok, I just wanna flag it.

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +428,156 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
+        bond->getStereoAtoms().size() < 2) {
+      // not a bond that can have stereo nor needs canonicalization
+      bond->setStereo(Bond::STEREONONE);
+      continue;
+    }
+
+    auto &currentNbrs = stereoBondNbrs[bond];
+    for (const auto *dblBondAtom :
+         std::array<Atom *, 2>{bond->getBeginAtom(), bond->getEndAtom()}) {
+      for (const auto *bond : mol.atomBonds(dblBondAtom)) {
```
d-b-w: This shadows the other `bond` name. Please change the name or change the scope of the other name

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +428,156 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
     }
+    std::ranges::sort(currentNbrs,
+                      [&molStackComparer](const Bond *aBnd, const Bond *bBnd) {
+                        // Reversing the bonds is intentional: molStackComparer
+                        // is a std::greater comparer (priority queue returns
+                        // the highest element), but here we want to sort in
+                        // increasing order, so we want a std::less comparer,
+                        // which can be achieved by reversing the std::greater
+                        // because we can have no ties here
+                        return molStackComparer(bBnd->getIdx(), aBnd->getIdx());
```
d-b-w: This looks like it should be 

    molStackComparer(bondVisitOrders[bBnd->getIdx()], bondVisitOrders[aBnd->getIdx()]);


If _not_, then maybe you can use a projection - 

    std::ranges::sort(currentNbrs, std::less{}, &Bond::getIdx)

Code/GraphMol/MMPA/MMPA_UnitTest.cpp
```diff
@@ -730,7 +730,7 @@ void testGithub8569() {
       });
   std::vector<std::pair<std::string, std::string>> expected = {
       {"null", "C/C=C/C=C/[*:1].C[*:1]"},
-      {"C(=C/[*:2])\\[*:1]", "C/C=C/[*:1].C[*:2]"},
+      {"C(=C\\[*:2])/[*:1]", "C/C=C/[*:1].C[*:2]"},
```
d-b-w, line 733: oh, interesting - the "mainline" branch is before the side one in assignment of direction

Code/GraphMol/Canon.cpp
d-b-w: Exception messages are part of the interface, so we should update this to something that might be meaningful

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +428,156 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
+        bond->getStereo() <= Bond::STEREOANY ||
+        bond->getStereoAtoms().size() < 2) {
+      // not a bond that can have stereo nor needs canonicalization
+      bond->setStereo(Bond::STEREONONE);
+      continue;
+    }
+
+    auto &currentNbrs = stereoBondNbrs[bond];
+    for (const auto *dblBondAtom :
+         std::array<Atom *, 2>{bond->getBeginAtom(), bond->getEndAtom()}) {
```
d-b-w: you don't need the cast to std::array here - `initializer_list` are iterable:

 
```suggestion
         {bond->getBeginAtom(), bond->getEndAtom()} {
```

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +428,156 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
-      if (tbond->getBondType() == Bond::DOUBLE &&
-          tbond->getStereo() > Bond::STEREOANY) {
-        dblBondPresent = true;
-      } else if (tbond->getBondType() == Bond::SINGLE &&
-                 tbond != firstFromAtom2) {
-        otherAtom3Bond = tbond;
-      }
+  // Now that we have bonds in the order we want to handle them,
+  // do the canonicalization
+  boost::dynamic_bitset<> seen_bonds(mol.getNumBonds());
```
d-b-w: maybe more idiomatic to use std::vector<bool> - there are no multi-bit operations here

Code/GraphMol/Canon.cpp
```diff
@@ -86,84 +154,14 @@ auto _possibleCompare = [](const PossibleType &arg1, const PossibleType &arg2) {
...
-    isFlipped = true;
-    dir = flipBondDir(dir);
-  }
-
-  return std::make_pair(dir, isFlipped);
-}
-
-}  // namespace
 // FIX: this may only be of interest from the SmilesWriter, should we
 // move it there?
```
d-b-w, line 162: Is this statement still accurate?

Code/GraphMol/Canon.cpp
```diff
@@ -233,6 +231,72 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
+  // When parsing a SMILES, we expect anchor1 to come before atom1,
+  // and atom2 before anchor2. But that is not always the case.
+  // Inverting the order has effects on the bond directions, so
+  // we define the variables below to help us find the right direction
+  // for the bond directions. These are interpreted as follows:
+  //
+  // - firstFromAtom1 and secondFromAtom1 are considered "flipped" if
+  // the anchor would come after the double bond start atom1, e.g.
+  // in the SMILES C(\N)(/O)=C(/N)\O they are both flipped. Note
+  // that, with this definition secondFromAtom1 is usually "flipped".
```
d-b-w, line 257: as always - spectacular comment about why this is needed with a clear & specific example. nice

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +428,156 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
-      if (isClosingRingBond(otherAtom3Bond)) {
-        dir = flipBondDir(dir);
+
+    std::queue<Bond *> connectedBondsQ;
+    connectedBondsQ.push(bond);
+
+    while (!connectedBondsQ.empty()) {
+      const auto currentBond = connectedBondsQ.front();
+      connectedBondsQ.pop();
+
```
d-b-w, line 588: I think that it is logically possible for connectedBondsQ to have the same bond added twice. In that situation, we'd need to do a `seen_bonds` check the second time.

I guess that any time that happens it might be an error in the priority queue? Basically, I'm not sure whether that `seen_bonds` should be an exception or a continue;

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +428,156 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
-                common_properties::_TraversalRingClosureBond)) {
-          // std::cerr<<"  switcheroo 1"<<std::endl;
-          switchBondDir(atom1ControllingBond);
+void canonicalizeDoubleBonds(ROMol &mol, const UINT_VECT &bondVisitOrders,
+                             const UINT_VECT &atomVisitOrders,
+                             UINT_VECT &bondDirCounts, UINT_VECT &atomDirCounts,
+                             const MolStack &molStack) {
+  auto getNeighboringStereoBond = [&mol](const Atom *dblBndAtom,
+                                         const Bond *nbrBnd) -> Bond * {
+    auto otherAtom = nbrBnd->getOtherAtom(dblBndAtom);
```
d-b-w, line 491: move this lamdba below the function implementtation description, please.

Code/GraphMol/Canon.cpp
ricrogz: yup, seems I forgot to delete it when I moved the exception here.

Code/GraphMol/Canon.cpp
ricrogz: Yeah, adding the exception is intentional. See my comment further below.

Code/GraphMol/Canon.cpp
ricrogz: Good catch!

Code/GraphMol/Canon.cpp
ricrogz: Actually, I removed the exception again.

ricrogz: I removed the exception again.

Code/GraphMol/Canon.cpp
ricrogz: I'm not sure. But if it was true before, then it still should be.

Code/GraphMol/Canon.cpp
ricrogz: Thanks

Code/GraphMol/Canon.cpp
ricrogz: No, it's definitely not an error. If there is a cycle involving conjugated double bonds such as  `OC1=C/C=C\C=C/C=C\1` we don't want to end in an endless loop, nor canonicalize the same bond twice. So we just skip.

Also, the same bonds we visit by iterating on `connectedBondsQ` will, sooner or later, pop up in `q` too, and we want to skip them too, since we already canonicalized them based on a neighbor.

Code/GraphMol/Canon.cpp
ricrogz: done

Code/GraphMol/Canon.cpp
ricrogz: Oh, sorry, I misunderstood your comment.

We check `seen_bonds` before pushing anything into the queue, but it doesn't hurt to check it again after popping.

d-b-w: It is _necessary_ to check it again after popping because the same bond could be added twice.

Code/GraphMol/Canon.cpp
```diff
@@ -17,10 +17,82 @@
...
+    case Bond::ENDUPRIGHT:
+      return Bond::ENDDOWNRIGHT;
+    case Bond::ENDDOWNRIGHT:
+      return Bond::ENDUPRIGHT;
+    default:
+      return bondDir;
+  }
+}
+
+void setDirectionFromNeighboringBond(Bond *sourceBond, bool isSourceBondFlipped,
```
greglandrum: please either take references or include PRECONDITIONS for pointer arguments.
We don't do it everywhere, but we probably should

Code/GraphMol/Canon.cpp
```diff
@@ -17,10 +17,82 @@
...
+  // one (and only one) of the bonds is flipped (if both were
+  // flipped, the flips in the direction would cancel out).
+  if (isSourceBondFlipped == isTargetBondFlipped) {
+    dir = flipStereoBondDir(dir);
+  }
+
+  targetBond->setBondDir(dir);
+}
+
+Bond::BondDir getReferenceDirection(const Bond *dblBond, const Atom *refAtom,
```
greglandrum: please either take references or include PRECONDITIONS for pointer arguments

ricrogz:
I downloaded ~49 million SMILES from enamine real that contained stereo double bonds (the selection criterion was that the SMILES had to contain either a forward or a backward slash. I ran the test set twice: once with Release 2025.09.5, and another one applying this patch. The results were identical (seems that enamine doesn't include any cases affected by this).

Code/GraphMol/Canon.cpp
```diff
@@ -369,146 +453,146 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
+    auto bond = msI.obj.bond;
+    Bond::BondDir dir = bond->getBondDir();
+    if (dir == Bond::ENDDOWNRIGHT || dir == Bond::ENDUPRIGHT) {
+      bond->setBondDir(Bond::NONE);
+    }
+
+    if (bond->getBondType() != Bond::DOUBLE ||
+        bond->getStereo() <= Bond::STEREOANY ||
+        bond->getStereoAtoms().size() < 2) {
+      // not a bond that can have stereo nor needs canonicalization
```
greglandrum:
```suggestion
      // not a bond that can have stereo or that needs canonicalization
```

greglandrum (changes requested):

greglandrum:
Sorry, forgot to include a comment: This basically looks fine to me. We also need an entry in the in the "Backwards incompatible changes" section about the SMILES for double bonds with stereo likely changing.

greglandrum:
@d-b-w how's this one look to you?

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6491,4 +6493,64 @@ M  END
...
+  }
+
+  // All samples should canonicalize to the same SMILES
+  REQUIRE(canonicalized_smiles.size() == 1);
+
+  // ... and the canonical SMILES should preserve stereo labels
+  auto m = v2::SmilesParse::MolFromSmiles(*canonicalized_smiles.begin());
+  REQUIRE(m);
+  CHECK(*labels_reference == get_bond_stereo_labels(*m));
+}
```
d-b-w, line 6556: If there is not one, could you make sure that we have a case for a molecule with conflicting bond directions? It would be good to document the behavoir.

(You mentioned that there are certain molecules for which there is no possible canonicalization)

Code/GraphMol/Canon.cpp
```diff
@@ -233,6 +235,110 @@ void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
...
+  // and check if both directions on each side are set.
+  // We hit this in cases with cycles like CO/C1=C/C=C\C=C/C=N\1.
+  if (dir1Set && dir2Set) {
+    // these are guaranteed to exist
+    bondDirCounts[firstFromAtom1->getIdx()] += 1;
+    bondDirCounts[firstFromAtom2->getIdx()] += 1;
+    atomDirCounts[atom1->getIdx()] += 1;
+    atomDirCounts[atom2->getIdx()] += 1;
+
+    // To do: check that the existing directions are consistent.
```
d-b-w, line 308: how, um, to do-y is this to do? Like, should we have a follow up case to address this? OR a test that demonstrates what happens if the check were to fail? Or could this be handled with an assertion/precondition

ricrogz: That depends on how likely you think we are to find rings where all vertexes have stereo bonds attached (and, of course, how important this issue is in comparison with other issues, e.g. #9064 and other ring finding issues), e.g.: 
image



In this case, I think there will be at least one double bond whose 2 adjacent ring bonds are conflicting due to having the same direction at this point in the algorithm. In this case, that bond *might* have its stereo inverted, and we'd have to figure out a way to handle this.

But this is all theoretical. I'd rather not speculate any further before having any real example I can study.

Code/GraphMol/catch_chirality.cpp
ricrogz: If you mean a test case that has conflicting bond directions on input, that's completely unrelated to this patch. This is about SMILES Writing. Conflicts in input should be handled by the parser.

If you mean an example of a case that ends up with conflicting directions during canonicalization, that's the case of the code you highlighted below.

Code/GraphMol/Canon.cpp
ricrogz: (just note that this was not introduced in the current patch, this is something that was already present before my changes)

d-b-w:
> I downloaded ~49 million SMILES from enamine real that contained stereo double bonds (the selection criterion was that the SMILES had to contain either a forward or a backward slash. I ran the test set twice: once with Release 2025.09.5, and another one applying this patch. The results were identical (seems that enamine doesn't include any cases affected by this).

could you bump this validation comment into the description at top?

ricrogz:
> > I downloaded ~49 million SMILES from enamine real that contained stereo double bonds (the selection criterion was that the SMILES had to contain either a forward or a backward slash. I ran the test set twice: once with Release 2025.09.5, and another one applying this patch. The results were identical (seems that enamine doesn't include any cases affected by this).
> 
> could you bump this validation comment into the description at top?

done

PR #8977, opened 2025-11-27
Embed Parameters to JSON

nmaeder:
#### Reference Issue
No Issue

Small functionality to convert the EmbedParameters object into a json string, optionally also with maps.

I extracted some of the code from `updateEmbedParamtersFromJSON` to prevent code duplication in listing the structs members.
Happy to also add this to the python wrapper at a later point in time.

Currently, only the coordMap is supported, but i plan to add functionality to also obtain the boundsmatrix / CPCI in the same manner, depending on the need adding single flags for each of the maps(?). 

#### Any other comments?
Very minimal tests so far, happy to take suggestions on how to further test this.


greglandrum (changes requested):
Code changes look fine.
I think it would be worth adding a test that uses `updateEmbedParametersFromJSON()`  on the output from `embedParametersToJSON()` and makes sure that you get the same structure back.

nmaeder:
added the test as proposed

PR #9005, opened 2025-12-12
Scsr writer

tadhurst-cdd:
#### Reference Issue
Need to take a fully atomistic mol structure/block/file and a list of SCSR templates, and convert it to an SCSR class item/block/file.

For a list of scsr templates , use them as patterns to find representatives in a fully atomistic molecule and produce an SCSR block or file from it


tadhurst-cdd:
re-writing the macromol part to be generic - not SCSR centric
So closing the PR for now

PR #9013, opened 2025-12-19
move molzip to its own file

greglandrum:
There's enough complexity there that it makes sense to have a separate file for molzip.

It's worth considering creating a `MolZip.h` too, but I wanted to keep this one minimal


greglandrum:
@bp-kelley this is purely mechanical - moving the code to a new file, there's no refactoring

PR #9016, opened 2025-12-19
Fixes #9014

### LINKED ISSUE #9014: SDMolSupplier enters an infinite loop if number of SGroups is incorrect
Tested with Release 2025.09.3:

```python
# There's more than one issue with this molblock...
mb = """
                    2D

  8  7  0  0  0  0            999 V3000
M  V30 BEGIN CTAB
M  V30 COUNTS 8 7 0 0 0
M  V30 BEGIN ATOM
M  V30 1 C 1.299038 0.750000 0.000000 0
M  V30 2 C 1.299038 2.250000 0.000000 0
M  V30 3 O 2.598076 3.000000 0.000000 0
M  V30 4 C 2.598076 -0.000000 0.000000 0
M  V30 5 C 0.000000 0.000000 0.000000 0
M  V30 6 H 3.897114 0.750000 0.000000 0
M  V30 7 H 2.598076 4.500000 0.000000 0
M  V30 8 H 2.598076 -1.500000 0.000000 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 1 1 2
M  V30 2 1 2 3
M  V30 3 1 1 4
M  V30 4 1 1 5
M  V30 5 1 4 6
M  V30 6 1 3 7
M  V30 7 1 4 8
M  V30 END BOND
M  V30 BEGIN SGROUP
M  V30 1 SRU 0 ATOMS=(8 1 2 3 4 5 6 7 8) LABEL=polyA BRKXYZ=(0 0 0)
M  V30 END SGROUP
M  V30 END CTAB
M  END
$$$$"""

from rdkit import Chem
s = Chem.SDMolSupplier()
s.SetData(mb, strictParsing=False)
next(s)
# Results in the next message, repeted ad infinitum:
# [13:32:34] BEGIN SGROUP  found but Sgroups NOT expected on line 26
```

ricrogz:
Fixes #9014

The issue was that once we hit the "BEGIN SGROUP" without expecting it, we got stuck on that line, parsing it once and again, and never moving to the next one.

This patch allows the parser to move ahead, and sets the number of SGroups to the maximum possible, so that we can actually parse them. It also checks for errors when parsing the beginning of an SGroup line, so that we can exit the loop when it looks like we don´t have more of them.

Please note that the new test builds on top of #9015 (please merge that on first).

### DIFF
--- a/Code/GraphMol/FileParsers/MolFileParser.cpp
+++ b/Code/GraphMol/FileParsers/MolFileParser.cpp
@@ -3298,24 +3298,25 @@ bool ParseV3000CTAB(std::istream *inStream, unsigned int &line, RWMol *mol,
           throw FileParseException(errout.str());
         } else {
           BOOST_LOG(rdWarningLog) << errout.str() << std::endl;
+          // Prepare to read a lot of sgroups
+          nSgroups = std::numeric_limits<unsigned int>::max();
+        }
+      }
+      sgroupFound = true;
+      tempStr =
+          ParseV3000SGroupsBlock(inStream, line, nSgroups, mol, strictParsing);
+      boost::to_upper(tempStr);
+      if (tempStr.length() < 10 || tempStr.substr(0, 10) != "END SGROUP") {
+        std::ostringstream errout;
+        errout << "END SGROUP line not found on line " << line;
+        if (strictParsing) {
+          throw FileParseException(errout.str());
+        } else {
+          BOOST_LOG(rdWarningLog) << errout.str() << std::endl;
         }
       } else {
-        sgroupFound = true;
-        tempStr = ParseV3000SGroupsBlock(inStream, line, nSgroups, mol,
-                                         strictParsing);
+        tempStr = getV3000Line(inStream, line);
         boost::to_upper(tempStr);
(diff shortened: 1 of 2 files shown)


Code/GraphMol/FileParsers/testMolSupplier.cpp
```diff
@@ -2627,3 +2629,67 @@ TEST_CASE("testGitHub3517") {
...
+M  END
+$$$$)CTAB";
+
+  v2::FileParsers::SDMolSupplier sdsup;
+  v2::FileParsers::MolFileParserParams p{.strictParsing = false};
+
+  sdsup.setData(molblock, p);
+
+  std::unique_ptr<ROMol> mol;
+  auto parser_thread = std::async(std::launch::async, [&sdsup, &mol]() {
```
greglandrum, line 2673: It's nice to have it, but in general I don't think it's necessary to do this effort to catch things that run forever. If the "loops forever" bug is still there, it will be obvious when running the tests by hand and will result in the CI builds timing out.

If it does seem like something important to do, I guess we could pull this "let a function run for a while before killing it and failing" logic out into a standalone function we could use in multiple places at some point.

PR #9021, opened 2025-12-22
Allow molzip to position fragments when joining them

greglandrum:
Adds an option to molzip to adjust the coordinates of the fragments so that they are aligned along the attachment point bonds with a bond length equal to one of the attachment-point bond lengths.

No minimization is done to remove clashes, so this isn't a general solution, particularly in 3D, but it should provide a reasonable starting point.

#9013 should be merged before this one is.

### DIFF
--- a/Code/GraphMol/ChemTransforms/catch_tests.cpp
+++ b/Code/GraphMol/ChemTransforms/catch_tests.cpp
@@ -737,4 +737,22 @@ O      3.100000    0.000000    0.000000
       }
     }
   }
+  SECTION("first fragment should not move") {
+    auto m3 =
+        "[1*]c1ccccc1.[1*]c1ncccc1 |(2.57143,0,;1.07143,0,;0.321429,-1.29904,;-1.17857,-1.29904,;-1.92857,0,;-1.17857,1.29904,;0.321429,1.29904,;12.5714,0,;11.0714,0,;10.3214,1.29904,;8.82143,1.29904,;8.07143,0,;8.82143,-1.29904,;10.3214,-1.29904,)|"_smiles;
+    MolzipParams params;
+    params.alignCoordinates = true;
+    params.label = MolzipLabel::Isotope;
+    auto res = molzip(*m3, params);
+    INFO(MolToCXSmiles(*res));
+    auto resconf = res->getConformer();
+    auto molconf = m3->getConformer();
+    for (unsigned int i = 0; i < 6; ++i) {
+      auto rap = resconf.getAtomPos(i);
+      auto map = molconf.getAtomPos(i + 1);
+      CHECK_THAT(rap.x, Catch::Matchers::WithinAbs(map.x, 1e-4));
+      CHECK_THAT(rap.y, Catch::Matchers::WithinAbs(map.y, 1e-4));
+      CHECK_THAT(rap.z, Catch::Matchers::WithinAbs(map.z, 1e-4));
+    }
+  }
 }


greglandrum:
@DavidACosgrove this is the PR we discussed. Feedback welcome if you have time to take a look at it.

Code/GraphMol/ChemTransforms/MolFragmenter.h
```diff
@@ -122,7 +122,14 @@ struct RDKIT_CHEMTRANSFORMS_EXPORT MolzipParams {
   std::vector<std::string> atomSymbols;
   std::string atomProperty;
   bool enforceValenceRules = true;
-  bool generateCoordinates = false;
+  bool generateCoordinates =
+      false;  //< if true, 2D coords will be generated for
+              //  the output molecule and assigned to the input fragments as
```
DavidACosgrove: Is the //< a typo, or some extra comment markup for the docs that I don't recognise?

greglandrum: It's a doxygen thing, but I think I mis-spelled it: should probably be: ///<

Code/GraphMol/ChemTransforms/MolZip.cpp
```diff
@@ -334,6 +335,73 @@ struct ZipBond {
...
+    templateTform.SetRotation(cosT, sinT, rotnAxis);
+  } else if (cosT == 1.0) {
+    RDGeom::Point3D normal(1, 0, 0);
+    if (fabs(Us.dotProduct(normal)) == 1.0) {
+      normal = RDGeom::Point3D(0, 1, 0);
+    }
+    RDGeom::Point3D rotnAxis = Us.crossProduct(normal);
+    templateTform.SetRotation(-1, 0, rotnAxis);
+  }
+
```
DavidACosgrove, line 386: I don't think the geometry is correct here.  For example,
```
  SECTION("on top") {
    auto m3 = "[1*]c1ccccc1.[1*]c1ncccc1 |(2.57143,0,;1.07143,0,;0.321429,-1.29904,;-1.17857,-1.29904,;-1.92857,0,;-1.17857,1.29904,;0.321429,1.29904,;12.5714,0,;11.0714,0,;10.3214,1.29904,;8.82143,1.29904,;8.07143,0,;8.82143,-1.29904,;10.3214,-1.29904,)|"_smiles;
    MolzipParams params;
    params.alignCoordinates = true;
    params.label = MolzipLabel::Isotope;
    auto res = molzip(*m3, params);
    std::cout << MolToCXSmiles(*res) << std::endl;
  }
```
gives

Image

greglandrum: Yep, I missed the edge case where the vectors were co-directional. Thanks for finding that.
Should be fixed now.

DavidACosgrove: I have several decades of experience in getting this sort of thing wrong, so I know what to look out for :-).

Code/GraphMol/Wrap/MolOps.cpp
```diff
@@ -1131,30 +1128,25 @@ ROMol *copyMolSubsetHelper2(const ROMol &mol,
...
-			    python::object path,
-			    SubsetInfo &selectionInfo,
-			    const SubsetOptions &options = SubsetOptions()) {
+ROMol *copyMolSubsetHelper4(const ROMol &mol, python::object path,
+                            SubsetInfo &selectionInfo,
+                            const SubsetOptions &options = SubsetOptions()) {
   auto pathvect = pythonObjectToVect<unsigned int>(path);
-  if (!pathvect.get())
-    pathvect = std::make_unique<std::vector<unsigned int>>();
+  if (!pathvect.get()) pathvect = std::make_unique<std::vector<unsigned int>>();
```
DavidACosgrove: Maybe whilst this is being re-formatted, these could be wrapped in braces according to current house style?

Code/GraphMol/Wrap/MolOps.cpp
```diff
@@ -3008,6 +3000,10 @@ Setting this to false allows assembling chemically incorrect fragments.")
             "generateCoordinates", &MolzipParams::generateCoordinates,
             "If true will add depiction coordinates to input molecules and\n\
 zipped molecule (for molzipFragments only)")
+        .def_readwrite(
+            "alignCoordinates", &MolzipParams::alignCoordinates,
+            "if true and the input fragments have coordinates, the fragments\n\
+will be aligned along connection vectors in the output molecule")
```
DavidACosgrove, line 3010: Maybe add the ` .def("__setattr__", &safeSetattr)` thing as well?

greglandrum:
@bp-kelley do you have some time to look at this one? @DavidACosgrove has already done a review, but I'd like an LGTM from another maintainer.

bp-kelley:
@greglandrum I do indeed.  I had to implement something similar in Python for that thing we talked about at the RDKit meeting.  I have a great test case for it.  I'll be able to look at it Monday.

Code/GraphMol/ChemTransforms/MolZip.cpp
```diff
@@ -334,6 +335,70 @@ struct ZipBond {
...
+  //    Psc: sidechain connection point
+  //    Psa: sidechain attachment point
+  //    Vm: Pmc-Pma (molecular attachment vector)
+  //    Vs: Psc-Psa (sidechain attachment vector)
+  //
+  //-----------------------------------------
+  const auto Pma = conf.getAtomPos(a.getIdx());
+  const auto Psa = conf.getAtomPos(b.getIdx());
+  const auto Pmc = conf.getAtomPos(a_dummy.getIdx());
+  const auto Psc = conf.getAtomPos(b_dummy.getIdx());
```
ricrogz: These should probably be references to avoid extra copies,

ricrogz (approved):
I'm a bit rusty with this math, but this looks good to me (one minor suggestion, though, see below).

Also, sorry for the late reply, I've been (very) off for a couple of weeks.

bp-kelley:
It's hard for me to tell here, but a general use case is to leave the first fragment fixed and only allow the other fragments to be aligned, at least when we are combining RGroups to a scaffold.  The molzip convention is the first fragment is the scaffold.

bp-kelley:
It looks like this is what it is doing for sure.  I didn't see any python tests though, I'm happy to add one that tests the "fixed" first fragment if you like.

greglandrum:
> It looks like this is what it is doing for sure. I didn't see any python tests though, I'm happy to add one that tests the "fixed" first fragment if you like.

There were no changes to the python wrapper, so I just added C++ tests. 
The first two C++ tests in the current PR check the explicit coordinates of the output molecules, so they are indirectly testing that the first fragment isn't moved, but I will add an explicit test for that (for documentation, if nothing else).

greglandrum:
@bp-kelley any concerns other than the "first fragment should not move" one (implicitly tested before, now explicitly tested) or can this be merged?

PR #9022, opened 2025-12-22
Github9007

### LINKED ISSUE #9007: SynthonSpace Substructure search misses some hits
This was found by @greglandrum 

**Describe the bug**
In the synthon space
```
SMILES	synton_id	synton#	reaction_id	release
[1*]c1nc[nH]c(=O)c1[2*]	1	1	r1	1
[1*]ccc(c[2*])[N+](=O)[O-]	10	2	r1	1
```

the query
q1 = Chem.MolFromSmarts('O=c1ncnc([a])c1[a]')
produces the 1 hit expected, but
q2 = Chem.MolFromSmarts('O=c1ncnc([c])c1[c]')
doesn't.

Since q2 is a sub-query of q1 it should also find it:

Image

**To Reproduce**
```
    SynthonSpace space;
    std::istringstream iss(
        R"(SMILES	synton_id	synton#	reaction_id	release
[1*]c1nc[nH]c(=O)c1[2*]	1	1	r1	1
[1*]ccc(c[2*])[N+](=O)[O-]	10	2	r1	1
)");
    bool cancelled = false;
    space.readStream(iss, cancelled);

    auto q1 = "O=c1ncnc([a])c1[a]"_smarts;
    REQUIRE(q1);
    auto res1 = space.substructureSearch(*q1);
    REQUIRE(res1.getHitMolecules().size() == 1);
    std::cout << MolToSmiles(*res1.getHitMolecules().front()) << std::endl;
    auto q2 = "O=c1ncnc([c])c1[c]"_smarts;
    REQUIRE(q2);
    auto res2 = space.substructureSearch(*q2);
    REQUIRE(res2.getHitMolecules().size() == 1);
```

 - RDKit version: master branch
 - OS: macOS, though this is a general thing
 - Are you using conda?
 - If you are not using conda: how did you install the RDKit? build from source

**Additional context**
I'm working on it.

DavidACosgrove:
#### Reference Issue
Fixes #9007

The search was missing hits where the query was two fragments of a ring (e.g. `cc1ccccc1c`) which should have matched ring forming reactions but didn't in databases like
```
SMILES	synton_id	synton#	reaction_id	release
[1*]C=CC=C[2*]	192	1	r2	1
O=C1NC=NC([2*])=C1[1*]	227	2	r2	1
```
image

`ac1ccccc1a` did work, but for the wrong reason.  The reason the hits were being missed is that the query was fragmented into `c[1*]`, `[1*]c1ccccc1[2*]`,`[2*]c` so the two "dangling" carbon atoms were never matched together against a single synthon.  Indeed, the potential hit was only 2 synthons so no match was attempted as there were more fragments than synthons.  The solution is to identify cases where this might be an issue, and combine fragments in all combinations before doing the match.  In this case that would give 2 fragments `[1*]c1ccccc1[2*]`,`[2*]c.][1*]c` which does find the correct hit.

#### Any other comments?
There is also a bit of re-factoring.  Some code common to the original search() function and the new one with the callback was pulled out into a separate function.

### DIFF
--- a/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSubstructureSearcher.cpp
+++ b/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSubstructureSearcher.cpp
@@ -10,11 +10,9 @@
 
 #include <GraphMol/MolBundle.h>
 #include <GraphMol/MolOps.h>
-#include <GraphMol/QueryAtom.h>
 #include <GraphMol/ChemTransforms/ChemTransforms.h>
 #include <GraphMol/FileParsers/FileWriters.h>
 #include <GraphMol/SmilesParse/SmartsWrite.h>
-#include <GraphMol/SmilesParse/SmilesWrite.h>
 #include <GraphMol/SmilesParse/SmilesParse.h>
 #include <GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.h>
 #include <GraphMol/SynthonSpaceSearch/SynthonSpaceSubstructureSearcher.h>


Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
```diff
@@ -631,6 +633,16 @@ void SynthonSpace::enumerateToStream(std::ostream &os) const {
   }
 }
 
+unsigned int SynthonSpace::getMaxNumConnectors() const {
+  unsigned int maxNumConns = 0;
+  for (const auto &[id, synSet] : d_reactions) {
+    if (synSet->getConnectors().count() > maxNumConns) {
+      maxNumConns = synSet->getConnectors().count();
+    }
```
greglandrum:
```suggestion
    maxNumConns = std::max(maxNumCons, synSet->getConnectors().count());
    
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.cpp
```diff
@@ -539,8 +539,7 @@ std::vector<std::vector<std::unique_ptr<ROMol>>> splitMolecule(
   if (maxNumFrags < 1) {
     maxNumFrags = 1;
   }
-  maxNumFrags =
-      std::min({maxNumFrags, MAX_CONNECTOR_NUM, query.getNumBonds() + 1});
+  maxNumFrags = std::min({maxNumFrags, query.getNumBonds() + 1});
```
greglandrum: I think you can just do:
```suggestion
  maxNumFrags = std::min(maxNumFrags, query.getNumBonds() + 1);
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -245,6 +229,27 @@ void processReactions(
...
+}
+
+std::vector<std::unique_ptr<SynthonSpaceHitSet>>
+SynthonSpaceSearcher::assembleHitSets(const TimePoint *endTime, bool &timedOut,
+                                      std::uint64_t &totHits) {
+  auto fragments = details::splitMolecule(
+      d_query, getNumQueryFragmentsRequired(), d_params.maxNumFragSets, endTime,
+      d_params.numThreads, timedOut);
+  if (timedOut || ControlCHandler::getGotSignal()) {
+    return std::vector<std::unique_ptr<SynthonSpaceHitSet>>();
```
greglandrum: I think you can do:
```suggestion
    return {};
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -245,6 +229,27 @@ void processReactions(
...
+  }
+  extraSearchSetup(fragments);
+  if (ControlCHandler::getGotSignal()) {
+    return std::vector<std::unique_ptr<SynthonSpaceHitSet>>();
```
greglandrum: same suggestion here:
```suggestion
    return {};
```

DavidACosgrove:
@greglandrum I'm afraid I've had to change the PR quite a lot.  The CI checks failed on a bug that should have shown up before.  It's a bit simpler now, if anything.  I have moved the merging of ring-forming fragments into extraSearchSetup, which means there is now no need to merge the pattern FPs as they are generated after that in extraSearchSetup.  The fragment merging will only happen once now, rather than once per reaction searched.

greglandrum:
> @greglandrum I'm afraid I've had to change the PR quite a lot. 

😨 
I will take another look tomorrow

DavidACosgrove:
> tomorrow

Yeah, sorry. It is actually the same as before but a bit less of it and in a different place.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSubstructureSearcher.cpp
```diff
@@ -15,6 +15,7 @@
 #include <GraphMol/FileParsers/FileWriters.h>
 #include <GraphMol/SmilesParse/SmartsWrite.h>
 #include <GraphMol/SmilesParse/SmilesWrite.h>
+#include <GraphMol/SmilesParse/SmilesParse.h>
```
greglandrum, line 16: I don't think you need this

DavidACosgrove: Can you live with it for now?  If I attend to it over the next couple of days it might be the last PR I ever do!

greglandrum: > Can you live with it for now?  If I attend to it over the next couple of days it might be the last PR I ever do!

Can't have that!

PR #9025, opened 2025-12-24
Fixes #9020

### LINKED ISSUE #9020: Hs labeled implicit or explicit depending on presence of 3D conformer. Chem.RemoveHs differs from MolFromMolBlock removeHs
Parsing a MolBlock with real hydrogens (as individual atoms) but without 3D coordinates results in all hydrogens being labeled as implicit.

The behavior when parsing a MolBlock with 3D positions differs for the following:
 1. `MolFromMolBlock()` with `removeHs=True`
 2. `MolFromMolBlock()` with `removeHs=False` followed by `Chem.RemoveHs()`

Method 1 produces explciit Hs, method 2 does not.

The labeling of hydrogens as explicit or implicit affects the outcome of a reaction.

```python
from rdkit import Chem
import rdkit
print(f"{rdkit.__version__=}")

def count_hs(mol):
    nr_impl = 0
    nr_expl = 0
    for atom in mol.GetAtoms():
        nr_impl += atom.GetNumImplicitHs()
        nr_expl += atom.GetNumExplicitHs()
    return f"{nr_impl=}, {nr_expl=}"


def molblock_roundtrip(mol, removeHs):
    molblock = Chem.MolToMolBlock(mol)
    mol = Chem.MolFromMolBlock(molblock, removeHs=removeHs)
    return mol

def does_enol_sanitize(mol):
    keto_to_enol = "[CX4:1][CX3:2]=[OX1:3]>>[CX4:1]=[CX3:2]-[OX1:3][#1]"
    rxn = rdChemReactions.ReactionFromSmarts(keto_to_enol)
    product_list = rxn.RunReactants((mol,))
    try:
        for product in product_list:
            Chem.SanitizeMol(product[0])
        return True
    except:
        return False

mol_1d = Chem.MolFromSmiles("CC(=O)")
mol_1d = Chem.AddHs(mol_1d)
mol_3d = Chem.Mol(mol_1d)
rdDistGeom.EmbedMolecule(mol_3d)

mol_1d_parser_removes_Hs = molblock_roundtrip(mol_1d, removeHs=True)
mol_3d_parser_removes_Hs = molblock_roundtrip(mol_3d, removeHs=True)
tmp = molblock_roundtrip(mol_3d, removeHs=False)
mol_3d_remove_Hs_after = Chem.RemoveHs(tmp)

print(f"No 3D positions, parser removed Hs:")
print(f"  - {count_hs(mol_1d_parser_removes_Hs)}")
print(f"  - can sanitize enol product? {does_enol_sanitize(mol_1d_parser_removes_Hs)}")
print()
print(f"With 3D positions, parser removed Hs:")
print(f"  - {count_hs(mol_3d_parser_removes_Hs)}")
print(f"  - can sanitize enol product? {does_enol_sanitize(mol_3d_parser_removes_Hs)}")
print()
print(f"With 3D positions, parser keeps Hs, Chem.RemoveHs() called after:")
print(f"  - {count_hs(mol_3d_remove_Hs_after)}")
print(f"  - can sanitize enol product? {does_enol_sanitize(mol_3d_remove_Hs_after)}")
```

Here's the output I get

```
rdkit.__version__='2025.09.3'
No 3D positions, parser removed Hs:
  - nr_impl=4, nr_expl=0
  - can sanitize enol product? True

With 3D positions, parser removed Hs:
  - nr_impl=1, nr_expl=3
[18:19:53] Explicit valence for atom # 0 C, 5, is greater than permitted
  - can sanitize enol product? False

With 3D positions, parser keeps Hs, Chem.RemoveHs() called after:
  - nr_impl=4, nr_expl=0
  - can sanitize enol product? True
```

greglandrum:
This ended up being simpler than I was afraid it might be.

When 3d coordinates are set, the call to `assignChiralTypesFrom3D()` in the mol file parsing code marks any four-coordinate atom that might be a chiral center as chiral.
`RemoveHs()` sees this and marks removed Hs as explicit instead of implicit.

The change here is mark removed Hs on "chiral" centers that have more than one H removed as implicit instead of explicit.

I would love to be able to snap my fingers and remove the stupid distinction we currently make between "explicit" implicit Hs and "implicit" implicit Hs, but doing so would no doubt break lots of other peoples' code in hard-to-detect ways. <sigh>

### DIFF
--- a/Code/GraphMol/FileParsers/molfile_stereo_catch.cpp
+++ b/Code/GraphMol/FileParsers/molfile_stereo_catch.cpp
@@ -752,4 +752,39 @@ M  END)CTAB";
       CHECK(mol3->getAtomWithIdx(0)->getNumImplicitHs() == 0);
     }
   }
+  SECTION("not chiral, one H") {
+    auto mb3 = R"CTAB(
+     RDKit          3D
+
+  6  5  0  0  0  0  0  0  0  0999 V2000
+   -0.3698    0.0026    0.0028 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.8980   -0.5748   -0.1191 O   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.6741   -0.1194    1.0508 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3142    1.0665   -0.3155 F   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.0830   -0.5246   -0.6430 F   0  0  0  0  0  0  0  0  0  0  0  0
+    1.5432    0.1497    0.0240 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0
+  1  3  1  0
+  1  4  1  0
+  1  5  1  0
+  2  6  1  0
+M  END)CTAB";
+    {
+      auto mol3 = v2::FileParsers::MolFromMolBlock(mb3);
+      REQUIRE(mol3);
+      CHECK(mol3->getAtomWithIdx(0)->getNumExplicitHs() == 0);
+      CHECK(mol3->getAtomWithIdx(0)->getNumImplicitHs() == 1);
+    }
+    {
+      v2::FileParsers::MolFileParserParams params;
+      params.removeHs = false;
+      auto mol3 = v2::FileParsers::MolFromMolBlock(mb3, params);
+      REQUIRE(mol3);
+      CHECK(mol3->getAtomWithIdx(0)->getNumExplicitHs() == 0);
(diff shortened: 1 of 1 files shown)


bp-kelley (approved):

bp-kelley:
This is remarkably subtle.  RDKit 3000 won't have this issue :)

greglandrum:
> This LGTM. Although I wonder if we should remove the "chiral" mark from atoms that have more than one explicit/implicit H.

I thought about this, but decided that's really one of those changes that belongs to the stereochem assignment code (where it currently lives).

ricrogz:
> I thought about this, but decided that's really one of those changes that belongs to the stereochem assignment code (where it currently lives).

Ok, makes sense.

PR #9036, opened 2025-12-31
Fixes #9033

### LINKED ISSUE #9033: Tversky for zero prototype string with alpha=1 and beta=0 returns incorrect output
If a prototype contains at least one bit 1 everything works fine. However, if a prototype has all 0 bits the calculated similarity is 1, while I would expect 0. Below is an example.

```python
from rdkit import DataStructs

def make_bitvector(bits):
    bv = DataStructs.ExplicitBitVect(len(bits))
    for i, b in enumerate(bits):
        if b:
            bv.SetBit(i)
    return bv

a = 1
b = 0

v = DataStructs.TverskySimilarity(make_bitvector([0,1,1,0]), make_bitvector([0,1,1,1]), a, b)
print(v)

v = DataStructs.TverskySimilarity(make_bitvector([1,0,1,0]), make_bitvector([0,1,1,1]), a, b)
print(v)

v = DataStructs.TverskySimilarity(make_bitvector([1,0,1,1]), make_bitvector([0,1,1,1]), a, b)
print(v)

v = DataStructs.TverskySimilarity(make_bitvector([1,0,0,0]), make_bitvector([0,1,1,1]), a, b)
print(v)

v = DataStructs.TverskySimilarity(make_bitvector([0,0,0,0]), make_bitvector([0,1,1,1]), a, b)
print(v)
```

output
```
1.0
0.5
0.6666666666666666
0.0
1.0   <- here I would expect 0
```

 - RDKit version: 2025.03.5
 - OS: Ubuntu 24.04
 - Python version (if relevant): 3.11.13
 - Are you using conda? yes

greglandrum:
Also fixes the other similarity metrics.
A very small amount of refactoring

### DIFF
--- a/Code/DataStructs/BitOps.cpp
+++ b/Code/DataStructs/BitOps.cpp
@@ -290,7 +290,7 @@ double TanimotoSimilarity(const T1 &bv1, const T2 &bv2) {
   }
   unsigned int total = bv1.getNumOnBits() + bv2.getNumOnBits();
   if (total == 0) {
-    return 1.0;
+    return 0.0;
   }
   unsigned int common = NumOnBitsInCommon(bv1, bv2);
   return (double)common / (double)(total - common);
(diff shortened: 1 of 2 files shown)


d-b-w:
This means that [0, 0, 0...] has 0 self-similarity! That's a surprising outcome for me. Is it a bug?

bp-kelley:
I think it’s fine, the intersection of two empty sets is an empty set

d-b-w:
That means that a molecule with an empty fingerprint doesn't match itself. I think that's a bit weird!

bp-kelley:
It’s an edge case for sure, but an empty molecule has no meaning IMO just as an empty set has no meaning.  That being said, I’m happy with either solution.

but mathematically you get 0/(0+0 - 0) so the answer is really undefined.  I would posit it’s really an error.  But keeping it as 1 probably keeps it as a proper metric so maybe that is the right solution but should be noted in the docs

PR #9051, opened 2026-01-14
cleanup of stereogroups and wedges for non-chiral sites

tadhurst-cdd:
#### Reference Issue
If a wedge bond starts on an atom that is not chiral and not part of an atropisomer bond, it should be cleaned up.  Also, if there is an enhanced stereo marker on an atom that is not chiral and not part of an atropisomer bond, it should be cleaned up.

The routine that reapplies the original wedging to a molecule now has a flag to make it NOT add an original wedge if the atom is not chiral nor part of an atropisomer.

Also if the routine that cleans up chirality (removes chirality markers from non chiral atoms) also cleans up StereoGroups if changes were made, and the routine that cleans up atropisomers (removes the markers from bonds that are not atropisomers) also cleans up the atropisomer bonds referenced in a StereoGroup


greglandrum (approved):
LGTM

greglandrum:
Will merge after the CI builds go through

PR #9052, opened 2026-01-15
Prevent CIP labels of bonds from being calculated twice.

ricrogz:
I just noticed a couple of bugs in #8582 (or rather, the same bug twice):

In the second run of the CIP calculation, we check stereo centers to see if they have already been labeled:
https://github.com/rdkit/rdkit/blob/4b92c2fa8c41410191cceae6f469b4b9fb980d2b/Code/GraphMol/CIPLabeler/CIPLabeler.cpp#L203-L206

But this check is wrong. It will not check stereo bonds or atropisomer bonds, since `conf->getFocus()` is always an atom. Fortunately, the bonds that were labeled twice are the ones that resolve quickly...

Also, while working on this, I noticed that the call were we reset CIP codes before the calculation also incurred on the same error: we only deleted stereo labels on atoms.


greglandrum (approved):
LGTM

PR #9056, opened 2026-01-16
Store CIP-ranked anchors after CIP labeling.

ricrogz:
This revisits https://github.com/rdkit/rdkit/pull/7825, adding atropisomers and fixing a couple of issues in the original PR. 

The most important change is acknowledging that, surprisingly, implicit Hs are not guaranteed to always have the lowest CIP ranking: according to the revised rule 2 in John Mayfield et al's paper on which the algorithm is based, when it comes to CIP ranking, H > 1H (!!!), which is something very awkward we have to take into account when building CIP-ranked lists of anchors, as skipping the implicit Hs in the lists might completely change the result of whatever we are doing with the list.

The CIP-ranked anchors will be stored under the new "_CIPNeighborRanks" property.

For tetrahedral stereo, indexes of the 4 neighboring atoms (exceptionally 3, in cases of pseudo-tetrahedral centers, e.g. trigonal pyramyds) will be stored. For Sp2 and atropisomer bonds, we will store 2 indexes, corresponding to the highest-ranking anchor on each side of the bond.

Note that, as mentioned above, these lists may include implicit hydrogens: 1 in tetrahedral configurations, or up to 2 in Sp2 and atropisomers, which, due to being implicit and therefore not having `RDKit::Atom` representation in the molecule's graph of which we can get an index, will be represented by the value of `std::numeric_limits<unsigned int>::max()`.

### DIFF
--- a/Code/GraphMol/Chirality.h
+++ b/Code/GraphMol/Chirality.h
@@ -118,11 +118,13 @@ struct RDKIT_GRAPHMOL_EXPORT StereoInfo {
   // REVIEW: absolute stereo data member?
 
   // used to mark missing atoms
-  inline static constexpr unsigned int NOATOM = Atom::NOATOM;
+  [[deprecated(
+      "please use Atom::NOATOM")]] inline static constexpr unsigned int NOATOM =
+      Atom::NOATOM;
 
   StereoType type = StereoType::Unspecified;
   StereoSpecified specified = StereoSpecified::Unspecified;
-  unsigned centeredOn = NOATOM;
+  unsigned centeredOn = Atom::NOATOM;
   StereoDescriptor descriptor = StereoDescriptor::None;
   unsigned permutation = 0;  // for the non-tetrahedral stereo cases
   std::vector<unsigned> controllingAtoms;  // all atoms around the atom or bond.
--- a/Code/GraphMol/Chirality.cpp
+++ b/Code/GraphMol/Chirality.cpp
@@ -1487,7 +1487,8 @@ bool atomIsCandidateForRingStereochem(
           ringNbrRanks.insert(atomRanks[nbr->getIdx()]);
         }
       }
-      // std::cerr << "!!!! " << atom->getIdx() << " " << ringNbrRanks.size() << " "
+      // std::cerr << "!!!! " << atom->getIdx() << " " << ringNbrRanks.size() <<
+      // " "
       //           << ringNbrs.size() << " " << nonRingNbrs.size() << std::endl;
       switch (nonRingNbrs.size()) {
         case 2:
@@ -1986,10 +1987,10 @@ void assignBondCisTrans(ROMol &mol, const StereoInfo &sinfo) {
   if (sinfo.type != StereoType::Bond_Double ||
       sinfo.specified != StereoSpecified::Unspecified ||
       sinfo.controllingAtoms.size() != 4 ||
-      ((sinfo.controllingAtoms[0] == StereoInfo::NOATOM &&
-        sinfo.controllingAtoms[1] == StereoInfo::NOATOM) ||
-       (sinfo.controllingAtoms[2] == StereoInfo::NOATOM &&
-        sinfo.controllingAtoms[3] == StereoInfo::NOATOM))) {
+      ((sinfo.controllingAtoms[0] == Atom::NOATOM &&
+        sinfo.controllingAtoms[1] == Atom::NOATOM) ||
+       (sinfo.controllingAtoms[2] == Atom::NOATOM &&
+        sinfo.controllingAtoms[3] == Atom::NOATOM))) {
     return;
   }
 
@@ -2003,7 +2004,7 @@ void assignBondCisTrans(ROMol &mol, const StereoInfo &sinfo) {
   if (begDir != Bond::BondDir::ENDDOWNRIGHT &&
       begDir != Bond::BondDir::ENDUPRIGHT) {
     begFirstNeighbor = false;
-    if (sinfo.controllingAtoms[1] != StereoInfo::NOATOM) {
+    if (sinfo.controllingAtoms[1] != Atom::NOATOM) {
       begBond = mol.getBondBetweenAtoms(dblBond->getBeginAtomIdx(),
                                         sinfo.controllingAtoms[1]);
       CHECK_INVARIANT(begBond, "no initial bond found");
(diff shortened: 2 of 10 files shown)


Code/GraphMol/CIPLabeler/configs/Configuration.h
```diff
@@ -24,6 +24,10 @@ class Bond;
 
 namespace CIPLabeler {
 
+// used to represent the index of implicit Hs
+inline constexpr unsigned int IMPLICITH =
```
d-b-w, line 28: This feels like a reasonable compromise.

I wonder if it needs to be exposed more broadly somehow.

Code/GraphMol/CIPLabeler/configs/Tetrahedral.cpp
```diff
@@ -133,14 +139,30 @@ Descriptor Tetrahedral::label(Node *node, const Rules &comp) const {
   // it must be different than the representation of the implicit H.
   auto ordered = std::vector<Atom *>(4, nullptr);
   int idx = 0;
+  d_ranked_anchors.reserve(4);
   for (const auto &edge : edges) {
     if (edge->getEnd()->isSet(Node::BOND_DUPLICATE) ||
         edge->getEnd()->isSet(Node::IMPL_HYDROGEN)) {
       continue;
     }
```
d-b-w, line 152: Is it possible to reach here with an implicit hydrogen? I see the `isSet(Node::IMPL_HYDROGEN) .. continue` above.

Code/GraphMol/CIPLabeler/catch_tests.cpp
```diff
@@ -21,20 +21,22 @@
...
 #include <GraphMol/SmilesParse/SmilesWrite.h>
 #include <GraphMol/FileParsers/FileParsers.h>
-#include <GraphMol/MarvinParse/MarvinParser.h>
-#include <GraphMol/test_fixtures.h>
 #include <GraphMol/Substruct/SubstructMatch.h>
 #include <GraphMol/test_fixtures.h>
 
-#include <RDGeneral/BoostStartInclude.h>
-#include <boost/algorithm/string.hpp>
 #include <RDGeneral/BoostEndInclude.h>
```
d-b-w, line 38: feels like BoostStartInclude/EndInclude should wrap the boost header, they don't after this re-order.

Code/GraphMol/CIPLabeler/configs/Tetrahedral.cpp
ricrogz: Good catch. In fact, in the case of chirality, we don't care about implicit Hs: since the implicit H is not physically there, we can't use its position to determine if neighbors are distributed CW/CCW around the chiral atom. So it gets ignored in the calculation of ordered vs carriers.

(Or that's what the code does, at least!)

Code/GraphMol/Chirality.h
```diff
@@ -116,12 +116,10 @@ enum class StereoSpecified {
...
   // REVIEW: absolute stereo data member?
-#ifdef _MSC_VER
-  static const unsigned NOATOM =
-      std::numeric_limits<unsigned>::max();  // used to mark missing atoms
-#else
-  static const unsigned NOATOM;  // used to mark missing atoms
-#endif
+
+  // used to mark missing atoms
+  inline static constexpr unsigned int NOATOM = Atom::NOATOM;
```
greglandrum: How about we mark this as `[[deprecated]]` and use Atom::NOATOM in the code directly?
We can then remove this in a few releases.
This will also need to be documented in the ReleaseNotes.md

greglandrum:
I can maybe see how having this information might be useful, but I think it's very likely to lead to confusion.

Here's an example showing why:
```
In [13]: m = Chem.MolFromSmiles('C[C@H](F)O[C@H](C)CC')

In [14]: rdCIPLabeler.AssignCIPLabels(m)

In [15]: list(m.GetAtomWithIdx(1).GetPropsAsDict()['_CIPNeighborRanks'])
Out[15]: [2, 3, 0]

In [16]: list(m.GetAtomWithIdx(4).GetPropsAsDict()['_CIPNeighborRanks'])
Out[16]: [3, 6, 5]

In [17]: m2 = Chem.MolFromSmiles('C[C@H](F)O[C@H](CC)C')

In [18]: rdCIPLabeler.AssignCIPLabels(m2)

In [19]: list(m2.GetAtomWithIdx(4).GetPropsAsDict()['_CIPNeighborRanks'])
Out[19]: [3, 5, 7]
```
Assuming I'm interpreting the results correctly (and if I'm not, then it's a further indication of how this is confusing), the O atom is rank 0 when we consider the first chiral center, but it's rank 3 when we consider the second one. In both cases it's the lowest rank neighbor, but the actual ranks are different.

Even if we document the hell out this and provide the explanation of why it's true, I suspect that many, many people are going to end up confused here.

Two possible suggestions:
1) rename the property to something like `_LocalCIPNeighborRanks` in the hopes that will cause people to pause and think.
2) rename the property to `_RelativeCIPNeighborRanks` and make the ranks relative. So the output of lines 15 and 16 above would be `[1,2,0]` and `[0,2,1]`, respectively.

ricrogz:
Renaming the property sound better to me. I have to admit I was more focused on the implementation than on giving the property a meaningful name. Would `_CIPPrioritizedAnchors` sound ok ? `_LocalCIPNeighborRanks` sounds like we are returning the Ranks instead of indexes.

Also, consider we are returning different things for chiral atoms and stereo/atropisomer bonds: 
- For chiral atoms, we return the indexes of neighboring atoms ordered by CIP priority.
- For stereo and atropisomer bonds we return 2 indexes, the ones of the "anchor" on each side of the bond with the highest CIP ranking.

greglandrum:
> * For chiral atoms, we return the indexes of neighboring atoms ordered by CIP priority.

In that case, I had totally misunderstood the output.

> Would _CIPPrioritizedAnchors sound ok ? _LocalCIPNeighborRanks sounds like we are returning the Ranks instead of indexes.

What about `_CIPNeighborOrder`?

ricrogz:
> What about `_CIPNeighborOrder`?

Sounds good.

greglandrum (changes requested):
one small suggestion.

Code/GraphMol/Chirality.h
ricrogz: Ok

greglandrum (approved):
LGTM

PR #9058, opened 2026-01-19
Move some more tests over to catch2

greglandrum:
Mostly this is switching more tests over from our old testing framework to catch2, but there's some modernization in there as well.

The majority of the work here was done using copilot


Code/GraphMol/Subgraphs/CMakeLists.txt
```diff
@@ -8,6 +8,6 @@ rdkit_headers(Subgraphs.h
 
 target_include_directories(Subgraphs PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/External/boost-numeric-bindings>)
 
-rdkit_test(testSubgraphs1 test1.cpp LINK_LIBRARIES Subgraphs SmilesParse)
+rdkit_catch_test(testSubgraphs1 test1.cpp LINK_LIBRARIES Subgraphs SmilesParse)
```
ricrogz: While you are here, would you mind renaming "test1.cpp" to something more easily identificable, like "testSubgraphs1.cpp" ?  Having a bunch of tests named "test.cpp" and similar makes it harder to find the source file for tests when they fail...

Code/GraphMol/Substruct/CMakeLists.txt
```diff
@@ -7,6 +7,6 @@ target_compile_definitions(SubstructMatch PRIVATE RDKIT_SUBSTRUCTMATCH_BUILD)
 rdkit_headers(SubstructMatch.h
               SubstructUtils.h DEST GraphMol/Substruct)
 
-rdkit_test(testSubstructMatch test1.cpp LINK_LIBRARIES FileParsers SmilesParse SubstructMatch)
+rdkit_catch_test(testSubstructMatch test1.cpp LINK_LIBRARIES FileParsers SmilesParse SubstructMatch)
```
ricrogz: Same as the other CMakefile:  can "test1.cpp" here be renamed to something like "testSubstructMatch.cpp" ?

Code/GraphMol/RGroupDecomposition/testRGroupDecomp.cpp
```diff
@@ -102,83 +89,121 @@ void DUMP_RGROUP(RGroupRows::const_iterator &it, std::string &result) {
...
+    // All Cl's should be labeled with the same rgroup
+    int i = 0;
+    for (RGroupRows::const_iterator it = rows.begin(); it != rows.end();
+         ++it, ++i) {
+      CHECK_RGROUP(it, "Core:c1ccc([*:1])cc1 R1:Cl[*:1]", mols[i].get());
+    }
+    delete core;
   }
-  delete core;
 }
```
ricrogz: This is mostly a duplicate of the section above (only `scoreMethod` and a `REQUIRE` vs `CATCH` differ). Maybe we could use a generator (https://github.com/catchorg/Catch2/blob/devel/docs/generators.md) or abstract the code into a function to avoid  the duplication.

Code/GraphMol/Substruct/test1.cpp
```diff
@@ -205,350 +180,244 @@ void test3() {
...
-  CHECK_INVARIANT(matchV.size() == 3, "");
+  REQUIRE(found);
+  REQUIRE(matchV.size() == 3);
   std::vector<MatchVectType> matches;
-  for (int i = 0; i < 30000; i++) {
-    n = SubstructMatch(*m, *q1, matches, true, true);
-    CHECK_INVARIANT(n == 1, "");
-    CHECK_INVARIANT(matches[0].size() == 3, "");
-    if (!(i % 500)) std::cout << i << std::endl;
+  for (int i = 0; i < 300000; i++) {
```
ricrogz: Do we need to repeat the matching that many times? should we reduce this number?

Code/GraphMol/Substruct/test1.cpp
```diff
@@ -712,900 +561,723 @@ void testMultiThread() {
...
   std::vector<std::future<void>> tg;
-  ROMol *query = SmartsToMol("[#6;$([#6]([#6])[!#6])]");
+  auto query = v2::SmilesParse::MolFromSmarts("[#6;$([#6]([#6])[!#6])]");
   boost::dynamic_bitset<> hits(mols.size());
   for (unsigned int i = 0; i < mols.size(); ++i) {
     MatchVectType matchV;
     hits[i] = SubstructMatch(*mols[i], *query, matchV);
   }
   unsigned int count = 4;
 #if 1
```
ricrogz: Maybe drop this `#if 1` while you are here.

Code/GraphMol/Substruct/test1.cpp
```diff
@@ -712,900 +561,723 @@ void testMultiThread() {
...
 }
 
-void testGithubIssue1489() {
-  BOOST_LOG(rdErrorLog) << "-------------------------------------" << std::endl;
-  BOOST_LOG(rdErrorLog) << "    Testing Github #1389: Substructure matching "
-                           "with chirality failure when query is built from "
-                           "SMARTS"
-                        << std::endl;
+TEST_CASE("testGithubIssue1489", "[substruct][github][issue1489]") {
 #if 1
```
ricrogz: Maybe drop this while you are here.

Code/GraphMol/Substruct/test1.cpp
greglandrum: Interesting... the AI added the extra zero there.

ricrogz: Maybe it wanted to test this more thoroughly :stuck_out_tongue_closed_eyes:

PR #9059, opened 2026-01-20
New RDMol flat-array backend

rietmann-nv:
This PR is more than a usual feature or bugfix, so I'll try to recap the goals.

The current RDKit has performance issues, partly due to excess allocations due to the object-based graph structure when creating and modifying molecules. Nia Dickson (formerly @ NVIDIA) started a new backend that is backed by a number of flat arrays that dramatically lowers allocations as well as has potential to improve caching and other performance factors.

The new backend has two major contributions:
1. A new RDMol molecule that directly uses the new array-based backend.
2. Compatibility layer between ROMol and RWMol to RDMol to maintain compatibility with existing code.

A [porting guide](https://github.com/rietmann-nv/rdkit/blob/minimal_rdmol_merge_master_rdkit_fixed/Docs/Code/PortingToRDMol.md) was also provided to port code using RO or RWMol to using RDMol, but existing functionality should be maintained.

As Nia Dickson wrote most of this port, and myself (Max Rietmann) and Matthias Noack provided some additional help to fix the final test failures to get this ready for a PR, we will do our best to work with the RDKit community to get this new backend into master.


Code/GraphMol/Atom.h
```diff
@@ -69,50 +73,81 @@ class AtomMonomerInfo;
...
+    CHI_SQUAREPLANAR = AtomEnums::ChiralType::CHI_SQUAREPLANAR,
+    //!< trigonal bipyramidal, use permutation flag
+    CHI_TRIGONALBIPYRAMIDAL = AtomEnums::ChiralType::CHI_TRIGONALBIPYRAMIDAL,
+    //!< octahedral, use permutation flag
+    CHI_OCTAHEDRAL = AtomEnums::ChiralType::CHI_OCTAHEDRAL
   };
 
+  // A "using" statement works for ValenceType, because it's an enum class,
+  // so needs ValenceType::IMPLICIT instead of just IMPLICIT, for example.
+  using ValenceType = AtomData::ValenceType;
```
greglandrum: This is a change to the public API.
It also breaks the SWIG builds.
I'd suggest reverting it... I will include this in my PR that gets the SWIG builds working

Code/GraphMol/RDMol.h
```diff
@@ -0,0 +1,2294 @@
...
+  MolType *d_mol;
+
+  friend class RDKit::RDMol;
+  IndexRange(MolType *mol) : d_mol(mol) {}
+
+ public:
+  IndexRange(const IndexRange<ISBOND, ISCONST> &) = default;
+  IndexRange &operator=(const IndexRange<ISBOND, ISCONST> &) = default;
+  Iter begin() const { return Iter(d_mol, 0); }
+  Iter end() const { return Iter(d_mol, d_mol->getNumAtoms()); }
```
d-b-w: This should be `getNumBonds()` if ISBOND (constexpr).

Code/GraphMol/RDMol.h
```diff
@@ -0,0 +1,2294 @@
...
+  uint32_t getWriteID(uint32_t groupIndex) const {
+    URANGE_CHECK(groupIndex, writeIds.size());
+    return writeIds[groupIndex];
+  }
+  void setWriteID(uint32_t groupIndex, uint32_t value) {
+    URANGE_CHECK(groupIndex, writeIds.size());
+    writeIds[groupIndex] = value;
+  }
+
+  std::pair<uint32_t*, uint32_t*> getAtoms(uint32_t groupIndex) {
```
d-b-w: I think this should be a `std::span<>`. That would allow ready use in range-based algorithms, and remove the requirement to write our own data structure

Code/GraphMol/RDMol.h
```diff
@@ -0,0 +1,2294 @@
...
+      if (ringBegins[ring + 1] - ringBegins[ring] == size) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  uint32_t minAtomRingSize(atomindex_t atomIndex) const {
+    uint32_t minSize = std::numeric_limits<uint32_t>::max();
+    const uint32_t* atomMembershipBegin =
```
d-b-w: feels like we want a helper method that returns a std::span from an iterator and two offsets. If we pretend `make_span()` exists, this function becomes:

    auto rings = make_span(atomMemberships.begin(), atomMembershipBegins[atomIndex], atomMembershipBegins[atomIndex + 1]);
    return std::ranges::min(rings, [&ringBegins](ring){return ringBegins[ring + 1] - ringBegins[ring];});

Code/GraphMol/RDMol.h
```diff
@@ -0,0 +1,2294 @@
...
+        minSize = size;
+      }
+    }
+    return minSize;
+  }
+
+  void initFusedInfoFromBondMemberships();
+};
+
+struct RDKIT_GRAPHMOL_EXPORT StereoGroups {
```
d-b-w, line 461: Stereogroups are quite rare - what profiling lead to the decision to include stereogroups in the SOA refactor?

d-b-w: Stereogroups are quite rare - what profiling lead to the decision to include stereogroups in the struct of arrays (SOA) refactor?

Code/GraphMol/RDMol.h
```diff
@@ -0,0 +1,2294 @@
...
+    return atomBondStarts.data();
+  }
+  const uint32_t* getOtherAtomIndices() const {
+    return otherAtomIndices.data();
+  }
+  const uint32_t* getBondDataIndices() const {
+    return bondDataIndices.data();
+  }
+
+  // For RDKit use only
```
d-b-w, line 1311: Does this mean for RDMol use only?

Code/GraphMol/RDMol.h
```diff
@@ -0,0 +1,2294 @@
...
+        }
+      }
+      if (!erased) {
+        ++it;
+      }
+    }
+  }
+
+  //! Clears all props for a given bond
+  void clearSingleBondAllProps(const std::uint32_t bondIndex, const bool onlyComputed = false) {
```
d-b-w: "single bond" has a different meaning in rdk. Maybe:

    clearPropsForBond
    clearPropsForBondWithIdx

(and rename the Atom, too)

d-b-w (commented):
I read through RDMol.h. I don't have much to say yet about the overall design and interaction with the rest of RDKit, that's gonna take a few more read-throughs. My comments are mostly about local things.

I do have a few general questions-

* Struct-of-Arrays (SOA) incurs significant bookkeeping complexity relative to AOS. What performance numbers do we have to justify that additional complexity?
* What do those performance gains look like cross-platform? 
* Are they visible to Python callers?
* This adds SOA to several independent layers - at least atom props, atoms, bonds, stereogroups. Do we want SOA for all of these?
* There are C++ STL implementations of some features here (std::iota, std::span, std::ranges::views come to mind). Were those avoided for performance reasons?

Schrödinger uses SOA for our internal Molecule data structure. But I've definitely seen some of the drawbacks of that approach, so I want us to be clear-eyed about making this transition.

rietmann-nv:
> I read through RDMol.h. I don't have much to say yet about the overall design and interaction with the rest of RDKit, that's gonna take a few more read-throughs. My comments are mostly about local things.
> 
> I do have a few general questions-
> 
> * Struct-of-Arrays (SOA) incurs significant bookkeeping complexity relative to AOS. What performance numbers do we have to justify that additional complexity?
> * What do those performance gains look like cross-platform?
> * Are they visible to Python callers?
> * This adds SOA to several independent layers - at least atom props, atoms, bonds, stereogroups. Do we want SOA for all of these?
> * There are C++ STL implementations of some features here (std::iota, std::span, std::ranges::views come to mind). Were those avoided for performance reasons?
> 
> Schrödinger uses SOA for our internal Molecule data structure. But I've definitely seen some of the drawbacks of that approach, so I want us to be clear-eyed about making this transition.

Thanks for all your comments. We don't have good performance data yet. Greg and I are porting some of the most used routines to use RDMol directly instead of the compatibility interface and will be able to make a better assessment of the performance impact of the new implementation over the old. The main reason for going with flat arrays in a SOA architecture was to reduce allocations, which were considered a driving factor in runtime cost for many common operations. Unfortunately, we don't yet have any hard timing numbers to back this assumption yet, but we should have some comparative benchmarks soon.

I will have to get back to you to justify why a SOA vs AOS structure would be better. In the HPC space, SOA typically produces faster code due to improved caching and vectorization, but RDKit operations are quite a bit different than most HPC codebases, so your point is well made, particularly regarding code complexity.

d-b-w:
> I will have to get back to you to justify why a SOA vs AOS structure would be better. In the HPC space, SOA typically produces faster code due to improved caching and vectorization...

I meant that RDKit current has AOS - a data structure of pointers to `Atom` objects. We use SOA internally in performance-critical spots, but it _can_ make keeping class invariants... invariant more laborious.

PR #9087, opened 2026-02-05
fix the python docs for CanonicalRankAtomsInFragment

greglandrum:
The docstring explanation of the `bondsToUse` argument was wrong.

some clang-format reformatting of rdmolfiles.cpp also is in here


ricrogz (approved):
LGTM

PR #9089, opened 2026-02-05
Don't silently ignore missing  or extra atoms/bonds in copyMolSubset

### LINKED ISSUE #9084: CopyMolSubset silently rewires bonds when bondIndices reference atoms not in atomIndices
**Describe the bug**
When using the new `CopyMolSubset` with explicit `atomIndices` and `bondIndices`, if a selected bond references an endpoint atom that is not included in `atomIndices`, this can silently produce an incorrect molecule by mapping the missing endpoint atom to subset atom at index 0. This seems like undesirable behavior.

It looks like this arises in `copySelectedAtomsAndBonds` in `Subset.cpp`:

```cpp
extracted_bond->setBeginAtomIdx(atomMapping[ref_bond->getBeginAtomIdx()]);
extracted_bond->setEndAtomIdx(atomMapping[ref_bond->getEndAtomIdx()]);
```

If an endpoint atom wasn’t selected, atomMapping[...] inserts the missing key with a default value (0 for unsigned int). That silently maps missing endpoints to subset atom 0.

The behavior is slightly different if the other bond endpoint also happens to be subset atom 0, or if both atom endpoints aren't included - in this case it attempts to create a self-bond (0 to 0), raising a pre-condition violation. This isn't silently incorrect, but still is not the most informative error message.

Relevant PR: https://github.com/rdkit/rdkit/pull/8811

**To Reproduce**
Here, `atomIndices` specifies a subset that skips a single atom (5). But `bondIndices` still includes the two bonds to it (4 and 5). In the resulting subset, those bonds are retained but are now connected to atom 0.

```python
>>> mol = Chem.MolFromSmiles("Oc1cccc(O)c1")
>>> sub = rdmolops.CopyMolSubset(mol, [0, 1, 2, 3, 4, 6, 7], [0, 1, 2, 3, 4, 5, 6, 7])
>>> Chem.MolToSmiles(sub)
'OO12:cccc1c:2'
>>> mol.Debug()
Atoms:
	0 8 O chg: 0  deg: 1 exp: 1 imp: 1 hyb: SP2
	1 6 C chg: 0  deg: 3 exp: 4 imp: 0 hyb: SP2 arom?: 1
	2 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
	3 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
	4 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
	5 6 C chg: 0  deg: 3 exp: 4 imp: 0 hyb: SP2 arom?: 1
	6 8 O chg: 0  deg: 1 exp: 1 imp: 1 hyb: SP2
	7 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
Bonds:
	0 0->1 order: 1 conj?: 1
	1 1->2 order: a conj?: 1 aromatic?: 1
	2 2->3 order: a conj?: 1 aromatic?: 1
	3 3->4 order: a conj?: 1 aromatic?: 1
	4 4->5 order: a conj?: 1 aromatic?: 1
	5 5->6 order: 1 conj?: 1
	6 5->7 order: a conj?: 1 aromatic?: 1
	7 7->1 order: a conj?: 1 aromatic?: 1
>>> sub.Debug()
Atoms:
	0 8 O chg: 0  deg: 4 exp: 1 imp: 1 hyb: SP2
	1 6 C chg: 0  deg: 3 exp: 4 imp: 0 hyb: SP2 arom?: 1
	2 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
	3 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
	4 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
	5 8 O chg: 0  deg: 1 exp: 1 imp: 1 hyb: SP2
	6 6 C chg: 0  deg: 2 exp: 3 imp: 1 hyb: SP2 arom?: 1
Bonds:
	0 0->1 order: 1 conj?: 1
	1 1->2 order: a conj?: 1 aromatic?: 1
	2 2->3 order: a conj?: 1 aromatic?: 1
	3 3->4 order: a conj?: 1 aromatic?: 1
	4 4->0 order: a conj?: 1 aromatic?: 1
	5 0->5 order: 1 conj?: 1
	6 0->6 order: a conj?: 1 aromatic?: 1
	7 6->1 order: a conj?: 1 aromatic?: 1
```

**Expected behavior**
Possible alternative behaviors:
1. Raise an error indicating that selected bonds reference atoms not in the selected atom set
2. Automatically extend the selected atom set to include all endpoints of selected bonds
3. Drop bonds whose atom endpoints are not selected

Option 1 might the the least likely to produce surprising/unexpected behavior (and maybe easiest to implement?). Option 2 is handled to some extent by the `path` variations of `CopyMolSubset` (implicitly including atoms from specified bonds). Option 3 seems less desirable to me.

 - RDKit version: 2025.09.5
 - OS: Rocky9
 - Are you using conda? No
 - If you are not using conda: how did you install the RDKit? Source install

### LINKED ISSUE #9088: CopyMolSubset ignores bondIndices when all atoms are included
**Describe the bug**
When calling `CopyMolSubset()` with both atom indices and bond indices, the `bondIndices` parameter is ignored if all atoms in the molecule are selected in `atomIndices`. This appears to be due to an optimization in `Subset.cpp` that returns a copy of the entire molecule when `atoms.size() == natoms` and `method == BONDS_BETWEEN_ATOMS` (the default).

**To Reproduce**
```python

# Select all atoms but no bonds
subset = Chem.CopyMolSubset(mol, [0, 1, 2], [])
print(f"Subset: {subset.GetNumAtoms()} atoms, {subset.GetNumBonds()} bonds")
# Expected: 3 atoms, 0 bonds
# Actual:   3 atoms, 2 bonds (entire molecule is copied)
```

Workaround: Setting `opts.method = rdmolops.SubsetMethod.BONDS` avoids triggering the optimization:

```python
opts = Chem.SubsetOptions()
opts.method = Chem.SubsetMethod.BONDS
subset = Chem.CopyMolSubset(mol, [0, 1, 2], [], opts)
print(f"Subset: {subset.GetNumAtoms()} atoms, {subset.GetNumBonds()} bonds")
# Now correctly outputs: 3 atoms, 0 bonds
```

This is because the optimization now switches to check `bonds.size() == nbonds`, and when all bonds are included in `bondIndices`, all atoms should always be included in `atomIndices` (and this will be enforced with a fix to #9084), so it is correct to return a copy of the original molecule.

**Expected behavior**
- `CopyMolSubset` should respect both `atomIndices` and `bondIndices` parameters, even when `atomIndices` contains all atoms in the molecule
- `SubsetOptions.method` should not affect the result in any way when both atom and bond selections are explicitly provided. (docstring claims this)

Suggested fix: In `Subset.cpp`, replace this:

```cpp
  if ((atoms.size() == natoms &&
       options.method == SubsetMethod::BONDS_BETWEEN_ATOMS) ||
      (bonds.size() == nbonds && options.method == SubsetMethod::BONDS) ||
      (atoms.size() == natoms && bonds.size() == nbonds)) {
    // optimization to copy the entire thing
    return std::make_unique<RDKit::RWMol>(mol);
  }
```

with

```cpp
  if (atoms.size() == natoms && bonds.size() == nbonds) {
    // optimization to copy the entire thing
    return std::make_unique<RDKit::RWMol>(mol);
  }
```

 - RDKit version: 2025.09.5
 - OS: Rocky Linux 9.5
 - Python version (if relevant): 3.10.7
 - Are you using conda? No
 - If you are not using conda: how did you install the RDKit? Source install

bp-kelley:
Fixes #9084
Fixes #9088

### DIFF
--- a/Code/GraphMol/Subset.cpp
+++ b/Code/GraphMol/Subset.cpp
@@ -298,10 +298,7 @@ std::unique_ptr<RDKit::RWMol> copyMolSubset(
     const SubsetOptions &options) {
   const auto natoms = mol.getNumAtoms();
   const auto nbonds = mol.getNumBonds();
-  if ((atoms.size() == natoms && bonds.size() == 0 &&
-       options.method == SubsetMethod::BONDS_BETWEEN_ATOMS) ||
-      (bonds.size() == nbonds && options.method == SubsetMethod::BONDS) ||
-      (atoms.size() == natoms && bonds.size() == nbonds)) {
+  if ((atoms.size() == natoms && bonds.size() == nbonds)) {
     // optimization to copy the entire thing
     return std::make_unique<RDKit::RWMol>(mol);
   }
--- a/Code/GraphMol/catch_molops.cpp
+++ b/Code/GraphMol/catch_molops.cpp
@@ -644,6 +644,13 @@ TEST_CASE("test_manual_atom_bond_subset", "[copyMolSubset]") {
     CHECK(m2->getNumBonds() == 2);
   }
   {
+    std::vector<unsigned int> atoms = {0,1,2};
+    std::vector<unsigned int> bonds = {};
+    auto m2 = copyMolSubset(*m, atoms, bonds);
+    CHECK(m2->getNumAtoms() == 3);
+    CHECK(m2->getNumBonds() == 0);
+  }
+  {
     std::vector<unsigned int> atoms = {0,1,2,3};
     std::vector<unsigned int> bonds = {0,1};
     REQUIRE_THROWS_AS(copyMolSubset(*m, atoms, bonds), IndexErrorException);


Code/GraphMol/Subset.cpp
```diff
@@ -294,8 +298,7 @@ std::unique_ptr<RDKit::RWMol> copyMolSubset(
     const SubsetOptions &options) {
   const auto natoms = mol.getNumAtoms();
   const auto nbonds = mol.getNumBonds();
-
-  if ((atoms.size() == natoms &&
+  if ((atoms.size() == natoms && bonds.size() == 0 &&
```
mcs07: At first glance this doesn't seem right to me. My instinct is that explicitly providing an empty bonds list (`bonds.size() == 0`) means don't include any bonds, not "infer bonds between atoms" like the path-based API.

It also seems to me that `options.method` shouldn't have any impact on the behavior of this function, so shouldn't be a factor in this check at all?

Seeing as the optimization is only valid when all atoms and all bonds are specified, can't the whole check here be replaced with just `atoms.size() == natoms && bonds.size() == nbonds`?

An example test for this could be:

```cpp
{
  std::vector<unsigned int> atoms = {0,1,2};  // all atoms
  std::vector<unsigned int> bonds = {};        // explicitly empty
  auto m2 = copyMolSubset(*m, atoms, bonds);
  CHECK(m2->getNumAtoms() == 3);
  CHECK(m2->getNumBonds() == 0);  // Right now I think this PR returns 2 here
}
```

bp-kelley: Done, that's way easier.

bp-kelley: And correct to boot.

mcs07 (approved):
Thanks, this looks great to me.

Tangentially related - just wanted to flag that I noticed that`SubsetOptions.conformerIdx` exists and is documented, but doesn't seem to actually be implemented.

PR #9092, opened 2026-02-06
The precondition in Bond::getOtherAtomIdx() is redundant

ricrogz:
The precondition tests exactly the same thing as the `if`/ `else if` in the function. So why check twice?

### DIFF
--- a/Code/GraphMol/Bond.cpp
+++ b/Code/GraphMol/Bond.cpp
@@ -85,7 +85,7 @@ unsigned int Bond::getOtherAtomIdx(const unsigned int thisIdx) const {
   }
   // This "precondition" would check exactly the same that is checked
   // above, but no need to be redundant, so just throw.
-  PRECONDITION(false, "bad index");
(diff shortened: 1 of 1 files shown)


Code/GraphMol/Bond.cpp
```diff
@@ -78,15 +78,14 @@ void Bond::setOwningMol(ROMol *other) {
...
   if (d_beginAtomIdx == thisIdx) {
     return d_endAtomIdx;
   } else if (d_endAtomIdx == thisIdx) {
     return d_beginAtomIdx;
   }
-  // we cannot actually get down here
-  return 0;
+  // This "precondition" would check exactly the same that is checked
+  // above, but no need to be redundant, so just throw.
+  PRECONDITION(false, "bad index");
```
greglandrum:
```suggestion
  POSTCONDITION(false, "bad index");
```

greglandrum (approved):
LGTM

PR #9093, opened 2026-02-06
WIP: MonomerMol, MonomerLibrary & Atomistic <-> Monomeric conversions

rachelnwalker:
This is a WIP and still needs a lot of cleaning up, posting to share & determine what we need to change to be compatible with SCSR implementation.


Code/GraphMol/MonomerMol/MonomerMol.h
```diff
@@ -0,0 +1,243 @@
...
+RDKIT_MONOMERMOL_EXPORT unsigned int getResidueNumber(const Atom* atom);
+
+//! MonomerMol is a molecule class for monomeric representations
+/*!
+    MonomerMol inherits from ROMol and provides methods for working with
+    monomeric molecules where atoms represent monomers and bonds represent
+    linkages between monomers. In the future I would like it to inherit from
+    RWMol instead, but that will require a bit more work to handle deletions.
+*/
+class RDKIT_MONOMERMOL_EXPORT MonomerMol : public ROMol {
```
PatrickPenner, line 82: I'm catching up on a few design decisions. There has surely been a lot of discussion on inheriting from R*Mol that I'm not aware of, but my initial reaction, and reading the tests where monomers are treated as atoms, make me feel like this is not a very intuitive way to model this.

I reached out to you on discord. I would love to talk in some depth with you about this on your preferred platform, if you can spare the time.

Code/GraphMol/MonomerMol/MonomerLibrary.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+// Static member definitions
+bool MonomerLibrary::s_useGlobalLibrary = true;
+std::unique_ptr<MonomerLibrary> MonomerLibrary::s_globalLibrary;
+std::once_flag MonomerLibrary::s_globalLibraryOnce;
+
+// Built-in monomer definitions (symbol -> SMILES)
+namespace {
+const std::unordered_map<std::string, std::string> builtin_monomer_data({
+    {"A", "C[C@H](N[H:1])C(=O)[OH:2] |atomProp:0.pdbName. CB :1.pdbName. CA "
```
PatrickPenner, line 33: I'm thinking about whether there is ever a case where we may need an OR operator in the monomer definition. An alternative way to solve that kind of problem is to enumerate all OR-ed versions and have a mechanism of mixtures on a higher level.

Code/GraphMol/MonomerMol/MonomerMol.h
greglandrum: > I'm catching up on a few design decisions. There has surely been a lot of discussion on inheriting from R*Mol that I'm not aware of, but my initial reaction, and reading the tests where monomers are treated as atoms, make me feel like this is not a very intuitive way to model this.
> 

There are multiple potential ways of thinking about this particular problem, each of them has advantages.

My two cents about a primary motivation for this particular design (I didn't do the design, but I do think it will be useful) and having a MonomerMol that is derived from an RWMol is to be able to take advantage of the rest of the RDKit code that knows how to work with RWMols without having to do additional work. For example: we'll be able to take advantage of all of the substructure searching machinery using MonomerMols as the queries and molecules, these will directly plug into the drawing code, etc.

PatrickPenner: I understand the argument of reusing existing code, but specifically in the substructure search example it's the graph both data structures have in common that's important. If we abstract at the level of the graph, we can probably still use existing substructure search algorithms by only reworking the interfaces a bit.

I think it will be repeatedly uncomfortable to explain to users that atoms are monomers sometimes. It feels like a weird legacy thing we are creating right now for expediency that will continue to be confusing.

There's also not a lot of interface beautification we can do about this. We can add wrappers called things like GetMonomer in addition to GetAtom, not really solving the confusion problem, but we also can't hide the mol interface, because then we can't reuse code. I guess we could just name the python interface functions something different, which would be a fun kind of confusing in its own way.

Inheriting from mol feels like pushing a lot of problems onto users that we could fix here (for example with a more loosely coupled Macromolecule class).

jenniferrbryant92548-jpg:
****
If@@ -22,6 +22,8 @@

PR #9096, opened 2026-02-06
Fix SMARTS for the strict definition of rotatable bonds

brje01:
#### Reference Issue
No open issue to reference.

The current SMARTS pattern for the strict definition for the number of rotatable bonds in a molecule excludes e.g. if either side of the bonds is C(F)(F)F, but includes CH3 when Hs are present in the molecular graph. 
Here's an example:
```
In [3]: rdMolDescriptors.CalcNumRotatableBonds(Chem.MolFromSmiles('CCCC'))
Out[3]: 1

In [4]: rdMolDescriptors.CalcNumRotatableBonds(Chem.AddHs(Chem.MolFromSmiles('CCCC')))
Out[4]: 3
```
The change in the SMARTS excludes the bond if there is CH3 on either side. 

#### Any other comments?
None.

### DIFF
--- a/Code/GraphMol/Descriptors/catch_tests.cpp
+++ b/Code/GraphMol/Descriptors/catch_tests.cpp
@@ -195,9 +195,8 @@ TEST_CASE(
     "molecule") {
   SECTION("basics") {
     auto m1 = "c1ccccc1c1ccc(CCC)cc1"_smiles;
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

greglandrum:
Thanks @brje01 
I will merge this after the CI builds finish.

PR #9100, opened 2026-02-10
Fix modern stereo performance issue for large structures

rachelnwalker:
#### Reference Issue
Our internal performance tests found that modern stereo took a very long time for large, ringy structures:

```
from rdkit import Chem
import time

mol = Chem.MolFromPDBFile("1FSK.pdb", removeHs=False, sanitize=False)

# turn on modern stereo

start = time.time()
Chem.AssignStereochemistry(mol, force=True, cleanIt=True)
end = time.time()
print(f"Time taken for AssignStereochemistry: {end - start:.2f} seconds")
```
Before: 13.04s
After these changes: 0.37s

I ran a profile on this and saw that the vast vast majority of time was spent in `compareRingAtomsConcerningNumNeighbors`, which is doing a quite a bit of `std::fill`-ing and `for (unsigned i = 0; i < nAtoms; ++i) {` loops. I asked claude to help refactor the function to track only the necessary indices, and it came up with this solution.

#### Other comments
Marking this as a draft because (1) I haven't done further performance tests and (2) claude wrote the majority of this so I still want to review closer.

### DIFF
--- a/Code/DataStructs/Wrap/DataStructs.cpp
+++ b/Code/DataStructs/Wrap/DataStructs.cpp
@@ -10,6 +10,10 @@
 //
 #define PY_ARRAY_UNIQUE_SYMBOL rddatastructs_array_API
 
+#ifdef RDK_BUILD_THREADSAFE_SSS
+#include <mutex>
+#endif
+
 #include <RDBoost/python.h>
 #include <RDBoost/Wrap.h>
 #include <DataStructs/BitVects.h>
@@ -32,9 +36,26 @@ void wrap_realValVect();
 void wrap_sparseIntVect();
 void wrap_FPB();
 
+#ifdef RDK_BUILD_THREADSAFE_SSS
+static std::once_flag s_ds_numpy_init_flag;
+#endif
+
+static void ds_ensure_numpy() {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+  std::call_once(s_ds_numpy_init_flag, rdkit_import_array);
+#else
+  static bool initialized = false;
+  if (!initialized) {
+    initialized = true;
+    rdkit_import_array();
+  }
+#endif
+}
+
 namespace {
 template <typename T, typename U>
 void converter(const T &v, python::object destArray, U func) {
+  ds_ensure_numpy();
   if (!PyArray_Check(destArray.ptr())) {
     throw_value_error("Expecting a Numeric array object");
   }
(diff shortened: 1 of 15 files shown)


Code/GraphMol/new_canon.cpp
```diff
@@ -168,10 +174,17 @@ void compareRingAtomsConcerningNumNeighbors(Canon::canon_atom *atoms,
...
+    // Sparse reset from previous atom's BFS
+    for (int i : visitedIndices) {
+      visited[i] = 0;
+    }
+    visitedIndices.clear();
+    for (int i : lastLevelIndices) {
+      lastLevelNbrs[i] = 0;
+    }
+    // currentLevelNbrs and revisitedNeighbors are reset within the BFS loop
+    lastLevelIndices.clear();
```
ricrogz, line 186: Instead of adding a second layer of vectors, is it possible to refactor `visited`, `lastLevelNbrs` and `currentLevelNbrs` from bitset-style to list-of-integers, and `revisitedNeighbors` into a map ?

rachelnwalker: Yeah this seems like a better approach, I think claude's original attempt was not great for smaller systems (I am see an fmcs timeout in the test failures)

ricrogz: Oh, the FMCS timeout happens from time to time. Might not be related to this change.

greglandrum: Yeah, the FMCS timeout is a thing that happens sometimes in the CI builds. You can ignore it.

greglandrum:
FWIW: I just compared the runtime of hanoitest.cpp:test7() on master with this PR.
That test does a bunch of calls to MolToSmiles(), which uses the canonicalization code, on small molecules.
It looks like things get a bit faster with this PR, even for the small molecules.

ricrogz:
How does performance look for this last version, @rachelnwalker ?

rachelnwalker:
> How does performance look for this last version, @rachelnwalker ?

@ricrogz a bit worse, 0.49s.

ricrogz:
> > How does performance look for this last version, @rachelnwalker ?
> 
> @ricrogz a bit worse, 0.49s.

Oh, ok. I thought this might be an improvement over the vectors due to allocating fewer variables and these being smaller, but it sounds like it's the other way around.

rachelnwalker:
> Oh, ok. I thought this might be an improvement over the vectors due to allocating fewer variables and these being smaller, but it sounds like it's the other way around.

Thanks for the suggestion though -- I reverted back to the original implementation.

ricrogz:
> Thanks for the suggestion though -- I reverted back to the original implementation.

Thanks for trying it out, I just wanted to see how that one performed.

Issue #9094, opened 2026-02-06
ChemDraw fails to compile with newer LLVM/Clang versions


reneeotten:
**Describe the bug**
Newer Clang versions add `-Wundef-prefix=TARGET_OS_` as default compiler argument and there are several instances in the ChemDraw code where things are not defined.  Click [here](https://build.macports.org/builders/ports-12_x86_64-builder/builds/149506/steps/install-port/logs/stdio) for an example build log on macOS 12, where "Clang 17.0.6" is used.

When trying to get around the error (by adding `-Wno-undef-prefix`) I noticed that [CMAKE_C_FLAGS flags are being set](https://github.com/rdkit/rdkit/blob/dfb34ec23c48899558a9fd12fbf51ba14627a76a/External/ChemDraw/CMakeLists.txt#L121-L127), whereas the code is C++ and these are thus not used. Instead one should likely use `CMAKE_CXX_FLAGS`; if I patch it do so then the flag is added and the build proceeds.

Of course ideally the code should be fixed so that the errors are gone and the additional flag to tell Clang to ignore the error is no longer needed.

**To Reproduce**
Compile the code with LLVM/Clang (not AppleClang) higher than version 11

**Expected behavior**
Code should compile

**Screenshots**

**Configuration (please complete the following information):**
 - RDKit version: 2025.09.5
 - OS: macOS (for example version 12, where AppleClang does not support the CXX20 standard and MacPorts installs an appropriate compiler (in this case LLVM/Clang 17).
 - Python version (if relevant): 3.14
 - Are you using conda? No
 - If you are not using conda: how did you install the RDKit? Through MacPorts

bp-kelley:
This does look like an oversight, the CXX flags should indeed be set, not the C flags.

bp-kelley:
I'm not able to reproduce this with condo-clang 18 at least, and I don't have access to macOS 12 which looks to not be updated anymore.  I'll try with Mac ports.  The C Flags should be CXX, but this doesn't seem to change the compilation much in any case.

bp-kelley:
You'll need to provide more details of your build environment, I can't duplicate the issue with llvm-18.  However, I do note that catch2 has compile issues:

````
/Users/brian/devel/rdkit/mac-build/_deps/catch2-src/src/catch2/catch_approx.cpp:73:24: error: identifier '_a' preceded by whitespace in a literal operator declaration is
      deprecated [-Werror,-Wdeprecated-literal-operator]
[...]

PR #9108, opened 2026-02-12
Use CXX flags not C flags in ChemDraw cmake config

bp-kelley:
Simple fix, this is intended to help with #9094 but I can't duplicate the issues on MacOS 15

### DIFF
--- a/External/ChemDraw/CMakeLists.txt
+++ b/External/ChemDraw/CMakeLists.txt
@@ -119,7 +119,6 @@ if(RDK_BUILD_CHEMDRAW_SUPPORT)
   endif()
 
   if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-    string(REPLACE " -Werror" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


bp-kelley:
@greglandrum if it builds, I'll just merge it.  It's just change _C_FLAG to _CXX_FLAGS like it should have in the first place.

greglandrum:
> @greglandrum if it builds, I'll just merge it. It's just change _C_FLAG to _CXX_FLAGS like it should have in the first place.

It failed...

bp-kelley:
Yep, I think the _C_ flags were accidentally protecting us from all the changes in LLVM compiler warning flags :)

I'm not a big fan of checking which warning flags work on a given day, I might just switch to -Wnoerror and just see the fur fly.

greglandrum:
> 'm not a big fan of checking which warning flags work on a given day, I might just switch to -Wnoerror and just see the fur fly.

Please do not do this. I think having `-Werror` set on the CI builds is a *significant* benefit.

bp-kelley:
this is JUST for the ChemDraw external build, btw.

greglandrum (approved):
LGTM

PR #9110, opened 2026-02-12
Cannot build JS bindings with MCS option enabled.

papillot:
#### Reference Issue
Trying to compile the MinimalLib WASM with the MCS option I encountered the following beep
```
394.7 [ 98%] Building CXX object Code/MinimalLib/CMakeFiles/RDKit_minimal.dir/minilib.cpp.o
397.6 /src/rdkit/Code/MinimalLib/minilib.cpp:893:19: error: no matching member function for call to 'push_back'
397.6   893 |       smartsArray.push_back(pair.first);
397.6       |       ~~~~~~~~~~~~^~~~~~~~~
397.6 /opt/boost/include/boost/json/array.hpp:1441:5: note: candidate function not viable: no known conversion from 'const std::string' to 'const value' for 1st argument
397.6  1441 |     push_back(value const& v);
397.6       |     ^         ~~~~~~~~~~~~~~
397.6 /opt/boost/include/boost/json/array.hpp:1465:5: note: candidate function not viable: no known conversion from 'const std::string' to 'value' for 1st argument
397.6  1465 |     push_back(value&& v);
397.6       |     ^         ~~~~~~~~~
397.6 /opt/boost/include/boost/json/array.hpp:1744:5: note: candidate function not viable: no known conversion from 'const std::string' to 'pilfered<value>' for 1st argument
397.6  1744 |     push_back(
397.6       |     ^
397.6  1745 |         pilfered<value> pv);
397.6       |         ~~~~~~~~~~~~~~~~~~
398.3 1 error generated.
398.3 make[3]: *** [Code/MinimalLib/CMakeFiles/RDKit_minimal.dir/build.make:92: Code/MinimalLib/CMakeFiles/RDKit_minimal.dir/minilib.cpp.o] Error 1
```

`boost::json::array.push_back` expects a `value`. This fix casts the std::string to a boost::json::value

#### Any other comments?
With this change the minimallib compilation succeeds (tested on MacOS with arm64 architecture, with the `RDK_BUILD_MINIMAL_LIB_MCS` flag, using the provided docker based script)

### DIFF
--- a/Code/MinimalLib/jswrapper.cpp
+++ b/Code/MinimalLib/jswrapper.cpp
@@ -452,8 +452,8 @@ emscripten::val get_mmpa_frags_helper(const JSMolBase &self,
                                       unsigned int maxCutBonds) {
   auto obj = emscripten::val::object();
   auto pairs = self.get_mmpa_frags(minCuts, maxCuts, maxCutBonds);
-  obj.set("cores", pairs.first);
-  obj.set("sidechains", pairs.second);
+  obj.set("cores", emscripten::val(pairs.first, emscripten::allow_raw_pointers()));
+  obj.set("sidechains", emscripten::val(pairs.second, emscripten::allow_raw_pointers()));
   return obj;
 }
 #endif


Code/MinimalLib/minilib.cpp
```diff
@@ -890,7 +890,7 @@ std::string get_mcs_as_json(const JSMolList &molList,
   if (!mcsResult.DegenerateSmartsQueryMolDict.empty()) {
     bj::array smartsArray;
     for (const auto &pair : mcsResult.DegenerateSmartsQueryMolDict) {
-      smartsArray.push_back(pair.first);
+      smartsArray.push_back(bj::value(pair.first));
```
greglandrum: I think you can just do this:
```suggestion
      smartsArray.emplace_back(pair.first);
```

greglandrum (changes requested):
@papillot Thanks for the contribution.
I've made a small suggested change.

While you're in this PR, can you please also change the file Dockerfile_3_rdkit_build so that it builds the MCS and molzip code by default? We might as well turn those on there so that the code gets tested in the CI builds.

papillot:
> @papillot Thanks for the contribution. I've made a small suggested change.
> 
> While you're in this PR, can you please also change the file Dockerfile_3_rdkit_build so that it builds the MCS and molzip code by default? We might as well turn those on there so that the code gets tested in the CI builds.

Thanks @greglandrum  for the feedback!

Regarding the changes to the Dockerfile, I have 2 observations:
- This will in effect override the options that are set-up from the main CMakeLists.txt. Is this desirable? https://github.com/rdkit/rdkit/blob/2a79115e9ce98def2a85d5aa9e6c1a2ff6ae2ddf/CMakeLists.txt#L76-L77
- There is also `RDK_BUILD_MINIMAL_LIB_MMPA` flag that is not exposed from the main CMakeList.txt. Shall I add it as well (either as an option in that file, or as an override in the DockerFile)?

greglandrum:
> > @papillot Thanks for the contribution. I've made a small suggested change.
> > While you're in this PR, can you please also change the file Dockerfile_3_rdkit_build so that it builds the MCS and molzip code by default? We might as well turn those on there so that the code gets tested in the CI builds.
> 
> Thanks @greglandrum for the feedback!
> 
> Regarding the changes to the Dockerfile, I have 2 observations:
> 
> * This will in effect override the options that are set-up from the main CMakeLists.txt. Is this desirable? 

Yes, I believe so. The primary use of these dockerfiles is for the CI builds and I think we want those to use/test all of the options. If someone wants to do partial local builds, they can do so.

> * There is also `RDK_BUILD_MINIMAL_LIB_MMPA` flag that is not exposed from the main CMakeList.txt. Shall I add it as well (either as an option in that file, or as an override in the DockerFile)?

Yes please.

papillot:
It was worth it: another issue found while compiling for MMPA support. For the record, here is the stack trace
```
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/wire.h:124:19: error: static assertion failed due to requirement '!std::is_pointer<JSMolList *>::value': Implicitly binding raw pointers is illegal.  Specify allow_raw_pointer<arg<?>>
352.5   124 |     static_assert(!std::is_pointer<T*>::value, "Implicitly binding raw pointers is illegal.  Specify allow_raw_pointer<arg<?>>");
352.5       |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/wire.h:147:26: note: in instantiation of template class 'emscripten::internal::TypeID<JSMolList *>' requested here
352.5   147 | struct TypeID<const T> : TypeID<T> {
352.5       |                          ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/wire.h:151:21: note: in instantiation of template class 'emscripten::internal::TypeID<JSMolList *const>' requested here
352.5   151 | struct TypeID<T&> : TypeID<T> {
352.5       |                     ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/wire.h:240:45: note: in instantiation of template class 'emscripten::internal::TypeID<JSMolList *const &>' requested here
352.5   240 |         static constexpr TYPEID types[] = { TypeID<Args>::get()... };
352.5       |                                             ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/wire.h:263:16: note: in instantiation of member function 'emscripten::internal::ArgArrayGetter<emscripten::internal::TypeList<emscripten::val, JSMolList *const &>>::get' requested here
352.5   263 |             >::get();
352.5       |                ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/val.h:601:86: note: in instantiation of member function 'emscripten::internal::WithPolicies<>::ArgTypeList<emscripten::val, JSMolList *const &>::getTypes' requested here
352.5   601 |     thread_local EM_INVOKER mc = _emval_create_invoker(argTypes.getCount(), argTypes.getTypes(), Kind);
352.5       |                                                                                      ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/val.h:353:20: note: in instantiation of function template specialization 'emscripten::val::internalCallWithPolicy<emscripten::internal::EM_INVOKER_KIND::CAST, emscripten::internal::WithPolicies<>, emscripten::val, JSMolList *const &>' requested here
352.5   353 |     new (this) val(internalCallWithPolicy<EM_INVOKER_KIND::CAST, WithPolicies<Policies...>, val>(nullptr, nullptr, std::forward<T>(value)));
352.5       |                    ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/val.h:629:12: note: in instantiation of function template specialization 'emscripten::val::val<JSMolList *const &>' requested here
352.5   629 |     return val(v, policies...);
352.5       |            ^
352.5 /opt/emsdk/upstream/emscripten/cache/sysroot/include/emscripten/val.h:488:74: note: in instantiation of function template specialization 'emscripten::val::val_ref<JSMolList *>' requested here
352.5   488 |     internal::_emval_set_property(as_handle(), val_ref(key).as_handle(), val_ref(value, policies...).as_handle());
352.5       |                                                                          ^
352.5 /src/rdkit/Code/MinimalLib/jswrapper.cpp:455:7: note: in instantiation of function template specialization 'emscripten::val::set<char[6], JSMolList *>' requested here
352.5   455 |   obj.set("cores", pairs.first);
352.5       |       ^
352.7 1 error generated.
352.7 make[3]: *** [Code/MinimalLib/CMakeFiles/RDKit_minimal.dir/build.make:77: Code/MinimalLib/CMakeFiles/RDKit_minimal.dir/jswrapper.cpp.o] Error 1
```

Also, from the logs I've noticed that chemdraw related libraries were taking quite some time even if they are not exposed in MinimalLib, so I've disabled them in the Dockerfile which seems to reduce the compilation time on my machine (-10%).

greglandrum:
Thanks @papillot !

Issue #9112, opened 2026-02-13
[pr plan]  remove `RDKit::Dict::getData` for a strict abstraction boundary


bddap:
In pursuit of speed-wins, I'm hoping to play with the internal data representation of `RDKit::Dict`. In order to keep the actual optimization prs small and digestible, I think it best to start with the following preparatory work:

`RDKit::Dict` [exposes its internal data structure](https://github.com/rdkit/rdkit/blob/4b92c2fa8c41410191cceae6f469b4b9fb980d2b/Code/RDGeneral/RDProps.h#L36) and some external callers access that structure directly. To allow changes to that internal representation, I'd like to remove the get `getData` method and replace it with interfaces that hide the underlying structure. I will, of course, modify any code inside the repo that calls `getData` to use the new interface.

Removing `getData` will technically be a change to the `c++` api, but not to other bindings. Hopefully that is acceptable; otherwise this optimization won't be possible.

I'm starting on a pr now. Your thoughts on the above plan are welcome. Do you have a stance on breaking changes to the c++ api?

bp-kelley:
It’s leaky because some ops (like serialization) need to get the the key, rdvalue pairs.

It feels like RDProps should be templates on dict for different implementations, or we need a new api point to support the serialization from rdvalues

bddap:
> or we need a new api point to support the serialization from rdvalues

I think that is the way since I'm not trying to make the dict implementation configurable.

PR #9113, opened 2026-02-14
Hide data representation inside RDKit::Dict

bddap:
implements https://github.com/rdkit/rdkit/issues/9112

### DIFF
--- a/Code/RDGeneral/testDict.cpp
+++ b/Code/RDGeneral/testDict.cpp
@@ -674,9 +674,9 @@ TEST_CASE("custom AnyTag data is destroyed through Dict lifecycle") {
     d.setVal<Bar>("mybar", b);
     REQUIRE(d.getVal<Bar>("mybar").x == 42);
   }
-  REQUIRE(count > 0);
+  REQUIRE(count == 3);
 
-  int prev = count;
+  count = 0;
   {
     Dict d;
     Bar b(7, &count);
@@ -684,7 +684,7 @@ TEST_CASE("custom AnyTag data is destroyed through Dict lifecycle") {
     Dict d2(d);
     REQUIRE(d2.getVal<Bar>("mybar").x == 7);
   }
-  REQUIRE(count > prev);
+  REQUIRE(count == 4);
 }
 
 TEST_CASE("testGithub2910") {


Code/RDGeneral/StreamOps.h
```diff
@@ -638,12 +630,11 @@ inline unsigned int streamReadProps(std::istream &ss, RDProps &props,
   if (reset) {
     dict.reset();  // Clear data before repopulating
   }
-  auto startSz = dict.getData().size();
-  dict.getData().resize(startSz + count);
   for (unsigned index = 0; index < count; ++index) {
-    CHECK_INVARIANT(streamReadProp(ss, dict.getData()[startSz + index],
-                                   dict.getNonPODStatus(), handlers),
+    Dict::Pair pair;
```
bp-kelley: This is likely to be much less performant since we don't have the call to "resize".

Code/RDGeneral/StreamOps.h
```diff
@@ -614,7 +607,6 @@ inline bool streamReadProp(std::istream &ss, Dict::Pair &pair,
       for (auto &handler : handlers) {
         if (propType == handler->getPropName()) {
           handler->read(ss, pair.val);
-          dictHasNonPOD = true;
           return true;
```
bp-kelley, line 610: You don't have CustomTag in the needsCleanup above.

Code/RDGeneral/StreamOps.h
bp-kelley: Perhaps read the Pairs into std::vector<Pair> as before and then we can hide the internal implementation behind something like:

```
void RDProps::append(std::vector<Pair>&& vec) {
    data_.insert(
        data_.end(),
        std::make_move_iterator(vec.begin()),
        std::make_move_iterator(vec.end())
    );
}
```

Code/RDGeneral/StreamOps.h
bp-kelley: After review, this might be converted to an AnyTag, we should add a test for this now that we don't have the dictHasNonPod anymore.

bp-kelley: See FooHandler in tesDict.cpp

Code/RDGeneral/Dict.h
```diff
@@ -136,14 +136,34 @@ class RDKIT_RDGENERAL_EXPORT Dict {
...
+  void appendPair(Pair &&pair) {
+    if (pair.val.needsCleanup()) {
+      _hasNonPodData = true;
+    }
+    _data.push_back(std::move(pair));
+  }
+
+  //! \brief Returns a const reference to the raw RDValue for a key.
+  //! Throws KeyErrorException if the key is not found.
+  const RDValue &getRawVal(const std::string_view what) const {
```
bp-kelley: Not a big change, but I'd be happier with "getRDValue"

Code/RDGeneral/StreamOps.h
bp-kelley: See note below

bp-kelley:
Nice work, I think it's a good idea to hide the implementation.  I have a few suggestions and the Custom handlers need to be explicitly tested so that we are sure they are destroyed.

Code/RDGeneral/Dict.h
```diff
@@ -136,14 +136,47 @@ class RDKIT_RDGENERAL_EXPORT Dict {
...
+  std::size_t size() const { return _data.size(); }
 
-  //----------------------------------------------------------
-  //! \brief Access to the underlying data.
-  const DataType &getData() const { return _data; }
-  DataType &getData() { return _data; }
+  //! \brief Returns whether the dictionary is empty
+  bool empty() const { return _data.empty(); }
+
+  using const_iterator = DataType::const_iterator;
```
bddap, line 144: Technically, future changes to this type would also be breaking changes. Could be remedied with a wrapper type but that seems overkill.

bp-kelley:
@greglandrum I think this is worth looking at now.

bp-kelley (approved):

Code/RDGeneral/Dict.h
```diff
@@ -136,14 +136,47 @@ class RDKIT_RDGENERAL_EXPORT Dict {
...
+  const_iterator begin() const { return _data.begin(); }
+  const_iterator end() const { return _data.end(); }
+
+  //! \brief Appends a populated Pair to the dictionary.
+  void appendPair(Pair &&pair) {
```
bddap: Self, change from `append` to `insert` to avoid communicating insertion order semantics.

Code/RDGeneral/Dict.h
```diff
@@ -136,14 +136,47 @@ class RDKIT_RDGENERAL_EXPORT Dict {
...
+    if (pair.val.needsCleanup()) {
+      _hasNonPodData = true;
+    }
+    _data.push_back(std::move(pair));
+  }
+
+  //! \brief Bulk-appends a vector of Pairs, moving them into the dictionary.
+  void append(std::vector<Pair> &&pairs) {
```
bddap: should be `extend`

Code/RDGeneral/testDict.cpp
```diff
@@ -652,6 +652,38 @@ TEST_CASE("testCustomProps") {
   delete handlers[1];
 }
 
+static int g_barDestructorCount = 0;
```
bddap: Rm this mutable global.

Code/RDGeneral/Dict.h
bddap: actually this method is never used, delete it

Code/RDGeneral/testDict.cpp
bddap: done in [7ab17ad](https://github.com/rdkit/rdkit/pull/9113/commits/7ab17ad063f195e38c5715181c0505c095c78862)

Code/RDGeneral/Dict.h
bddap: done in [7ab17ad](https://github.com/rdkit/rdkit/pull/9113/commits/7ab17ad063f195e38c5715181c0505c095c78862)

Code/RDGeneral/Dict.h
bddap: done in [7ab17ad](https://github.com/rdkit/rdkit/pull/9113/commits/7ab17ad063f195e38c5715181c0505c095c78862)

Code/RDGeneral/Dict.h
bp-kelley: This will break the ABI so that we can't replace one rdkit.dll with another, but we have never been able to do that on MSVC builds anyway, so I wouldn't worry about it too much.

This is compile safe (mostly) which is good enough.

Code/RDGeneral/Dict.h
bddap: done

Code/RDGeneral/StreamOps.h
bddap: I've added a new bulk-append method. 

```
  void extend(std::vector<Pair> &&pairs) {
    for (auto &p : pairs) {
      _hasNonPodData |= p.val.needsCleanup();
    }
    _data.insert(_data.end(), std::make_move_iterator(pairs.begin()),
                 std::make_move_iterator(pairs.end()));
  }
```

It's still one more allocation than `master`. Since we are moving elements to a new vec.

bddap:
> Nice work, I think it's a good idea to hide the implementation. I have a few suggestions and the Custom handlers need to be explicitly tested so that we are sure they are destroyed.

[5116890](https://github.com/rdkit/rdkit/pull/9113/commits/51168907e75a681173b24bc7b76aeaefb0c2a645) adds a test to exercise the `DataStructsExplicitBitVecPropHandler` custom handler. valgrind reports no leaks.

Code/RDGeneral/testDict.cpp
```diff
@@ -652,6 +652,41 @@ TEST_CASE("testCustomProps") {
...
+  Bar(const Bar &o) = default;
+  Bar &operator=(const Bar &) = default;
+  ~Bar() {
+    if (dtor_count) {
+      ++(*dtor_count);
+    }
+  }
+};
+
+TEST_CASE("custom AnyTag data is destroyed through Dict lifecycle") {
```
greglandrum, line 669: The `count` tests here pass even if you never call `setVal`.
I think this would be more precise if you checked the expected value of `count` instead of simply testing that it's > a target value.

greglandrum:
@bddap any chance that you can make this change relatively quickly?

This one would need to be in a major release and we're doing one of those next week. Otherwise we'll have it on master, but it won't make it out in a release until September.

Code/RDGeneral/testDict.cpp
bddap: done in [0a31cf2](https://github.com/rdkit/rdkit/pull/9113/commits/0a31cf22459307591f4db66c5a4e8c17ee159f9a)

greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @bddap!

PR #9118, opened 2026-02-17
Exclude Zero order bonds  from FindRingFamilies() and add arguments for includeDativeBonds, includeHydrogenBonds

ricrogz:
Also:
- adds the `includeHydrogenBonds` argument to the Python wrappers for `findSSSR()` and `symmetrizeSSSR()` 
- fixes some typos
- Removes an `#if 0`'d declaration for `findSpanningTree ()` that is never implemented.


greglandrum (changes requested):
Code changes look good, please add tests for the new args to `findRingFamilies()`

PR #9120, opened 2026-02-17
Add valence 1 as possible option for sulfur

### LINKED ISSUE #9114: DetermineBonds fails for thiolate
**Describe the bug**
`DetermineBonds` fails with `ValueError: Unable to determine valence of atom 11 with atomic number 16 and 1 bonds.` on the $$S^-$$ atom of negative cysteine. It was working fine until RDKit version 2025.09.3, so I suspect the change comes from the github pull #8888 from [ChiCheng45](https://github.com/ChiCheng45).

**To Reproduce**
```
file = "neg_cys_zwitterion_termini.pdb"
mol = MolFromPDBFile(str(file), removeHs=False, sanitize=False)
DetermineBonds(mol, charge=-1)
```
See the file [neg_cys_zwitterion_termini.pdb](https://github.com/user-attachments/files/25326491/neg_cys_zwitterion_termini.pdb) (negative cysteine side chain with thiolate, NH3+/COO- termini).
I also tried the negative cysteine with the neutral NH2/COOH termini ([neg_cys_neutral_termini.pdb](https://github.com/user-attachments/files/25326614/neg_cys_neutral_termini.pdb)), but the problem is the same.
And using .xyz files instead of .pdb also gives the same problem.
Setting the arguments `removeHs` and `sanitize` to `True` doesn't get rid of the problem either (and wouldn't be the behaviour I need in my case anyway).

**Expected behavior**
`DetermineBonds` shouldn't fail in this case but should determine that with a molecular charge of -1, the $$S^-$$ has only one bond, which is correct in the input file.
This was the behaviour until RDKit version 2025.09.3, and fails in the latest versions 2025.09.4 and 2025.09.5.

**Configuration:**
 - RDKit version: 2025.09.5
 - OS: Ubuntu 22.04.1
 - Python version: 3.11.0
 - Are you using conda? yes

LaurianeJD:
Fixes #9114

As described in the corresponding issue, `DetermineBonds` was working for thiolates until version 2025.09.3. The implementation of `DetermineBonds` was changed in PR #8888, where the possible number of bonds an atom can form is now defined by
https://github.com/rdkit/rdkit/blob/8fc241495c7859727ae00db0eb29801055c9a7d2/Code/GraphMol/DetermineBonds/DetermineBonds.cpp#L100-L103
and a singly bonded sulfur ($$R-S^-$$) isn't accepted anymore and therefore makes `DetermineBonds(mol, charge=-1)` fail.

This PR adds `1` as an option for sulfur (atomic number `16`) so that thiolates can again be recognised by `DetermineBonds` and don't raise the error `Unable to determine valence of atom ... with atomic number 16 and 1 bonds.`

If this isn't a desired change, a workaround would be to set the formal charge of sulfur beforehand:
```
sulfur = [a for a in mol.GetAtoms() if a.GetSymbol() == "S"][0]
sulfur.SetFormalCharge(-1)
DetermineBonds(mol, charge=-1)
```

### DIFF
--- a/Code/GraphMol/DetermineBonds/catch_tests.cpp
+++ b/Code/GraphMol/DetermineBonds/catch_tests.cpp
@@ -501,6 +501,33 @@ TEST_CASE(
   }
 }
 
+TEST_CASE("github #9114: DetermineBonds fails for thiolate") {
+  SECTION("as reported") {
+    RWMol m;
+    bool updateLabel = true;
+    bool takeOwnership = true;
+    m.addAtom(new Atom(6), updateLabel, takeOwnership);
+    m.addAtom(new Atom(16), updateLabel, takeOwnership);
+    m.addAtom(new Atom(1), updateLabel, takeOwnership);
+    m.addAtom(new Atom(1), updateLabel, takeOwnership);
+    m.addAtom(new Atom(1), updateLabel, takeOwnership);
+    m.addBond(0, 1, Bond::UNSPECIFIED);
+    m.addBond(0, 2, Bond::UNSPECIFIED);
+    m.addBond(0, 3, Bond::UNSPECIFIED);
+    m.addBond(0, 4, Bond::UNSPECIFIED);
+
+    int charge = -1;
+    bool allowChargedFragments = true;
+    bool embedChiral = false; 
+    CHECK_NOTHROW(determineBondOrders(m, charge, allowChargedFragments, embedChiral));
+
+    for (auto bnd : m.bonds()) {
+      CHECK(bnd->getBondType() == Bond::SINGLE);
+    }
+    CHECK(m.getAtomWithIdx(1)->getFormalCharge() == -1);
+  }
+}
+
 TEST_CASE("problems with H2") {
   SECTION("as reported") {
     std::string xyz = R"XYZ(2


greglandrum (changes requested):
Thanks for the contribution!
The code changes look good, but please add a test in catch_tests.cpp that demonstrates that the fix is actually working.

Feel free to ask if you have questions about how to do this.

ChiCheng45:
The value 4 for sulfur and phosphorus is missing in `atomicValence` compared to table 1 of https://onlinelibrary.wiley.com/doi/10.1002/bkcs.10334.

From my understanding, the original values in rdkit were set to whatever was in

https://github.com/jensengroup/xyz2mol/blob/9ec591dd01bb4793fc221c526d5f7a19c05d0aca/xyz2mol.py#L52-L64

but for whatever reason, the values 4 for sulfur and phosphorus are commented out. Do we know why? Is it worth adding them back in here?

LaurianeJD:
I've now added a test in catch_tests.cpp @greglandrum 

---

Concerning the valence 4 @ChiCheng45 : 

- for sulfur I think it would indeed make sense to add it because currently, molecules like sulfur tetrafluoride fail:
```
>>> f = "SF4.xyz"
>>> m = MolFromXYZFile(f)
>>> DetermineConnectivity(m)
>>> DetermineBonds(m)
  File "<stdin>", line 1, in <module>
ValueError: Unable to determine valence of atom 1 with atomic number 16 and 4 bonds. Check the input molecules bonding.
```

- for phosphorus though I don't think so, as I don't see any common molecules where the phosphorus would have a valence of 4? Phosphorus can be bounded to 4 atoms, but its valency is generally 3 or 5 (e.g. $$PO_4^{3-}$$, with a valence of 5, which already passes in `DetermineBonds`)

What do you think @greglandrum ?

greglandrum:
Thanks for the contribution @LaurianeJD

PR #9123, opened 2026-02-19
swap zlib download to use github

greglandrum:
zlib.net apparently only keeps the most recent version around, so when they update our CI pipelines break

also bumps the version


greglandrum:
This is a simple infrastructure change and was holding up every PR, so I'm doing a self merge

PR #9125, opened 2026-02-19
Deterministic kekulize, independent of atom and bond order

### LINKED ISSUE #9122: Kekulization depends on atom/bond ordering, affects tautomer enumeration and canonicalization
**Describe the bug**
Tautomer enumeration/canonicalization can be non-deterministic for the same input molecule because it depends on the particular kekule assignment chosen for aromatic rings. Then certain rules will or won't apply, causing different enumerations and possibly canonical tautomer.

**To Reproduce**
When we shuffle atom orders, then the kekulization leads to a tautomer rule not matching on the intermediate state, so you don't sample via the transformation.
```
from rdkit import Chem

smi = "OC1=Nc2ccccc2C1=Cc1ccc[nH]1"
m = Chem.MolFromSmiles(smi)

# Renumber atoms (same chemistry, different internal order)
order = list(range(m.GetNumAtoms()))
order = order[::-1]  # deterministic example; random shuffles also reproduce
m2 = Chem.RenumberAtoms(m, order)

te = rdMolStandardize.TautomerEnumerator()
te.SetMaxTautomers(2000)
te.SetMaxTransforms(2000)

tauts1 = sorted({Chem.MolToSmiles(x) for x in te.Enumerate(m)})
tauts2 = sorted({Chem.MolToSmiles(x) for x in te.Enumerate(m2)})

print("counts:", len(tauts1), len(tauts2))
print("only_in_m2:", sorted(set(tauts2) - set(tauts1))[:10])

c1 = Chem.MolToSmiles(te.Canonicalize(m))
c2 = Chem.MolToSmiles(te.Canonicalize(m2))
print("canonical:", c1)
print("canonical_renum:", c2)
```
```
input_smiles: OC1=Nc2ccccc2C1=Cc1ccc[nH]1
nAtoms: 16
renumber_order: [15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
counts: 7 10
only_in_m2_count: 3
only_in_m1_count: 0
only_in_m2: ['O=C1N=c2ccccc2=C1C=C1CCC=N1', 'O=C1N=c2ccccc2=C1CC1=CCC=N1', 'O=C1N=c2ccccc2=C1Cc1ccc[nH]1']
only_in_m1: []
canonical: O=C1Nc2ccccc2C1=Cc1ccc[nH]1
canonical_renum: O=C1N=c2ccccc2=C1Cc1ccc[nH]1
```

**Expected behavior**
Tautomer enumeration should be invariant to atom/bond ordering (renumbering) for the same molecule. Which will also make the canonical tautomer invariant to atom/bond order.

**Screenshots**

Image

 - RDKit version: Release_2025_09_5
 - OS: [e.g. Ubuntu 20.04] Ubuntu 24.04
 - Python version (if relevant): n/a
 - Are you using conda? no
 - If you are not using conda: how did you install the RDKit? manual builds

**Additional context**
There might be a larger issue here about the determinism (or not) of kekulization.

pechersky:
#### Reference Issue
Fixes #9122.

Walk atoms in canonical order, thread that down to kekulization code. Order we pick atoms changes too, pick end atoms of wedge/hashes, otherwise, pick lowest by rank. Neighbor precollection, then sorted by rank.

Switch to `vector<int>` because we need sorting.

The test changes are
- main test capturing regression is `test13bTautomerOrderIndependence` in `Code/GraphMol/MolStandardize/Wrap/testMolStandardize.py`
- other test changes are flipping indices to where the double bonds are now

I tried to be as surgical and minimal-diff as possible. And to commit one set of test files at a time.

### DIFF
--- a/Code/GraphMol/RascalMCES/mces_cluster_catch.cpp
+++ b/Code/GraphMol/RascalMCES/mces_cluster_catch.cpp
@@ -81,7 +81,7 @@ TEST_CASE("ChEMBL 1907596") {
   clusOpts.similarityCutoff = 0.7;
   auto clusters = RDKit::RascalMCES::rascalCluster(mols, clusOpts);
   REQUIRE(clusters.size() == 21);
-  std::vector<size_t> expSizes{342, 71, 64, 33, 23, 11, 10, 6, 6, 5, 5,
+  std::vector<size_t> expSizes{343, 71, 64, 33, 23, 11, 10, 6, 6, 5, 5,
                                4,   3,  3,  3,  3,  3,  2,  2, 2, 14};
   for (size_t i = 0; i < 21; ++i) {
     REQUIRE(clusters[i].size() == expSizes[i]);


greglandrum:
This is great. I have had doing something like this on my ToDo list for a *really* long time!

Code/GraphMol/Depictor/catch_tests.cpp
```diff
@@ -863,6 +865,29 @@ M  END
...
+
+    ROMol baseMol(*wedgedMol);
+    Chirality::reapplyMolBlockWedging(baseMol);
+
+    auto getBondDirBetween = [](const ROMol &mol, unsigned int a1,
+                                unsigned int a2) {
+      const auto bond = mol.getBondBetweenAtoms(a1, a2);
+      REQUIRE(bond);
+      return bond->getBondDir();
+    };
```
ricrogz: These two lambdas are duplicated in the next test too. It might be worth abstracting them into functions to avoid the duplication.

Code/GraphMol/Depictor/catch_tests.cpp
```diff
@@ -828,6 +828,8 @@ M  END
  28 29  1  0
  21 27  1  0
 )CTAB";
+  (void)originalWedges;
+  (void)invertedWedges;
```
ricrogz: If these are not used, please delete them.

Code/GraphMol/Kekulize.cpp
```diff
@@ -223,11 +224,34 @@ void markDbondCands(RWMol &mol, const INT_VECT &allAtms,
...
+  };
+
+  // Prefer starting traversal at atoms which are the *end* of wedged/dashed
+  // bonds. This heuristic helps avoid assigning double bonds to wedged bonds.
+  boost::dynamic_bitset<> wedgeEndAtoms(mol.getNumAtoms());
+  for (const auto bond : mol.bonds()) {
+    if (bond->getBondDir() == Bond::BondDir::BEGINWEDGE ||
+        bond->getBondDir() == Bond::BondDir::BEGINDASH) {
+      const auto endIdx = bond->getEndAtomIdx();
+      if (endIdx < wedgeEndAtoms.size() && inAllAtms.test(endIdx)) {
```
ricrogz: Since `wedgeEndAtoms` covers all atoms, isn't the `endIdx < wedgeEndAtoms.size()` check redundant here?

Code/GraphMol/Kekulize.cpp
```diff
@@ -223,11 +224,34 @@ void markDbondCands(RWMol &mol, const INT_VECT &allAtms,
...
+  }
+
+  auto lessByRank = [&atomRanks](int a, int b) {
+    const auto ra = atomRanks.at(static_cast<unsigned int>(a));
+    const auto rb = atomRanks.at(static_cast<unsigned int>(b));
+    return (ra < rb) || (ra == rb && a < b);
+  };
+
+  // Prefer starting traversal at atoms which are the *end* of wedged/dashed
+  // bonds. This heuristic helps avoid assigning double bonds to wedged bonds.
```
ricrogz: Why do we prefer wedge ends to atoms that are not wedged at all ?

Code/GraphMol/Kekulize.cpp
```diff
@@ -254,10 +278,17 @@ bool kekulizeWorker(RWMol &mol, const INT_VECT &allAtms,
...
+      curr = -1;
+      bool currIsWedgeEnd = false;
       for (int allAtm : allAtms) {
-        if (std::find(done.begin(), done.end(), allAtm) == done.end()) {
+        if (std::find(done.begin(), done.end(), allAtm) != done.end()) {
+          continue;
+        }
+        const bool isWedgeEnd = wedgeEndAtoms.test(allAtm);
+        if (curr < 0 || (isWedgeEnd && !currIsWedgeEnd) ||
+            (isWedgeEnd == currIsWedgeEnd && lessByRank(allAtm, curr))) {
```
ricrogz: Would it make sense to pre-sort `allAtms` based on `lessByRank` ? This way, the first wedge end we see should be the one with the lowest rank, and we don't need to further traverse `allAtms`.

Code/GraphMol/Kekulize.cpp
```diff
@@ -278,18 +309,25 @@ bool kekulizeWorker(RWMol &mol, const INT_VECT &allAtms,
       CHECK_INVARIANT(opts.size() > 0, "");
     } else {
       INT_DEQUE lstack;
-      INT_DEQUE wedgedOpts;
+      std::vector<int> optsV;
+      std::vector<int> wedgedOptsV;
+      std::vector<int> nbrs;
       for (const auto &nbrIdx : boost::make_iterator_range(
                mol.getAtomNeighbors(mol.getAtomWithIdx(curr)))) {
```
ricrogz: While you are here, would you mind updating this to:
```suggestion
for (auto nbrIdx : mol.atomNeighbors(mol.getAtomWithIdx(curr))) {
```

Code/GraphMol/Kekulize.cpp
```diff
@@ -278,18 +309,25 @@ bool kekulizeWorker(RWMol &mol, const INT_VECT &allAtms,
...
         // ignore if the neighbor has already been dealt with before
         if (std::find(done.begin(), done.end(), nbrIdx) != done.end()) {
           continue;
         }
         // ignore if the neighbor is not part of the fused system
-        if (std::find(allAtms.begin(), allAtms.end(), nbrIdx) ==
-            allAtms.end()) {
+        if (!inAllAtms.test(nbrIdx)) {
           continue;
         }
```
ricrogz, line 330: Maybe move this before the other check, to save us some traversals of `done`.

Code/GraphMol/Kekulize.cpp
```diff
@@ -315,15 +353,23 @@ bool kekulizeWorker(RWMol &mol, const INT_VECT &allAtms,
...
+      // Deterministic option ordering: prefer lower-ranked neighbors,
+      // but keep wedged/dashed bonds at the end.
+      std::sort(optsV.begin(), optsV.end(), lessByRank);
+      std::sort(wedgedOptsV.begin(), wedgedOptsV.end(), lessByRank);
+      for (int v : optsV) {
+        opts.push_back(v);
+      }
+      for (int v : wedgedOptsV) {
+        opts.push_back(v);
+      }
```
ricrogz: What's the difference between sorting `optsV` and `wedgedOptsV` separately, and then inserting them into `opts` one after the other and inserting them both and then sorting `opts` ? or merging them into a single vector, then sort, then insert into `opts` ?

ricrogz:
Oh, also, have you checked the effect of your changes on performance?

greglandrum:
I think we're going to have to be very careful about the wording here. 
The "canonical" atom ranks that we construct are going to be canonical given the chemistry of molecule we kekulize.
If we are working with a sanitized molecule (i.e. kekulizing before tautomerization or writing an SDF), then we should be ok.

However the kekulization code is also used on partially sanitized molecules when we are doing the initial sanitization. Here the stereochemistry won't have been perceived and the chemistry will be inconsistent. Here things will only be canonical in a very specific sense (and one requiring some explanation). This isn't a problem, but it will need to be documented.

I also have some concerns about the performance impact of doing this during sanitization. The canonicalization doesn't bring much (if any) benefit there. I think we should make the canonicalization optional and avoid using it during sanitization  (you can just use the atom indices as ranks there).

Code/GraphMol/Kekulize.cpp
pechersky: Because we want to retain the stereo as long as possible, so starting with ends of such bonds, the double bonds of those atoms are more likely to go somewhere that _wasn't_ the wedge bond.

Code/GraphMol/Depictor/catch_tests.cpp
pechersky: 0cbc0ff

Code/GraphMol/Depictor/catch_tests.cpp
pechersky: c0819ae

Code/GraphMol/Kekulize.cpp
pechersky: 57db56a

Code/GraphMol/Kekulize.cpp
pechersky: efb6c64

Code/GraphMol/Kekulize.cpp
pechersky: 24f063a

Code/GraphMol/Kekulize.cpp
pechersky: ceb410d

Code/GraphMol/Kekulize.cpp
pechersky: 171b4a5

pechersky:
> Oh, also, have you checked the effect of your changes on performance?

On my laptop, did 50 runs of kekulizing 5003 mols (NCI first_5K + SAMPLES) after scrambling atom order

| Commit | Description | Mean | vs parent |
|--------|-------------|------|-----------|
| `8fc241495` (parent) | No canonical ranking | 53.2 ms | 1.00x |
| `b338a0ed0` (original PR before shortcut path) | Always canonical | 122.3 ms | 2.30x |
| `67638bd10` (current) canonical=false | Sanitization path | 55.3 ms | 1.04x |
| `67638bd10` (current) canonical=true | Deterministic path | 118.4 ms | 2.23x |

Callgrind instruction counts

| Commit | Total Ir | KekulizeFragment | rankFragmentAtoms |
|--------|----------|-----------------|-------------------|
| master | 5.30B | 50.1M | — |
| original PR before shortcut path | 6.86B | 53.0M | 8.8M |
| current (both paths) | 7.96B | 83.9M | 8.8M |

I also did MolFromSmiles benchmarking for a similar set

| Commit | Description | Mean (ms) | Std dev | vs parent |
|--------|-------------|-----------|---------|-----------|
| 8fc241495 (parent) | No canonical ranking | 229.4 | 11.3 | 1.00x |
| b338a0ed0 (original PR) | Always canonical in sanitize | 235.4 | 12.4 | 1.03x |
| 67638bd10 (current) | canonical=false in sanitize | 234.1 | 7.5 | 1.02x |

benchmark code:
```
std::vector<std::string> load_smi_strings(const std::filesystem::path &relative) {
  char *rdbase = std::getenv("RDBASE");
  REQUIRE(rdbase);
  std::filesystem::path path(rdbase);
  path /= relative;
  std::ifstream in(path);
  REQUIRE(in.good());

  std::vector<std::string> ret;
  std::string line;
  while (std::getline(in, line)) {
    std::istringstream iss(line);
    std::string smi;
    if (!(iss >> smi) || smi.empty() || smi[0] == '#') {
      continue;
    }
    ret.push_back(std::move(smi));
  }
  return ret;
}

std::vector<RDKit::ROMol> load_smi_file(const std::filesystem::path &relative) {
  char *rdbase = std::getenv("RDBASE");
  REQUIRE(rdbase);
  std::filesystem::path path(rdbase);
  path /= relative;
  std::ifstream in(path);
  REQUIRE(in.good());

  std::vector<RDKit::ROMol> ret;
  std::string line;
  while (std::getline(in, line)) {
    std::istringstream iss(line);
    std::string smi;
    if (!(iss >> smi) || smi.empty() || smi[0] == '#') {
      continue;
    }
    try {
      auto mol = RDKit::v2::SmilesParse::MolFromSmiles(smi);
      if (mol) {
        ret.push_back(std::move(*mol));
      }
    } catch (...) {
      continue;
    }
  }
  return ret;
}

namespace {
// Scramble atom order and re-aromatize molecules for kekulization benchmark.
// Molecules that fail any step are silently dropped.
std::vector<RWMol> scramble_and_aromatize(
    const std::vector<ROMol> &mols) {
  std::mt19937 rng(42);
  std::vector<RWMol> ret;
  ret.reserve(mols.size());
  for (auto &mol : mols) {
    try {
      std::vector<unsigned int> perm(mol.getNumAtoms());
      std::iota(perm.begin(), perm.end(), 0u);
      std::shuffle(perm.begin(), perm.end(), rng);
      auto pmol = MolOps::renumberAtoms(mol, perm);
      RWMol rw(*pmol);
      MolOps::setAromaticity(rw);
      RWMol test(rw);
      MolOps::Kekulize(test);
      ret.push_back(std::move(rw));
    } catch (...) {
      continue;
    }
  }
  return ret;
}
}  // namespace

TEST_CASE("MolOps::Kekulize", "[molops][kekulize]") {
  auto nci = bench_common::load_smi_file("Data/NCI/first_5K.smi");
  auto inline_samples = bench_common::load_samples();
  for (auto &mol : inline_samples) {
    nci.push_back(std::move(mol));
  }

  auto scrambled = scramble_and_aromatize(nci);
  WARN("Benchmark molecules (after filtering): " << scrambled.size()
       << " / " << nci.size());

  BENCHMARK("Kekulize canonical=true") {
    auto total = 0u;
    for (auto mol : scrambled) {
      MolOps::Kekulize(mol, true, true);
      total += mol.getNumAtoms();
    }
    return total;
  };

  BENCHMARK("Kekulize canonical=false") {
    auto total = 0u;
    for (auto mol : scrambled) {
      MolOps::Kekulize(mol, true, false);
      total += mol.getNumAtoms();
    }
    return total;
  };
}

TEST_CASE("MolFromSmiles (sanitization)", "[molops][sanitize]") {
  auto smiles = bench_common::load_smi_strings("Data/NCI/first_5K.smi");
  for (auto s : bench_common::SAMPLES) {
    smiles.push_back(s);
  }
  WARN("SMILES count: " << smiles.size());

  BENCHMARK("MolFromSmiles") {
    auto total = 0u;
    for (auto &smi : smiles) {
      try {
        auto mol = v2::SmilesParse::MolFromSmiles(smi);
        if (mol) total += mol->getNumAtoms();
      } catch (...) {
        continue;
      }
    }
    return total;
  };
}
```

ricrogz:
> On my laptop, did 50 runs of kekulizing 5003 mols (NCI first_5K + SAMPLES) after scrambling atom order
 
That looks good, thanks for checking!

Code/GraphMol/MolOps.h
```diff
@@ -746,6 +755,7 @@ RDKIT_GRAPHMOL_EXPORT void adjustHs(RWMol &mol);
 
 */
 RDKIT_GRAPHMOL_EXPORT void Kekulize(RWMol &mol, bool markAtomsBonds = true,
+                                    bool canonical = true,
                                     unsigned int maxBackTracks = 100);
```
ricrogz, line 766: I think we should update the Python wrappers to include this argument too.

Code/GraphMol/MolOps.h
```diff
@@ -733,6 +733,15 @@ RDKIT_GRAPHMOL_EXPORT void adjustHs(RWMol &mol);
    settings on both the Bonds and Atoms are turned to false following the
    Kekulization, otherwise they are left alone in their original state.
 
+   \param canonical  if true (the default), uses canonical atom ranking so
```
ricrogz: Should the default be `false` ? Setting it to true is both slower and also a change in behavior. And, as the comment says, results will be an artifact if stereo hasn't been perceived or is outdated (I'm concerned about the effects on existing code).

Code/GraphMol/Kekulize.cpp
```diff
@@ -278,18 +319,25 @@ bool kekulizeWorker(RWMol &mol, const INT_VECT &allAtms,
...
-        if (std::find(allAtms.begin(), allAtms.end(), nbrIdx) ==
-            allAtms.end()) {
+        // ignore if the neighbor has already been dealt with before
+        if (std::find(done.begin(), done.end(), nbrIdx) != done.end()) {
           continue;
         }
+        nbrs.push_back(nbrIdx);
+      }
+
+      std::sort(nbrs.begin(), nbrs.end(), lessByRank);
```
ricrogz: Do we need to consider wedge end here too ?

Code/GraphMol/Kekulize.cpp
ricrogz: Pushing front will revert the order of `nbrs` in the stacks. I don't think that's a problem, but it's not consistent with the intention in the rest of the code.

Code/GraphMol/MolOps.h
pechersky: How about in file writing like SmilesWriter, MarvinWriter, etc?

pechersky: Also, `false` is also a change in behavior. It uses `iota` for atom indices, not SSSR based traversal order.

ricrogz: Ok, fair point. If behavior is going to change anyway, then there isn't much point in making `false` the default.

greglandrum (changes requested):
I'm excited to have this in the RDKit!

I did a PR against your branch with the following suggested changes:
1. Add the `canonical` flag to the python wrappers
2. document the changes in ReleaseNotes.md

greglandrum:
Quick performance test on my machine using the current state of this PR, merged with master.
This is running with 10K molecules from ChEMBL
master:
```
In [4]: %timeit ms = [Chem.MolFromSmiles(x) for x in ls]
829 ms ± 3.07 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: ms = [Chem.MolFromSmiles(x) for x in ls]

In [9]: %timeit [Chem.Kekulize(Chem.Mol(m)) for m in ms]
170 ms ± 495 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [10]: %timeit [Chem.MolToV3KMolBlock(m) for m in ms]
1.69 s ± 18.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
and with this PR the runtimes are:
```
822 ms ± 4.27 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
252 ms ± 1.58 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.76 s ± 3.14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So there's a noticeable slowdown in kekulization (expected), but that's a small enough part of generating a mol block that it doesn't make a noticeable difference in actual run time.

Also, as expected, there's no impact on parsing time.

Nice!

pechersky:
> I'm excited to have this in the RDKit!
> 
> I did a PR against your branch with the following suggested changes:
> 
> 1. Add the `canonical` flag to the python wrappers
> 2. document the changes in ReleaseNotes.md

merged

pechersky:
Just checking, we want canonical=True to be the default, correct?

ricrogz:
> Just checking, we want canonical=True to be the default, correct?

If behavior is going to change anyway, then I think defaulting to true is ok.

PR #9127, opened 2026-02-20
Defer numpy initialization to first use

### LINKED ISSUE #9126: Slow imports due to numpy initialization, affects cold start
**Describe the bug**
`from rdkit import Chem` unconditionally bootstraps numpy (~120ms+ import timing), even when it won't actually utilized.

**To Reproduce**
```
$ python -X importtime -c "from rdkit import Chem" 2>&1 | grep numpy
import time:       674 |     119564 |   numpy
```
```
from rdkit import Chem
print("numpy loaded:", "numpy" in sys.modules)
```

**Expected behavior**
`from rdkit import Chem` should not load numpy. Numpy should only be initialized when an API that actually returns/consumes numpy arrays is called (e.g. GetDistanceMatrix, GetPositions, ConvertToNumpyArray).

 - RDKit version: 2024.09+
 - OS: Ubuntu 24.04, also AWS lambda AL2023
 - Are you using conda? no
 - If you are not using conda: how did you install the RDKit? manual build

**Additional context**
The root cause is three C++ extension modules calling `rdkit_import_array()` and/or `boost::python::numpy::initialize()` in their `BOOST_PYTHON_MODULE` init:
- rdchem.so — numpy::initialize() + rdkit_import_array()
- rdmolops.so — rdkit_import_array()
- cDataStructs.so — rdkit_import_array()
 
These run unconditionally at import time. Making them lazy (guard with a std::call_once on first actual numpy use) drops from rdkit import Chem from ~385ms to ~35ms in cold-start benchmarks.

Note: the DataStructs top level import in `__init__.py` and `Geometry/__init__.py` is 20 years old!

pechersky:
#### Reference Issue
Fixes #9126.

`from rdkit import Chem` unconditionally bootstrapped numpy (~120ms) via import_array()/boost::python::numpy::initialize() in module init functions, even when no numpy-dependent APIs were called. This is costly in cold-start environments like AWS Lambda.

Move numpy initialization behind lazy guards (static bool + first-call init) in rdchem.so, rdmolops.so, and cDataStructs.so. Numpy now loads only when an API that actually needs it is invoked (GetDistanceMatrix, GetPositions, SetPositions, GetAdjacencyMatrix, ConvertToNumpyArray, etc.).

Also change Conformer::SetPos to accept python::object instead of np::ndarray to prevent Boost.Python from requiring numpy type conversion before the lazy guard runs.

Adds test_lazy_numpy.py with subprocess-based tests verifying:
- `from rdkit import Chem` does not load numpy
- SmilesToMol/MolToSmiles work without numpy
- numpy loads on demand when array APIs are called

### DIFF
--- a/Code/DataStructs/Wrap/DataStructs.cpp
+++ b/Code/DataStructs/Wrap/DataStructs.cpp
@@ -10,6 +10,10 @@
 //
 #define PY_ARRAY_UNIQUE_SYMBOL rddatastructs_array_API
 
+#ifdef RDK_BUILD_THREADSAFE_SSS
+#include <mutex>
+#endif
+
 #include <RDBoost/python.h>
 #include <RDBoost/Wrap.h>
 #include <DataStructs/BitVects.h>
@@ -32,12 +36,20 @@ void wrap_realValVect();
 void wrap_sparseIntVect();
 void wrap_FPB();
 
-struct ds_numpy_init {
-  ds_numpy_init() { rdkit_import_array(); }
-};
+#ifdef RDK_BUILD_THREADSAFE_SSS
+static std::once_flag s_ds_numpy_init_flag;
+#endif
 
 static void ds_ensure_numpy() {
-  static ds_numpy_init init;
+#ifdef RDK_BUILD_THREADSAFE_SSS
+  std::call_once(s_ds_numpy_init_flag, rdkit_import_array);
+#else
+  static bool initialized = false;
+  if (!initialized) {
+    initialized = true;
+    rdkit_import_array();
+  }
+#endif
 }
 
 namespace {
--- a/Code/GraphMol/Wrap/rdchem.cpp
+++ b/Code/GraphMol/Wrap/rdchem.cpp
@@ -21,21 +21,34 @@
 
 #include <sstream>
 #include <utility>
+#ifdef RDK_BUILD_THREADSAFE_SSS
+#include <mutex>
+#endif
 
 #include "seqs.hpp"
 namespace python = boost::python;
 using namespace RDKit;
 
-struct rdchem_numpy_init {
-  rdchem_numpy_init() {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+static std::once_flag s_rdchem_numpy_init_flag;
+#endif
+
+void rdkit_rdchem_ensure_numpy() {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+  std::call_once(s_rdchem_numpy_init_flag, []() {
+    const bool register_scalar_converters = false;
+    boost::python::numpy::initialize(register_scalar_converters);
+    rdkit_import_array();
+  });
+#else
+  static bool initialized = false;
+  if (!initialized) {
+    initialized = true;
     const bool register_scalar_converters = false;
     boost::python::numpy::initialize(register_scalar_converters);
     rdkit_import_array();
   }
-};
-
-void rdkit_rdchem_ensure_numpy() {
-  static rdchem_numpy_init init;
+#endif
 }
 
 namespace RDKit {
--- a/Code/GraphMol/Wrap/rdmolops.cpp
+++ b/Code/GraphMol/Wrap/rdmolops.cpp
@@ -17,16 +17,27 @@
 #include <RDBoost/import_array.h>
 #include <RDGeneral/Exceptions.h>
 #include <GraphMol/SanitException.h>
+#ifdef RDK_BUILD_THREADSAFE_SSS
+#include <mutex>
+#endif
 
 namespace python = boost::python;
 using namespace RDKit;
 
-struct rdmolops_numpy_init {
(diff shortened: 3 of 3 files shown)


Code/GraphMol/Wrap/test_lazy_numpy.py
```diff
@@ -0,0 +1,162 @@
...
+
+      # Substructure match
+      query = Chem.MolFromSmarts("[OH]")
+      assert mol.HasSubstructMatch(query)
+
+      # MolToMolBlock (V2000)
+      mb = Chem.MolToMolBlock(mol)
+      assert "V2000" in mb
+
+      # InChI round-trip
```
greglandrum: The inchi code is not always present. If you're going to use it in a test, you need to first check if it's there by checking `rdkit.Chem.inchi.INCHI_AVAILABLE`

greglandrum: This is why the CI builds failed

Code/DataStructs/Wrap/DataStructs.cpp
```diff
@@ -32,9 +32,19 @@ void wrap_realValVect();
 void wrap_sparseIntVect();
 void wrap_FPB();
 
+static bool s_ds_numpy_initialized = false;
+
+static void ds_ensure_numpy() {
```
greglandrum, line 43: I believe that this isn't thread safe and that you may end up calling `rdkit_import_array()` simultaneously in multiple threads. I'm not sure what the consequences of that are.

Mabye it's better to use std::call_once?

pechersky: a926c65

Code/GraphMol/Wrap/test_lazy_numpy.py
pechersky: 8a4c834

Code/GraphMol/Wrap/rdchem.cpp
```diff
@@ -26,6 +26,17 @@
 namespace python = boost::python;
 using namespace RDKit;
 
+static bool s_rdchem_numpy_initialized = false;
+
+void rdkit_rdchem_ensure_numpy() {
```
greglandrum: Sorry, I should have mentioned this in the previous comment: but I think all of these need to be thread-safe.

We were safe before because a Python module can only be imported once, but now we don't have that guarantee

pechersky: Yeah, my mistake, was doing too many things as the same time. Will add.

Code/DataStructs/Wrap/DataStructs.cpp
```diff
@@ -32,9 +36,26 @@ void wrap_realValVect();
...
+  std::call_once(s_ds_numpy_init_flag, rdkit_import_array);
+#else
+  static bool initialized = false;
+  if (!initialized) {
+    initialized = true;
+    rdkit_import_array();
+  }
+#endif
+}
+
```
bp-kelley, line 54: Another way of doing the same thing without the #ifdef's is to use a "magic static"

This works on all C++11 and beyond compilers.

```
struct rdkit_init {
  rdkit_init() { rdkit_import_array(); }
}

static void ds_ensure_numpy() {
  static rdkit_init init;
};
```

bp-kelley: This is an excellent write up of why it works:  https://stackoverflow.com/questions/26013650/threadsafe-lazy-initialization-static-vs-stdcall-once-vs-double-checked-locki

pechersky: prefer this new style vs the older existing pattern? do this just for the touched files, or across the board elsewhere?

bp-kelley: I think it's cleaner, but I'll let @greglandrum be the judge.  We use a mix of both so either is fine.

Code/GraphMol/Wrap/rdchem.cpp
pechersky: ifdef style
a5c7eef

Code/DataStructs/Wrap/DataStructs.cpp
pechersky: magic static style 7300188, can always revert

greglandrum: > I think it's cleaner, but I'll let @greglandrum be the judge.  We use a mix of both so either is fine.

I prefer call_once since it's explicit and immediately clear what is intended. (Thus a bit easier to quickly understand while reading the code)

However, I don't have a problem with using statics instead; either way is fine.

pechersky: Reverted back to call_once style

greglandrum:
I'm going to defer this to the 2026.03.1 release since there may be some cases where it affects other people's scripts (?).
That's only a month away, so it's not a huge difference.

PR #9130, opened 2026-02-24
Fix: TautomerEnumerator drops CHI on unmodified sp3 atoms

### LINKED ISSUE #9129: TautomerEnumerator.Enumerate does not match tautomer's stereo that Canonicalize gives
**Describe the bug**
TautomerEnumerator.Canonicalize(mol) returns a tautomer with stereo annotation ([C@@H]) that is not present in TautomerEnumerator.Enumerate(mol). The two APIs disagree on the stereochemistry of the canonical tautomer.

**To Reproduce**
```
import sys
from rdkit import Chem

mol = Chem.MolFromSmiles("O=C1C(O)=C([C@H]2CC[C@H](C)CC2)C(=O)c2ccccc21")
te = rdMolStandardize.TautomerEnumerator()
te.SetRemoveSp3Stereo(False)

canon_smi = Chem.MolToSmiles(te.Canonicalize(mol))
tauts = list(te.Enumerate(mol))
enum_smiles = {Chem.MolToSmiles(t) for t in tauts}

canon_noste = Chem.MolToSmiles(Chem.MolFromSmiles(canon_smi), isomericSmiles=False)
enum_noste = {Chem.MolToSmiles(t, isomericSmiles=False): Chem.MolToSmiles(t) for t in tauts}

print(f"canonical   : {canon_smi}")
print(f"in enum     : {canon_smi in enum_smiles}")
if canon_smi not in enum_smiles:
    same_conn = enum_noste.get(canon_noste)
    print(f"enum has    : {same_conn or '(connectivity not found either)'}  ← same tautomer, stereo stripped")
    # show all enumerate entries for this connectivity (may be >1 epimer)
    matches = [Chem.MolToSmiles(t) for t in tauts
               if Chem.MolToSmiles(t, isomericSmiles=False) == canon_noste]
    if len(matches) > 1:
        print("all enum entries with same connectivity:")
        for s in matches:
            print(f"  {s}")

sys.exit(0 if canon_smi in enum_smiles else 1)
```

**Expected behavior**
I think we should make sure that Canonicalize always gives something that you'd get out of Enumerate, and vice versa, that Enumerate should make sure to include the canonical tautomer in its list. I don't know which is the correct stereo choice to make here, because of the complex non-atom-center stereo. Should the distal methyl lose the stereo flag since it became "meso"?

**Screenshots**
Image

 - RDKit version: Release_2025_09_5, also master
 - OS:  Ubuntu 24.04
 - Are you using conda? no
 - If you are not using conda: how did you install the RDKit? manual build

**Additional context**
Part of regression testing (making sure Enumerate and Canonicalize agree) on registry molecules.

pechersky:
#### Reference Issue
Fixes #9129

When tautomerRemoveSp3Stereo=false, TautomerEnumerator.Enumerate could produce a set that does not contain Canonicalize's result. The root cause: inside setTautomerStereoAndIsoHs, MolOps::assignStereochemistry (cleanIt=true) correctly cleared the CHI tag on an sp3 atom that is not part of the tautomeric path, but whose CIP environment temporarily looked symmetric in an exo-methylene intermediate tautomer. That CHI-stripped tautomer was stored first; when the stereo-bearing version arrived it was deduplicated away (state key is stereo-blind), so the chirality was lost for the rest of the enumeration.

Fix: after assignStereochemistry, for each sp3 atom that (a) was not modified by the tautomeric transform and (b) had its CHI cleared, tentatively restore the CHI from the input molecule and re-run assignStereochemistry. Only keep the restored CHI if the atom receives a _CIPCode or _ringStereoAtoms property, meaning it is truly chiral in this tautomer; otherwise discard it. The round-trip ensures we never hallucinate chirality.

Test: testEnumeratePreservesUnmodifiedSp3Stereo() uses 1-(trans-4-methylcyclohexyl)naphthalene-1,4-dione enol; the trans-4-methylcyclohexyl stereocenter is outside all tautomeric paths but was being silently dropped.

#### Any other comments?
The choice being made here is to make Enumerate more like Canonicalize. The other option would have been to make Canonicalize also drop the stereo, to behave more like Enumerate.


Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -257,6 +257,49 @@ bool TautomerEnumerator::setTautomerStereoAndIsoHs(
...
+    //
+    // Fix: tentatively restore the input-mol CHI for any sp3 atom not in
+    // d_modifiedAtoms whose CHI was cleared, then re-run
+    // assignStereochemistry.  Only keep restored CHI if the center gets a
+    // _CIPCode or _ringStereoAtoms property (truly chiral in this
+    // tautomer); otherwise leave it cleared.
+    if (!d_removeSp3Stereo) {
+      bool anyRestored = false;
+      for (unsigned int idx = 0; idx < mol.getNumAtoms(); ++idx) {
+        if (res.d_modifiedAtoms.test(idx)) continue;
```
greglandrum: We always put brackets around if. Please update here and below (clang-tidy can also do this for you).

Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -257,6 +257,49 @@ bool TautomerEnumerator::setTautomerStereoAndIsoHs(
...
+        auto *tautAtom = taut.getAtomWithIdx(idx);
+        if (molAtom->getChiralTag() == Atom::CHI_UNSPECIFIED) continue;
+        if (tautAtom->getChiralTag() != Atom::CHI_UNSPECIFIED) continue;
+        if (tautAtom->getHybridization() != Atom::SP3) continue;
+        tautAtom->setChiralTag(molAtom->getChiralTag());
+        anyRestored = true;
+      }
+      if (anyRestored) {
+        MolOps::assignStereochemistry(taut, cleanIt, force);
+        for (unsigned int idx = 0; idx < mol.getNumAtoms(); ++idx) {
```
greglandrum: I believe that this loop is not needed.  You're calling `assignStereochemistry()` with the cleanIt argument set, so it should take care of removing stereo flags from any invalid centers. Commenting out this loop does not cause your new test (or any of the other tests) to fail, so I think it's safe to remove.

As an informational point: the "new" (non-legacy) stereochem perception code no longer sets `_CIPCode`, so relying upon that flag to be present is fragile under any circumstances.

greglandrum (changes requested):
There's a logic question/correction and some small formatting changes here.

Code/GraphMol/MolStandardize/testTautomer.cpp
```diff
@@ -1785,6 +1827,10 @@ int main() {
   testTautomerEnumeratorResult_const_iterator();
   testGithub3430();
   testGithub3755();
+<<<<<<< fix/tautomer-enumerate-drops-stereo-of-unmodified-sp3
+  testEnumeratePreservesUnmodifiedSp3Stereo();
+=======
   testCanonicalizePreservesNonTautomericBondStereo();
+>>>>>>> master
   return 0;
```
pechersky:
```suggestion
  testEnumeratePreservesUnmodifiedSp3Stereo();
  testCanonicalizePreservesNonTautomericBondStereo();
  return 0;
```

PR #9132, opened 2026-02-25
Fix some minor annoyances in Python tests

ricrogz:
I usually work on a Qt-based out-of-tree installation, and this patch addresses some issues in Python tests I have been ignoring for a (long) while:

- The `github4823` test is configured to run in-tree, and fails due to a rdkit import error when using an out-of-tree installation with `RDK_INSTALL_INTREE` disabled and `RDK_INSTALL_PYTHON_TESTS` enabled.

~- `UnitTestDraw.py::_testMolToFile()` and `UnitTestDraw.py::_testMolToImage()` were not being run due to the test names being prefixed with an underscore. I renamed them and skipped them when Cairo is not available, since they both require `MolToImage()` at some point. `MolToImage()` is not available on Qt builds due to Cairo/Qt incompatibility.~ Whoops, these aren't meant to be tests, they are functions used by other tests...

- `PandasTools.py::testGithub1507()` (both instances): As explained by the comment, `SaveXlsxFromFrame()` uses `MolToImage()` to generate PNG images that are embedded into the Xlsx file, and due to the incompatibility mentioned above, it won't work in Qt-based installations.

- `UnitTestPandasTools.py::test_github2380()`: when run under `ctest -R pythonSourceTests`, this test fails in a Qt installation due to a test-test interaction with `Chem/Draw/UnitTestIPython.py`: when running `pythonSourceTests`, 
`Chem/Draw/UnitTestIPython.py` runs before `UnitTestPandasTools.py::test_github2380()`. It imports `IPythonConsole`, which sets up the IPython renderer (which allows avoiding the Qt/Cairo incompatibility by doing `Draw.MolsToGridImage = IPythonConsole.ShowMols` and drawing SVGs instead of PNGs), and then uninstalls it as the test ends. Since tests are all run in the same process, when `UnitTestPandasTools.py::test_github2380()` runs, the `from rdkit.Chem.Draw import IPythonConsole` call does not set up the renderer because the `IPythonConsole` module is still loaded, so the init won't trigger again, but since the rendered was uninstalled, `Draw.MolsToGridImage()` has been restored to the default implementation, which uses `MolToImage()`, triggering the failure. My fix ensures that the IPython rendered is installed before `PandasTools.FrameToGridImage()` is called, and then cleans it up after it runs.

### DIFF
--- a/rdkit/Chem/Draw/UnitTestDraw.py
+++ b/rdkit/Chem/Draw/UnitTestDraw.py
@@ -87,8 +87,7 @@ class TestCase(unittest.TestCase):
     # We avoid checking in the code with development flag set
     self.assertFalse(self.showAllImages)
 
-  @unittest.skipIf(not hasattr(rdMolDraw2D, 'MolDraw2DCairo'), 'Skipping cairo test')
-  def testMolToFile(self):
+  def _testMolToFile(self):
     try:
       fhdl, fn = tempfile.mkstemp(suffix='.png')
       # mkstemp returns a file handle that we don't need; close it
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

PR #9133, opened 2026-02-26
Add explicit operator= and copy c'tors to Transform3D and its base classes.

DavidACosgrove:
#### Reference Issue
No issue.

#### Any other comments?
operator= for Transform3D was explicitly removed. I needed it in the GaussianShape code.


DavidACosgrove:
The failure is in rdFMCS, again.  Maybe I'll try and find time to work out why this happens so often!

greglandrum:
> The failure is in rdFMCS, again.  Maybe I'll try and find time to work out why this happens so often!

It's a timeout since the CI machines can be super slow; we just need to adjust that timeout

greglandrum (approved):
LGTM

PR #9135, opened 2026-02-26
Small refactor before PR #8968 follow up

ricrogz:
This is a small refactor ahead of a follow up to #8968 I'm working on. It does:

- Refactor & modernize `clearBondDirs()` and `removeRedundantBondDirSpecs()`. It also removes the `&bondVisitOrders` argument from them, which isn't used.

- Abstract the check in `catch_canon.cpp`'s "Canonicalization issues watch" test into a separate function, and add a stricter test that checks actual CIP labels instead of just the feature counts (this is the version I'm using during development). Both these are useful to set up smaller and stricter tests when debugging/developing.


greglandrum (approved):
LGTM

PR #9136, opened 2026-02-27
Small bug fix in PR #8968

ricrogz:
This small bug slipped through in the review of #8968:

https://github.com/rdkit/rdkit/blob/8e76c2778436b3fa368dcaf2fae86fd5f6e233d7/Code/GraphMol/Canon.cpp#L301-L314

In Line 303 we increment `bondDirCounts[firstFromAtom1->getIdx()]`, and then we check it in line 310. Obviously, the check will never be triggered (even when it should), due to the increment. There's a similar check for `firstFromAtom2` further down. The fix is just to increment after the checks have taken place.

Luckily, it's infrequent to hit this block (only when directions on both sides have been set already), and even if that's the case, it should only trigger when `bondDirCounts[firstFromAtom1->getIdx()] == 0`.


greglandrum (approved):
LGTM

PR #9138, opened 2026-03-02
Add RDLog::CaptureLog for capturing log messages

cdvonbargen:

Adds an RAII `CaptureErrorLog` class to `namespace RDLog` (alongside the existing `LogStateSetter`) that redirects `rdErrorLog`'s output to an internal `std::stringstream` for the duration of its lifetime. On destruction the original stream destination and enabled state are fully restored. Nesting is supported: an inner capture shadows the outer one and each collects its own messages independently. The log is always enabled while a capture is active, even if a `LogStateSetter` has silenced it.

The Python bindings expose the class as `rdBase.CaptureErrorLog`, following the same context-manager pattern as the existing `BlockLogs`. The `messages` property is accessible both inside and after the context.

The implementation uses the stream-swap trick already present in `rdLogger`: on construction the logger's `dp_dest` and `teestream` pointers are swapped with an internal `stringstream`, and swapped back on destruction.

### DIFF
--- a/Code/RDGeneral/RDLog.cpp
+++ b/Code/RDGeneral/RDLog.cpp
@@ -149,24 +149,28 @@ void InitLogs() {
   rdErrorLog = std::make_shared<boost::logging::rdLogger>(&std::cerr);
 }
 
-CaptureErrorLog::CaptureErrorLog()
-    : d_savedDest(&d_messages), d_savedTeestream(nullptr) {
-  if (!rdErrorLog) {
-    InitLogs();
+CaptureLog::CaptureLog(RDLogger log) : d_log(std::move(log)) {
+  if (!d_log) {
+    return;
   }
-  d_logWasEnabled = rdErrorLog->df_enabled;
-  rdErrorLog->df_enabled = true;
-  std::swap(rdErrorLog->dp_dest, d_savedDest);
-  std::swap(rdErrorLog->teestream, d_savedTeestream);
+  d_logWasEnabled = d_log->df_enabled;
+  d_log->df_enabled = true;
+  d_savedDest = d_log->dp_dest;
+  d_log->dp_dest = &d_messages;
+  d_savedTeestream = d_log->teestream;
+  d_log->teestream = nullptr;
 }
 
-CaptureErrorLog::~CaptureErrorLog() {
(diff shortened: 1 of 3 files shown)


bp-kelley:
Previously, the way to do this in python was:

```
>>> sys.stderr = io.StringIO()
>>> from rdkit.Chem import rdBase
>>> rdBase.WrapLogs()
>>> Chem.MolFromSmiles("asdfasdfs")
>>> print(sys.stderr.getvalue())
[12:22:37] SMILES Parse Error: syntax error while parsing: asdfasdfs
[12:22:37] SMILES Parse Error: check for mistakes around position 1:
[12:22:37] asdfasdfs
[12:22:37] ^
[12:22:37] SMILES Parse Error: Failed parsing SMILES 'asdfasdfs' for input: 'asdfasdfs'
```

But this had the issues that the errors went to C stderr as well.  Another issue was that, on windows, we never knew when things were going to be flushed.  Hopefully this solves both problems.

bp-kelley:
It might be useful to also have
```
capture.error_messages
capture.debug_message
...
```

if this ends up working.

cdvonbargen:
> capture.error_messages
> capture.debug_message

My instinct is to leave this to leave this an error capture log for now, keep this minimal, and we can extend if the need arises. Does that make sense?

bp-kelley:
I think that’s fine, a user can always disable/enable various logs using a the standard call

cdvonbargen:
> I like the change and can definitely see how it will be useful. I think it would be improved if the new capture class wasn't specific to `rdErrorLog` but allowed the log to be captured to be supplied in the ctor. So you could do: `CaptureLog cl{rdErrorLog};` or `CaptureLog cl{rdWarningLog};`

Done. I included a convenience subclass CaptureErrorLog since I suspect that is still the most useful -- but lemme know if you think I should remove it

greglandrum:
> Done. I included a convenience subclass CaptureErrorLog since I suspect that is still the most useful -- but lemme know if you think I should remove it

Nope! I was actually going to suggest exactly that but forgot it in my review. Thanks!

greglandrum (approved):
LGTM

PR #9154, opened 2026-03-05
Fixes #9153

### LINKED ISSUE #9153: Incorrect double bond CIP label for bond with adjacent double bond
Example:
```python
   ...: m = Chem.MolFromSmiles(r'S=P(=N\C)/C')
   ...: m.Debug()
   ...: Chem.AssignCIPLabels(m)
   ...: print(m.GetBondWithIdx(1).GetProp('_CIPCode'))
   ...: 
# Atoms:
#     0 16 S chg: 0  deg: 1 exp: 2 imp: 0 hyb: SP2
#     1 15 P chg: 0  deg: 3 exp: 5 imp: 0 hyb: SP2
#     2 7 N chg: 0  deg: 2 exp: 3 imp: 0 hyb: SP2
#     3 6 C chg: 0  deg: 1 exp: 1 imp: 3 hyb: SP3
#     4 6 C chg: 0  deg: 1 exp: 1 imp: 3 hyb: SP3
# Bonds:
#     0 0->1 order: 2
#     1 1->2 order: 2 stereo: Z ats: (0 3)     <----- STEREOZ with stereo atoms (C, S)
#     2 2->3 order: 1 dir: \
#     3 1->4 order: 1
# Z     <----- CIP label is Z (ok)

In [2]: smi = Chem.MolToSmiles(m)
   ...: print(smi)
   ...: 
# C/N=P(/C)=S

In [3]: m2 = Chem.MolFromSmiles(smi)
   ...: m2.Debug()
   ...: Chem.AssignCIPLabels(m2)
   ...: print(m2.GetBondWithIdx(1).GetProp('_CIPCode'))
   ...: 
# Atoms:
#     0 6 C chg: 0  deg: 1 exp: 1 imp: 3 hyb: SP3
#     1 7 N chg: 0  deg: 2 exp: 3 imp: 0 hyb: SP2
#     2 15 P chg: 0  deg: 3 exp: 5 imp: 0 hyb: SP2
#     3 6 C chg: 0  deg: 1 exp: 1 imp: 3 hyb: SP3
#     4 16 S chg: 0  deg: 1 exp: 2 imp: 0 hyb: SP2
# Bonds:
#     0 0->1 order: 1 dir: /
#     1 1->2 order: 2 stereo: Z ats: (0 4)     <----- Still STEREOZ with same stereo atoms (C, S)
#     2 2->3 order: 1
#     3 2->4 order: 2
# E     <----- CIP label switched to E !!!!
```

The initial SMILES comes from importing https://pubchem.ncbi.nlm.nih.gov/compound/16129835 as mol file, exporting to SMILES, and replacing the rings with "C" for simplicity.

I'm not quite sure we should see double bond stereo here (Pubchem says "Defined Bond Stereocenter Count = 0" /  "Undefined Bond Stereocenter Count = 0"), but since we do, the CIP labels should be correct...

ricrogz:
This fixes #9153 .

The issue is that in the CIP labeler graphs, double bonds are represented by a duplicate edge. So, in this case, when the edges are sorted by CIP priority, the double bond to S (i.e. the two duplicate edges) are ranked first, meaning that 
`edges2[0] == edges2[1] != edges2[2]`  (2 edges to S, 1 edge to C), so that by checking against `edges2[1]` we aren't actually checking against the second atom, but still against the first atom at the end of second edge of the double bond.

The fix is as simple as, instead of checking for equality against the second atom, check for inequality to the first one.

I think this bug has been here since the CIP labeler was ported from Java, since we don't have any tests that check this case. Interestingly, the [CIP Validation Suite](https://github.com/CIPValidationSuite/ValidationSuite)  doesn't include any examples similar to this one, which makes me wonder whether this should be detected as a stereo bond or not.


ricrogz:
> Your code change looks good and I'd be fine with merging it (since it could maybe be relevant for some other system), but I would do it without any additional tests.

I removed them!

PR #9155, opened 2026-03-05
More refactoring in Canon.cpp

ricrogz:
This does:

- Convert `bondDirCounts` and `atomDirCounts` into a ~`vector<char>`~ `vector<int8_t>`, since the counts are expected to be low (they should be 2 at most, but the accounting is bugged, and they can go higher; I'll fix this in an upcoming PR that I'm still working on).

-  Remove the `cycles` variable, which is passed around, but never used or changed.

- Move `ringStereoChemAdjusted` into the `if (doIsomericSmiles)`, since it's never used outside of it.

- Move the declaration of `cyclesAvailable` into `canonicalizeFragment`, change it's type to a bitset (it's always 0/1), and change `MAX_CYCLES` to `1024`, to make full use of the bitset. At some point, we should make this allocation dynamic, or at least, add a check on the number of rings, just in case someone tries to work with something with LOTS of rings...

- Fix the way we create a copy of `colors` in `canonicalDFSTraversal()`: the way it was done, we ended up allocating twice as many `tcolors` as we need, leaving half of them with the default value (this was probably a typo, using `tcolors.resize(colors.size());` instead of tcolors.reserve(colors.size());).

- Remove `directTravList` and `cycleEndList` in `dfsBuildStack()`, which are never used.
 
- modernizing the loop in `isUnsaturated()`

- assigning `q->getDescription()` to a `const std::sting &` in `hasSingleHQuery()` to avoid making copies of the description string.

- Some other minor changes.

### DIFF
--- a/Code/GraphMol/Canon.cpp
+++ b/Code/GraphMol/Canon.cpp
@@ -163,8 +163,8 @@ auto _possibleCompare = [](const PossibleType &arg1, const PossibleType &arg2) {
 //
 void canonicalizeDoubleBond(Bond *dblBond, const UINT_VECT &bondVisitOrders,
                             const UINT_VECT &atomVisitOrders,
-                            std::vector<char> &bondDirCounts,
-                            std::vector<char> &atomDirCounts) {
+                            std::vector<int8_t> &bondDirCounts,
+                            std::vector<int8_t> &atomDirCounts) {
   PRECONDITION(dblBond, "bad bond");
   PRECONDITION(dblBond->getBondType() == Bond::DOUBLE, "bad bond order");
   PRECONDITION(dblBond->getStereo() > Bond::STEREOANY, "bad bond stereo");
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

Code/GraphMol/Canon.h
```diff
@@ -17,9 +17,9 @@ class Atom;
 class Bond;
 
 namespace Canon {
-const int MAX_NATOMS = 5000;  //!< used in the canonical traversal code
-const int MAX_CYCLES = 1000;  //!< used in the canonical traversal code
-const int MAX_BONDTYPE = 32;  //!< used in the canonical traversal code
+constexpr int MAX_NATOMS = 5000;  //!< used in the canonical traversal code
+constexpr int MAX_CYCLES = 1024;  //!< used in the canonical traversal code
```
greglandrum, line 21: :-)

Issue #9151, opened 2026-03-04
PandasTools.SaveXlsxFromFrame is ignoring the remaining data in the dataframe when writing the .xlsx file


lmsigmund:
Hi RDKit team,

**Describe the bug**
When an MS Excel file is generated with images of the Lewis structures of molecules and additional columns are present in this dataframe, their data will not show up in the resulting .xlsx file.

**To Reproduce**
```python
from rdkit import Chem
from rdkit.Chem import PandasTools
import pandas as pd

smis = ["NCCO", "c1ccccc1O", "c1ncccc1"]
df = pd.DataFrame(smis, columns=["smiles"])
df["structure"] = df["smiles"].apply(lambda x: Chem.MolFromSmiles(x))
df["names"] = ["mol1", "mol2", "mol3"]

PandasTools.SaveXlsxFromFrame(df, "molecules.xlsx", molCol="structure")
```
When the above given code is run, the resulting Excel file includes the images of the Lewis structures as expected but not the other data. The headers of all columns are present, though.

Thank you. :-)

**Expected behavior**
The additional data, that is the "smiles" column and the "name" column data should be part of the Excel file.

**Configuration (please complete the following information):**
 - RDKit version: 2025.09.6
 - pandas version: 3.0.1
 - OS: Rocky Linux 8.10 (Green Obsidian)
 - Python version: 3.13.3
 - Are you using conda? yes
 - If you are using conda, which channel did you install the rdkit from? pip
 - If you are not using conda: how did you install the RDKit?

ptosco:
I cannot reproduce that with `pandas` 2.2.3, so I guess something has broken with `pandas` 3. I will try later on a `pandas` 3 environment and provide a fix.

ptosco:
Confirmed, I can reproduce the problem with `pandas` 3,0.1.

greglandrum:
Fixed in #9160

PR #9160, opened 2026-03-08
Fix GitHub #9151

ptosco:
Fix GitHub #9151 by making `PandasTools.SaveXlsxFromFrame()` work with pandas 3 and `str` `dtype`.


bp-kelley:
@greglandrum this was a simple fix, I'll go ahead and merge it.

PR #9162, opened 2026-03-10
Fixes a boundary issue when indexing SDF

### LINKED ISSUE #9101: SDMolSupplier length/indexing regression in 2025.09.5 when "$$$$\n" crosses 64KiB chunk boundary
Hi RDKit team,

I found a bug in `SDMolSupplier` on `2025.09.5` that appears related to the buffered indexing change from PR #9010.

## Summary

When using:

```python
sup = Chem.SDMolSupplier(path, removeHs=False)
for i, mol in zip(range(len(sup)), sup):
    ...
```

`len(sup)` triggers index building, and if an SDF record separator `$$$$` is at the end of a 65536-byte chunk while its trailing `\n` is in the next chunk, parsing of a later molecule can fail with:

- `ERROR: Counts line too short: ''`
- `ERROR: moving to the beginning of the next molecule`

## Reproduction
`rdkit_bug_buffer.sdf` is attached (pls remove the .txt suffix).

```python
from rdkit import Chem
import rdkit

print(rdkit.__version__)  # 2025.09.5

fp = "rdkit_bug_buffer.sdf"

# A: direct streaming iteration (works)
sup = Chem.SDMolSupplier(fp, removeHs=False)
print([i for i, m in enumerate(sup) if m is None])  # []

# B: len()+iteration (fails)
sup = Chem.SDMolSupplier(fp, removeHs=False)
L = len(sup)
print(L)  # 2
print([i for i, m in zip(range(L), sup) if m is None])  # [1]
```

## Version comparison

- `2025.09.5`: reproduces
- `2025.09.4`: does not reproduce with the same file and code
- `2025.09.3`: does not reproduce with the same file and code

## Suspected root cause

In `SDMolSupplier::buildIndexTo()`, when `$$$$` is found near the end of a buffer:

```cpp
char *nlPos = match + 4;
while (nlPos < bufEnd && *nlPos != '\n') ++nlPos;
if (nlPos < bufEnd) ++nlPos;
```

If `\n` is in the next chunk, `nlPos == bufEnd`, so `++nlPos` is not executed. Then the newline start is "\n", not the real new line start. This causes downstream Mol block misalignment and the `"Counts line too short"` parse error.

[rdkit_bug_buffer.sdf.txt](https://github.com/user-attachments/files/25203459/rdkit_bug_buffer.sdf.txt)

bp-kelley:
Fixes #9101

### DIFF
--- a/Code/GraphMol/Wrap/rough_test.py
+++ b/Code/GraphMol/Wrap/rough_test.py
@@ -8687,7 +8687,7 @@ M  END
 
   def testGithub9101(self):
     fileN = os.path.join(RDConfig.RDBaseDir, 'Code', 'GraphMol', 'FileParsers', 'test_data',
-                         'rdkit_chunk_boundary_bug.sdf.txt')
+                         'rdkit_chunk_boundary_bug.sdf')
     sdSup = Chem.SDMolSupplier(fileN)
     mols = list(sdSup)
     self.assertEqual(len(mols[0].GetPropsAsDict()["comment"]), 65369)


bp-kelley:
This was a regression caused in #9010  to trigger the bug takes a particularly well crafted SDF file and was lucky to have been spotted by @kongexp

Code/GraphMol/FileParsers/testMolSupplier.cpp
```diff
@@ -2704,4 +2704,16 @@ TEST_CASE("Read SD properties till last '>'") {
   auto m2 = supplier.next();
   REQUIRE(m2);
   CHECK(m2->hasProp(prop_name));
-}
\ No newline at end of file
+}
+
+TEST_CASE("github9101 - $$$$ at buffer end") {
+  std::string infile = rdbase + "/Code/GraphMol/FileParsers/test_data/rdkit_chunk_boundary_bug.sdf.txt";
```
ricrogz:
```suggestion
  std::string infile = rdbase + "/Code/GraphMol/FileParsers/test_data/rdkit_chunk_boundary_bug.sdf";
```

Code/GraphMol/Wrap/rough_test.py
```diff
@@ -8685,6 +8685,14 @@ def testVerifyWedges(self):
                     if b.GetBondDir() in (Chem.BondDir.BEGINWEDGE, Chem.BondDir.BEGINDASH))
     self.assertEqual(numWedges, 1)
 
+  def testGithub9101(self):
+    fileN = os.path.join(RDConfig.RDBaseDir, 'Code', 'GraphMol', 'FileParsers', 'test_data',
+                         'rdkit_chunk_boundary_bug.sdf.txt')
```
ricrogz:
```suggestion
    fileN = os.path.join(RDConfig.RDBaseDir, 'Code', 'GraphMol', 'FileParsers', 'test_data',
                         'rdkit_chunk_boundary_bug.sdf')
```

Code/GraphMol/FileParsers/testMolSupplier.cpp
bp-kelley: Indeed I uploaded the wrong one...

bp-kelley:
@ricrogz I uploaded the wrong one :)

PR #9168, opened 2026-03-12
Follow up to PR #8968

ricrogz:
This implements the (hopefully) last batch of changes in the stereo bond canonicalization that started with PR #8968. Some of these changes, and especially the new tests I added are based on a benchmark I ran on a subset of Pubchem (all mols with double bonds with stereo).

This patch implements:
- stereo bond dir conflict detection and handling (this was the "TO DO" @d-b-w mentioned in #8968). Note that not all conflicts are fixable. We'll fix then when we can, otherwise, we'll log a warning and continue.

- Removal of "unwanted" bond directions: as the comment above `removeUnwantedBondDirSpecs()` says, trying to add all possible directions may force a stereo label on  STEREONONE / STEREOANY bonds. I added some code that attempts to repair such cases. Again, that's not always possible.

- Fix atom/dir counts: after PR #8968 these counts were not always correct, causing `removeRedundantBondDirSpecs()` to remove some directions it shouldn't remove, and stripping stereo from some bonds. Pubchem 101176040 was one of such cases.

- Added the  `ENABLE_EXTRA_CHECKS` macro at the top of `Canon.cpp`. Switching it on enables extra checks on the atom / dir counts, and causes unsolvable direction conflicts to throw. These checks are helpful during debugging and development.

### DIFF
--- a/Code/GraphMol/Canon.cpp
+++ b/Code/GraphMol/Canon.cpp
@@ -195,7 +195,7 @@ bool handleDirConflictsAcrossDoubleBond(
         firstFromAtom2, isFirstFromAtom2Flipped);
     return expectedFirstFromAtom2Dir == firstFromAtom2.getBondDir();
 
-  } else if (!atom2DirsAreConsistent) {
+  } else if (!atom2DirsAreConsistent && atom1DirsAreConsistent) {
     // atom2 has conflicting directions, which means we must have a
     // secondFromAtom2. We don't know anything about secondFromAtom1:
     // it might be present or not, but we don't care about it, since
@@ -205,7 +205,7 @@ bool handleDirConflictsAcrossDoubleBond(
         secondFromAtom2, isSecondFromAtom2Flipped, atom1, firstFromAtom1,
         isFirstFromAtom1Flipped, bondDirCounts, atomDirCounts);
 
-  } else if (!atom1DirsAreConsistent) {
+  } else if (!atom1DirsAreConsistent && atom2DirsAreConsistent) {
     // atom1 has conflicting directions, which means we must have a
     // secondFromAtom1. We don't know anything about secondFromAtom2:
     // it might be present or not, but we don't care about it, since


Code/GraphMol/Canon.cpp
```diff
@@ -36,9 +40,60 @@ static constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
   }
 }
 
-void setDirectionFromNeighboringBond(Bond &sourceBond, bool isSourceBondFlipped,
-                                     Bond &targetBond,
-                                     bool isTargetBondFlipped) {
+#if ENABLE_EXTRA_CHECKS
+static void checkDirCounts(const ROMol &mol,
```
greglandrum: this is in an anonymous namespace, I think it doesn't need to be static
This also applies to all the other static functions in this PR (if they aren't in an anonymous namespace they should be so that we don't need the static keyword)

Code/GraphMol/Canon.cpp
```diff
@@ -36,9 +40,60 @@ static constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
...
+                           const std::vector<int8_t> &bondDirCounts,
+                           const std::vector<int8_t> &atomDirCounts) {
+  // atoms at the end of double bonds can only have 2 bons with directions
```
greglandrum:
```suggestion
  // atoms at the end of double bonds can only have 2 bonds with directions
```

Code/GraphMol/Canon.cpp
```diff
@@ -36,9 +40,60 @@ static constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
...
+          "Conflicting single bond directions around double bond at index ") +
+      std::to_string(idx) + ".\n  Stereochemistry may be incorrect.";
+  BOOST_LOG(rdWarningLog) << msg << std::endl;
+
+#if ENABLE_EXTRA_CHECKS
+  CHECK_INVARIANT(false, msg);
+#endif
+}
+
+static void setDirectionFromNeighboringBond(Bond &sourceBond,
```
greglandrum: I know you didn't change this one but:

```suggestion
static void setDirectionFromNeighboringBond(const Bond &sourceBond,
```

Code/GraphMol/Canon.cpp
```diff
@@ -91,6 +145,144 @@ Bond::BondDir getReferenceDirection(const Bond &dblBond, const Atom &refAtom,
...
+    // atom1 has conflicting directions, which means we must have a
+    // secondFromAtom1. We don't know anything about secondFromAtom2:
+    // it might be present or not, but we don't care about it, since
+    // we know it is not problematic.
+    return fixConflictAcrossDoubleBond(
+        dblBond, atom1, firstFromAtom1, isFirstFromAtom1Flipped,
+        secondFromAtom1, isSecondFromAtom1Flipped, atom2, firstFromAtom2,
+        isFirstFromAtom2Flipped, bondDirCounts, atomDirCounts);
+
+  } else {
```
greglandrum, line 218: I don't think you ever get to this else?
You'd need to change 197 to:
```
  } else if (!atom2DirsAreConsistent && atom1DirsAreConsistent) {
```
and 207 to:
```
  } else if (!atom1DirsAreConsistent && atom2DirsAreConsistent) {
```

greglandrum:
@ricrogz did you check to see if this has a noticeable performance impact on SMILES writing?

Code/GraphMol/Canon.cpp
ricrogz: Oh, right!

PR #9169, opened 2026-03-13
switch the Query infrastructure to use std::function instead of function pointers

greglandrum:
The use of function pointers goes back to the very beginning; switching to `std::function` gets us additional flexibility.

### DIFF
--- a/Code/Query/Query.h
+++ b/Code/Query/Query.h
@@ -159,7 +159,6 @@ class RDKIT_QUERY_EXPORT Query {
   CHILD_VECT d_children;
   bool df_negate{false};
   std::function<bool(MatchFuncArgType)> d_matchFunc;
-  // bool (*d_matchFunc)(MatchFuncArgType);
 
   // MSVC complains at compile time when TypeConvert(MatchFuncArgType what,
   // Int2Type<false>) attempts to pass what (which is of type MatchFuncArgType)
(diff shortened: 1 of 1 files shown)


Code/Query/Query.h
```diff
@@ -153,16 +158,19 @@ class RDKIT_QUERY_EXPORT Query {
   std::string d_queryType = "";
   CHILD_VECT d_children;
   bool df_negate{false};
-  bool (*d_matchFunc)(MatchFuncArgType);
+  std::function<bool(MatchFuncArgType)> d_matchFunc;
+  // bool (*d_matchFunc)(MatchFuncArgType);
```
ricrogz: Can this be deleted?

Code/Query/Query.h
```diff
@@ -153,16 +158,19 @@ class RDKIT_QUERY_EXPORT Query {
...
   // as parameter of d_dataFunc() (which should be of type DataFuncArgType). The
   // union is but a trick to avoid silly casts and keep MSVC happy when building
   // DLLs
   union {
-    MatchFuncArgType (*d_dataFunc)(DataFuncArgType);
-    MatchFuncArgType (*d_dataFuncSameType)(MatchFuncArgType);
+    std::function<MatchFuncArgType(DataFuncArgType)> d_dataFunc;
+    std::function<MatchFuncArgType(MatchFuncArgType)> d_dataFuncSameType;
+    // MatchFuncArgType (*d_dataFunc)(DataFuncArgType);
+    // MatchFuncArgType (*d_dataFuncSameType)(MatchFuncArgType);
```
ricrogz: Can we delete this?

Issue #9143, opened 2026-03-02
ETKDGv3 generating 90 degree twisted amides


dkranthi221:
Hello,

I am using RDKit ETKDGv3() to generate conformations of some macrocycles. Here are the two smiles strings I am using:
O=C1NCC=CC[C@@H](NC(=O)c2c[nH]cc2Br)CC(=O)N2CCCC2C2CCN(C2)C(=O)c2csc1n2 smiles_1
CC1(C)CN2CCC1NC(=O)C[C@H](NC(=O)c1cnc(N)cn1)CC=CCOc1cccc(c1)C2=O smiles_2

When I use the following code to generate conformations of these molecules, I get very strange amides twisted 90 degrees out of plane. I used an orthogonal method (Conformator) to generate conformations of these molecules and got planar amides all around. I am curious as to why this is happening and how I could fix it. Also attached are some photos of the twisted amide.

Code:
```python
smiles = 'O=C1NCC=CC[C@@H](NC(=O)c2c[nH]cc2Br)CC(=O)N2CCCC2C2CCN(C2)C(=O)c2csc1n2'
mol_init_test = Chem.MolFromSmiles(smiles)
mol_elem_test = Chem.AddHs(mol_init_test)
params0 = AllChem.ETKDGv3()   
params0.useBasicKnowledge = True      
params0.useMacrocycleTorsions = True  #these are always true but just in case...
params0.ETversion = 2
params0.randomSeed = 210186
AllChem.EmbedMolecule(mol_elem_test,params0)
``` 

Twisted amide images of the same macrocycle:
![Image](https://github.com/user-attachments/assets/167b2800-48ff-4c72-bc13-12f7325bd5fe)

<img width="396" height="436" alt="Image" src="https://github.com/user-attachments/assets/a7f3c123-d742-4c41-8261-7e48310d1ed1" />

I would be grateful if anyone has any ideas on how to fix this!
Thanks!!

greglandrum:
Thanks for the bug report @dkranthi221.

The problem here is that the ET term that we use to provide reasonable amide geometries in macrocycles unfortunately does not match the amide bonds where the N is in a five ring. We should be able to fix this relatively easily for the next release.

[...]

PR #9172, opened 2026-03-14
Add 'k' extension to SMARTS to support ringsize queries

greglandrum:
The `r` primitive in SMARTS is for minimum SSSR ring size, so for this molecule:
image
we get results like this:
```
In [5]: m = Chem.MolFromSmiles('C1CC2N1CCCCCC2')

In [6]: m.GetSubstructMatches(Chem.MolFromSmarts('[r8]'))
Out[6]: ((4,), (5,), (6,), (7,), (8,), (9,))
```
i.e. atoms 2 and 3 do not match the `r8` search; they would match an `r4` search:
```
In [3]: m.GetSubstructMatches(Chem.MolFromSmarts('[r4]'))
Out[3]: ((0,), (1,), (2,), (3,))
```
This PR adds the new primitive `k` that just uses SSSR ring size:
```
In [7]: m.GetSubstructMatches(Chem.MolFromSmarts('[k8]'))
Out[7]: ((2,), (3,), (4,), (5,), (6,), (7,), (8,), (9,))
```

This is part of fixing #9143, but I think it will be generally useful.

### DIFF
--- a/Code/GraphMol/QueryOps.cpp
+++ b/Code/GraphMol/QueryOps.cpp
@@ -1003,36 +1003,45 @@ Atom *replaceAtomWithQueryAtom(RWMol *mol, Atom *atom) {
   return mol->getAtomWithIdx(idx);
 }
 
+enum class RangeQueryType : char {
+  EQUAL,
+  LESS,
+  GREATER,
+  RANGE
+};
 void finalizeAtomRingSizeQuery(Queries::Query<int, Atom const *, true> *query,
-                               const std::string &prefix) {
-  if (prefix.empty()) {
-    auto tgt = static_cast<ATOM_EQUALS_QUERY *>(query)->getVal();
-    query->setDataFunc(
-        [tgt](Atom const *at) { return queryAtomIsInRingOfSize(at, tgt); });
-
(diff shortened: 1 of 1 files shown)


Code/GraphMol/QueryOps.cpp
```diff
@@ -1046,28 +1003,61 @@ Atom *replaceAtomWithQueryAtom(RWMol *mol, Atom *atom) {
...
+    throw ValueErrorException("Do not know how to handle prefix: '" + prefix +
+                              "'");
+  }
+}
+
 void finalizeQueryFromDescription(
     Queries::Query<int, Atom const *, true> *query, Atom const *) {
   std::string descr = query->getDescription();
 
+  std::string prefix;
```
ricrogz: Given the limited number of choices here, and that the prefix is never shown to the user, would it make sense to use an enum class for this, like:
```c++
enum class : char  { EQUAL, LESS, GREATER, RANGE };
```

greglandrum:
> LGTM. Just a small suggestion to use an enum instead of a string.

Makes sense. The suggested changes are in

Issue #9023, opened 2025-12-22
Legend position


PaulC61:
Very simple really, I'd like to be able to move the legend position to an associated molecule to the side or on top of the molecule instead of directly below. I think it would be useful in terms of annotation for plots for mols/fragments.

greglandrum:
@DavidACosgrove will have to confirm, but I think adding the option to put it at the top should be relatively straightforward, but off the side is probably more difficult.

DavidACosgrove:
It would definitely be useful and I’ve thought about it before. I don’t think it’s as straightforward as it should be. Having said that, we’ve changed things a bit since I last thought about it. I’ll try and find time in the next couple of weeks to have a look at it unless someone else wants to have a go.

PaulC61:
Thanks for the response guys! For now, I'm okay - I'm essentially trying to create "fuzzy logic" models of sub-structures.
In this case a substructure can be accompanied by some logical AND or AND NOT operator. It's okay to put them in the legend below the molecule but I think

"SubStructure1 AND SubStructure2 AND NOT SubStructure7" with the legend on the side containing the logic is more "sentence-like", neat and tidy. Might be helpful for others working with logic-based models or CART. So not essential but "nice to have"

DavidACosgrove:
Hi @PaulC61.  What do you mean by "on the side"? I interpreted that as down the left or right margin, with a vertical string:
```
S
u
b
s
t
.
.
.
```
That would be achievable but not straightforward.  It is possible to put text directly on the canvas yourself, as described at
https://greglandrum.github.io/rdkit-blog/posts/2025-03-07-drawing-on-drawings.html
I'm not suggesting this is as or more convenient than a "LEGEND_TOP" type flag, but it is a potential workround for now.

greglandrum:
Fixed by #9183

PR #9183, opened 2026-03-17
MolDraw2D: configurable legend position and vertical side legends (Issue #9023)

Brandon-Cole:
#### Reference Issue
Addresses issue #9023: legend position request (Top / Left / Right / Bottom, optional vertical).

This adds configurable legend placement for 2D drawings so the legend can be shown
**above, below, to the left, or to the right** of a molecule, with optional
**vertical text** on the left/right. This is useful for annotating plots of
molecules or fragments where a bottom-only legend overlaps other content.

- **API / options**
  - `MolDrawOptions.legendPosition` (`Bottom` [default], `Top`, `Left`, `Right`)
  - `MolDrawOptions.legendVerticalText` (for Left/Right, one character per line)
  - JSON: `"legendPosition"` and `"legendVerticalText"` keys

- **Layout behavior**
  - Left/Right legends use a dedicated side strip (controlled by `legendFraction`).
  - Horizontal side legends reserve a **small horizontal gutter** between the
    molecule and the legend so text no longer touches the drawing.
  - Vertical side legends scale the font so a long stacked name still fits within
    the drawing panel.

- **Bindings / tests / docs**
  - Exposed options in the Python `rdMolDraw2D` bindings and `Draw` helpers.
  - C++ Catch tests and Python `UnitTestDraw.testLegendPosition` exercise the new options.
  - `ReleaseNotes.md` entry documents the feature.

grid_legend_left_vertical
grid_legend_top
single_legend_right

#### Any other comments?

# Python usage example:

```python
from rdkit import Chem
from rdkit.Chem import Draw, rdMolDraw2D

opts = rdMolDraw2D.MolDrawOptions()
```
# Default: legend at bottom
```python
svg_bottom = Draw.MolToSVG(mol, legend="Ethanol", drawOptions=opts)
```
# Legend at top
```python
opts.legendPosition = rdMolDraw2D.LegendPosition.Top
svg_top = Draw.MolToSVG(mol, legend="Ethanol", drawOptions=opts)
```
# Legend on the left, drawn vertically
```python
opts.legendPosition = rdMolDraw2D.LegendPosition.Left
opts.legendVerticalText = True
svg_left_vertical = Draw.MolToSVG(mol, legend="Ethanol", drawOptions=opts)
```
# Legend on the right, horizontal, with a small gutter to avoid overlap
```python
opts.legendPosition = rdMolDraw2D.LegendPosition.Right
opts.legendVerticalText = False
svg_right = Draw.MolToSVG(mol, legend="Ethanol", drawOptions=opts)
```

Thanks I saw the issue and it's something I had played around with before for powerpoint presentations of MolsToGridImage exports! Hope this is helpful!

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawMol.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawMol.cpp
@@ -1197,8 +1197,11 @@ void DrawMol::calculateScale() {
          drawOptions_.legendPosition == MolDrawOptions::LegendPosition::Right) &&
         !drawOptions_.legendVerticalText;
     if (sideLegendHoriz && molWidth_ > 0) {
+      // Keep a minimum absolute side gap of 8 px so labels do not crowd the
+      // horizontal side legend on small canvases.
       const double g =
           std::max(8.0, marginPadding_ * static_cast<double>(drawWidth_));
+      // Require at least ~24 px of drawable molecule width after the gap.
       if (double(molWidth_) > g + 24.0) {
         widthForScale = double(molWidth_) - g;
       }
@@ -1850,8 +1853,12 @@ double calc_line_gap(const DrawText &textDrawer, bool vertText,
                      double relFontScale) {
   const double font_px = textDrawer.fontSize() * relFontScale;
   if (vertText) {
+    // Vertical legend text needs more inter-glyph breathing room.
+    // 0.35*font_px was tuned to avoid glyph overlap across common fonts.
     return std::max(1.0, font_px * 0.35);
   }
+  // Horizontal legend lines can be tighter; keep a 2 px minimum so short
+  // legends remain readable on small canvases.
   return std::max(2.0, font_px * 0.15);
 }
 
@@ -2089,7 +2096,7 @@ void DrawMol::extractLegend() {
                drawOptions_.legendPosition ==
                    MolDrawOptions::LegendPosition::Top) {
       double extra_padding = total_height * marginPadding_;
-      extra_padding = extra_padding < 2.0 ? 2.0 : extra_padding;
+      extra_padding = std::max(extra_padding, 2.0);
       legendHeight_ = int(total_height + extra_padding);
       drawHeight_ += legendHeight_;
       height_ += legendHeight_;
@@ -3165,6 +3172,8 @@ void DrawMol::getDrawTransformers(Point2D &trans, Point2D &scale,
     case MolDrawOptions::LegendPosition::Left:
       if (!legend_.empty() && !drawOptions_.legendVerticalText &&
           molWidth_ > 0) {
+        // Same minimum side gap used in initDrawMolecule() so scaling and final
+        // placement stay consistent for horizontal side legends.
         const double g =
             std::max(8.0, marginPadding_ * static_cast<double>(drawWidth_));
         // Drawn x spans [toCentre.x, toCentre.x + scaledRanges.x]; leave g past
@@ -3177,6 +3186,7 @@ void DrawMol::getDrawTransformers(Point2D &trans, Point2D &scale,
     case MolDrawOptions::LegendPosition::Right:
       if (!legend_.empty() && !drawOptions_.legendVerticalText &&
           molWidth_ > 0) {
+        // Same minimum side gap used on the left side for symmetry.
         const double g =
             std::max(8.0, marginPadding_ * static_cast<double>(drawWidth_));
         xCentre = basePanelX + double(molWidth_) - g - scaledRanges.x;
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -82,7 +82,6 @@ wszqkzqk, Diogo, dehaenw
   optional argument toggling the canonical Kekulization.
 
 ## New Features and Enhancements:
-  - **MolDraw2D / Drawing (GitHub #9023)**: Legend position is now configurable (Top, Left, Right, Bottom). For Left/Right, text can be drawn vertically (one character per line) via `legendVerticalText`. Options: `legendPosition`, `legendVerticalText` on MolDrawOptions; JSON keys `"legendPosition"` (string: "Top", "Left", "Right", "Bottom") and `"legendVerticalText"` (boolean). Applies to single-molecule drawing (e.g. MolToSVG, MolToImage); grid drawing can pass the same drawOptions so the position applies to each cell's legend.
   - MolFromSmiles scales quadratically with chain length
  (github issue #8776 from i-tub)
   - Add extract mol fragment api


Code/GraphMol/MolDraw2D/Wrap/rdMolDraw2D.cpp
```diff
@@ -919,6 +926,12 @@ BOOST_PYTHON_MODULE(rdMolDraw2D) {
       .def_readwrite(
           "legendFraction", &RDKit::MolDrawOptions::legendFraction,
           "fraction of the draw panel to be used for the legend if present")
+      .def_readwrite(
+          "legendPosition", &RDKit::MolDrawOptions::legendPosition,
+          "legend position: 0=Bottom, 1=Top, 2=Left, 3=Right")
```
DavidACosgrove: Should these values refer to the enum?  It would be good if you mentioned the default value, as well.

Code/GraphMol/MolDraw2D/catch_tests.cpp
```diff
@@ -4757,6 +4757,90 @@ TEST_CASE("vary proportion of panel for legend", "[drawing]") {
   }
 }
 
+TEST_CASE("legend position Top Left Right and vertical text", "[drawing]") {
+  auto m1 = "CCO"_smiles;
```
DavidACosgrove, line 4761: These tests only check if there is a legend in the picture, not where it is.  The easiest way to do this is to write the file, as other tests do, so someone can check them by eye.  A better way would be to use a regex to find one of the letters in the legend and make sure it's in the right place.  There are examples of the sort of regex required in the catch_tests.cpp  for example in `TEST_CASE("Github 6397 - chiral tag overlapping atom label")`.

rdkit/Chem/Draw/UnitTestDraw.py
```diff
@@ -162,6 +162,35 @@ def testGithubIssue86(self):
     nbds = [x.GetBondDir() for x in mol.GetBonds()]
     self.assertEqual(obds, nbds)
 
+  def testLegendPosition(self):
```
DavidACosgrove: These tests would be better in MolDraw2D/Wrap/testMolDraw2D.py.  The tests can be more lightweight as the correct functioning will be checked in catch_tests.cpp.  There's no need to check the legend is in the correct place here as well as there, for example, so these are fine otherwise.

Code/GraphMol/MolDraw2D/catch_tests.cpp
```diff
@@ -4757,6 +4757,90 @@ TEST_CASE("vary proportion of panel for legend", "[drawing]") {
...
+  REQUIRE(m1);
```
DavidACosgrove, line 4762: I don't see a test for horizontal left or right legend.  If there isn't one, can you add it, please?

Code/GraphMol/MolDraw2D/catch_tests.cpp
```diff
@@ -4757,6 +4757,90 @@ TEST_CASE("vary proportion of panel for legend", "[drawing]") {
...
+    MolDraw2DSVG drawer(200, 200, -1, -1, NO_FREETYPE);
+    CHECK(drawer.drawOptions().legendPosition ==
+          MolDrawOptions::LegendPosition::Bottom);
+    MolDraw2DUtils::prepareAndDrawMolecule(drawer, *m1, legend);
+    drawer.finishDrawing();
+    auto text = drawer.getDrawingText();
+    CHECK(text.find("class='legend'") != std::string::npos);
+  }
+  SECTION("Long vertical side legend fits panel height") {
+    const std::string longName(48, 'M');
```
DavidACosgrove, line 4883: For this test, I don't see a legend in the picture, though it is in the SVG.  Is that correct?  If so, a comment in the test would be helpful so people in future know what to expect.

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -1819,82 +1864,307 @@ void DrawMol::extractLegend() {
...
-    } else {
-      next_piece += c;
+  if (vertText) {
+    for (unsigned int i = 0; i < legend_.size(); ++i) {
+      if (legend_[i] != '\n') {
+        legend_bits.push_back(std::string(1, legend_[i]));
+      }
+    }
+  } else {
+    std::string next_piece;
```
DavidACosgrove: You could do this using boost::split.  An example from rxn_test1.cpp: `boost::split(lines, indata, boost::is_any_of("\n"));`

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -1819,82 +1864,307 @@ void DrawMol::extractLegend() {
...
-  if (!flexiCanvasY_) {
-    auto adjLegHt = drawHeight_ * drawOptions_.legendFraction;
-    // subtract off space for the padding.
-    if (total_height > adjLegHt) {
-      relFontScale *= double(adjLegHt) / total_height;
-      calc_legend_height(legend_bits, relFontScale, total_width, total_height);
+  const double baseX = xOffset_ + width_ * marginPadding_;
+  const double baseY = marginPadding_ * height_ + yOffset_;
+
+  auto place_bottom = [&]() {
```
DavidACosgrove: House style is generally to use lambdas only for small functions.  We prefer these to be in anonymous namespace and defined above the function that uses them.  It makes the parent function easier to follow.

Brandon-Cole:
Hi @DavidACosgrove, thanks again for the helpful feedback! I pushed a follow-up commit with the requested updates:

- Updated Python `legendPosition` binding docs to reference enum names and the default.
- Strengthened the MolDraw2D Catch legend test: writes SVG and uses regex-based placement checks (including horizontal Left/Right and vertical stacked ordering).
- Moved the Python legend-position unit test to `MolDraw2D/Wrap/testMolDraw2D.py` as a lightweight wrapper check.
- Added explicit horizontal Left/Right legend coverage and clarified the long-vertical legend expectation in the test comment.
- Refactored legend parsing/placement support in `DrawMol.cpp` to use `boost::split` and anonymous-namespace helpers, per house style.

Please let me know if I missed anything else or there's something I can improve! Thanks for maintaining RDKit!

Brandon-Cole:
Sorry had to resolve an issue where some old tests asserted hardcoded legend positions! Thanks again and please let me know if there's anything else I can change to improve the modifications!

Code/GraphMol/MolDraw2D/catch_tests.cpp
```diff
@@ -4757,6 +4757,207 @@ TEST_CASE("vary proportion of panel for legend", "[drawing]") {
...
+  double right_x = 0.0, right_y = 0.0;
+  SECTION("Top") {
+    MolDraw2DSVG drawer(200, 200, -1, -1, NO_FREETYPE);
+    drawer.drawOptions().legendPosition =
+        MolDrawOptions::LegendPosition::Top;
+    MolDraw2DUtils::prepareAndDrawMolecule(drawer, *m1, legend);
+    drawer.finishDrawing();
+    auto text = drawer.getDrawingText();
+    CHECK(text.find("class='legend'") != std::string::npos);
+    CHECK(get_legend_xy(text, top_x, top_y));
```
DavidACosgrove, line 4813: Sorry, I wasn't clear.  What I meant before was that, having extracted the coordinates of a character from the legend, you should check that it means the legend was in the correct place on the canvas.  In this case, for example, something like:
`CHECK(top_y < 30.0);`

Code/GraphMol/MolDraw2D/catch_tests.cpp
```diff
@@ -4757,6 +4757,207 @@ TEST_CASE("vary proportion of panel for legend", "[drawing]") {
...
+    drawer.finishDrawing();
+    auto text = drawer.getDrawingText();
+    CHECK(text.find("class='legend'") != std::string::npos);
+    CHECK(get_legend_xy(text, bottom_x, bottom_y));
+    std::ofstream outs("testLegendPosition_bottom.svg");
+    outs << text;
+    outs.flush();
+  }
+  SECTION("Top appears above Bottom") {
+    MolDraw2DSVG drawerTop(200, 200, -1, -1, NO_FREETYPE);
```
DavidACosgrove: If you check the positions of the characters as suggested above, this test becomes redundant.  Or you could write the SVGs here and keep this test instead.  We probably don't need both.

Code/GraphMol/MolDraw2D/test1.cpp
```diff
@@ -3854,19 +3854,7 @@ void testGithub3112() {
     std::ofstream outs("testGithub3112_1.svg");
     outs << text;
     outs.close();
-#ifdef RDK_BUILD_FREETYPE_SUPPORT
```
DavidACosgrove, line 3858: These tests were doing rather more than just checking there is a legend.  Between them they check, albeit rather crudely, that the newline in the legend is being honoured.  In the first pair, for example, it's looking for the 'b' because in the 2nd case that's on a different line so the y coordinate is different and also that the font size is smaller because it has to fit 2 lines into the legend box.
Can you put them back like they were, please, and update the string to reflect the new coordinates.  That's a pain, I know, because you have to run the tests twice, once with Freetype support, once without, and copy the relevant values into the C++ from the SVG.  
At some point it would be a good idea for someone (me, probably) to go through these very old tests and make them fit for modern C++, but that's a separate job.

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -1781,20 +1795,253 @@ void DrawMol::calcMeanBondLength() {
...
     legendHeight_ = 0;
-  } else {
-    if (!flexiCanvasY_) {
-      legendHeight_ = int(drawOptions_.legendFraction * float(drawHeight_));
-      molHeight_ = drawHeight_ - legendHeight_;
-    } else {
+    return;
+  }
+  switch (drawOptions_.legendPosition) {
+    case MolDrawOptions::LegendPosition::Bottom:
```
DavidACosgrove, line 1809: Can the Bottom and Top cases be collapsed into 1, as you've done with Left and Right?  They look like duplicate code to me.

Brandon-Cole:
Awesome thanks again David! I implemented the following changes based on the feedback:

- Added a simple bounds check after parsing the legend so we assert the legend sits in the right part of the panel, not just that it exists.

- Removed the redundant Top vs Bottom and Left vs Right sections since the per-position checks already cover that.

- Put the old testGithub3112 checks back and updated the FreeType path coordinates (and the expected SVG hashes for those four files since the output changed).

- Merged the duplicate Top and Bottom branches in partitionForLegend()

Brandon-Cole:
Let me know if it all looks ok to you @DavidACosgrove. Thanks again!

ReleaseNotes.md
```diff
@@ -55,6 +55,7 @@ Paul, paconius, spparel, wszqkzqk, Diogo, dehaenw
   optional argument toggling the canonical Kekulization.
 
 ## New Features and Enhancements:
+  - **MolDraw2D / Drawing (GitHub #9023)**: Legend position is now configurable (Top, Left, Right, Bottom). For Left/Right, text can be drawn vertically (one character per line) via `legendVerticalText`. Options: `legendPosition`, `legendVerticalText` on MolDrawOptions; JSON keys `"legendPosition"` (string: "Top", "Left", "Right", "Bottom") and `"legendVerticalText"` (boolean). Applies to single-molecule drawing (e.g. MolToSVG, MolToImage); grid drawing can pass the same drawOptions so the position applies to each cell's legend.
```
greglandrum: This section of the release notes is constructed automatically when I'm preparing the release. It will end up referencing this PR like the other entries.

Please remove this change

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -1189,8 +1190,21 @@ void DrawMol::calculateScale() {
...
+        !legend_.empty() &&
+        (drawOptions_.legendPosition == MolDrawOptions::LegendPosition::Left ||
+         drawOptions_.legendPosition == MolDrawOptions::LegendPosition::Right) &&
+        !drawOptions_.legendVerticalText;
+    if (sideLegendHoriz && molWidth_ > 0) {
+      const double g =
+          std::max(8.0, marginPadding_ * static_cast<double>(drawWidth_));
+      if (double(molWidth_) > g + 24.0) {
+        widthForScale = double(molWidth_) - g;
+      }
```
greglandrum, line 1207: It's a good idea to comment on magic constants like the 8 and 24 here so that future maintainers can understand what they are and where they came from.

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -1781,20 +1795,246 @@ void DrawMol::calcMeanBondLength() {
...
+  }
+}
+
+double calc_line_gap(const DrawText &textDrawer, bool vertText,
+                     double relFontScale) {
+  const double font_px = textDrawer.fontSize() * relFontScale;
+  if (vertText) {
+    return std::max(1.0, font_px * 0.35);
+  }
+  return std::max(2.0, font_px * 0.15);
```
greglandrum, line 1862: Again, please document the meaning and origin of the constants

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -1804,97 +2044,131 @@ void DrawMol::extractLegend() {
...
+    if (!flexiCanvasY_) {
+      relFontScale *= maxHeight / total_height;
+      calc_legend_dims(legend_bits, relFontScale, drawOptions_.legendColour,
+                       textDrawer_, total_width, total_height);
+    } else if (drawOptions_.legendPosition ==
+                   MolDrawOptions::LegendPosition::Bottom ||
+               drawOptions_.legendPosition ==
+                   MolDrawOptions::LegendPosition::Top) {
+      double extra_padding = total_height * marginPadding_;
+      extra_padding = extra_padding < 2.0 ? 2.0 : extra_padding;
```
greglandrum: Isn't this just:

```suggestion
      extra_padding = std::max(extra_padding, 2.0);
```

Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -2876,10 +3150,40 @@ void DrawMol::getDrawTransformers(Point2D &trans, Point2D &scale,
...
+    case MolDrawOptions::LegendPosition::Bottom:
+      break;
+    case MolDrawOptions::LegendPosition::Top:
+      yCentre += legendHeight_;
+      break;
+    case MolDrawOptions::LegendPosition::Left:
+      if (!legend_.empty() && !drawOptions_.legendVerticalText &&
+          molWidth_ > 0) {
+        const double g =
+            std::max(8.0, marginPadding_ * static_cast<double>(drawWidth_));
```
greglandrum, line 3178: Please document the meaning/origin of the 8.0 here and below on l3181

Brandon-Cole:
Thanks Greg for the feedback and sorry for the delay! 


I’ve addressed the final review items:

- Removed the Release Notes entry for this PR (`ReleaseNotes.md`) 

- Added comments documenting the magic constants in `DrawMol.cpp` (the `8.0`, `24.0`, and line-gap factors) so hopefully their purpose/origin is more clear for future maintainers

- Replaced the manual min expression with `std::max()` for the extra legend padding logic


Let me know if there's anything else I can fix to improve the changes, thanks again!

greglandrum:
@Brandon-Cole thanks for the contribution!
@DavidACosgrove thanks for doing the review.

PR #9189, opened 2026-03-19
Add getSGroupDataLabels() to expose DAT SGroup label positions

### LINKED ISSUE #7829: Consider a function returning the data calculated by DrawMol::extractSGroupData()
We'd like to have a way to display SGroups labels (the stuff in FIELDDATA and FILEDDISP) in a way that's consistent with RDKit's placement, using the information calculated in DrawMol::extractSGroupData().  I think this boils down to the **text** of the label and the **coordinates** where it should be rendered.
A function that returns those two would be great

@cdvonbargen in case I'm missing something

thanks!

cdvonbargen:
#### Reference Issue

Fixes #7829

Adds `MolDraw2D_detail::getSGroupDataLabels()`, a function that returns the text and molecule-coordinate positions of DAT SGroup labels using the same placement logic as the drawing code. This allows external renderers to display SGroup labels consistently with RDKit's placement without reimplementing the FIELDDISP parsing and coordinate transformation logic.

The function returns a vector of `SGroupDataLabel` structs, each containing the label text, position in molecule coordinates, a `positioned` flag indicating whether the position came from an explicit FIELDDISP entry (vs. a fallback to the associated atom's conformer position), and the associated atom index.

`DrawMol::extractSGroupData()` is refactored to call `getSGroupDataLabels()` internally. Behavior is preserved exactly for all existing cases; the one edge case that changes (`!located` with no associated atom and no FIELDDISP) previously silently placed an annotation at (0,0), and now logs a warning and skips — an improvement over the old behavior.

#### Any other comments?

Co-authored with [Claude Sonnet 4.6](https://www.anthropic.com/claude) (claude-sonnet-4-6).

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawMol.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawMol.cpp
@@ -602,95 +602,21 @@ void DrawMol::extractSGroupData() {
   if (!includeAnnotations_) {
     return;
   }
-  const auto &sgs = getSubstanceGroups(*drawMol_);
-  if (sgs.empty()) {
-    return;
-  }
-
-  // details of this transformation are in extractAtomCoords
-  double rot = drawOptions_.rotate * M_PI / 180.0;
-  RDGeom::Transform2D tform;
-  tform.SetTransform(Point2D(0.0, 0.0), rot);
-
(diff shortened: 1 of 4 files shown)


greglandrum (approved):
LGTM

PR #9190, opened 2026-03-20
Fix issue #8108: assignStereochemistry should detect atropisomers

### LINKED ISSUE #8108: assignStereochemistry does not handle Atropisomers
Atropsiomer detection is not part of the usual `assignStereochemistry()` workflow. This means that Atropisomers will only be detected at import time, or when an explicit call to `detectAtropisomerChirality()` is made. 

This means that atropisomers won't be detected, updated or cleared (since `assignStereochemistry(cleanIt=true)` also does not clear atropisomer bond stereo markers either) after that, even if we made changes to the mol that directly affect the atropisomer groups (e.g. reverting the bond/parity that defines the atropisomer), leaving the mol with invalid stereo.

It feels like the atropisomer detection should be triggered inside `assignStereochemistry()`.

cdvonbargen:
#### Reference Issue

Fixes #8108.

`assignStereochemistry()` now detects atropisomers during normal stereo perception while respecting `cleanIt`: existing tags are preserved when false and replaced from available wedge or 3D evidence when true. Detection also honors `replaceExistingTags` in the 2D and 3D helper APIs, retains macrocyclic candidates, and preserves enumerated atrop configurations when 2D coordinates are stale. Since perception now happens centrally, the individual file format parsers no longer trigger atropisomer detection themselves.

Because detection runs after sanitization, it repeats the ring check that `cleanupAtropisomers()` would otherwise have performed, so that bonds which cannot actually rotate don't end up tagged. That check uses the same definition of a macrocycle as the rest of the codebase: rings of nine or more members.

#### Any other comments?


Code/GraphMol/Atropisomers.cpp
```diff
@@ -596,6 +596,11 @@ void detectAtropisomerChirality(ROMol &mol, const Conformer *conf) {
         bondToTry->getEndAtom()->getHybridization() != Atom::SP2) {
       continue;
     }
+    // Ring bonds cannot be atropisomers (no free rotation)
+    if (mol.getRingInfo()->isInitialized() &&
+        mol.getRingInfo()->numBondRings(bondToTry->getIdx()) > 0) {
+      continue;
+    }
```
ricrogz: I'd move this into the loop that populates `bondsToTry` to discard bonds as soon as possible.

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6554,3 +6554,60 @@ TEST_CASE(
   REQUIRE(m);
   CHECK(*labels_reference == get_bond_stereo_labels(*m));
 }
+
+TEST_CASE("Github #8108: assignStereochemistry should handle atropisomers",
+          "[chirality][atropisomers]") {
+  std::string rdbase = getenv("RDBASE");
+  std::string fName =
+      rdbase +
+      "/Code/GraphMol/FileParsers/test_data/atropisomers/RP-6306_atrop1.sdf";
```
ricrogz: This is not used in the second section; it can be moved into the first.

Code/GraphMol/Chirality.cpp
```diff
@@ -2426,6 +2427,11 @@ void legacyStereoPerception(ROMol &mol, bool cleanIt,
         }
       }
     }
+    if (!hadAtropisomers) {
+      const Conformer *conf =
+          mol.getNumConformers() ? &mol.getConformer() : nullptr;
+      Atropisomers::detectAtropisomerChirality(mol, conf);
+    }
```
ricrogz: This feels strange: we only run atropisomer detection when we are running a "clean" detection where stereo is reset before the detection. Same thing in `stereoPerception()`.

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6554,3 +6554,60 @@ TEST_CASE(
...
+  SECTION("atropisomer stereo is preserved by assignStereochemistry") {
+    auto m = v2::FileParsers::MolFromMolFile(fName);
+    REQUIRE(m);
+    auto *atropBond = m->getBondWithIdx(3);
+    REQUIRE(atropBond->getStereo() == Bond::BondStereo::STEREOATROPCW);
+
+    bool cleanIt = true, force = true;
+    MolOps::assignStereochemistry(*m, cleanIt, force);
+    CHECK(atropBond->getStereo() == Bond::BondStereo::STEREOATROPCW);
+  }
```
ricrogz: This test passes without the patch. It would be more useful if the initial mol didn't have the atropisomer, which should be added afterwards, making sure that `assignStereochemistry()` detects it with and without the `cleanIt` flag (this is mostly what the issue is about).

The test should also run with both legacy and modern stereochemistry perception.

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6554,3 +6554,60 @@ TEST_CASE(
...
+    // Use CXSMILES with 2D coords+wedges so wedge bonds are retained after
+    // parsing (unlike SDF parsing which clears them).
+    auto m = v2::SmilesParse::MolFromSmiles(
+        "OC1=C(C)C(N2C3=C(C(C(=O)N)=C2N)C=C(C)C(C)=N3)=C(C)C=C1 "
+        "|(2.0992,3.5589,;2.2694,2.7517,;1.6554,2.2008,;0.8712,2.4571,;"
+        "1.8255,1.3935,;1.0637,0.4452,;0.8122,-0.3353,;-0.0126,-0.3353,;"
+        "-0.264,0.4452,;-1.0482,0.7015,;-1.6621,0.1505,;-1.2182,1.5087,;"
+        "0.3998,0.9266,;0.3998,1.7516,;-0.425,-1.0498,;-0.0126,-1.7642,;"
+        "-0.425,-2.4786,;0.8122,-1.7642,;1.2247,-2.4786,;1.2247,-1.0498,;"
+        "2.6096,1.1372,;2.7797,0.33,;3.2236,1.6882,;3.0535,2.4954,),wD:4.3|");
```
ricrogz: Can you add a third section in with the CXSMILES doesn't include coordinates? (see https://github.com/rdkit/rdkit/blob/master/Docs/Book/RDKit_Book.rst#interpreting-atropisomers-without-coordinates)

Also, as the one above, both this section and the one without coordinates should run under both stereo perception algorithms.

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6554,3 +6554,60 @@ TEST_CASE(
...
+        atropBond = bond;
+        break;
+      }
+    }
+    REQUIRE(atropBond);
+    auto originalStereo = atropBond->getStereo();
+
+    // Clear the detected stereo to simulate a programmatic molecule
+    // where detection hasn't been run yet
+    atropBond->setStereo(Bond::BondStereo::STEREONONE);
```
ricrogz: This is what the `CleanIt` argument is expected to do. It would be useful to check that `CleanIt=true` can clear invalid/outdated atropisomer stereo, and that `CleanIt=false` will not. (that's the other part of the issue).

ricrogz (changes requested):

Code/GraphMol/Atropisomers.cpp
```diff
@@ -533,6 +533,11 @@ void detectAtropisomerChirality(ROMol &mol, const Conformer *conf) {
         if (nbrBond == bond) {
           continue;  // a bond is NOT its own neighbor
         }
+        // Ring bonds cannot be atropisomers (no free rotation)
+        if (mol.getRingInfo()->isInitialized() &&
+            mol.getRingInfo()->numBondRings(nbrBond->getIdx()) > 0) {
+          continue;
+        }
```
ricrogz: This is an interesting check. But I don't thin it holds for macrocycles. This is a silly example, but it's the first thing I came up with (I just added a huge ring around an existing example of an atropisomer):

Image

Code/GraphMol/Chirality.cpp
```diff
@@ -3498,6 +3514,7 @@ void assignChiralTypesFrom3D(ROMol &mol, int confId, bool replaceExistingTags) {
       atom->setProp<int>(common_properties::_NonExplicit3DChirality, 1);
     }
   }
+  Atropisomers::detectAtropisomerChirality(mol, &conf);
```
ricrogz: I think `detectAtropisomerChirality()` needs a `replaceExistingTags` mechanism to honor the argument in this function.

Code/GraphMol/Chirality.cpp
```diff
@@ -3769,6 +3783,7 @@ void assignStereochemistryFrom3D(ROMol &mol, int confId,
 void assignChiralTypesFromBondDirs(ROMol &mol, const int confId,
                                    const bool replaceExistingTags) {
   if (!mol.getNumConformers()) {
+    Atropisomers::detectAtropisomerChirality(mol, nullptr);
```
ricrogz: This needs `replaceExistingTags` too.

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6554,3 +6554,133 @@ TEST_CASE(
...
+        "/Code/GraphMol/FileParsers/test_data/atropisomers/RP-6306_atrop1.sdf";
+    auto m = v2::FileParsers::MolFromMolFile(fName);
+    REQUIRE(m);
+    auto *atropBond = m->getBondWithIdx(3);
+    REQUIRE(atropBond->getStereo() == Bond::BondStereo::STEREOATROPCW);
+
+    // Stereo detected at parse time should be preserved (not overwritten or
+    // cleared) when assignStereochemistry is called on the parsed molecule.
+    MolOps::assignStereochemistry(*m, true, true);
+    CHECK(atropBond->getStereo() == Bond::BondStereo::STEREOATROPCW);
```
ricrogz: It would be interesting to flip the atropisomer before triggering the stereo detection, to make sure we are actually re-detecting it. Otherwise, we don't know if this is the already existing label that escaped the `cleanIt=true`.

Code/GraphMol/catch_chirality.cpp
```diff
@@ -6554,3 +6554,133 @@ TEST_CASE(
...
+    // Clear stereo: assignStereochemistry should re-detect with cleanIt=false
+    atropBond->setStereo(Bond::BondStereo::STEREONONE);
+    MolOps::assignStereochemistry(*m, false, true);
+    CHECK(atropBond->getStereo() == correctStereo);
+
+    // Reset and verify detection also works with cleanIt=true
+    atropBond->setStereo(Bond::BondStereo::STEREONONE);
+    MolOps::assignStereochemistry(*m, true, true);
+    CHECK(atropBond->getStereo() == correctStereo);
+  }
```
ricrogz: Isn't this the same as the test case above, except for the input SMILES not having coordinates? Both cases could be collapsed into one test function, and then call it on the two inputs.

greglandrum:
@cdvonbargen do you plan to come back to this, or should I just close it?

cdvonbargen:
I'll come back to this!

ricrogz:
Test failures are real.

Code/GraphMol/Chirality.cpp
```diff
@@ -3749,6 +3787,7 @@ void assignStereochemistryFrom3D(ROMol &mol, int confId,
 void assignChiralTypesFromBondDirs(ROMol &mol, const int confId,
                                    const bool replaceExistingTags) {
   if (!mol.getNumConformers()) {
+    Atropisomers::detectAtropisomerChirality(mol, nullptr, replaceExistingTags);
```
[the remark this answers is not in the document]
greglandrum: I'm not convinced that this is actually a problem. Please provide an example C++ test that demonstrates it.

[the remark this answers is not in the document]
greglandrum: I don't think that the logic here is correct, but I'm going to have to take some time to construct the counter-example.
In the meantime, look at the specification for how atropisomers without coordinates are interpreted, the documentation is here https://github.com/rdkit/rdkit/blob/master/Docs/Book/RDKit_Book.rst#interpreting-atropisomers-without-coordinates , and confirm that the logic behind the test above is correct.

Code/GraphMol/Chirality.cpp
```diff
@@ -2240,6 +2240,39 @@ std::ostream &operator<<(std::ostream &oss, const StereoSpecified &s) {
...
+    return;
+  }
+  bool removedAny = false;
+  for (auto bond : mol.bonds()) {
+    // bonds in macrocycles are left alone, since they can link actual
+    // atropisomeric portions
+    if ((bond->getStereo() == Bond::BondStereo::STEREOATROPCW ||
+         bond->getStereo() == Bond::BondStereo::STEREOATROPCCW) &&
+        ri->numBondRings(bond->getIdx()) > 0 &&
+        ri->minBondRingSize(bond->getIdx()) < 8) {
```
greglandrum, line 2265: We use 9 elsewhere in the code

Code/GraphMol/FileParsers/CDXMLParser.cpp
greglandrum, line 1: The new Chemdraw parser, in `External/ChemDraw/chemdraw.cpp` also needs to be updated. I think the correct diff for that is:
```
diff --git a/External/ChemDraw/chemdraw.cpp b/External/ChemDraw/chemdraw.cpp
index 89d168018..867289578 100644
--- a/External/ChemDraw/chemdraw.cpp
+++ b/External/ChemDraw/chemdraw.cpp
@@ -168,10 +168,6 @@ void visit_children(
         } else {
           MolOps::assignChiralTypesFromBondDirs(*res, confidx, true);
         }
-        Atropisomers::detectAtropisomerChirality(*res,
-                                                 &res->getConformer(confidx));
-      } else {  // no Conformer
-        Atropisomers::detectAtropisomerChirality(*res, nullptr);
       }
 
       // now that atom stereochem has been perceived, the wedging
```
Certainly the tests pass for me when I try that, but I don't think we have great coverage in the public tests. @bp-kelley will need to confirm that it's actually correct.

greglandrum (changes requested):
One small change about macrocycle sizes and a larger one to also patch the new ChemDraw parser.

greglandrum (approved):
LGTM

greglandrum:
@ricrogz this now looks good to me. Are you 👍 ?

ricrogz:
> @ricrogz this now looks good to me. Are you 👍 ?

Yes. I've had a quick look at it and it looks ok to me.

PR #9195, opened 2026-03-22
Fixes #9143

### LINKED ISSUE #9143: ETKDGv3 generating 90 degree twisted amides

greglandrum:
Some of the macrocycle ET definitions needed to be loosened to remove the `r{9-}`.

There's a larger thing here where we probably really should duplicate all the macrocycle ET definitions and have one with:
`[...k{9-}:2]@;-[...r:3]` and one with `[...r:2]@;-[...k{9-}:3]`
but that requires a more significant rewrite and a bunch of validation.

There is still some weirdness in the matching of ET bonds in macrocycles, but it is not connected to this change.

### DIFF
--- a/Code/GraphMol/DistGeomHelpers/catch_tests.cpp
+++ b/Code/GraphMol/DistGeomHelpers/catch_tests.cpp
@@ -1614,7 +1614,7 @@ TEST_CASE("Github #9143: ETKDGv3 generating twisted amides") {
                Catch::Matchers::WithinAbs(-180, 10) ||
                    Catch::Matchers::WithinAbs(180, 10));
     CHECK_THAT(MolTransforms::getDihedralDeg(conf, 31, 30, 28, 29),
-               Catch::Matchers::WithinAbs(0, 10));
+               Catch::Matchers::WithinAbs(0, 12));
     CHECK_THAT(MolTransforms::getDihedralDeg(conf, 19, 18, 20, 21),
                Catch::Matchers::WithinAbs(-180, 20) ||
                    Catch::Matchers::WithinAbs(180, 20));


greglandrum:
I talked to @sriniker about this and she agrees that this is a reasonable fix for at least some of the problems.
We will address the broader point as part of getting ETKDGv4 out the door

PR #9198, opened 2026-03-22
Leak fixes for 2026.03.1

ricrogz:
Mostly mols leaked in tests.

Theres  two  things here that' are "real":
- The first one, trivial, is the `FingerprintGenerator` that wasn't taking ownership of the atom/bond invariant generators. Adding the `true`, `true` flags to the `FingerprintGenerator<std::uint64_t>()` call makes it take ownership.

- The second one, more complicated, is about `MorganFeatureAtomInvGenerator` not cleaning up `dp_patterns` on termination. I added this, plus making copies of the patterns when creating new instances, so that each one can own its own set of patterns.

### DIFF
--- a/Code/GraphMol/Fingerprints/MorganGenerator.cpp
+++ b/Code/GraphMol/Fingerprints/MorganGenerator.cpp
@@ -70,9 +70,6 @@ MorganAtomInvGenerator *MorganAtomInvGenerator::clone() const {
 MorganFeatureAtomInvGenerator::MorganFeatureAtomInvGenerator(
     const std::vector<const ROMol *> *patterns) {
   if (patterns) {
-    if (dp_patterns) {
-      cleanUpPatterns();
-    }
     dp_patterns = new std::vector<const ROMol *>;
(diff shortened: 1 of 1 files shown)


Code/GraphMol/Fingerprints/MorganGenerator.cpp
```diff
@@ -94,6 +117,9 @@ void MorganFeatureAtomInvGenerator::fromJSON(
     const boost::property_tree::ptree &pt) {
   if (pt.get_child_optional("patternSMARTS")) {
     const auto &patternsNode = pt.get_child("patternSMARTS");
+    if (dp_patterns) {
```
greglandrum: you don't really need this check since `cleanupPatterns()` is a no-op if dp_patterns is null

ricrogz:
> One small suggestion, I'm fine merging as-is if you don't think it's worth making the change.

I'll make the change, plus another one I thought you would flag: the other `if (dp_patterns) {  cleanUpPatterns(); }` is called in a constructor where `dp_patterns` can't possibly be initialized.

PR #9202, opened 2026-03-24
Fix typo in _calculateBeta: check nb1 instead of nb2 twice

evasnow1992:
#### Reference Issue

N/A — found during code review.

In `_calculateBeta` (TorsionFingerprints.py, line 395), the guard for filtering
non-terminal bonds checks `len(nb2) > 1` twice instead of `len(nb1) > 1 and len(nb2) > 1`.

This means `nb1` (the begin-atom neighbors) is never checked, so bonds where the
begin-atom is terminal but the end-atom is not are incorrectly included when
computing `dmax` for torsion weights.

The fix changes the first `nb2` to `nb1`.

#### Any other comments?

The existing tests in `UnitTestMol3D.py` should still pass. The bug is hard to trigger
in practice because RDKit's bond iteration order and typical test molecules rarely
have a terminal begin-atom paired with a non-terminal end-atom, but the logic is
clearly wrong as written.


greglandrum:
Yeah, this is clearly a bug. Thanks for the fix!

PR #9203, opened 2026-03-25
Enable templating for macrocycles

ZontaNicola:
This enables the templating algorithm to be run for single macrocycles too


greglandrum:
I'm not sure it's possible, but it would be good to add a test for this.

ZontaNicola:
> I'm not sure it's possible, but it would be good to add a test for this.

added a test

greglandrum:
The CI failure here is not real

greglandrum:
> Shouldn't the workflow have made the templating changes PR for you?

The workflow only runs automatically once a week, so it hadn't run.
I just triggered it manually.

PR #9204, opened 2026-03-26
Add spiro flipping to 2D coord generation

ZontaNicola:
this adds the ability to flip one of the two rings of a spiro compound to solve clashes (the same way we flip around a single bond)

before and after pic of the structure added as a test:
Screenshot 2026-03-26 at 14 22 43
Screenshot 2026-03-26 at 14 15 36


rachelnwalker (commented):
@ZontaNicola I haven't had time to look at this super closely but at a high level this seems reasonable. Any chance you could do a quick perf test before and after this change? 

Also look like there are some builds failing, just from tests though

greglandrum:
@ZontaNicola the CI failures here look real

ZontaNicola:
> @ZontaNicola I haven't had time to look at this super closely but at a high level this seems reasonable. Any chance you could do a quick perf test before and after this change?
> 
> Also look like there are some builds failing, just from tests though

on my system performance seems to be comparable for general molecules. For the molecule in the test (which is a rare example of structure that needs a spiro inversion) it looks it's gotten slightly faster, maybe because it's skipping more expensive clash detection down the road

ZontaNicola:
@greglandrum The remaining CI errors seem unrelated to me, what do you think?

Code/GraphMol/Depictor/EmbeddedFrag.h
```diff
@@ -337,9 +343,9 @@ class RDKIT_DEPICTOR_EXPORT EmbeddedFrag {
                                         double mimicDmatWt = 0.0,
                                         bool permuteDeg4Nodes = false);
 
-  //! Remove collisions in a structure by flipping rotatable bonds
+  //! Remove collisions in a structure by flipping rotatable bonds and spiro centers
   //! along the shortest path between two colliding atoms
-  void removeCollisionsBondFlip();
+  void removeCollisionsBondAndSpiroFlip();
```
greglandrum, line 353: This is formally a backwards compatiblity break. I doubt that anyone is actually using this function in their code, but we should still account for it. I'd suggest the following (not tested):

```suggestion
  void removeCollisionsBondAndSpiroFlip();
  
  [[deprecated("please use removeCollisionsBondAndSpiroFlip()"]] void removeCollisionsBondFlip() { removeCollisionsBondAndSpiroFlip(); };
```

Code/GraphMol/Depictor/DepictUtils.cpp
```diff
@@ -402,6 +402,92 @@ RDKit::INT_VECT getRotatableBonds(const RDKit::ROMol &mol, unsigned int aid1,
...
+  // Check that the two rings share ONLY this atom (spiro)
+  std::set<int> ring1(rings[0].begin(), rings[0].end());
+  std::set<int> ring2(rings[1].begin(), rings[1].end());
+
+  std::vector<int> shared;
+  std::set_intersection(ring1.begin(), ring1.end(),
+                       ring2.begin(), ring2.end(),
+                       std::back_inserter(shared));
+
+  if (shared.size() != 1 || shared[0] != static_cast<int>(aid)) {
```
greglandrum: Do we need the second part of this test? Aren't we guaranteed that both rings contain `aid`?

Code/GraphMol/Depictor/DepictUtils.cpp
```diff
@@ -402,6 +402,92 @@ RDKit::INT_VECT getRotatableBonds(const RDKit::ROMol &mol, unsigned int aid1,
...
+    // Rings share more than just this atom - not a spiro
+    return false;
+  }
+
+  // Spiro center should have exactly 4 neighbors (typical for carbon)
+  unsigned int degree = mol->getAtomWithIdx(aid)->getDegree();
+  if (degree != 4) {
+    return false;
+  }
+
```
greglandrum: This test is cheap and should be done at the very beginning of the function.

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1446,6 +1446,24 @@ void _recurseAtomOneSide(unsigned int endAid, unsigned int begAid,
...
+
+  // Collect the 2 rings containing this spiro atom
+  for (const auto &ring : atomRings) {
+    if (std::find(ring.begin(), ring.end(), static_cast<int>(spiroAid)) !=
+        ring.end()) {
+      result.push_back(ring);
+    }
+  }
+
+  CHECK_INVARIANT(result.size() == 2, "Spiro must have exactly 2 rings");
```
greglandrum: It's great to have this test here, but I think it's a postcondition:
```suggestion
  POSTCONDITION(result.size() == 2, "Spiro must have exactly 2 rings");
```

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1873,6 +1891,62 @@ void EmbeddedFrag::flipAboutBond(unsigned int bondId, bool flipEnd) {
   }
 }
 
+void EmbeddedFrag::flipAboutSpiroCenter(unsigned int spiroAid) {
+  PRECONDITION(dp_mol, "");
+  PRECONDITION(spiroAid < dp_mol->getNumAtoms(), "");
+  PRECONDITION(isSpiroCenter(spiroAid, dp_mol), "Must be spiro center");
```
greglandrum: This is relatively expensive to calculate; it's probably better to not have it as a precondition

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1873,6 +1891,62 @@ void EmbeddedFrag::flipAboutBond(unsigned int bondId, bool flipEnd) {
...
+  const auto &targetRing = rings[0];
+
+  // Find the two neighbors of the spiro atom that are in the target ring
+  // (must be bonded to the spiro atom, not just in the same ring)
+  std::set<int> targetRingSet(targetRing.begin(), targetRing.end());
+  std::vector<unsigned int> ringNeighbors;
+
+  for (auto nbr : dp_mol->atomNeighbors(dp_mol->getAtomWithIdx(spiroAid))) {
+    unsigned int nbrIdx = nbr->getIdx();
+    if (targetRingSet.find(nbrIdx) != targetRingSet.end()) {
```
greglandrum:
```suggestion
    if (targetRingSet.contains(nbrIdx)) {
```

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1977,60 +2051,138 @@ void EmbeddedFrag::openAngles(const double *dmat, unsigned int aid1,
...
+    const std::pair<unsigned int, unsigned int> &cAids, unsigned int ncols,
+    double prevDensity, std::map<int, unsigned int> &doneSpiros,
+    const std::set<unsigned int> &spiroCenters, const double *dmat) {
+  // Find spiro centers on the path using our cached set (avoid expensive
+  // re-checks)
+  RDKit::INT_LIST path =
+      RDKit::MolOps::getShortestPath(*dp_mol, cAids.first, cAids.second);
+  std::vector<unsigned int> spiros;
+  for (auto aid : path) {
+    if (spiroCenters.find(aid) != spiroCenters.end()) {
```
greglandrum:
```suggestion
    if (spiroCenters.contains(aid)) {
```

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1977,60 +2051,138 @@ void EmbeddedFrag::openAngles(const double *dmat, unsigned int aid1,
...
+      // Didn't help - undo the flip
+      flipAboutSpiroCenter(spiroAid);
+    }
+  }
+  return false;
+}
+
+void EmbeddedFrag::removeCollisionsBondAndSpiroFlip() {
+  // Pre-compute which atoms are spiro centers (expensive check, so cache it)
+  std::set<unsigned int> spiroCenters;
```
greglandrum: We generally use a `boost::dynamic_bitset<>` for things like this. At the very least it would be better to use `std::unordered_set`

greglandrum (changes requested):
Some suggestions

Code/GraphMol/Depictor/DepictUtils.cpp
ZontaNicola: moved

Code/GraphMol/Depictor/DepictUtils.cpp
ZontaNicola: yeah, I got rid of it

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: ok gone

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: implemented dynamic_bitset

greglandrum (approved):
LGTM

ZontaNicola:
@greglandrum I resolved the conflicts and merged... I think we are all set to merge this PR, sorry it took so long

greglandrum:
@ZontaNicola that last commit - [c9853f4 ](https://github.com/rdkit/rdkit/pull/9204/commits/c9853f49e99f3db0badd051f0a93d3fa88f6e2bc)- introduced a bunch of formatting changes in files that have nothing to do with this PR and which are not present on master.

I don't know how that happened, but that needs to be cleaned up before we can merge this.

greglandrum (approved):
LGTM

PR #9208, opened 2026-03-27
Normalize rings

ricrogz:
This PR doesn't fix anything. Its intention is to "normalize" rings in an attempt to keep test changes to a minimum when we make changes to ring finding in upcoming PRs.

The code changes in this PR are just only a small change in `Code/GraphMol/FindRings.cpp`. The rest of the changes are just updates to tests. Disclaimer: since there is no real functionality change here, I haven't reviewed the test changes too thoroughly.

The "normalization" of the rings consists in updating the "atom rings" so that the first atom index in the vector is the smallest one in the ring, and the second one is the smallest index between the 2 neighbors to the first atom. Once these two conditions are met, the order of the rest of the atoms in the ring is predetermined. The idea is that these normalized ring definitions should stay the same as long as the indexes of the atoms in the molecule doesn't change, no matter how rings were discovered.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,5 +1,5 @@
-# Release_2026.03.1
-(Changes relative to Release_2025.09.1)
+# Release_2026.09.1
+(Changes relative to Release_2026.03.1)
 
 ## Acknowledgements
 (Note: I'm no longer attempting to manually curate names. If you would like to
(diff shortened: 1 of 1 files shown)


ricrogz:
> I forgot in my initial review: please add something to the "Backwards incompatible changes" column of ReleaseNotes.md for this one.

Done. Note that I updated the "# Release" line because there were 2 lines for 2026.03.1 and none for 2026.09.1 :)

PR #9216, opened 2026-03-30
Adds SVE to BFGSOpt

ajeyabsf:
#### Reference Issue
No issue raised.

This PR introduces SVE (Scalable Vector Extension) kernels to accelerate the BFGS update algorithm.

The implementation analyzes multiple independent loops within the BFGS workflow and selectively applies SVE vectorization only to those that demonstrated measurable performance improvements. Key optimizations include:

* Vectorized initialization of the search direction and norm computation
* Fused matrix-vector multiplications and dot-product accumulations
* Optimized symmetric rank-1 updates of the inverse Hessian using SVE FMA instructions
* Efficient computation of the new search direction

Benchmarking on an AWS m7g.8xlarge instance (32 cores, ARM-based) shows that the per-molecule processing time for AllChem.EmbedMolecule decreased by ~34%, improving from ~500 ms to ~330 ms.

#### Any other comments?

This PR aims to improve performance on ARM systems.

To ensure correctness and portability:

* Runtime detection is used to check for SVE support via getauxval
* SVE-specific code paths are only executed on supported hardware
* A fully functional scalar fallback path is preserved for non-SVE systems

This guarantees that:

* There is no impact on existing x86 or non-SVE ARM systems
* The behavior remains unchanged aside from performance improvements on supported hardware


shubham-fujitsu:
@greglandrum This work is in line with the [discussion item](https://github.com/rdkit/rdkit/discussions/9065)

greglandrum:
@ajeyabsf thanks for the contribution.
This make take a while for me to get my head around.

ajeyabsf:
Thanks for taking the time out @greglandrum . 

If it would be helpful, we’d be happy to provide a concise rundown of the implementation or walk through any specific parts in more detail to make the review easier. Please let us know if there’s anything we can do.

greglandrum:
Just to see what would happen, I tried compiling this on my M1 Mac Mini and get the following error:
```
/Users/glandrum/RDKit_git/Code/Numerics/Optimizer/BFGSOpt.h:18:12: fatal error: 'sys/auxv.h' file not found
   18 |   #include <sys/auxv.h>
18      | :           ^~~~~~~~~~~~12
: fatal error: 'sys/auxv.h' file not found
   18 |   #include <sys/auxv.h>
      |            ^~~~~~~~~~~~
```

It seems like the M1 doesn't have the relevant instructions anyway, but the code does still need to compile. :-)

scal444:
I'm curious if maybe Eigen (which is an optional dep anyway) would provide SIMD intrinsics at the vector level under the hood without the need for hand-rolling. Alternatively, there are portable SIMD libraries like [EVE](https://github.com/jfalcou/eve) which would detect ARM vs x86 at compile time, and provide nicer interfaces than inline intrinsics. If rdkit was willing to take a dependency on a SIMD lib, I'd be happy to see if it's applicable to a bunch of places in the codebase

greglandrum:
> I'm curious if maybe Eigen (which is an optional dep anyway) would provide SIMD intrinsics at the vector level under the hood without the need for hand-rolling. Alternatively, there are portable SIMD libraries like [EVE](https://github.com/jfalcou/eve) which would detect ARM vs x86 at compile time, and provide nicer interfaces than inline intrinsics. If rdkit was willing to take a dependency on a SIMD lib, I'd be happy to see if it's applicable to a bunch of places in the codebase

If we can use it to get a measurable increase in performance (and I strongly suspect that we can), I am definitely open to the idea of adding at least an optional dependency on a SIMD library (as long as that doesn't come along with a really complex runtime dependency). We currently are not making efficient usage of modern CPUs and I think it would be great to start doing so.

ajeyabsf:
Sorry for the delay in the change @greglandrum . 

The tests failing seem to be unrelated to the changes done?

bp-kelley:
@ajeyabsf I restarted the failed tests just in case, but the fails are not related to your code.

ajeyabsf:
Thanks for rerunning the tests @bp-kelley .

What I notice is, it does not show any reviewer assigned in the reviewers tab, and I don't see a way of adding reviewers. Is that not needed?

@greglandrum @bp-kelley if you did get time to review, do you have any feedback on the code?

rageshh-fj:
@greglandrum @bp-kelley @scal444: Hi everyone, please let us know for any further improvement feedback on this patch and if it looks good, we kindly request to approve and merge. Thanks.

greglandrum (changes requested):
I can't evaluate the SVE code, so I'm going to assume that, as long as the tests pass, everything is fine there. I've also checked that the performance of embedding under linux on an intel-based machine stays more or less the same with the changes here (as you'd expect since that code path hasn't changed).

I would, however, prefer to have the SVE-specific code that's at the top here (down to line 203) pulled out into a separate header: `BFGSOpt_SVE.h` (or something similar) that is then `#include`d into `BFGSOPt.h`. Having the CPU-specific code isolated in a different header seems better.
Please don't forget to add the new header to the `rdkit_headers()` section in `CMakeLists.txt`

Finally: please run clang-format over the code so that the formatting is consistent with the rest of the RDKit.

greglandrum (approved):
LGTM

As mentioned elsewhere: I have to assume that the SVE code has been tested and is correct.

greglandrum:
@ajeyabsf @rageshh-fj Thanks for the contribution! I will merge this after the CI builds go through.

ajeyabsf:
Hello @greglandrum , thank you for the approval. Will this be merged at a later date, or is it ready to be merged now?

PR #9217, opened 2026-03-30
If templates match, skip ring number check

ZontaNicola:
for symmetric rings the ring number check was failing and rejecting good templates. Something to do with SSSR vs getNumRings() vs symmetrizeSSSR() but I'm not exactly sure. Extracting the ring system's molecule just to calculate the SSSR seems like a waste of time since this check is only for performance (to avoid calls to substructureMatch)


Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -495,8 +495,6 @@ bool EmbeddedFrag::matchToTemplate(const RDKit::INT_VECT &ringSystemAtoms,
     // and bond count first
     if (mol->getNumBonds() != numBonds) {
       continue;
-    } else if (mol->getRingInfo()->numRings() != ring_count) {
-      continue;
```
rachelnwalker, line 499: is `ring_count` used after this change?

rachelnwalker (commented):
Does this make a performance difference? If not lgtm, but looks like you have a couple red builds.

rachelnwalker:
@ZontaNicola the test failure seems unrelated? 

Would you mind doing a quick perf test by calculation coordinates on all the templates?

ZontaNicola:
> @ZontaNicola the test failure seems unrelated?
> 
> Would you mind doing a quick perf test by calculation coordinates on all the templates?

I ran the whole [macrocycle_db](https://macro-db.cn):


Total molecules:     50653

BEFORE
Total time:          207.004s
Mean time:           4.09ms
Median time:         2.01ms
Std dev:             7.92ms
Min time:            0.06ms
Max time:            185.57ms
Throughput:          244.7 molecules/sec

AFTER
Total time:          207.864s
Mean time:           4.10ms
Median time:         2.03ms
Std dev:             7.92ms
Min time:            0.06ms
Max time:            184.23ms
Throughput:          243.7 molecules/sec

greglandrum (changes requested):
The code changes look good, but I don't think that TemplateSmarts.h should be part of this PR.

ZontaNicola:
should be ok now

greglandrum (approved):
LGTM

greglandrum:

PR #8878, opened 2025-10-18
add more benchmarking

bddap:
Adds more catch2 c++ performance benchmarks, builds on https://github.com/rdkit/rdkit/pull/8865

<details>

<summary>
results
</summary>

Note these are no-longer measuring single molecule performance, each bench runs against a batch of molecules at a time.

```
> time ./Code/Bench/bench

Randomness seeded to: 1122338737

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
bench is a Catch2 v3.4.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
Descriptors::calcNumSpiroAtoms
-------------------------------------------------------------------------------
___/rdkit/Code/Bench/descriptors.cpp:11
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Descriptors::calcNumSpiroAtoms                 100            11     2.8963 ms 
                                        2.51843 us    2.47692 us    2.58097 us 
                                        255.696 ns    185.782 ns    348.875 ns 
                                                                               

-------------------------------------------------------------------------------
Descriptors::calcNumBridgeheadAtoms
-------------------------------------------------------------------------------
___/rdkit/Code/Bench/descriptors.cpp:22
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Descriptors::calcNumBridgeheadAtoms            100            13     2.8496 ms 
                                        2.24245 us    2.20897 us     2.2913 us 
                                        204.137 ns    154.599 ns    268.711 ns 
                                                                               

-------------------------------------------------------------------------------
MorganFingerprints::getFingerprint
-------------------------------------------------------------------------------
___/rdkit/Code/Bench/fingerprint.cpp:11
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
MorganFingerprints::getFingerprint             100             1     31.856 ms 
                                        298.787 us    296.895 us    301.641 us 
                                        11.7257 us    8.36815 us    17.5023 us 
                                                                               

-------------------------------------------------------------------------------
bench_common::nth_random
[...]

PR #9223, opened 2026-04-06
Add AVX2/SSSE3 SIMD vectorization for bitmap similarity functions

HeyangQin:
## Summary

- Vectorize all 6 `CalcBitmap*` functions in `BitOps.cpp` using the Mula vpshufb nibble-lookup popcount algorithm
- Three-tier runtime dispatch: **AVX2** (256-bit) > **SSSE3** (128-bit) > existing scalar fallback
- Runtime CPU detection via CPUID (MSVC) / `__builtin_cpu_supports` (GCC/Clang) — no new CMake flags needed
- `CalcBitmapAllProbeBitsMatch` optimized with ANDNOT+TESTZ (avoids popcount entirely)
- Zero impact on non-x86 platforms: all SIMD code behind `#ifdef` on `__x86_64__ || _M_X64 || __i386__ || _M_IX86`
- Existing scalar POPCNT/LUT code completely preserved as fallback
- Also fixes the strict-aliasing UB noted in #4689 for the SIMD paths (uses `_mm256_loadu_si256` instead of `(uint64_t*)` casts)

## Performance

Measured on 2048-bit Morgan fingerprints, 10K-pair database, best-of-5 trials:

| Path | Tanimoto pairs/sec | Speedup vs scalar LUT |
|------|-------------------:|----------------------:|
| Scalar (byte lookup table) | 2.0 M | 1.0x |
| SSSE3 (128-bit) | 10.5 M | **5.3x** |
| AVX2 (256-bit) | 16.3 M | **8.2x** |

A large-scale virtual screening computing billions of Tanimoto similarities would see proportional wall-clock reduction.

## Implementation

The Mula algorithm uses `vpshufb` / `pshufb` as a parallel 4-bit LUT to count bits per nibble, accumulates per-byte popcounts, then reduces via `vpsadbw`. Byte accumulators are flushed every 31 iterations to prevent overflow (max popcount per byte = 8, 8 × 31 = 248 < 255).

On GCC/Clang, `__attribute__((target("avx2")))` enables per-function ISA selection without global `-mavx2` flags. On MSVC, AVX2 intrinsics compile unconditionally and CPUID gates execution at runtime.

## Test plan

- [x] All existing Catch2 tests pass on **MSVC 19.44** (Windows): 1069 assertions, 25 test cases
- [x] All existing Catch2 tests pass on **GCC 13.3** (Linux/WSL): 1069 assertions, 25 test cases
- [x] Standalone SIMD-vs-scalar correctness tests: 11,218 checks across 8 fingerprint sizes (1–512 bytes), including edge cases (all-zero, all-one, non-aligned), verified on MSVC, GCC, and Clang 18.1
- [x] Benchmark included (`benchBitOps.cpp`) for reproducible performance measurement


greglandrum:
@HeyangQin Thanks for the contribution. I will review as soon as I get the chance.

It looks like this PR had some AI help. We don't have any problems with that, but would like to see it acknowledged in the PR description (we still need to write this down somewhere).

If you would like to add benchmarking to the PR itself, #8878 is a good example of how to do so

ashvardanian:
Hi @greglandrum and @HeyangQin,

Just saw this and wanted to add my two cents. If you're looking for more speedups, it'd be worth leveraging VPOPCNTQ on Ice Lake and newer — it's a single instruction for 512-bit popcount, no nibble-lookup emulation needed. For batch calculations on Arm, SME-capable machines have BMOPA/BMOPS instructions that I use for tiled matrix-matrix Hamming & Jaccard. Let me know if that'd be of interest — I've implemented both (and quite a few other backends) over the last three years in my [NumKong](https://github.com/ashvardanian/NumKong) numerics library.

PS: Thanks for this project! I've been using RDKit for years and just forked it (fork number 1000 🥳 ) to accelerate some other part of the codebase 🤗

PR #9224, opened 2026-04-06
mention AI tools in the contrib guidelines

greglandrum:
I'm not sure this was the best place to do this, but we should having something.

### DIFF
--- a/Docs/Book/GettingStartedWithContributing.md
+++ b/Docs/Book/GettingStartedWithContributing.md
@@ -14,7 +14,10 @@ If you make significant use of AI tools beyond auto-complete when creating your
 contribution, please acknowledge that, and mention which tool you used, in the
 description of your PR. We encourage everyone to use the tools that make them
 maximally productive, but we also want to be completely transparent about the
-source of contributions.
+source of contributions. Your name will be associated with the code change, so 
+please review any generated code before posting it and make sure that you 
+understand what it is doing. Please do not submit any code that has not been 
+reviewed by a human first.
 
 ## What and How to Contribute
 


ricrogz:
Should we also add something along the lines of "You are responsible of the AI code you submit. Please review any generated code before posting it, and make sure you understand what it is doing. Do not post any code that has not been reviewed by a human first." ?

greglandrum:
> Should we also add something along the lines of "You are responsible of the AI code you submit. Please review any generated code before posting it, and make sure you understand what it is doing. Do not post any code that has not been reviewed by a human first." ?

That makes sense. What about "Your name will be associated with the code change, so please review any generated code before posting it and make sure that you understand what it is doing. Please do not submit any code that has not been reviewed by a human first."

ricrogz:
> That makes sense. What about "Your name will be associated with the code change, so please review any generated code before posting it and make sure that you understand what it is doing. Please do not submit any code that has not been reviewed by a human first."

Sounds good to me!

bp-kelley:
Isn't it more complicated?  Code (purely) generated by AI cannot be
copyrighted, hence it can't fall under the RDKIt BSD license and is
considered public domain.

I'm not sure what this entails with respect to PRs.

greglandrum:
> Isn't it more complicated? Code (purely) generated by AI cannot be copyrighted, hence it can't fall under the RDKIt BSD license and is considered public domain. I'm not sure what this entails with respect to PRs.

Though this is likely true, mentioning it is an invitation to techies arguing about law on the internet, which is a genre of “discourse” that I would like to stay as far away from as possible. I’d rather just stick with this more “intended for normal humans” take on the argument.

PR #9228, opened 2026-04-07
Fix bug in inversion term for UFF, add finite difference checker.

### LINKED ISSUE #9229: Incorrect UFF Inversion term gradient for P/As/Sb/Bi centers
Note: This was found by nvMolKit's AI code review tool while I was porting UFF over to CUDA kernels, and I had cursor help me with checking where the derivation went wrong. 

**Describe the bug**
The gradient in [inversion](https://github.com/rdkit/rdkit/blob/master/Code/ForceField/UFF/Inversion.cpp#L113) has an incorrect sign for the C2 contribution. Derivation:
```
E = K (C0 + C1 cosW + C2 cos 2W)
```
because sinY = cosW.

```
d/dW [C1 cosW] = -C1 sinW
d/dW [C2 cos 2W] = -2 C2 sin 2W = -4 C2 sinW cosW
dE/dW = -K (C1 sinW + 4 C2 sinW cosW)
```
and substituting back to the code variables:
```
sinW = cosY
cosW = sinY
```
gives
```
dE/dW = -K (C1 cosY + 4 C2 cosY sinY)
```

The C2 term is only nonzero for group 5, code [here](https://github.com/rdkit/rdkit/blob/master/Code/ForceField/UFF/Utils.cpp#L54-L78)

**To Reproduce**
Check out [PR 9228](https://github.com/rdkit/rdkit/pull/9228), undo the fix, should reproduce the energy/gradient inconsistency.

scal444:
There was an incorrect sign in the inversion gradient that only applied to phosphorus centers and other elements in that periodic table group. 

Added a finite difference calculator to demonstrate the bug and the fix. This may be generally useful but I wanted feedback on how best to incorporate, and how the regression test should be implemented. 

Implementation drafted by cursor before being improved manually

Fixes #9229

### DIFF
--- a/Code/ForceField/CMakeLists.txt
+++ b/Code/ForceField/CMakeLists.txt
@@ -15,7 +15,6 @@ target_compile_definitions(ForceField PRIVATE RDKIT_FORCEFIELD_BUILD)
 
 rdkit_headers(Contrib.h
               ForceField.h
-              FiniteDifference.h
               AngleConstraint.h
               AngleConstraints.h
               DistanceConstraint.h
--- a/Code/ForceField/CMakeLists.txt
+++ b/Code/ForceField/CMakeLists.txt
@@ -1,6 +1,7 @@
 
 rdkit_library(ForceField
-              ForceField.cpp AngleConstraint.cpp AngleConstraints.cpp 
+              ForceField.cpp FiniteDifference.cpp
+              AngleConstraint.cpp AngleConstraints.cpp 
               DistanceConstraint.cpp DistanceConstraints.cpp
               PositionConstraint.cpp TorsionConstraint.cpp
               UFF/AngleBend.cpp UFF/BondStretch.cpp UFF/Nonbonded.cpp
@@ -14,6 +15,7 @@ target_compile_definitions(ForceField PRIVATE RDKIT_FORCEFIELD_BUILD)
 
 rdkit_headers(Contrib.h
               ForceField.h
+              FiniteDifference.h
               AngleConstraint.h
               AngleConstraints.h
               DistanceConstraint.h
--- a/Code/ForceField/UFF/testUFFForceField.cpp
+++ b/Code/ForceField/UFF/testUFFForceField.cpp
@@ -27,11 +27,9 @@
 #include <ForceField/UFF/PositionConstraint.h>
 
 #include <GraphMol/RDKitBase.h>
-#include <GraphMol/DistGeomHelpers/Embedder.h>
 #include <GraphMol/FileParsers/FileParsers.h>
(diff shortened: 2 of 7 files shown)


scal444:
This is a fix for #9229 but I wanted feedback on how best to test the fix, and whether or not a general finite difference checker over a series of molecules would be a good test in general for these forcefields. If not wanted, the test I added could be changed to use golden values, etc.

Code/ForceField/UFF/testUFFForceField.cpp
```diff
@@ -1656,6 +1659,28 @@ M  END
...
 
+void testFiniteDifference() {
+  std::cerr << "-------------------------------------" << std::endl;
+  std::cerr << "    Test finite difference gradient check (P(F)(F)F)."
+            << std::endl;
+
+  std::unique_ptr<ROMol> mol(SmilesToMol("P(F)(F)F"));
+  TEST_ASSERT(mol);
+  std::unique_ptr<ROMol> molH(MolOps::addHs(*mol));
+  TEST_ASSERT(DGeomHelpers::EmbedMolecule(*molH) >= 0);
```
greglandrum: Instead of doing embedding here (and adding the DGeomHelpers dependency to the UFF tests), I think it would be clearer and cleaner to provide the test molecule with coordinates. So either using a mol block or CXSMILES with coordinates. (Don't forget to make sure that the Hs arent removed when you parse it).

Code/ForceField/CMakeLists.txt
```diff
@@ -14,6 +15,7 @@ target_compile_definitions(ForceField PRIVATE RDKIT_FORCEFIELD_BUILD)
 
 rdkit_headers(Contrib.h
               ForceField.h
+              FiniteDifference.h
```
greglandrum: Since we're currently just using this in tests, I don't think we need to make it part of the public API by putting it in rdkit_headers

greglandrum (changes requested):
The fix itself looks good, but I would suggest a small change to the way the test is done.

I think it's fine to leave the finite difference calculator where you've put it for now, but I'd prefer that we not add it to the public API by including it in rdkit_headers (these are the files that are installed on client machines). If we end up needing it elsewhere at some point, we can think about moving it and adding it to the code that is installed with the rdkit then.

Code/ForceField/UFF/testUFFForceField.cpp
scal444: Done

Code/ForceField/CMakeLists.txt
scal444: Done

greglandrum (approved):
LGTM

PR #9230, opened 2026-04-08
Support using iterators with MolSuppliers

greglandrum:
Allows using MolSuppliers in range algorithms and range-based for loops.
So you can now do:
```
for(const auto mol : supplier) {
}
```
Suppliers also work with view and range algorithms. Here's grabbing the first 5 molecules from a supplier that match a substructure query:
```
    std::vector<std::shared_ptr<RWMol>> mols;
    std::ranges::copy(supplier |
                  std::views::filter([&query, &params](const auto &mol) {
                    return SubstructMatch(*mol, *query, params).empty();
                  }) |
                  std::views::take(5), 
                std::back_inserter(mols));
```
views are lazy, so this will only read molecules until it finds 5 matches.

To allow the random-access suppliers to be used instead of separate vectors of molecules, I also added the option to have them cache molecules so that they are only read once. This is a simple shorthand for maintaining the vector of smart pointers to molecules yourself. Of course it still will use a ton of memory if you read a giant SD file with caching turned on, just like it will use a ton of memory if you read those molecules into a vector.

I also modified the random-access suppliers so that they are thread safe and can work with STL parallel executors. Here's a simple demo (from the testing code) of using the STL parallel executors. This is using a caching mol supplier where I'd already looped over the supplier once to prime the cache:
```
    // sequential execution
    constexpr unsigned int numIters = 200;
    for (unsigned int iter = 0; iter < numIters; ++iter) {
      std::vector<unsigned int> nAts(reader.length());
      std::transform(std::execution::seq, reader.begin(), reader.end(),
                     nAts.begin(),
                     [](const auto mol) { return MolToSmiles(*mol).size(); });
      accum += nAts.size();
    }

    // parallel execution
    for (unsigned int iter = 0; iter < numIters; ++iter) {
      std::vector<unsigned int> nAts(reader.length());
      std::transform(std::execution::par, reader.begin(), reader.end(),
                     nAts.begin(),
                     [](const auto mol) { return MolToSmiles(*mol).size(); });
      accum += nAts.size();
    }

```
Output on my laptop (running on battery, so it will only use a few threads):
```
Base transform of 500x200 molecules took 5927 ms
Parallel transform of 500x200 molecules took 1350 ms
```
If the cache isn't being used and the molecules all actually need to be read from disk, the parallel execution may actually be slower.


greglandrum:
Looks like I need to try compiling with apple clang to figure out what's going on with `std::execution`

Code/GraphMol/FileParsers/fileParsersIterTest.cpp
```diff
@@ -0,0 +1,432 @@
...
+  CHECK(reader.length() == len);
+  std::vector<unsigned int> expected;
+  expected.reserve(reader.length());
+  for (unsigned int i = 0; i < reader.length(); ++i) {
+    expected.push_back(reader[i]->getNumAtoms());
+  }
+  std::vector<unsigned int> actual;
+  std::ranges::transform(
+      reader, std::back_inserter(actual),
+      [](const auto &mol) { return (size_t)mol->getNumAtoms(); });
```
ricrogz: The `size_t` cast here is redundant, `getNumAtoms()` already returns an `unsigned int`, which is the same type `actual` contains.
```suggestion
      [](const auto &mol) { return mol->getNumAtoms(); });
```

Code/GraphMol/FileParsers/fileParsersIterTest.cpp
```diff
@@ -0,0 +1,432 @@
...
+  CHECK(actual == expected);
+}
+
+template <typename Supplier>
+void forwardIterTest(Supplier &reader, size_t len) {
+  std::vector<unsigned int> actual;
+  std::ranges::transform(reader, std::back_inserter(actual),
+                         [](const auto &mol) {
+                           REQUIRE(mol);
+                           return (size_t)mol->getNumAtoms();
```
ricrogz: `size_t` is redundant too.
```suggestion
                           return mol->getNumAtoms();
```

Code/GraphMol/FileParsers/fileParsersIterTest.cpp
```diff
@@ -0,0 +1,432 @@
...
+  SECTION("with caching") { cacheTest(reader, 16); }
+  SECTION("reverse iteration") {
+    std::vector<unsigned int> expected;
+    std::ranges::transform(
+        std::begin(reader), std::end(reader), std::back_inserter(expected),
+        [](const auto &mol) { return (size_t)mol->getNumAtoms(); });
+    std::vector<unsigned int> actual;
+    std::ranges::transform(
+        std::rbegin(reader), std::rend(reader), std::back_inserter(actual),
+        [](const auto &mol) { return (size_t)mol->getNumAtoms(); });
```
ricrogz:
```suggestion
    std::ranges::transform(
        std::begin(reader), std::end(reader), std::back_inserter(expected),
        [](const auto &mol) { return mol->getNumAtoms(); });
    std::vector<unsigned int> actual;
    std::ranges::transform(
        std::rbegin(reader), std::rend(reader), std::back_inserter(actual),
        [](const auto &mol) { return mol->getNumAtoms(); });
```

Code/GraphMol/FileParsers/MolSupplier.h
```diff
@@ -249,34 +443,21 @@ class RDKIT_FILEPARSERS_EXPORT SmilesMolSupplier : public MolSupplier {
   /**************************************************************************
    * Lazy file parser for Smiles table file, similar to the lazy SD
    * file parser above
-   * - As an when new molecules are read using "next" their
-   *    positions in the file are noted.
-   *  - A call to the "length" will automatically parse the entire
-   *    file and cache all the mol block positions
+   * - When new molecules are read using "next()" their
```
ricrogz: Sorry abut this one, it's just  my OCD...
```suggestion
   *  - When new molecules are read using "next()" their
```

Code/GraphMol/FileParsers/MolSupplier.h
```diff
@@ -341,26 +537,22 @@ class RDKIT_FILEPARSERS_EXPORT TDTMolSupplier : public MolSupplier {
   /**************************************************************************
    * Lazy file parser for TDT files, similar to the lazy SD
    * file parser above
-   * - As an when new molecules are read using "next" their
+   * - When new molecules are read using "next()" their
```
ricrogz: OCD again
```suggestion
   *  - When new molecules are read using "next()" their
```

Code/GraphMol/FileParsers/SDMolSupplier.cpp
```diff
@@ -256,9 +259,47 @@ void SDMolSupplier::moveTo(unsigned int idx) {
...
+  if (d_cacheMolecules) {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+    const std::lock_guard<std::mutex> guard(d_cacheMutex);
+#endif
+    auto len = length();
+    if (d_molCache.size() != len) {
+      d_molCache.clear();
+      d_molCache.resize(len);
+    }
+    d_molCache[idx] = res;
```
ricrogz: Hm, this feels strange to me. Why do we need to clear the cache? Also, I expect the resize to only take place once, since `length()` will always return the full size of inputs (i.e. the second time we hit this code, the size will still be the same).

I think the length check / clear / resize strategy makes sense if we expect the input data to change, but I think that's not the case (or, at least not without hitting the `init()` function, which will already clear the mol cache and the cached length.

I'd, instead, do something like this to avoid having to parse the full input to calculate length (may be slow for large input files), and lazily allocate the cache without triggering a  realloc on each call:
```suggestion
    if (d_molCache.size() <= idx) {
      constexpr unsigned int molCacheAllocChunkSize = 1000;
      d_molCache.resize(idx + molCacheAllocChunkSize);
    }
    d_molCache[idx] = res;
```

Code/GraphMol/FileParsers/SmilesMolSupplier.cpp
```diff
@@ -466,6 +470,38 @@ std::unique_ptr<RWMol> SmilesMolSupplier::operator[](unsigned int idx) {
...
+  }
+  if (d_cacheMolecules) {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+    const std::lock_guard<std::mutex> guard(d_cacheMutex);
+#endif
+    auto len = length();
+    if (d_molCache.size() != len) {
+      d_molCache.clear();
+      d_molCache.resize(len);
+    }
```
ricrogz: Same comment as above in `SDMolSupplier::getShared()`

Code/GraphMol/FileParsers/TDTMolSupplier.cpp
```diff
@@ -372,11 +375,44 @@ void TDTMolSupplier::moveTo(unsigned int idx) {
...
+  if (d_cacheMolecules) {
+#ifdef RDK_BUILD_THREADSAFE_SSS
+    const std::lock_guard<std::mutex> guard(d_cacheMutex);
+#endif
+    auto len = length();
+    if (d_molCache.size() != len) {
+      d_molCache.clear();
+      d_molCache.resize(len);
+    }
+    d_molCache[idx] = res;
```
ricrogz: And again, the same as in `SDMolSupplier::getShared()`

Code/GraphMol/FileParsers/MolSupplier.h
```diff
@@ -118,6 +162,45 @@ class RDKIT_FILEPARSERS_EXPORT MolSupplier {
...
+  const_value_type operator*() const { return current.value(); }
+  ForwardSupplierIter &operator++() {
+    if (supplier->atEnd()) {
+      current.reset();
+      return *this;
+    }
+    current = supplier->nextShared();
+    if (!current.value() && supplier->atEnd()) {
+      current.reset();
+    }
```
ricrogz, line 204: Am I understanding this wrong? It feels like that if we failed parsing (`!current.value()`) the last molblock in the stream (`supplier->atEnd()` after the updating `current`), we return the same thing as if the stream was already at EOF (see block above), a  `ForwardSupplierIter` with `current = nullopt`.

Unless I'm understanding this wrong, I think if the last block in the input failed parsing, we should return a `ForwardSupplierIter` where `current.value() = nullptr` (failed parsing), not `current = nullopt`.

Code/GraphMol/FileParsers/fileParsersIterTest.cpp
```diff
@@ -0,0 +1,432 @@
...
+  params.smilesColumn = 0;
+  params.nameColumn = -1;
+  v2::FileParsers::SmilesMolSupplier reader(infile, params);
+  SECTION("basics") { iterTest(reader, 200); }
+  SECTION("with caching") { cacheTest(reader, 200); }
+  SECTION("reverse iteration") {
+    std::vector<unsigned int> expected;
+    std::ranges::transform(
+        std::begin(reader), std::end(reader), std::back_inserter(expected),
+        [](const auto &mol) { return (size_t)mol->getNumAtoms(); });
```
ricrogz:
```suggestion
        [](const auto &mol) { return mol->getNumAtoms(); });
```

Code/GraphMol/FileParsers/fileParsersIterTest.cpp
```diff
@@ -0,0 +1,432 @@
...
+    std::vector<unsigned int> actual;
+    std::ranges::transform(
+        std::rbegin(reader), std::rend(reader), std::back_inserter(actual),
+        [](const auto &mol) { return (size_t)mol->getNumAtoms(); });
```
ricrogz:
```suggestion
        [](const auto &mol) { return mol->getNumAtoms(); });
```

Code/GraphMol/FileParsers/fileParsersIterTest.cpp
ricrogz: Also, given the number of times this lambda is used, it might be worth moving it out to a separate function. Same for the one that gets the pointer addresses.

Code/GraphMol/FileParsers/MolSupplier.h
```diff
@@ -341,26 +537,22 @@ class RDKIT_FILEPARSERS_EXPORT TDTMolSupplier : public MolSupplier {
   /**************************************************************************
    * Lazy file parser for TDT files, similar to the lazy SD
    * file parser above
-   * - As an when new molecules are read using "next" their
+   * - When new molecules are read using "next()" their
    *    positions in the file are noted.
-   *  - A call to the "length" will automatically parse the entire
+   *  - A call to "length()" will automatically parse the entire
    *    file and cache all the mol block positions
```
ricrogz: Use same wording as in `SmilesMolSupplier` ?
  
```suggestion
   *    file and store all the mol block positions
```

ricrogz: Wording in this one too
```suggestion
   *    file and store all the mol block positions
```

ricrogz (commented):
Somehow this one didn't make it into the other review.

Code/GraphMol/FileParsers/SDMolSupplier.cpp
greglandrum: > Hm, this feels strange to me. Why do we need to clear the cache? Also, I expect the resize to only take place once, since `length()` will always return the full size of inputs (i.e. the second time we hit this code, the size will still be the same).

Yeah, you're right. The data can only change with `setData()`, and that calls `init()`, which clears the cache.

ricrogz (approved):
LGTM

PR #9232, opened 2026-04-09
Macrocycle template matching

ZontaNicola:

If we cannot template a macrocycle, try again using a more relaxed approach, allowing for substituents inside the ring. Return the position that has the fewest atoms inside the ring and does not invert stereochemistry on the macrocycle bonds. The templating works even if small rings are fused to the macrocycle, sharing 2, 3 or 4 atoms

[here's a comparison table with the generation times.](https://zontanicola.github.io/sketcher/galleries/macrocycle_comparison_100.html
) Notice how fused rings are placed correctly when a template is found (even though the shapes, especially pentagons,  will need to be refined in a future PR)

if a template is not found we revert to the regular polygon shape.


Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -543,16 +570,556 @@ bool EmbeddedFrag::matchToTemplate(const RDKit::INT_VECT &ringSystemAtoms,
...
+  std::shared_ptr<RDKit::ROMol>
+      relaxed_query;              // Query without degree constraints
+  std::vector<bool> is_internal;  // Which atoms have degree constraints
+};
+
+// Get cached template info (build cache on first access)
+// Cache is keyed by template pointer for identity-based lookup
+static const CachedTemplateInfo &getCachedTemplateInfo(
+    const std::shared_ptr<RDKit::ROMol> &tmpl) {
+  static std::map<const RDKit::ROMol *, CachedTemplateInfo> cache;
```
rachelnwalker: The way the templates currently work, you could technically clear the template library which would make the keys in this map dangling pointers. I think you could do `std::map<std::shared_ptr<const RDKit::ROMol>, CachedTemplateInfo> cache;` instead.

There are also some thread safety issues here - not sure if that is a major issue though (I will defer to @greglandrum on that).

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -543,16 +570,556 @@ bool EmbeddedFrag::matchToTemplate(const RDKit::INT_VECT &ringSystemAtoms,
...
+
+  // Build cached info for this template
+  CachedTemplateInfo info;
+
+  // Convert template to SMARTS and remove degree constraints
+  std::string tmpl_smarts = RDKit::MolToSmarts(*tmpl);
+  std::regex degree_pattern("&D[0-9]");
+  std::string relaxed_smarts =
+      std::regex_replace(tmpl_smarts, degree_pattern, "");
+  info.relaxed_query.reset(RDKit::SmartsToMol(relaxed_smarts));
```
rachelnwalker: regex operations directly scare me - maybe you could `ReplaceAtom` with plain query atom of the same atomic number?

Code/GraphMol/Depictor/DepictUtils.cpp
```diff
@@ -164,8 +164,22 @@ RDKit::INT_VECT setNbrOrder(unsigned int aid, const RDKit::INT_VECT &nbrs,
 
 int pickFirstRingToEmbed(const RDKit::ROMol &mol,
                          const RDKit::VECT_INT_VECT &fusedRings) {
-  // ok this is what we will do here
-  // we will pick the ring with the smallest number of substituents
+  // If there are any macrocycles, pick the largest one
+  int macrocycle_idx = -1;
+  unsigned int max_macrocycle_size = 0;
+  for (int i = 0; i < static_cast<int>(fusedRings.size()); ++i) {
+    if (fusedRings[i].size() > 8) {
```
rachelnwalker: can we make 8 a constant somewhere?

Code/GraphMol/Depictor/DepictUtils.cpp
```diff
@@ -218,6 +235,9 @@ RDKit::VECT_INT_VECT findCoreRings(const RDKit::VECT_INT_VECT &fusedRings,
         if (currRingId == otherRingId || removedRings[otherRingId]) {
           continue;
         }
+        // Don't skip macrocycles - we need to count all intersections to avoid
+        // removing rings that have significant fusion (3+ atoms) with a
+        // macrocycle
         RDKit::INT_VECT commmonAtoms;
         RDKit::Intersect(fusedRings[currRingId], fusedRings[otherRingId],
                          commmonAtoms);
```
rachelnwalker: `commmonAtoms` - did I do this? the extra `m`?

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -444,22 +454,39 @@ static bool checkStereoChemistry(const RDKit::ROMol &mol,
...
+    new_at.df_fixed = true;
+    d_eatoms.emplace(mol_aidx, new_at);
+  }
+
+  // Setup neighbors and attachment points
+  this->setupNewNeighs();
+  this->setupAttachmentPoints();
+
+  return true;
+}
```
rachelnwalker: Looks like this always returns true

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -543,16 +570,556 @@ bool EmbeddedFrag::matchToTemplate(const RDKit::INT_VECT &ringSystemAtoms,
...
+    const RDKit::VECT_INT_VECT &allRings) {
+  std::vector<FusedRingInfo> fused_rings;
+
+  if (allRings.empty()) {
+    return fused_rings;
+  }
+
+  for (const auto &ring : allRings) {
+    // Skip any macrocycles
+    if (ring.size() > 8) {
```
rachelnwalker, line 758: another `8` :)

Code/GraphMol/Depictor/EmbeddedFrag.h
```diff
@@ -375,6 +379,24 @@ class RDKIT_DEPICTOR_EXPORT EmbeddedFrag {
...
+  //! Copy coordinates from a template match into embedded atoms
+  /*!
+    \param template_mol the template molecule with conformer coordinates
+    \param match the mapping from template atom indices to molecule atom indices
+    \return true after successfully copying coordinates and setting up neighbors
+  */
+  bool applyTemplateCoordinates(
+      const std::shared_ptr<RDKit::ROMol> &template_mol,
+      const RDKit::MatchVectType &match);
+
```
rachelnwalker, line 396: Do these need to be in the header?

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -543,16 +570,556 @@ bool EmbeddedFrag::matchToTemplate(const RDKit::INT_VECT &ringSystemAtoms,
...
+static bool validateFusionGeometry(
+    const std::shared_ptr<RDKit::ROMol> &template_mol,
+    const RDKit::MatchVectType &match, const RDKit::INT_VECT &macrocycleRing,
+    const std::vector<FusedRingInfo> &fusedRings) {
+  const double ANGLE_TOLERANCE = 20.0 * M_PI / 180.0;  // ~20 degrees in radians
+  const double TARGET_60_DEG = 60.0 * M_PI / 180.0;    // 60 degrees in radians
+
+  // Determine which angle sign is convex (majority) vs concave (minority)
+  bool positive_is_convex =
+      determineConvexSign(template_mol, match, macrocycleRing);
```
rachelnwalker: I think this could be factored out of this function and passed in instead

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -653,11 +1220,50 @@ void EmbeddedFrag::embedFusedRings(const RDKit::VECT_INT_VECT &fusedRings,
   if (useRingTemplates) {
     RDKit::INT_VECT coreRingsIds;
     auto coreRings = findCoreRings(fusedRings, coreRingsIds, *dp_mol);
-    if (coreRings.size() > 1 && coreRings.size() < fusedRings.size()) {
+
+    bool systemIsSimplified = coreRings.size() < fusedRings.size();
+    bool hasMultipleCoreRings = coreRings.size() > 1;
```
rachelnwalker: I think you could collapse this into 1 variable

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -653,11 +1220,50 @@ void EmbeddedFrag::embedFusedRings(const RDKit::VECT_INT_VECT &fusedRings,
...
+        [](const RDKit::INT_VECT &ring) { return ring.size() > 8; });
+
+    if ((hasMultipleCoreRings && systemIsSimplified) || hasMacrocycle) {
       // look for a template that matches the core ring system
       RDKit::Union(coreRings, funion);
-      bool found_template = matchToTemplate(funion, coreRings.size());
-      if (found_template) {
+      bool found_template = false;
+      if (hasMultipleCoreRings && systemIsSimplified)
+        found_template = matchToTemplate(funion, coreRings.size());
```
rachelnwalker: brackets please :)

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -653,11 +1220,50 @@ void EmbeddedFrag::embedFusedRings(const RDKit::VECT_INT_VECT &fusedRings,
...
+            // This ring is not in core
+            size_t ring_size = fusedRings[i].size();
+            if (ring_size < 4 || ring_size > 6) {
+              all_small_rings = false;
+              break;
+            }
+          }
+        }
+
+        if (all_small_rings) {
```
rachelnwalker: General comment: RDKit's convention for variable names is camelCase - hopefully claude can be useful to make renaming things less painful

Code/GraphMol/Depictor/EmbeddedFrag.cpp
rachelnwalker: Second thought - maybe all of this information should be stored in the templates class at instantiation

rachelnwalker (commented):
I'm seeing an unused parameter complaint 

```
/home/vsts/work/1/s/Code/GraphMol/Depictor/EmbeddedFrag.cpp:1071:28: error: unused parameter ‘ringSystemAtoms’ [-Werror=unused-parameter]
 1071 |     const RDKit::INT_VECT &ringSystemAtoms,
```

At a high level: I think that your library looks good. I haven't been able to fully verify the implementation, but the approach makes sense. It would be nice if we could improve the scoring and 'picking best match' performance a little, if possible

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: done

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: gone

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: removed

Code/GraphMol/Depictor/DepictUtils.cpp
ZontaNicola: fixed typo

Code/GraphMol/Depictor/DepictUtils.cpp
ZontaNicola: added constant

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: now using rdkit queries

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: used shared_ptr

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: you are right

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: added!

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: ok I think I've updated most of them, but I see a lot even in pre-existing code

Code/GraphMol/Depictor/EmbeddedFrag.h
ZontaNicola: not sure, but I've done a lot of modifications to this code in the next review, so let's defer to that

PR #9236, opened 2026-04-10
Add option to use precompiled headers

greglandrum:
This is disabled by default, but when enabled leads to significant reductions in compile times for me on window and the mac. They seem to slow things down for me with g++ on linux.

I've enabled them for the Windows and Mac (non-SWIG) CI builds.


skearnes:
Can I help get this over the line? Claude marked it as its top-ranked suggestion for reducing compile times and noted that the g++ regression is a known issue.

greglandrum:
> Can I help get this over the line? Claude marked it as its top-ranked suggestion for reducing compile times and noted that the g++ regression is a known issue.

Sure, I'm happy to have the help!
I did this as an experiment and my enthusiasm kind of dried up when I discovered that it didn't help with g++ (which is what I use most of the time). It could still be really useful with the other compilers though.

PR #9171, opened 2026-03-14
CIP labeller performance: Don't calculate auxiliary descriptors unnecessarily

d-b-w:
The first 3 rules (the constitutional rules) are pretty easy to understand. After rule 3, we need to calculate auxiliary stereo descriptors to break ties.

However, we _were actually_ calculating auxiliary stereodescriptors for all centers! We should only need to calculate auxiliary stereocenters for sites that are needed to break ties.

This cost time - it also caused errors if the auxiliary descriptors needed a graph expansion, because bonds in the digraph might be pointed in the wrong direction.

Example case PDB ID 4AXM: 14s -> 0.036s

Before this commit, errored with "Could not calculate parity! Carrier mismatch" after **14s**. After this commit, completes successfully in **0.036s**. Labelled centers all match (for the centers that had labels in the failure case). There are many protein-sized molecules that see similar speedups.

Includes a test that I can imagine breaking with a flawed implementation of this optimization. The reference labels are from before this change.

PR #9239, opened 2026-04-15
Make setDoubleBondNeighborDirections() more consistent

ricrogz:
This fixes an issue that makes the double bond directions assigned `setDoubleBondNeighborDirections()` potentially inconsistent for the same inputs.

The issue here is that when sorting `orderedBondsInPlay`, which is a `std::vector<std::pair<unsigned int, Bond *>>`, if the first elements of the pair are the same (which isn't infrequent), we will disambiguate using the address of the `Bond*`, which is arbitrary. This opens the door to different direction assignments for the exact same input molecule based on the pattern of the addresses where the bonds are allocated.

In my case, I hit this when running some checks for https://github.com/rdkit/rdkit/pull/9171: I was going to generate CIP labels from pubchem mols with and without the patch, but accidentally ran the baseline (without the patch) twice, and, to my surprise, I found mismatched labels between two runs that should be identical. Reproducing and debugging the issue was a real pain, but I ended up tracing it back to this issue.

This patch avoids using the addresses for the sorting by using the indexes instead. This doesn't guarantee that the results will always be consistent for the same molecule (e.g. same atoms / bonds but different ordering), as the order in which the bonds are assigned can change what is `ENDUPRIGHT`/`ENDDOWNRIGHT`, and even create conflicts, but the results should now at least should be the same when the inputs are identical (same atoms, same bonds, and in the same order).

I'm not adding a test because I haven't been able to find a reasonably simple way to reproduce the issue.

### DIFF
--- a/Code/GraphMol/Chirality.cpp
+++ b/Code/GraphMol/Chirality.cpp
@@ -26,6 +26,7 @@
 #include <algorithm>
 #include <cstdlib>
 #include <optional>
+#include <ranges>
 #include <set>
 #include <sstream>
 #include <utility>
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

PR #9242, opened 2026-04-16
Add optional default_val parameter to GetProp() 

### LINKED ISSUE #9241: Add optional default value to Mol.GetProp() in Python
Every call to `Mol.GetProp()` raises `KeyError` when the requested property is absent, requiring callers to wrap every lookup in a try/except block. This is verbose for the common case of wanting a fallback value.

Add an optional default parameter to `Mol.GetProp()`:                                                                                                          
```                                                                                                                                         
  mol.GetProp(key, autoConvert=False, default=<default value>)                                                                                                      
```                                                                                                                                              
If `default` is provided and the property is absent, it is returned directly. If `default` is omitted, the existing `KeyError` behavior is preserved. Where `default=None` is supported and distinct from "not provided".   

https://github.com/rdkit/rdkit/pull/8372

**Additional context**
This mirrors the Python standard library convention (e.g. dict.get(key, default)).

emilyrrhodes:
#### Reference Issue

Fixes #9241 

Add an optional default parameter to Mol.GetProp():

```
mol.GetProp(key, autoConvert=False, default=<default value>)  
```


Code/GraphMol/Wrap/props.hpp
```diff
@@ -40,6 +40,15 @@
...
 
+inline python::object &getPropSentinel() {
+  static python::object sentinel;
+  return sentinel;
+}
+inline void initPropSentinel() {
+  getPropSentinel() = python::object(python::handle<>(
+      PyObject_CallObject((PyObject *)&PyBaseObject_Type, nullptr)));
+}
+
```
emilyrrhodes, line 51: Here I was trying to make a default value that is distinctly different from a python object that a caller could pass in. 

Curious what others think about this and if there might be a better way to implement this?

bp-kelley:
I think it would be easier just to have two GetProp functions with different signatures.  I think this would work:

```
template <class RDOb>
PyObject *GetPyProp(const RDOb *obj, const std::string &key, PyObject *default_value, bool autoConvert) {
  if(obj->hasProp(key)) {
    return GetPyProp(obj, key, autoConvert);
  }
  return default_value;
}
```

These will need to be wrapped as well:

```
        .def(
            "GetProp", GetPyProp<Bond>,
            (python::arg("self"), python::arg("key"), python::arg("default_value"),
             python::arg("autoConvert") = false),
            "Returns the value of the property.\n\n"
            "  ARGUMENTS:\n"
            "    - key: the name of the property to return (a string).\n\n"
            "    - default_value: a value to return if the key doesn't exist\n\n
            "    - autoConvert: if True attempt to convert the property into a python object\n\n"
            "  RETURNS: a string\n\n",
            boost::python::return_value_policy<return_pyobject_passthrough>())
```

I'll note if you add to GetProp you'l probably want to have the same for GetIntProp, etc, etc.

Code/GraphMol/Wrap/props.hpp
greglandrum: Since the type of the default value and the return type are the same, and each possible type is wrapped separately, I think you can just provide the default value as an extra argument with a default of the appropriate type.

It's unfortunate that `GetProp()` already takes an optional second argument, so we have to add a new function to allow `mol.GetProp("foo","bar")`, but we're kind of stuck with that in order to maintain backwards compatibility.

So you'd do something like this for `GetProp()` (untested):
```c++
        .def(
            "GetProp", GetPyProp<ROMol>,
            (python::arg("self"), python::arg("key"),
             python::arg("autoConvert")),
            "Returns the value of the property.\n\n"
            "  ARGUMENTS:\n"
            "    - key: the name of the property to return (a string).\n\n"
            "    - autoConvert: if True attempt to convert the property into a python object\n\n"
            "  RETURNS: a string\n\n"
            "  NOTE:\n"
            "    - If the property has not been set, a KeyError exception will be raised.\n",
            boost::python::return_value_policy<return_pyobject_passthrough>())
        .def(
            "GetProp", GetStringProp<ROMol>,
            (python::arg("self"), ,
             python::arg("default")=""),
            "Returns the value of the property.\n\n"
            "  ARGUMENTS:\n",
            boost::python::return_value_policy<return_pyobject_passthrough>())

```
The first function is what was there before, but I removed the default value for `autoConvert`.
I think this kind of overloading will work.

The others can be done like this:
```
        .def("GetDoubleProp", GetProp<ROMol, double>,
             (python::arg("self"), python::arg("key"), python::arg("default")=0.0),
             "Returns the double value of the property if possible.\n\n"
             "  ARGUMENTS:\n"
             "    - key: the name of the property to return (a string).\n\n"
             "  RETURNS: a double\n\n"
             "  NOTE:\n"
             "    - If the property has not been set, a KeyError exception "
             "will be raised.\n",
             boost::python::return_value_policy<return_pyobject_passthrough>())
```
(of course the docstrings and functions being called would also need to be updated)

emilyrrhodes:
I think there is some ambiguity with something like: 
```
mol.GetProp("key", True)
```

*Issue*
This currently means: 
```
mol.GetProp("key", True)
mol.GetProp("key", autoConvert = True)
```
so in order to keep backwards compatibility we would need to retain this behavior. 

However, something like: 
```
mol.GetProp("key", "string default value")
mol.GetProp("key", default_val = "string default value")
```
indicates that the second argument is the default value. 

Therefore, in order to pass in a boolean as the default value we would have to use the keyword: 
```
mol.GetProp("key", default_val = True)
```
which seems like a great place for bug to hide if someone thinks `mol.GetProp("key", True)` is a way to provide the default value when it is really a way to provide `autoConvert`. 


If there is a way to work around this that resolves the ambiguity, please let me know.

greglandrum:
> If there is a way to work around this that resolves the ambiguity, please let me know.

The boost::python wrappers can resolve overloads based on types, so I think there is a possibility that it could work and have:
`mol.GetProp("foo", True)` call the version with `auto_convert` while `mol.GetProp("foo","bar")` calls the version where `"bar"` is the default value, but we'd definitely need to test that.

We definitely could only have one version of the function accept a single argument (i.e. both forms can't have a default value for the second arg). I would suggest having this be the one that accepts the default value (and always returns strings)

Code/GraphMol/Wrap/props.hpp
bp-kelley: Setting the default to 0.0 means you can't return (or always return) a default of 0.0.

I'm beginning to think we just might need a python version of the C++ GetPropIfPossible for cleanliness but this returns None or the value.

i.e.

```
mol.GetPropIfPossible("foo") -> None
mol.GetPropIfPossible("foo", "bar") -> "bar"
mol.GetPropIfPossible("foo", 2.0) -> 2.0
```

And the other variants
```
mol.GetIntPropIfPossible("foo")
```

bp-kelley: I'll note that we already have a way of optional returns.

```
mol.GetPropsAsDict().get("foo")
```

Or we could just give an optional prop name string to GetData if we are only looking up a single value.  It's not pretty though.

```
mol.GetPropsAsDict("foo").get("foo")
```

Perhaps we could leverage this somehow.

greglandrum: > Setting the default to 0.0 means you can't return (or always return) a default of 0.0.

Ah, shit, right.

greglandrum: > > Setting the default to 0.0 means you can't return (or always return) a default of 0.0.
> 
> Ah, shit, right.

So then we can have the `GetTYPEProp()` functions take an optional value for the default that defaults to None. 
How about resolving the problem with `GetProp()` by leaving it as is, but adding `GetStringProp()` that takes an optional value for the default that defaults to None?

rachelnwalker: Just chiming in :) 

Is there a way to do a sentinel w/o using cpython directly? Claude recommended 
```
  void initPropSentinel() {
    python::object builtins = python::import("builtins");
    getPropSentinel() = builtins.attr("object")();
  }
  ```
It looks like we already import `builtins` here https://github.com/rdkit/rdkit/blob/e35f7db0093afaeda39155f9ea232a8daa47c525/Code/RDBoost/Wrap.cpp#L58

I realize this would mean that to set the default value you'd need to do either
```
mol.GetProp("key", True, my_val) # or
mol.GetProp("key", default=my_val)
```
But I like that better than making `mol.GetProp("key", True)` ambiguous (as @emilyrrhodes mentioned) 

> I'm beginning to think we just might need a python version of the C++ GetPropIfPossible for cleanliness but this returns None or the value.

I like this idea, but `getPropIfPresent` returns a bool in C++ and that difference seems odd.

> How about resolving the problem with GetProp() by leaving it as is, but adding GetStringProp() that takes an optional value for the default that defaults to None?

I assume we'd add the default for all the other `Get[Type]Prop` functions as well?

Comment (bp-kelley) [reply Code/GraphMol/Wrap/props.hpp]: I think GetPropIfPresent works for two reasons, (1) we can't store a python None as a prop and (2) there is no real way to implement `bool getPropIfPresent(string, T&)` in python.

I think `val = getPropIfPresent(string)` where val = None means it isn't there is pythonic and `val = getPropIfPresent(string, default)` works well but is pretty verbose, it has the same semantics as dict.get which actually inspired getPropIfPresent

Comment (emilyrrhodes) [reply Code/GraphMol/Wrap/props.hpp]: I tried to change the review to match what @bp-kelley is describing here. Let me know if this matches everyone's expectations.

Comment (d-b-w) [reply Code/GraphMol/Wrap/props.hpp]: It sounds like some of this discussion is about whether the default value should influence the return type. I think that it should not - a very common "default" return type is None.

I appreciate @bp-kelley's examples that showed what this would look like in Python. I think that we should discuss the _interface that we want_, and we can get however we need to.

I think that the most robust and clear way to do this at the calling site would be:

```
mol.GetProp('k') # throws KeyError if 'k' is not present
mol.GetProp('k', default=None) # returns None if 'k' is not present
mol.GetProp('k', autoconvert=True, default=None)

mol.GetIntProp('k') # throws KeyError if 'k' is not present
mol.GetIntProp('k', default=-5) # returns -5 if 'k' is not present
```

I guess the concern is - what if I see this? What did the author intend?
```
mol.GetProp('k', True) # actually means autoconvert=True, but is ambiguous)
```

Python has keyword-only arguments - and **nanobind supports them** https://nanobind.readthedocs.io/en/latest/functions.html#keyword-only-parameters . I think we should make default a keyword-only argument once we switch to nanobind.

boost::python doesn't natively support keyword-only arguments. I think we can fake it for a few months until we switch to nanobind. If you think we absolutely need to lock this to keyword-only in boost::python... we can do it but it's not gonna be pretty (I think we'd need to use raw_function and manually unbind the arguments)

Comment (rachelnwalker) [reply Code/GraphMol/Wrap/props.hpp]: What @d-b-w is suggesting here (and what @emilyrrhodes originally implemented) would be my preference in terms of interface

Comment (greglandrum) [reply Code/GraphMol/Wrap/props.hpp]: > What @d-b-w is suggesting here (and what @emilyrrhodes originally implemented) would be my preference in terms of interface

I'm ok with this too, but it will require that the `default=` keyword be provided in many cases, which I somehow thought was undesirable

**Comment by emilyrrhodes** [comment]
To summarize where we are at: 

The desired user interface looks like:
```
mol.GetProp("prop_key", default="default_value")
```

Since `default` is the third argument you *cannot* do something like: 
```
mol.GetProp("prop_key", "default_value")
```

Does this work for everyone?

**Comment by emilyrrhodes** [comment]
@bp-kelley I am curious what your thoughts are on this implementation and if there is anything you would like to see changed.

File: Code/GraphMol/Wrap/SubstanceGroup.cpp [hunk 2]
```diff
@@ -260,26 +260,54 @@ struct sgroup_wrap {
             "    - If the property has not been set, a KeyError exception "
             "will be raised.\n",
             boost::python::return_value_policy<return_pyobject_passthrough>())
+        .def(
+            "GetProp", GetPyPropOrDefault<SubstanceGroup>,
+            (python::arg("self"), python::arg("key"),
+             python::arg("autoConvert") = false,
```
greglandrum, line 266: I think you shouldn't provide the default value here.

Code/GraphMol/Wrap/Mol.cpp
```diff
@@ -757,6 +757,18 @@ struct mol_wrapper {
             "  NOTE:\n"
             "    - If the property has not been set, a KeyError exception will be raised.\n",
             boost::python::return_value_policy<return_pyobject_passthrough>())
+        .def(
+            "GetProp", GetPyPropOrDefault<ROMol>,
+            (python::arg("self"), python::arg("key"),
+             python::arg("autoConvert") = false,
```
greglandrum, line 763: I think you shouldn't have the default value here

Code/GraphMol/Wrap/Conformer.cpp
```diff
@@ -240,6 +240,18 @@ struct conformer_wrapper {
             "    - If the property has not been set, a KeyError exception "
             "will be raised.\n",
             boost::python::return_value_policy<return_pyobject_passthrough>())
+        .def(
+            "GetProp", GetPyPropOrDefault<Conformer>,
+            (python::arg("self"), python::arg("key"),
+             python::arg("autoConvert") = false,
```
greglandrum, line 246: I think you shouldn't have the default value here.

Code/GraphMol/Wrap/Bond.cpp
```diff
@@ -216,6 +216,18 @@ struct bond_wrapper {
             "    - If the property has not been set, a KeyError exception "
             "will be raised.\n",
             boost::python::return_value_policy<return_pyobject_passthrough>())
+        .def(
+            "GetProp", GetPyPropOrDefault<Bond>,
+            (python::arg("self"), python::arg("key"),
+             python::arg("autoConvert") = false,
```
greglandrum, line 222: I think you shouldn't have the default value here.

Code/GraphMol/Wrap/Atom.cpp
```diff
@@ -344,6 +344,18 @@ struct atom_wrapper {
             "    - If the property has not been set, a KeyError exception "
             "will be raised.\n",
             boost::python::return_value_policy<return_pyobject_passthrough>())
+        .def(
+            "GetProp", GetPyPropOrDefault<Atom>,
+            (python::arg("self"), python::arg("key"),
+             python::arg("autoConvert") = false,
```
greglandrum, line 350: I think you shouldn't have the default value here.

Code/GraphMol/Wrap/testGetPropDefault.py
greglandrum, line 1: The tests in here are never run as part of running `ctest`,
You do call `add_pytest()` in `CMakeLists.txt`, but that creates a test that can is run as:
`python testGetPropDefault.py`
Doing that with this file doesn't run anything.

[RDKitUtils.cmake](https://github.com/rdkit/rdkit/blob/master/Code/cmake/Modules/RDKitUtils.cmake) also defines a command `add_pythonpytest()` which is designed for use with pytest, but that takes a directory as an argument and uses pytest to discover all pytestable scripts in that directory. When I try that with your branch it ends up running all the *other* tests in the directory Code/GraphMol/Wrap, but not yours. I don't understand pytest discovery at all.

I can think of three options, maybe you can come up with others:
1. Convert this file to use python's unittest framework, like all of the other tests of the python wrappers in the Code subtree.
2. Figure out how to modify this file so that it uses pytest to actually run tests when invoked as `python testGetPropDefault.py`
3. Update `Code/GraphMol/Wrap/CMakeLists.txt` to remove the individual tests definitions in this directory and make pytest run them all (including this one).

I'm fine with any of these, but have a slight preference for 2 or 3 since we could use those as a pattern to use pytest when adding future tests.

> Aside: I was resisting expanding our use of pytest, but after seeing how you have done these, and how compact and expressive the tests are, I now get it and my resistance has evaporated.

greglandrum (changes requested):
Some small code changes suggested.
The big thing is fixing the problem with the tests.

Code/GraphMol/Wrap/Atom.cpp
emilyrrhodes: I am curious why we should not have a default for `GetProp`, but should for `GetIntProp`, `GetUnsignedProp`, `GetDoubleProp`, and `GetBoolProp`. 

Is it mainly because this will create confusion about what is the second argument? (The second argument being `autoConvert` for `GetProp` and `default` for everything else.) I think not providing a `default` for `GetProp` could also create some confusion and am curious why omitting it is better than the argument becoming the third, rather than the second, argument.

Code/GraphMol/Wrap/testGetPropDefault.py
emilyrrhodes: Yes, I'll work on making these changes.

Code/GraphMol/Wrap/Atom.cpp
greglandrum: Sorry I was too terse with my confusing comment.

You provide a default value for `autoConvert` - `python::arg("autoConvert") = false` -  but don't provide one for `default` - `python::arg("default")`. I'm actually surprised that this even compiles and works, but maybe I'm underestimating the magic that boost::python does.

emilyrrhodes: Oh I see! Good point - I'll change that.

emilyrrhodes: If I remove the default value for `autoConvert` then I can't do something like: `atom.GetProp("missing_prop", default="provided_default_value")`.

greglandrum: > If I remove the default value for `autoConvert` then I can't do something like: `atom.GetProp("missing_prop", default="provided_default_value")`.

You're right. 
I just did a bit of playing with the code to see what it does and I was definitely underestimating the magic that boost::python does. Though you can't define a python function where arguments with defaults come before arguments without defaults, this clearly isn't a problem with boost::python. It looks like it just uses the extra arguments to resolve overloads when things are called from Python.

So my comments about removing the default value from the autoConvert argument were all wrong. I'm going to mark them as resolved. Sorry for raising the non-issue.

emilyrrhodes: No problem! Great opportunity for me to learn as well.

Code/GraphMol/Wrap/testGetPropDefault.py
emilyrrhodes: From what I understand, I was missing one of two options: 

1. Adding a `__main__` block where `pytest.main([__file__])` is called. Source: https://docs.pytest.org/en/stable/how-to/usage.html
2. Renaming the file to test_*.py or *_test.py (pytest file-discovery rules). Source: https://docs.pytest.org/en/stable/explanation/goodpractices.html

I went with option 1 because the renaming of the file seemed non-obvious and difficult to know to copy if someone were to write other pytest tests and did not match the naming of the other test files.

greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @emilyrrhodes!

PR #9218, opened 2026-04-01
Add some std::ranges support

greglandrum:
This adds std::ranges support for: `ROMol::atoms()` (random access), `ROMol::bonds()` (bidirectional), and `ROMol::atomNeighbors()` (random access).

This allows using range algorithms and views, so you do stuff like loop over atoms/bonds in either direction:
```c++
    {
      std::vector<unsigned int> atomDegrees;
      std::ranges::transform(atoms, std::back_inserter(atomDegrees),
                             [](const auto atom) { return atom->getDegree(); });
      CHECK(atomDegrees == std::vector<unsigned int>{1, 3, 1, 2, 1});
    }
    {
      std::vector<unsigned int> atomDegrees;
      std::ranges::transform(atoms | std::views::reverse,
                             std::back_inserter(atomDegrees),
                             [](const auto atom) { return atom->getDegree(); });
      CHECK(atomDegrees == std::vector<unsigned int>{1, 2, 1, 3, 1});
    }
```
and use filters:
```c++
    std::ranges::transform(atoms | std::views::filter([](const auto atom) {
                             return atom->getAtomicNum() == 6;
                           }) | std::views::take(2),
                           std::back_inserter(atomIndices),
                           [](const auto atom) { return atom->getIdx(); });
    CHECK(atomIndices == std::vector<unsigned int>{0, 2});
```
The iterators are exposed so that you cannot modify the graph itself through them - the Atoms and Bonds can be changed, but the underlying connectivity cannot be.

There's some simple benchmarking code in the tests. My laptop isn't the best place to run this kind of stuff, but this shows that there can be a small penalty for using ranges, but that's measured in a loop that is essentially doing no work. I don't think it will be noticeable with real calculations. 

I've tested it with g++, clang++, and MSVC++

I believe that this should be safe to go in a patch release, but would appreciate feedback on that.

## Benchmarking info added post-merge
I did some simple benchmarking to see what the fastest way is to loop over atoms in a molecule and do some cheap operation on each.

The core benchmark was (this is the version using `mol->atoms()`):
```c++
    int accum = 0;
    for (auto &m : mols) {
      for (unsigned int i = 0; i < 100000; ++i) {
        for (auto a : m->atoms()) {
          accum += a->getFormalCharge();
        }
      }
    }
```
`mols` had 1000 molecules in it

Running on my linux box (build with g++) the times for the various approaches were:
```
m->atoms() elapsed time: 545 ms
accumulate elapsed time: 542 ms
[...]

PR #9243, opened 2026-04-17
Cleanup/get atoms and bonds

greglandrum:
The benchmarking I did after finishing #9218 showed that the old pattern of iterating between `mol.getBeginAtoms()`/`mol.getEndAtoms()` in for loops is quite a bit slower than using a range-based for over `mol.atoms()`. The same thing is true for looping over bonds.

This PR replaces all occurrences of the first pattern with the second in the non-testing RDKit code. I will come back later and do the same thing for the testing code, but that seems less important.

I also did a bit of modernization in the Gasteiger charge calculation code while I was in there.


bp-kelley:
I'm glad to finally see this change.  It's way easier on the eyes :)  Also happy to see that #9218 makes it faster.

bp-kelley (approved):

PR #9244, opened 2026-04-20
Make DiGraph::getNodes() only visit already seen nodes

d-b-w:
This makes the bug that Ricardo pointed out in #9171 deterministic - programming errors in auxiliary labelling make it happen every time.


d-b-w:
Looks like this found some mismatches in CIP ranks from the reference! I'm looking through them.

Issue #9248, opened 2026-04-23
Multiple Bugs in ETKDG Torsion Angle Contribs


scal444:
**Describe the bug**

3 bugs found:

1. Early exit of contrib loop when encountering one incalculable contrib. A zero length vector causes all contribs to be skipped with an [early return](https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/ForceFieldHelpers/CrystalFF/TorsionAngleContribs.cpp#L116), rather than skipping the one problematic torsion.

2. Incorrect force constant for [6th term in gradient](https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/ForceFieldHelpers/CrystalFF/TorsionAngleContribs.cpp#L139). This is https://github.com/rdkit/rdkit/issues/8715, which I think we lost track of.

3. Missing cosPhi term in final gradient. The final part of the [energy term](https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/ForceFieldHelpers/CrystalFF/TorsionAngleContribs.cpp#L35) is a function of (`cos6(phi) cos4(phi) cos2(phi)`), the final part of the [gradient](https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/ForceFieldHelpers/CrystalFF/TorsionAngleContribs.cpp#L139-L141) should be function of (`cos5(phi) cos3(phi) cos(phi)`)

scal444:
These issues were discovered or re-discovered when expanding finite difference checker to assess all forcefields (#9249). 

@greglandrum back when we discussed in issue #8715, maybe the missing piece was that there was another incorrect part of that 6th gradient term?

PR #9249, opened 2026-04-23
Fix CrystalFF ETKDG bugs ( bug #9248) and expand Finite difference checker to all forcefields

scal444:
Expands finite difference checker to run ETKDG, MMFF, UFF on the MMFF dataset. UFF doesn't support all of the molecules, a few cases are skipped based on invalid assignment or high energy.

ETKDG forcefield header is exposed under a testing namespace so to warn against general use. 

Addresses #9248


scal444:
On my laptop, this takes 5-10 second

greglandrum:
@scal444 it looks like the CI test failures here are real

scal444:
A reference molecule needed to be regenerated

scal444:
Remaining failure is in conformer pruning code. The expected number of conformers after pruning is hard-coded. 

I'll do a deeper analysis and report back on whether this is a regression or if we're unmasking more downstream effects of the bug.

PR #9251, opened 2026-04-24
Pandastools improvements

marcobICR:
#### Reference Issue
Adds functionality to PandasTools.LoadSDF function to automatically convert read properties into Python objects (see #9235).

1. I have changed the extraction of the properties from a tuple comprehension to the rdkit `mol.GetPropsAsDict` function, as this allows for passing the "autoConvertStrings" option to convert the properties to Python objects. I have also added the same option to the `LoadSDF` function, allowing the conversion to take place only when needed to avoid breaking current behaviour
2. I have cleaned up and added some details to the `LoadSDF` docstring
3. I have slightly changed the way the check for gzipped files is done. Previously, it checked for the extension ".gz" (notice the full stop) and this didn't allow reading in ".sdfgz" files as they were missing the full stop before the "gz". What I did is change the check to only look for the "gz" portion at the end of the string, so that it will match both ".sdf.gz" and ".sdfgz" extensions.

Happy to discuss these changes further and revert them if needed.

#### Any other comments?

I have only modified the Python code. All tests related to PandasTools passed.

### DIFF
--- a/rdkit/Chem/PandasTools.py
+++ b/rdkit/Chem/PandasTools.py
@@ -240,7 +240,7 @@ else:
 
   def LoadSDF(filename, idName='ID', molColName='ROMol', includeFingerprints=False,
               isomericSmiles=True, smilesName=None, embedProps=False, removeHs=True,
-              strictParsing=True, sanitize=True):
+              strictParsing=True, sanitize=True, autoConvertStrings=False):
     '''Read file in SDF format and return as Pandas data frame.
       If embedProps=True all properties also get embedded in Mol objects in the molecule column.
       If molColName=None molecules would not be present in resulting DataFrame (only properties
@@ -268,7 +268,7 @@ else:
                                   strictParsing=strictParsing)):
       if mol is None:
         continue
-      row = dict((k, mol.GetProp(k)) for k in mol.GetPropNames())
+      row = mol.GetPropsAsDict(autoConvertStrings=autoConvertStrings)
       if molColName is not None and not embedProps:
         for prop in mol.GetPropNames():
           mol.ClearProp(prop)
--- a/rdkit/Chem/PandasTools.py
+++ b/rdkit/Chem/PandasTools.py
@@ -246,9 +246,9 @@ else:
       Arguments:
       
        - filename: path to the SDF file or a file-like object.
-       - idName: name of the column to be used for molecule title. Defaults to "ID".
-       - molColName: name of the column to be used for RDKit molecule objects. If None, molecules will not be included in resulting DataFrame. Defaults to "ROMol".
-       - includeFingerprints: if True, precompute Avalon fingerprints and store them within the molecule objects to accelerate substructure matching. Defaults to False.
+       - idName: name of the column to be used for the molecule title. Defaults to "ID".
+       - molColName: name of the column to be used for the RDKit molecule objects. If None, molecules will not be included in resulting DataFrame. Defaults to "ROMol".
+       - includeFingerprints: if True, precompute fingerprints and store them within the molecule objects to accelerate substructure matching. Defaults to False.
        - isomericSmiles: if True, generated SMILES will include isomeric information. Defaults to True.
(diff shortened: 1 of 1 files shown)


rdkit/Chem/PandasTools.py
```diff
@@ -240,17 +240,35 @@ def RenderImagesInAllDataFrames(images=True):
...
+        A pandas DataFrame containing the data from the SDF file.
+      
+      
+      Note:
+      
       If neither molColName nor smilesName are set, sanitize=false.
       '''
     if isinstance(filename, str):
-      if filename.lower()[-3:] == ".gz":
+      if filename.lower()[-2:] == "gz":
```
greglandrum: This is a bit more explicit and readable:
```suggestion
      if filename.lower().endswith("gz"):
```

rdkit/Chem/PandasTools.py
```diff
@@ -416,7 +434,7 @@ def WriteSDF(df, out, molColName='ROMol', idName=None, properties=None, allNumer
     '''
   close = None
   if isinstance(out, str):
-    if out.lower()[-3:] == ".gz":
+    if out.lower()[-2:] == "gz":
```
greglandrum:
```suggestion
    if out.lower().endswith("gz"):
```

rdkit/Chem/PandasTools.py
```diff
@@ -240,17 +240,35 @@ def RenderImagesInAllDataFrames(images=True):
...
-      would be read).
-
-      Sanitize boolean is passed on to Chem.ForwardSDMolSupplier sanitize.
+      
+      Arguments:
+      
+       - filename: path to the SDF file or a file-like object.
+       - idName: name of the column to be used for molecule title. Defaults to "ID".
+       - molColName: name of the column to be used for RDKit molecule objects. If None, molecules will not be included in resulting DataFrame. Defaults to "ROMol".
+       - includeFingerprints: if True, precompute Avalon fingerprints and store them within the molecule objects to accelerate substructure matching. Defaults to False.
```
greglandrum: Details of which fingerprint are calculated are not important:
```suggestion
       - includeFingerprints: if True, precompute fingerprints and store them within the molecule objects to accelerate substructure matching. Defaults to False.
```

rdkit/Chem/PandasTools.py
```diff
@@ -240,17 +240,35 @@ def RenderImagesInAllDataFrames(images=True):
...
-      If molColName=None molecules would not be present in resulting DataFrame (only properties
-      would be read).
-
-      Sanitize boolean is passed on to Chem.ForwardSDMolSupplier sanitize.
+      
+      Arguments:
+      
+       - filename: path to the SDF file or a file-like object.
+       - idName: name of the column to be used for molecule title. Defaults to "ID".
+       - molColName: name of the column to be used for RDKit molecule objects. If None, molecules will not be included in resulting DataFrame. Defaults to "ROMol".
```
greglandrum:
```suggestion
       - molColName: name of the column to be used for the RDKit molecule objects. If None, molecules will not be included in resulting DataFrame. Defaults to "ROMol".
```

rdkit/Chem/PandasTools.py
```diff
@@ -240,17 +240,35 @@ def RenderImagesInAllDataFrames(images=True):
...
-      If embedProps=True all properties also get embedded in Mol objects in the molecule column.
-      If molColName=None molecules would not be present in resulting DataFrame (only properties
-      would be read).
-
-      Sanitize boolean is passed on to Chem.ForwardSDMolSupplier sanitize.
+      
+      Arguments:
+      
+       - filename: path to the SDF file or a file-like object.
+       - idName: name of the column to be used for molecule title. Defaults to "ID".
```
greglandrum:
```suggestion
       - idName: name of the column to be used for the molecule title. Defaults to "ID".
```

greglandrum (changes requested):
Thanks for the contribution!
I've made a few small suggestions.

marcobICR (commented):
Thanks for the review! Everything here looks good.

greglandrum (approved):
LGTM

greglandrum:
@marcobICR don't worry about the minimallib build failure; that's not because of this code change.

PR #9255, opened 2026-04-28
Extended fix for #9101

### LINKED ISSUE #9101: SDMolSupplier length/indexing regression in 2025.09.5 when "$$$$\n" crosses 64KiB chunk boundary

RaulSofia:
#### Reference Issue
Fixes #9101 (extends what was already done)
Original fixing was done in PR #9162 
Original source of the bug was PR #9010 

After version 2025.9.5, buffered reading was introduced for better performance with SDMolSupplier. This did not account for a boundary condition where a delimiter was present at the very end of the buffer. PR #9162 fixed this (somewhat) and introduced testing. However, while the test passed, similar problems still persisted on some real files I was using. The problem was that the fix was using the older `checkForEnd`, which changes the pointer position, instead of `peekCheckForEnd` (introduced in PR #9010 ) that restores the pointer position and avoids drift. The test case was not catching this because it only contained 2 molecules (drift would only be observed after that).

I changed the supplier to always use the peek version of the function, and added an extended test case with three molecules, that catches this issue.

#### Any other comments?
I did not remove the old test (although it may be redundant now, it really is just one more test).

Some other QoL changes:
- Some tests were failing on Windows, due to the "spaces in paths" issue. It is a common issue on Windows since the username folder itself usually contains a space, and i figured it wouldn't hurt to update that, it may help other people building there.
- Due to garbage collection delays (also probably only visible on Windows), a test was failing. I forced garbage collection on that one, it is just a test, doesn't hurt performance, and makes it more deterministic.
- gitignored a test file that was being generated

Ran the changes through clang too.

### DIFF
--- a/Code/GraphMol/FileParsers/SDMolSupplier.cpp
+++ b/Code/GraphMol/FileParsers/SDMolSupplier.cpp
@@ -358,11 +358,8 @@ void SDMolSupplier::buildIndexTo(unsigned int targetIdx) {
               posHold = posHold + std::streamoff(1);
               needEOL = false;
             }
-            this->checkForEnd();
-          } else {
-            this->peekCheckForEnd(nlPos, bufEnd,
-                                  posHold);  // the optimized peek version
           }
+          this->peekCheckForEnd(nlPos, bufEnd, posHold);  // the optimized peek version
           if (!this->df_end) {
             d_molpos.push_back(posHold);
             ++d_last;
--- a/Code/GraphMol/FileParsers/SDMolSupplier.cpp
+++ b/Code/GraphMol/FileParsers/SDMolSupplier.cpp
@@ -359,7 +359,8 @@ void SDMolSupplier::buildIndexTo(unsigned int targetIdx) {
               needEOL = false;
             }
           }
-          this->peekCheckForEnd(nlPos, bufEnd, posHold);  // the optimized peek version
+          this->peekCheckForEnd(nlPos, bufEnd,
+                                posHold);  // the optimized peek version
           if (!this->df_end) {
             d_molpos.push_back(posHold);
             ++d_last;
--- a/Code/GraphMol/FileParsers/testMolSupplier.cpp
+++ b/Code/GraphMol/FileParsers/testMolSupplier.cpp
@@ -2753,7 +2753,8 @@ TEST_CASE("chunk boundary stream drift with 3+ molecules") {
       "M  END\n"
       "$$$$\n";
 
-  // first separator must hit exactly at the 65536 byte boundary (the others could too but no need)
+  // first separator must hit exactly at the 65536 byte boundary (the others
+  // could too but no need)
   size_t paddingSize = 65536 - m1.size() - 5;
   std::string padding(paddingSize, 'x');
   std::string data = m1 + padding + m1tail + m2 + m3;
--- a/Code/GraphMol/FileParsers/testMolSupplier.cpp
+++ b/Code/GraphMol/FileParsers/testMolSupplier.cpp
@@ -2725,33 +2725,33 @@ TEST_CASE("github9101 - $$$$ at buffer end") {
 // (although the second created drift it only took effect on the next molecule
 // which would be the third). this creates the edge case but with three
 TEST_CASE("chunk boundary stream drift with 3+ molecules") {
-  std::string m1 =
-      "mol1\n"
-      "     RDKit          3D\n"
-      "\n"
-      "  1  0  0  0  0  0  0  0  0  0999 V2000\n"
-      "    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0\n"
-      "M  END\n"
-      "> <comment>\n";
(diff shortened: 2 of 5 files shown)


RaulSofia:
You may have noticed that the two last commits are empty, just for retriggering the CI pipeline. First, a Java test (in DiversityPickerTests.java, test3) wasn't passing, due to randomness in the test I believe, so I relaunched. On the second attempt, some dependencies failed to download (I was seeing 404s during the build phase) and so it couldn't properly build some instances.
I finally tried a third time, no changes, and it now passes all checks

greglandrum:
> You may have noticed that the two last commits are empty, just for retriggering the CI pipeline. First, a Java test (in DiversityPickerTests.java, test3) wasn't passing, due to randomness in the test I believe, so I relaunched. On the second attempt, some dependencies failed to download (I was seeing 404s during the build phase) and so it couldn't properly build some instances. I finally tried a third time, no changes, and it now passes all checks

Hi @RaulSofia just as an FYI for the future: you don't need to worry about doing this kind of dance to make the CI builds work. We know which tests occasionally fail and either just ignore them or trigger a re-run of the failing tests if there are any questions.

Code/GraphMol/FileParsers/testMolSupplier.cpp
```diff
@@ -2719,3 +2719,68 @@ TEST_CASE("github9101 - $$$$ at buffer end") {
   REQUIRE(mol);
   delete mol;
 }
+
+// extends the "github9101 - $$$$ at buffer end" test case
+// just two molecules werent enough to reliably trigger the issue
+// (although the second created drift it only took effect on the next molecule
+// which would be the third). this creates the edge case but with three
+TEST_CASE("chunk boundary stream drift with 3+ molecules") {
+  std::string m1 =
```
greglandrum: Please use a raw string for this so that you don't have to have all of the extra quotation marks and `\n`s

If you have questions about how to do that, take a look at the test that starts on line 2627 of this file

RaulSofia:
@greglandrum got it

Projects/DbCLI/UnitTestDbCLI.py
greglandrum, line 1: I think the changes in this file are not connected to an actual issue and don't reflect anything that is a common problem. They do create something of a mystery though.

Please remove this from this PR. If it's important to you, please create a separate PR for it and explain when it's relevant and what it does there.

greglandrum (changes requested):
A couple of small changes requested

Code/GraphMol/FileParsers/SDMolSupplier.cpp
```diff
@@ -358,11 +358,9 @@ void SDMolSupplier::buildIndexTo(unsigned int targetIdx) {
               posHold = posHold + std::streamoff(1);
               needEOL = false;
             }
-            this->checkForEnd();
```
bp-kelley, line 361: I think this is the right fix, I was trying to remember why it wasn't done in the first place but couldn't think of anything.

greglandrum (approved):
LGTM

PR #9258, opened 2026-04-29
Ensure that StereoGroups don't have duplicate atoms or bonds

### LINKED ISSUE #9259: Heap-use-after-free in RDKit `cleanupStereoGroups` via duplicate CXSMILES stereo group atom index
## Summary

RDKit's CXSMILES parser allows stereo group atom indices to be specified as a comma-separated list (e.g., `|o1:0,0|`), with no deduplication. When the same atom index appears twice, the stereo group's internal `d_atoms` vector contains two copies of the same raw `Atom*` pointer. During SMILES sanitization, `removeHs` calls `removeAtomFromGroups` to purge the freed H atom from any stereo groups — but `removeAtomFromGroups` uses `std::find` and erases only the **first** occurrence, leaving the second dangling pointer behind. The group is non-empty after this partial cleanup and is not discarded. Moments later, `assignStereochemistry` calls `cleanupStereoGroups`, which iterates the group's atom list and reads `atom->getChiralTag()` through what is now freed heap memory, producing a deterministic heap-use-after-free.

The bug is reachable via any unauthenticated SMILES parsing path: HTTP REST APIs, the RDKit PostgreSQL cartridgpossible Heap-use-after-free in RDKit `cleanupStereoGroups` functione, Python bindings, and Jupyter notebook servers that accept user-supplied SMILES strings. A single crafted SMILES string reliably crashes the process, and in production (non-ASAN) builds the freed memory may be recycled before `cleanupStereoGroups` runs, creating a plausible path to information disclosure or memory corruption beyond denial of service.

### Background

RDKit's `MolFromSmiles` function processes SMILES strings in two phases. First, the Flex/Bison parser allocates `Atom` objects on the heap and builds the molecular graph. The CXSMILES suffix (`|...|`) parser runs as part of this phase; it handles enhanced stereo group descriptors such as `|o1:N,...|` (OR groups) and `|&1:N,...|` (AND groups), storing each group as a `StereoGroup` object whose `d_atoms` field is a `std::vector<Atom*>` of non-owning raw pointers. Second, after parsing, `MolFromSmiles` calls `removeHs` (which frees explicit hydrogen atoms) and `assignStereochemistry` (which processes stereo groups).

### Vulnerable Code — `parse_enhanced_stereo` in CXSmilesOps.cpp

https://github.com/rdkit/rdkit/blob/cb251343b9448601963cbcec5057a76ff02b27c2/Code/GraphMol/SmilesParse/CXSmilesOps.cpp#L1357-L1404

For input `|o1:0,0|`, this loop executes twice for index `0`, pushing the same `Atom*` twice: `d_atoms = [H_atom, H_atom]`.

### First Faulty Condition — `removeAtomFromGroups` in StereoGroup.cpp

https://github.com/rdkit/rdkit/blob/cb251343b9448601963cbcec5057a76ff02b27c2/Code/GraphMol/StereoGroup.cpp#L76-L93

When `d_atoms = [H_atom, H_atom]`:
1. `std::find` locates the first copy at index 0 and erases it.
2. `d_atoms` is now `[H_atom]` — one stale pointer remains.
3. The group has `getAtoms().size() == 1`, so the empty-group check does NOT discard it.

After `removeAtomFromGroups` returns, `RWMol::removeAtom` calls `delete atom` (RWMol.cpp:496), physically freeing the 80-byte heap region. The second copy of `H_atom*` in `d_atoms` is now a dangling pointer.

### The UAF — `cleanupStereoGroups` in Chirality.cpp

https://github.com/rdkit/rdkit/blob/cb251343b9448601963cbcec5057a76ff02b27c2/Code/GraphMol/Chirality.cpp#L2182-L2212

`getChiralTag()` is an inline 1-byte read at `Atom.h:245` that accesses `d_chiralTag` at offset +48 within the freed 80-byte `Atom` region. ASAN reports this as `READ of size 1 at ... 48 bytes inside of 80-byte region`.

## Recommended Fix

Deduplicate atom pointers when building stereo groups (`parse_enhanced_stereo`) inline or using a separate functions. Errors can also be thrown if duplicates have been found.

```diff
--- a/Code/GraphMol/SmilesParse/CXSmilesOps.cpp
+++ b/Code/GraphMol/SmilesParse/CXSmilesOps.cpp
@@ -1355,7 +1355,9 @@ bool parse_enhanced_stereo(...) {
     if (read_int(first, last, aidx)) {
       if (VALID_ATIDX(aidx)) {
         Atom *atom = mol.getAtomWithIdx(aidx - startAtomIdx);
-        atoms.push_back(atom);
+        if (std::find(atoms.begin(), atoms.end(), atom) == atoms.end()) {
+          atoms.push_back(atom);
+        }
       }
     }
```

## Attribution
This vulnerability was discovered by Claude, Anthropic's AI assistant, and triaged manually by Ada Logics in collaboration with Anthropic Research.

greglandrum:
At the moment it's possible to create a StereoGroup that has duplicate atoms or bonds; this PR fixes that.

I also added explicit checks for this situation in the CXSMILES and Mol block parsers.

Fixes #9259

### DIFF
--- a/Code/GraphMol/FileParsers/MolFileParser.cpp
+++ b/Code/GraphMol/FileParsers/MolFileParser.cpp
@@ -245,7 +245,21 @@ std::string parseEnhancedStereo(std::istream *inStream, unsigned int &line,
       for (size_t i = 0; i < count; ++i) {
         ss >> index;
         // atoms are 1 indexed in molfiles
-        atoms.push_back(mol->getAtomWithIdx(index - 1));
+        auto atom = mol->getAtomWithIdx(index - 1);
+        if (std::ranges::find(atoms, atom) != atoms.end()) {
+          std::string message =
+              (boost::format(
+                   "Atom %1% appears more than once in stereo group specification on line %2%!") %
+               index % line)
+                  .str();
+          if (strictParsing) {
+            throw FileParseException(message);
+          } else {
+            BOOST_LOG(rdWarningLog) << message << std::endl;
+          }
+        } else {
+          atoms.push_back(atom);
+        }
       }
       std::vector<Bond *> newBonds;
       groups.emplace_back(grouptype, std::move(atoms), std::move(newBonds),
--- a/Code/GraphMol/FileParsers/file_parsers_catch.cpp
+++ b/Code/GraphMol/FileParsers/file_parsers_catch.cpp
@@ -4107,7 +4107,9 @@ M  V30 1 1 1 2
 M  V30 END BOND
 M  V30 END CTAB
 M  END)CTAB";
-    { REQUIRE_THROWS_AS(MolBlockToMol(ctab), FileParseException); }
+    {
+      REQUIRE_THROWS_AS(MolBlockToMol(ctab), FileParseException);
+    }
     {
       bool sanitize = true;
       bool removeHs = true;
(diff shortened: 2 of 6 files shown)


ricrogz (approved):
LGTM

PR #9264, opened 2026-05-01
Docs: fix CosineSimilarity formula and clarify similarity metric names in BitOps.h

rezabagher:
Reference Issue
Related to Issue #8317
(similarity metric naming inconsistencies).

This PR updates documentation in BitOps.h to clarify similarity metric definitions and correct one formula typo.

Changes include:

Fixed CosineSimilarity denominator (* instead of +).
Added standard literature names for Kulczynski, Sokal, Asymmetric, and AllBit similarities.
Added brief clarification of a/b/c/d notation.
Any other comments?
Docs only; no behavior changes.


greglandrum (changes requested):
Thanks for the contribution @rezabagher!

I agree that the definition provided for the cosine similarity is correct, and think it's fine to provide alternate names for some of the similarity measures, but I disagree with the use of the term "standard literature names". Most of the terms/definitions currently used in the RDKit are from the cheminformatics literature like https://doi.org/10.1023/a:1016387816342  
Specifically, I don't find the clarifications about "Kulczynski 2" or "Sokal-Sneath 2" (terms which do *not* seem to be broadly used) helpful.

I also don't think that the introduction of the a, b, c, and d terms is a good idea. Different symbols are used for those in different papers. The current documentation is explicit about what the terms are without needing to introduce new symbols; I find being explicit better.

Please drop the mention of the a,b,c,d terms, Kulczynski 2, and Sokal-Sneath 2 and I think this is great.

rezabagher:
Thanks for the feedback, I’ve updated the PR to remove the a/b/c/d notation and the references to “Kulczynski 2” and “Sokal–Sneath 2”, while keeping the cosine formula correction.

greglandrum (approved):
LGTM
Thanks for the contribution @rezabagher!

PR #9265, opened 2026-05-02
Multi conf gaussian shape

DavidACosgrove:
#### Reference Issue
No issue

This adds multi-conformer support to the GaussianShape shape overlays.  It doesn't add a great deal of functionality compared with creating individual shapes for each conformer, but makes for a more compact dataset.

There are 2 extra API points - [bB]estSimilarity, which returns the best similarity between the shapes in 2 multi-conformer shape objects, and ScoreMoleculeAllConformers() which does similar for 2 input molecules but also returns the scores matrix for all alignments.

#### Any other comments?
This is a necessary pre-cursor for searching synthon spaces by shape, where the more compact data structure was essential.  That PR will come once this is merged in.


Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -140,125 +173,228 @@ class RDKIT_GAUSSIANSHAPE_EXPORT ShapeInput {
 #endif
   }
 
-  // Note that the coords returned is a vector size 4*getNumAtoms()
-  // with the 4th value per atom being the alpha paramter.
-  const std::vector<double> &getCoords() const { return d_coords; }
+  std::string getSmiles() const { return d_smiles; }
```
bp-kelley: I find the "getSmiles" function odd.  What is the purpose here?  Is this for external book keeping?

Code/GraphMol/GaussianShape/ShapeInput.cpp
```diff
@@ -166,52 +263,151 @@ ShapeInput &ShapeInput::operator=(const ShapeInput &other) {
   return *this;
 }
 
+void ShapeInput::merge(ShapeInput &other) {
+  if (!d_coords.empty() &&
```
bp-kelley, line 269: Shouldn't d_smiles be the same here as well?

Code/GraphMol/GaussianShape/ShapeInput.h
DavidACosgrove: Yes.  Used in the synthon shape search to keep track of the shapes produced by fragments.

Code/GraphMol/GaussianShape/ShapeInput.cpp
DavidACosgrove: Good spot. I've made it a PRECONDITION.

DavidACosgrove:
I think the failure on Windows is unrelated to this PR.

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -64,14 +65,26 @@ void serialize(Archive &ar, boost::dynamic_bitset<Block, Allocator> &bs,
...
-    std::vector<std::tuple<unsigned int, RDGeom::Point3D, double>>;
+
+struct CustomFeature {
+  CustomFeature(
+      const unsigned int t, const RDGeom::Point3D &p, const double r,
+      const std::vector<unsigned int> &a = std::vector<unsigned int>())
+      : type(t), pos(p), rad(r), atoms(a) {}
+  unsigned int type;
+  RDGeom::Point3D pos;
+  double rad;
```
bp-kelley, line 84: Why is this a radii and not, say, in gaussian form (prefactor/width)?  This somewhat limits the utility.

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -82,39 +95,54 @@ struct ShapeInputOptions {
...
       true};  //! Whether to use carbon radii for all atoms (which is quicker
               //! but less accurate) or vdw radii appropriate for the elements.
+  double shapePruneThreshold{-1.0};  //! If there is more than 1 conformer for
+                                     //! the input molecule, prune the shapes so
+                                     //! that none of them are more similar to
+                                     //! each other than the threshold.  Default
+                                     //! -1.0 means no pruning.
+  bool sortShapes{true};  //! If true, the shapes are sorted in descending order
+                          //! of total volume.
+  bool includeDummies{true};  //! Whether to include dummy atoms in the shape
```
bp-kelley, line 123: What radii are dummy atoms given?

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -140,125 +173,229 @@ class RDKIT_GAUSSIANSHAPE_EXPORT ShapeInput {
 #endif
   }
 
-  // Note that the coords returned is a vector size 4*getNumAtoms()
-  // with the 4th value per atom being the alpha paramter.
-  const std::vector<double> &getCoords() const { return d_coords; }
+  std::string getSmiles() const { return d_smiles; }
```
bp-kelley: could be `const std::string getSmiles() const { return d_smiles; }` to avoid the copy if necessary.

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -140,125 +173,229 @@ class RDKIT_GAUSSIANSHAPE_EXPORT ShapeInput {
...
+  //! which will toggle whether the atom/feature is used in the volume
+  //! calculation or not.  For temporarily "turning off" an atom or feature.
+  void negateAlpha(unsigned int alphaNum);
+
   //! Fetch the coordinates of the atoms and optionally features.
   std::vector<RDGeom::Point3D> getAtomPoints(bool includeColors = false) const;
-  bool getNormalized() const { return d_normalized; }
+  //! Return whether the coordinates for the current active shape are
+  //! normalized.
+  bool getNormalized() const { return d_normalizeds[d_activeShape]; }
```
bp-kelley: `getIsNormalized` might make more sense at a glance.

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -140,125 +173,229 @@ class RDKIT_GAUSSIANSHAPE_EXPORT ShapeInput {
...
+  //! Return the feature types of all atoms/features in the shape.  Atoms
+  //! have type 0.
   const std::vector<int> &getTypes() const { return d_types; }
```
bp-kelley: maybe `getFeatureTypes`

bp-kelley:
I'm not too sure I like the activeShape portion of the collection.  I somewhat understand it in the context of molecule conformers, which seems like what you are getting at.   But there is no api to use a shape in the collection without setting the active shape which looks to lock out things like multithreading on the same ShapeInput if one needs to set the current active index.  I would rather the ShapeIndex be a collection of ShapeObjects where you can grab any active shape by `const Shape &shp = shapeinput.getShapeAtIndex(idx);`.

bp-kelley:
I know this is a big change, so take it with a grain of salt but it looks like one can't multithread over conformers just shapeinputs at this stage.

Code/GraphMol/GaussianShape/ShapeInput.h
DavidACosgrove: DUMMY_RAD is defined at line 72, along with CARBON_RAD.

Code/GraphMol/GaussianShape/ShapeInput.h
DavidACosgrove: Because that's how the interface to the PubChem code did it, I guess.  People understand radii.  It's converted to an alpha value on import.  I guess we could make providing the alpha directly an option.

DavidACosgrove:
> I'm not too sure I like the activeShape portion of the collection. I somewhat understand it in the context of molecule conformers, which seems like what you are getting at. But there is no api to use a shape in the collection without setting the active shape which looks to lock out things like multithreading on the same ShapeInput if one needs to set the current active index. I would rather the ShapeIndex be a collection of ShapeObjects where you can grab any active shape by `const Shape &shp = shapeinput.getShapeAtIndex(idx);`.

There is a constructor that takes a ShapeInput and a shape number and returns a new ShapeInput with just that conformation in it which would be functionally equivalent to this although not as fast or efficient.

The thinking behind the design of the object was to minimise the memory footprint for multi-conformer molecules/shapes.  There's a lot of data constant across all conformations and that seemed wasteful, especially for large collections of shapes such as the synthon databases.  The multi-threading is an issue that I have already banged against in the synthon space searching.  I am open to re-thinking the object model so long as the memory use doesn't go up dramatically.

DavidACosgrove:
> I know this is a big change, so take it with a grain of salt but it looks like one can't multithread over conformers just shapeinputs at this stage.

I don't think multi-threading over conformers is a big deal.  The individual overlays are fast enough that unless you were doing proteins or such like or enormous conformer ensembles I don't think the overhead of threading it would be worth it.  What _is_ true is that you can't use the same target multi-conformer shape against multiple queries in different threads, you need a copy for each thread and that might be more of an issue.

bp-kelley:
For pose matching, we were shape matching thousands (10k)  of conformers at a go ,so there is a (niche) use case for it.   The same target in multiple threads is more of an issue, again for niche but important cases.  I won't argue (too much) if there is a good use case for the activeShape book-keeping, but I personally don't think it's necessary and can be limiting in certain cases.   I also don't want to give you too much extra work because I want to start using this!

DavidACosgrove:
> For pose matching, we were shape matching thousands (10k) of conformers at a go ,so there is a (niche) use case for it. The same target in multiple threads is more of an issue, again for niche but important cases. I won't argue (too much) if there is a good use case for the activeShape book-keeping, but I personally don't think it's necessary and can be limiting in certain cases. I also don't want to give you too much extra work because I want to start using this!

@bp-kelley Will you be at the UGM?  We can thrash it out there if so.  By that point we will have had real-world experience of using it and will have seen the bottlenecks and problems.  As I said, I'm not averse to changing it for something better.  It it's just changing the underlying datastructures and adds to rather than changes the API then reviewing and improving it is pretty sensible.

bp-kelley:
I am planning on being at the UGM.

Code/GraphMol/GaussianShape/GaussianShape.h
```diff
@@ -92,30 +92,64 @@ RDKIT_GAUSSIANSHAPE_EXPORT std::array<double, 3> AlignMolecule(
     const ShapeOverlayOptions &overlayOpts = ShapeOverlayOptions(),
     int refConfId = -1, int fitConfId = -1);
 
+//! Calculate scores for the alignment of all conformers of one molecule
+//! onto another.  Returns a matrix of the combination scores, the conformer
+//! numbers of the two molecules that gave the best overlay and the
+//! transformation matrix for that overlay if requested.  The molecules
+//! themselves are not altered.  scores[0][1] is the score of aligning
+//! fit conformation 1 onto ref conformation 0
```
greglandrum:
```suggestion
//! fit conformer 1 onto ref conformer 0
```

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -82,39 +95,54 @@ struct ShapeInputOptions {
...
-  // definitions.
   bool useColors{
       true};  //! Whether to build the color features.  By default, it will
               //! create features using the RDKit pharmacophore definitions.
 
-  CustomFeatures customFeatures;  //! Custom color features used verbatim.  A
-                                  //! vector of tuples of integer type, Point3D
-                                  //! coords, double radius.
+  std::vector<std::vector<CustomFeature>>
+      customFeatures{};  //! Custom color features used verbatim.  One outer
```
greglandrum:
```suggestion
      customFeatures;  //! Custom color features used verbatim.  One outer
```

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -82,39 +95,54 @@ struct ShapeInputOptions {
...
+                         //! vector for each conformation in the molecule.
   std::vector<unsigned int>
-      atomSubset;  //! If not empty, use just these atoms in the molecule to
-                   //! form the ShapeInput object.
+      atomSubset{};  //! If not empty, use just these atoms in the molecule to
```
greglandrum:
```suggestion
      atomSubset;  //! If not empty, use just these atoms in the molecule to
```

Code/GraphMol/GaussianShape/ShapeInput.h
```diff
@@ -82,39 +95,54 @@ struct ShapeInputOptions {
...
+                     //! form the ShapeInput object.
   std::vector<std::pair<unsigned int, double>>
-      atomRadii;  //! Use these non-standard radii for these atoms. The int is
-                  //! for the atom index in the molecule, not the atomic number.
-                  //! Not all atoms need be specified, just some radii can be
-                  //! over-ridden, with the rest left as standard.
+      atomRadii{};  //! Use these non-standard radii for these atoms. The int is
```
greglandrum:
```suggestion
      atomRadii;  //! Use these non-standard radii for these atoms. The int is
```

greglandrum (changes requested):
some small changes to go along with the PR I submitted against your branch

greglandrum:
@bp-kelley are you ok with us merging this as is?

Code/GraphMol/GaussianShape/ShapeInput.h
bp-kelley: The only reason I bring it up is that you can only customize one variable her, the radii not the full operation.  For example a color atom with high amplitude  and low radius.  But we could always add that later I guess.  I prefer Gaussians being exposed since that is the underlying optimization.  But YMMV

bp-kelley:
I think this is fine as a first pass, but I am finding it a bit over complicated.  I think in the future we can look at composition of concerns versus forcing everything together.  That being said, I'm glad it's here.

bp-kelley (approved):

Code/GraphMol/GaussianShape/ShapeInput.h
greglandrum: I completely agree that it would be nice to have the flexibility to vary the weights of features as well as their radii. But maybe this is best done at the ShapeInput level, not the feature level?

greglandrum (approved):
LGTM

PR #9267, opened 2026-05-05
Fix WedgeMolBonds stealing wiggly bonds from adjacent chiral atoms

cdvonbargen:

The standard re-wedging pattern (`clearSingleBondDirFlags` then `WedgeMolBonds`) was stealing wiggly bonds from adjacent chiral centers. `clearSingleBondDirFlags` saves `_UnknownStereo=1` and clears `BondDir` to `NONE`, but `countChiralNbrs` only checked `BondDir` to decide whether a chiral atom's stereo was already expressed — missing the marker. With the chiral atom not pre-skipped, `pickBondToWedge` would pick the wiggly bond itself (terminal neighbors are scored lower), and the caller's subsequent restore of `BondDir::UNKNOWN` would erase the wedge, leaving the chiral atom with no visible stereo.

Fix: extend `countChiralNbrs` to recognize `_UnknownStereo=1` as equivalent to `BondDir::UNKNOWN`. Chiral atoms with wiggly-bond annotations are pre-skipped the same way they already are for bonds that still have `BondDir::UNKNOWN` set.

### DIFF
--- a/Code/GraphMol/WedgeBonds.cpp
+++ b/Code/GraphMol/WedgeBonds.cpp
@@ -54,11 +54,23 @@ namespace detail {
 std::pair<bool, INT_VECT> countChiralNbrs(const ROMol &mol, int noNbrs) {
   INT_VECT nChiralNbrs(mol.getNumAtoms(), noNbrs);
 
-  // start by looking for bonds that are already wedged
+  // start by looking for bonds that are already wedged. A wiggly bond with
+  // BondDir::NONE plus _UnknownStereo=1 (the state left by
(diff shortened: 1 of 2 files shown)


ricrogz:
I think the goal here should be to make `WedgeMolBonds()` handle wiggly bonds the same way as wedge and dash bonds, not to just ignore them when `pickBondToWedge()` picks a bond to be assigned a direction.

cdvonbargen:
Updated `WedgeMolBonds` instead.

greglandrum (approved):
LGTM

greglandrum:
Rolled back in #9293

PR #9272, opened 2026-05-07
Fixes #9270

### LINKED ISSUE #9270: Segfault when calling MolToSmiles on submol
Someone reported that this code results in a segmentation fault:
```python3
from rdkit import Chem

smi = 'CC1/C=C(/C)[O][Ir]23(<-[O]=1)([c]1ccccc1c1cncc[n]->21)[c]1ccccc1c1cncc[n]->31'
mol = Chem.MolFromSmiles(smi)

atomidx = 7
rad = 3
env = Chem.FindAtomEnvironmentOfRadiusN(mol, rad, atomidx)
submol = Chem.PathToSubmol(mol, env)

# all good (apparently)

print(f'{Chem.MolToSmiles(submol)=}')  # segfault
```

They report that with 2025.09.6 this code printed `Chem.MolToSmiles(submol)='c[c](c)[Ir]1([O]/C=C\\C(C)=[O]->1)([c](c)c)(<-[n](c)c)<-[n](c)c'`

ricrogz:
This fixes #9270 and also the handling of double bond stereo in `copyMolSubset()`.

There were multiple issues with the old code here:
- If the bond's stereo atoms were copied into the subset, the stereo atom indexes were mapped twice (once in the `if (toms.size() == 2)` block, and a second one in the `for (auto &atomidx : atoms)` loop after it). this means that the stereo atoms ended up being incorrect, and might point to atoms that are completely unrelated to the double bond, which would cause the seg fault reported in the issue.
 - The second issue was triggered when the at least one of the atoms is not mapped: in such case, the stereo atoms were cleared, but the double bond's stereo was left as-is. This might be fine for E/Z bonds (since these are defined based on CIP priorities, we could still find the stereo atoms), but CIS/TRANS labels would be lacking a reference, which would also cause issues.
 - The code didn't update the stereo labels depending on which atoms/bonds were selected for the copy. On one side, even if the original stereo atoms were not selected, the opposite pair of neighbors might, so we could still translate stereo.
 Also, depending on which atoms/bonds were selected, the CIP priorities might change, so E/Z might potentially flip the stereo on the bond.
 
This patch addresses these issues by implementing a new `handleBondStereo()` function that translates double bond stereo and stereo atoms considering which atoms are selected for the copy.

### DIFF
--- a/Code/GraphMol/Subset.cpp
+++ b/Code/GraphMol/Subset.cpp
@@ -25,6 +25,108 @@ inline void copyComputedProps(const ROMol &src, ROMol &dst) {
   }
 }
 
+atomindex_t getOtherAtomIdx(const ROMol &ref_mol, const Bond &ref_bond,
+                            atomindex_t dblBndAtomIdx,
+                            atomindex_t stereoAtomIdx) {
+  auto ref_atom = ref_bond.getBeginAtom();
+  auto other_atom = ref_bond.getEndAtom();
+  if (other_atom->getIdx() == dblBndAtomIdx) {
+    std::swap(ref_atom, other_atom);
+  }
+  POSTCONDITION(ref_atom->getIdx() == dblBndAtomIdx,
+                "dblBndAtomIdx should be one of the bond's atoms");
+
+  for (auto nbr : ref_mol.atomNeighbors(ref_atom)) {
+    auto nbrIdx = nbr->getIdx();
+    if (nbrIdx != stereoAtomIdx && nbr != other_atom) {
+      return nbrIdx;
+    }
+  }
+
+  return RDKit::Atom::NOATOM;
+}
+
+void handleBondStereo(Bond &extracted_bond, const Bond &ref_bond,
+                      const ROMol &ref_mol,
+                      const std::map<unsigned int, unsigned int> &atomMapping) {
+  auto &atoms = extracted_bond.getStereoAtoms();
+  if (atoms.size() != 2) {
+    return;
+  }
+
+  bool needsSwap = false;
+  auto map1 = atomMapping.find(atoms[0]);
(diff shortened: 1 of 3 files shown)


Code/GraphMol/Subset.cpp
```diff
@@ -25,6 +25,108 @@ inline void copyComputedProps(const ROMol &src, ROMol &dst) {
...
 
+atomindex_t getOtherAtomIdx(const ROMol &ref_mol, const Bond &ref_bond,
+                            atomindex_t dblBndAtomIdx,
+                            atomindex_t stereoAtomIdx) {
+  auto ref_atom = ref_bond.getBeginAtom();
+  auto other_atom = ref_bond.getEndAtom();
+  if (other_atom->getIdx() == dblBndAtomIdx) {
+    std::swap(ref_atom, other_atom);
+  }
+  POSTCONDITION(ref_atom->getIdx() == dblBndAtomIdx,
```
greglandrum: This isn't really a post condition on the function:
```suggestion
  CHECK_INVARIANT(ref_atom->getIdx() == dblBndAtomIdx,
```

bp-kelley:
I think the fixes are correct (I had to wrap my head around if both endpoint atoms are swapped, then the bond stereo should be unchanged). 

In fact it might fix a confusing issue I saw last August when I was doing a larger test.  I'll re-run that and see if they are related.

greglandrum (changes requested):
minor suggestion for which INVARIANT is used

Code/GraphMol/Subset.cpp
ricrogz: I never know with of these checks to use :upside_down_face:

greglandrum (approved):
LGTM

PR #9273, opened 2026-05-07
Fixes #9231: improve escaping values in CXSMILES

### LINKED ISSUE #9231: CXSmiles doesn't round-trip if there's a comma in a property string.
**Describe the bug**
If there's more than 1 property on an atom and the first property has a comma in it, the CXSMILES that's written can't be read back in.

**To Reproduce**
```
mol = Chem.MolFromSmiles("c1ccccc1")
atom = mol.GetAtomWithIdx(0)
atom.SetProp("Prop1", "1,2")
atom.SetProp("Prop2", "0")
smi = Chem.MolToCXSmiles(mol)
print(smi)
newMol = Chem.MolFromSmiles(smi)
print(newMol)
```
gives
```
c1ccccc1 |atomProp:2.Prop1.1,2:2.Prop2.0|
None
```
Adding the properties the other way round gives a molecule:
```
mol = Chem.MolFromSmiles("c1ccccc1")
atom = mol.GetAtomWithIdx(0)
atom.SetProp("Prop2", "0")
atom.SetProp("Prop1", "1,2")
smi = Chem.MolToCXSmiles(mol)
print(smi)
newMol = Chem.MolFromSmiles(smi)
print(newMol)
```
gives
```
c1ccccc1 |atomProp:2.Prop2.0:2.Prop1.1,2|
<rdkit.Chem.rdchem.Mol object at 0x7396800a8970>
```
**Expected behavior**
Both SMILES strings should be parsed correctly.

 - RDKit version: 26.03.1 and current master.
 - OS: [e.g. Ubuntu 20.04] Ubuntu 22.04
 - Python version (if relevant): 3.13.11
 - Are you using conda?mamba

greglandrum:
Uses the info from the [CXSMILES docs on which characters need to be escaped where](https://docs.chemaxon.com/latest/formats_chemaxon-extended-smiles-and-smarts-cxsmiles-and-cxsmarts.html#escaping) to improve our CXSMILES writing and round-tripping.

Note that this undoes the code change made when fixing #5466: that was actually a case where we should not have been escaping the `.` character.

### DIFF
--- a/Code/GraphMol/SmilesParse/CXSmilesOps.cpp
+++ b/Code/GraphMol/SmilesParse/CXSmilesOps.cpp
@@ -1,5 +1,5 @@
 //
-//  Copyright (C) 2016-2021 Greg Landrum and other RDKit contributors
+//  Copyright (C) 2016-2026 Greg Landrum and other RDKit contributors
 //
 //   @@ All Rights Reserved @@
 //  This file is part of the RDKit.
@@ -9,7 +9,6 @@
 //
 #include <RDGeneral/BoostStartInclude.h>
 #include <boost/algorithm/string.hpp>
-#include <boost/lexical_cast.hpp>
 #include <boost/format.hpp>
 #include <RDGeneral/BoostEndInclude.h>
 #include <GraphMol/RDKitBase.h>
@@ -28,6 +27,7 @@
 #include <array>
 #include <map>
 #include <string_view>
+#include <string>
 
 namespace SmilesParseOps {
 using namespace RDKit;
@@ -124,7 +124,7 @@ bool read_int(Iterator &first, Iterator last, unsigned int &res) {
   if (num.empty()) {
     return false;
   }
-  res = boost::lexical_cast<unsigned int>(num);
+  res = std::atoi(num.c_str());
   return true;
 }
 template <typename Iterator>
@@ -137,7 +137,7 @@ bool read_int_list(Iterator &first, Iterator last,
       ++first;
     }
     if (!num.empty()) {
-      res.push_back(boost::lexical_cast<unsigned int>(num));
+      res.push_back(std::atoi(num.c_str()));
     }
     if (first >= last || *first != sep) {
       break;
@@ -182,7 +182,7 @@ std::string read_text_to(Iterator &first, Iterator last, std::string delims) {
       }
       if (next > first + 2) {
         std::string blk = std::string(first + 2, next);
-        res += (char)(boost::lexical_cast<int>(blk));
+        res += (char)(std::atoi(blk.c_str()));
       }
       first = next + 1;
       start = first;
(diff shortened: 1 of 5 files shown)


Code/GraphMol/SmilesParse/CXSmilesOps.cpp
```diff
@@ -441,13 +441,13 @@ bool parse_coords(Iterator &first, Iterator last, RDKit::RWMol &mol,
         std::vector<std::string> tokens;
         boost::split(tokens, tkn, boost::is_any_of(std::string(",")));
         if (tokens.size() >= 1 && tokens[0].size()) {
-          pt.x = boost::lexical_cast<double>(tokens[0]);
+          pt.x = std::atof(tokens[0].c_str());
```
bp-kelley, line 444: Are we ok with the failure modes of std::atof i.e. returns 0.0 if it can't convert.  lexical_cast throws an exception.

Code/GraphMol/SmilesParse/CXSmilesOps.cpp
```diff
@@ -1621,17 +1621,24 @@ getSortedStereoGroupsAndIndices(
...
-std::string quote_atomprop_string(const std::string &txt) {
-  // at a bare minimum, . needs to be escaped
+// from:
+// https://docs.chemaxon.com/latest/formats_chemaxon-extended-smiles-and-smarts-cxsmiles-and-cxsmarts.html#escaping
+constexpr const char *sgroupAllowedSpecialChars = "><\"!@#$%()[]./\\?-+*^_~= ";
+constexpr const char *atomPropAllowedSpecialChars =
+    "><\"!@#$%()[]./\\?-+*^_~= ";
+constexpr const char *labelAllowedSpecialChars = "><\"!@#%()[]./\\?-+*^_~=,: ";
+std::string quote_string(const std::string &txt,
+                         std::string allowedSpecialChars) {
```
bp-kelley, line 1635: It's cheap, but there are a lot of copies of allowedSpecialChars made.

Code/GraphMol/SmilesParse/CXSmilesOps.cpp
```diff
@@ -1621,17 +1621,24 @@ getSortedStereoGroupsAndIndices(
...
   std::string res;
   for (auto c : txt) {
```
bp-kelley, line 1637: Seems like a good use of string_view

```
constexpr const char *sgroupAllowedSpecialChars = "><\"!@#$%()[]./\\?-+*^_~= ";
constexpr std::string_view sgroupAllowed = sgroupAllowedSpecialChars;
```

and then quote_string could just take a string_view by value.

Code/GraphMol/SmilesParse/CXSmilesOps.cpp
greglandrum: I think that's reasonable; but we can stick with the current behavior if you prefer

Code/GraphMol/SmilesParse/CXSmilesOps.cpp
greglandrum: Seems like excessive complexity to just have things be constexpr.
I'll just switch to using const std::strings

greglandrum:
@bp-kelley the requested changes are made. Can you please take another look?

Code/GraphMol/SmilesParse/CXSmilesOps.cpp
bp-kelley: It's fine with me.

bp-kelley (approved):

Issue #9266, opened 2026-05-05
RDKit ETKDGv3 twisted tertiary amides


dkranthi221:
Hello,

I noticed that RDKit (ver 2026.03.1) ETKDGv3() is generating twisted tertiary amides in macrocycles. I am not sure why this is happening but here is an example of this:

<img width="349" height="505" alt="Image" src="https://github.com/user-attachments/assets/b25f0a85-732e-4e62-9536-d6d8b86d7ffd" />

<img width="317" height="236" alt="Image" src="https://github.com/user-attachments/assets/c3a042ef-858a-4569-ba9c-037cf4f9191f" />

Here are some SMILES strings for the macrocycles where I have noticed this:
1. `CC1=CSC2=C1CCN(C(=O)[C@@H]1CCN(C)C(=O)[C@@H](CCO)NC3=CC(=CC(N4C=CC=C4)=C3)C(=O)N1C)C2` molecule_1 (pictured above, amide angle -118 deg)
2. `C=CC1=CC=C(NC(=O)[C@@H]2CCN(C)C(=O)[C@@H](CCO)NC3=CC(=CC(OCC(F)(F)F)=C3)C(=O)N2C)C=C1` molecule_2 (pictured below, amide angle -116 deg)

<img width="832" height="718" alt="Image" src="https://github.com/user-attachments/assets/f3c9536e-4430-45dd-905f-ca49e645177a" />

PR #9279, opened 2026-05-13
Synthon space shape search

DavidACosgrove:
#### Reference Issue
No issue.

This adds shape searching via rdGaussianShape to the synthon space searching.  There has been significant overhaul of the other searches as well for performance and also to make them sit better alongside the new search.
An extra thing is the ability to dump out a file of synthon combinations of potential hits for easier parallelisation of hit verification.

#### Any other comments?
This incorporates PR #9265.


Code/GraphMol/GaussianShape/GaussianShape.h
```diff
@@ -26,6 +26,7 @@ class ROMol;
 class Conformer;
 
 namespace GaussianShape {
+RDKIT_GAUSSIANSHAPE_EXPORT
```
bp-kelley: You can remove this line, it looks to have crept in by accident

DavidACosgrove:
@bp-kelley @greglandrum The CI failure is now down to a link failure in Windows_VS2022_x64_dll, in the test code at that.  I will keep plugging at it, but I don't think it's a blocker to review when you have the time.

rdkit/Chem/Draw/IPythonConsole.py
```diff
@@ -172,14 +172,12 @@ def drawMols3D(mols, view=None, confIds=None, drawAs=None, bgColor=None, size=No
...
   for i in range(len(mols)):
-    style = {
-          'colorscheme': colors[i % len(colors)]
-        }
-    if drawAs[i] == "sphere":
-      style["radius"] = 0.5
     view.setStyle({
       'model': i,
-    }, {drawAs[i]: style})
+    }, {drawAs[i]: {
```
d-b-w: is this related? If not, could you pull it out into a different PR? This one is... long

Code/GraphMol/GaussianShape/ShapeInput.cpp
```diff
@@ -166,52 +260,151 @@ ShapeInput &ShapeInput::operator=(const ShapeInput &other) {
   return *this;
 }
 
+void ShapeInput::merge(ShapeInput &other) {
```
d-b-w: is this slurping out the data from `other`? It looks like it may be. If so, I'd prefer the signature encode that:


```suggestion
void ShapeInput::merge(ShapeInput &&other) {
```

Code/GraphMol/GaussianShape/GaussianShape.cpp
```diff
@@ -85,7 +84,7 @@ std::array<double, 4> getInitialRotationPlain(
     int index, const ShapeInput &refShape, const ShapeInput &fitShape,
     const RDGeom::Point3D &refDisp, const ShapeOverlayOptions &overlayOpts,
     double &score) {
-  static const double sinpi_4 = std::sin(std::atan(1.0));
+  static const double sinpi_4 = std::sin(std::numbers::pi / 4.0);
```
d-b-w, line 87: constexpr in c++26!!

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -79,72 +108,217 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
...
+  mzparams.label = MolzipLabel::Isotope;
+  mzparams.alignCoordinates = false;
+
+  while (true) {
+    const std::int64_t nextLine = ++mostRecentLine;
+    if (nextLine >= lastLine) {
+      break;
+    }
+    auto &[smiles, name] = checkLines[nextLine];
+    auto mol = v2::SmilesParse::MolFromSmiles(smiles);
```
d-b-w, line 217: could be null - this is reading from a file

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceShapeSearcher.cpp
```diff
@@ -0,0 +1,1277 @@
...
+#include <GraphMol/GaussianShape/ShapeOverlayOptions.h>
+#include <GraphMol/GaussianShape/SingleConformerAlignment.h>
+#include <GraphMol/SynthonSpaceSearch/ProgressBar.h>
+#include <GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h>
+#include <GraphMol/SynthonSpaceSearch/SynthonSpaceShapeSearcher.h>
+#include <RDGeneral/ControlCHandler.h>
+#include <RDGeneral/RDThreads.h>
+#include <boost/dynamic_bitset/dynamic_bitset.hpp>
+
+std::mutex myMutex;
```
d-b-w: this should at least be `static`

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+    const std::string &smiles, unsigned int numConformers)>;
+
+// This the maximum number of connectors that we can deal with at the moment.
+// In reality, there may be fewer than this.  However, the key limit is in
+// The symbols used for the connectors in Enamine REAL etc.
+const std::vector<std::string> CONNECTOR_SYMBOLS{"[U]", "[Np]", "[Pu]", "[Am]"};
+constexpr unsigned int MAX_CONNECTOR_NUM{4};
+
+struct RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpaceSearchParams {
+  std::int64_t maxHits{1000};  // The maximum number of hits to return.
```
d-b-w, line 38: This is a lot of parameters!

It's good to give users options, but it's almost impossible to _remove_ params - we don't have good visibility into which parameters are and are not used.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+  unsigned int minHitHeavyAtoms{0};  // Minimum number of heavy atoms in a hit.
+  int maxHitHeavyAtoms{-1};          // Maximum number of heavy atoms in a hit.
+  // -1 means no maximum.
+  double minHitMolWt{0};  // Minimum molecular weight for a hit.
+  double maxHitMolWt{0};  // Maximum molecular weight for a hit.  0.0 means
+  // no maximum.
+  unsigned int minHitChiralAtoms{
+      0};                           // Minimum number of chiral atoms in a hit.
+  int maxHitChiralAtoms{-1};        // Maximum number of chiral atoms in a hit.
+                                    // -1 means no maximum.
```
d-b-w, line 94: these feel like candidates for removal

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+  unsigned int numConformers{100};  // When doing a shape search, the number of
+                                    // conformers to use for each molecule.
+  double confRMSThreshold{0.5};  // When doing a shape search, the RMS threshold
+                                 // to use when pruning conformers.  Passed
+                                 // directly to EmbedMultipleConfs.
+  GaussianShape::ShapeOverlayOptions
+      shapeOverlayOptions;  // Options to use when shape searching.
+  bool bestHit{false};  // If true, when doing a shape search it will return the
+                        // hit conformer with the best shape match to a query
+                        // conformer.  If false it will just return the first
```
d-b-w, line 104: what's the speed differential here?

Code/SimDivPickers/catch_tests.cpp
```diff
@@ -83,3 +84,33 @@ TEST_CASE(
...
+    double operator()(unsigned int i, unsigned int j) { return d_dists[i][j]; }
+    const std::vector<std::vector<double>> &d_dists;
+  };
+  RDPickers::LeaderPicker leaderPicker;
+  DistFunctor distFunctor(dists);
+
+  auto picks = leaderPicker.lazyPick(distFunctor, 10, 0, 1.8);
+  std::cout << "Picks : " << picks.size() << " :: ";
+  for (auto p : picks) {
+    std::cout << p << " ";
```
d-b-w: I don't see any assertions in this test.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
```diff
@@ -30,19 +31,22 @@
...
+#include <GraphMol/SynthonSpaceSearch/ProgressBar.h>
 #include <GraphMol/SmilesParse/SmilesParse.h>
 #include <RDGeneral/ControlCHandler.h>
 #include <RDGeneral/RDThreads.h>
 #include <RDGeneral/StreamOps.h>
+#include <sys/stat.h>
 
-namespace RDKit::SynthonSpaceSearch {
-
+namespace RDKit {
```
d-b-w, line 44: why does this go back to old-style nested namespaces?

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
```diff
@@ -133,68 +139,224 @@ SearchResults SynthonSpace::substructureSearch(
...
+  SynthonSpaceRascalSearcher ssrs(query, rascalOptions, params, this);
+  return ssrs.checkPossibleHits(startLine, finishLine);
+}
+
+namespace {
+std::unique_ptr<ROMol> addZeroOrderBondsBetweenFrags(
+    const ROMol &query, const std::vector<std::vector<int>> &frags) {
+  auto retMol = std::make_unique<RWMol>(query);
+  retMol->beginBatchEdit();
+  for (unsigned int i = 1; i < frags.size(); ++i) {
```
d-b-w: this feels like a double loop to find the a close point for each fragment.

* It's a bit hard to read
* I think it's order-dependent - differing orders of fragments could get different topologies, I'm not sure if that is matters
* the sanitize step feels out of place
* should this be a k-d tree of the whole fragment space?

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
```diff
@@ -133,68 +139,224 @@ SearchResults SynthonSpace::substructureSearch(
...
+}
+}  // namespace
+
+SearchResults SynthonSpace::shapeSearch(
+    const ROMol &query, const SynthonSpaceSearchParams &params) {
+  PRECONDITION(query.getNumAtoms() != 0, "Search query must contain atoms.");
+
+  // It there's more than 1 fragment in the query, join them together with
+  // a zero-order bond between the closest atoms in each.  It makes the
+  // fragmentation work, but otherwise doesn't affect the final outcome.
```
d-b-w, line 314: oh. Does it _matter_ where the bonds are? If not, then it might be nice to make them strictly in order - that way you wouldn't need to sanitize.

Alternatively - should we disallow fragmented inputs?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -79,72 +108,217 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
...
+    BOOST_LOG(rdWarningLog)
+        << "No possible hits file given, so no results generated." << std::endl;
+    return SearchResults();
+  }
+  auto checkLines =
+      readPossHitsLines(getParams().possibleHitsFile, startLine, finishLine);
+  if (checkLines.size() < finishLine) {
+    finishLine = checkLines.size();
+  }
+  std::cout << "Checking " << checkLines.size() << " lines from " << startLine
```
d-b-w: woah, I don't think I'd want to see this cout every time I ran the search

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -79,72 +108,217 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
...
+                  return lhs->getNumAtoms() < rhs->getNumAtoms();
+                }
+                return lsim > rsim;
+              });
+  }
+}
+}  // namespace
+
+SearchResults SynthonSpaceSearcher::checkPossibleHits(
+    std::uint64_t startLine, std::uint64_t finishLine) {
```
d-b-w, line 260: Is this a feature you were using while testing ? Is it possible that it could be a separate commit?

d-b-w (commented):
Thanks for this feature - it's super exciting. I'd like to work with you to get it in.

That being said... This is a pretty big PR! 

I think it would be easier to review (and faster to land!) if you split it up. Some ideas

* wait for #9279 or mark it as an upstream
* remove changes in extra files that aren't really part of the feature (e.g. IPythonConsole, maybe more?)
* Is it possible to split out the refactor of SynthonSpace from the addition of the new search type?

rdkit/Chem/Draw/IPythonConsole.py
DavidACosgrove: No, that was a mistake.  It would be nice to make the 3D display more flexible, but that was just a hack to get me going.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
DavidACosgrove: The OpenBind folk, who funded this, have fragmented inputs as a key requirement.  In FBDD, the really hard part is joining 2 fragments together so their relative positions are maintained in the joined molecule.  There are very few cases in the literature where this has been claimed as being done successfully - we tried to find some for validation purposes :-).  Here, you can put both fragments as 3D structures into the shape search and it will return whole molecules with bits very similar to the fragments in the right place.  The few tests we've run show that it looks very promising.  We think it could be very helpful for FBDD.  Obviously as a Brit "game-changing" isn't in my vocabulary, but...

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
DavidACosgrove: This function is helpful whilst testing but that wasn't the main aim.  I think it will make the synthon searching more feasible for many people.   It came about in part from feedback from people who have been using the current substructure and fingerprint searching extensively.  The first stage, getting the synthon combinations that might give hits, requires access to the whole synthon space so needs a reasonably large amount of memory and, for shape searching especially, a large number of threads all accessing the same memory is beneficial - a large, expensive machine.  My Mac mini runs the first stage of a shape search for a typical molecule against FreedomSpace 5.0 in about 4 hours on 13 threads so a larger machine would clearly help the impatient.

The second stage, taking that list of synthon combos and checking each one against the query is often more time-consuming and is also embarrassingly parallel.  This function is part of a workflow that allows you to dump the synthon combos into a file and process them as a second step.  Each line in the file just needs the synthons zipping into the product molecule and then the substructure, fingerprint or shape query running against them.  That can be done by spreading the file over a large number of small memory, low thread count machines with a much higher throughput and probably more cheaply.  Whereas the original all in one go workflow was restricted to maybe a couple of hundred threads (AWS tops out at ~300 IIRC),  this can be run on thousands of threads simultaneously.  

Another advantage of doing it this way is that, for the shape search, you can provide your own conformer generator to use instead of the RDKit one.  If you are running in Python and provide a Python function as the conformer generator, it captures the GIL when generating the conformers for potential hits, which is a sizeable part of the verification time, and the parallel performance collapses.  So for processing the synthon combos you can use this workflow to run multiple single-thread processes in parallel on the same or multiple nodes and get better throughput.

And finally, given the synthon combos file, you could process the hits with completely different code - Nvidia's nvMol for fingerprints or their GPU port of the conformation generator would be 2 obvious things to try.

Having said all that, it probably could be pulled out into a separate PR if that would make your life easier for reviewing.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
DavidACosgrove: I've made the code that reads the file more robust, so there should be no empty entries in checkLines.  The name might be a blank string.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceShapeSearcher.cpp
DavidACosgrove: It was worse than that.  It was only being used in the shape search, it should have been used in all 4 searches.  I have moved it to be a member variable of SynthonSpaceSearcher.

Code/SimDivPickers/catch_tests.cpp
DavidACosgrove: Sorry, that was just a convenient place to run an experiment.  Taken it out.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
DavidACosgrove: @greglandrum prefers it.  It must have been 1 I missed when taking them out in the first release.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
DavidACosgrove: The idea is to join each fragment to its neighbours by the shortest zero-order bond possible.  But yes, it is only going to do that for 2 fragments.  It will get very messy for more than 2, including introducing a ring I think.  I have only tested it for 2 fragments which is the expected use case.  I'll re-work it.

DavidACosgrove: Actually, after chatting with @greglandrum I think it's fair to restrict the query to only 2 fragments for now, which will make the code simpler.

greglandrum:
> Thanks for this feature - it's super exciting. I'd like to work with you to get it in.
> 
> That being said... This is a pretty big PR!

Yeah, it's huge, but most of the synthon-search PRs have been huge. The way that Dave and I have been handling these is to assume that the code in the synthon directory is under development and subject to change from release to release. We wanted to get something out there and in real users' hands so that we could (hopefully) better understand the use cases and usage patterns. The idea has always been that there will be large changes that come with new functionality. Given that, I've more or less only been reviewing the API to make sure it "feels" ok.
It's awesome that you're really digging into the code. :-)

> 
> I think it would be easier to review (and faster to land!) if you split it up. Some ideas
> 
> * wait for #9279 or mark it as an upstream

I think  you mean #9265? I wanted to give @bp-kelley a chance to make any further comments he might have on that one, but I'll merge it tomorrow at the latest.

> * remove changes in extra files that aren't really part of the feature (e.g. IPythonConsole, maybe more?)

👍 

> * Is it possible to split out the refactor of SynthonSpace from the addition of the new search type?

I bet that's going to be impossible. :-)
Hopefully this code stabilizes enough in the "near" future that we can do things like that

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
bp-kelley: We did a lot of this style back in the day, it worked quite well.  This implementation of shape really doesn't care about bonds, they only exist to get the conformations right.

bp-kelley:
@greglandrum I approved #9266 with some reservations, but I can talk to Dave at the UGM to see if we want to rethink the api a touch.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.cpp
DavidACosgrove: > We did a lot of this style back in the day, it worked quite well. This implementation of shape really doesn't care about bonds, they only exist to get the conformations right.

Yeah, the zero-order bonds are just so the splitting the query up for comparison with the synthons works correctly.  There are 2 different uses of the word "fragment" which is unhelpful.

DavidACosgrove:
> Thanks for this feature - it's super exciting. I'd like to work with you to get it in.
> 
> That being said... This is a pretty big PR!
> 
> I think it would be easier to review (and faster to land!) if you split it up. Some ideas
> 
> * wait for #9279 or mark it as an upstream
> * remove changes in extra files that aren't really part of the feature (e.g. IPythonConsole, maybe more?)
> * Is it possible to split out the refactor of SynthonSpace from the addition of the new search type?

Hi Dan,

Many thanks for looking at this.  Apologies for the size of it.  I was treating it more as v2.0 than an incremental change, so I'm afraid I was a bit undisciplined with changes to the existing search modes.  The re-factoring was largely done because the shape searching required it, and then if necessary back-porting to the other search classes.  Unpicking it will be very unpleasant if possible at all.  I think that anything outside the SynthonSpaceSearch directory should be covered by other PRs (the GaussianShape one being the most obvious) or is a mistake such as the IPythonConsole.py changes.  

I will resolve the merge conflicts tomorrow.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
DavidACosgrove: That's hard to say, but it is real.  Given N conformers per potential hit, it will on average not do N/2 shape comparisons for what turns out to be an actual hit.  If the potential hit doesn't produce any conformers that match the query, there's no saving.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
DavidACosgrove: They were already in the first version.  The ChemSpace people said they thought they would be useful.

greglandrum:
@DavidACosgrove if you can fix the merge conflicts, I will look at this this week

External/pubchem_shape/PubChemShape.hpp
greglandrum, line 1: I don't think these changes are connected to this PR?

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -398,6 +494,8 @@ class RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpace {
   // Just do the lookup, and return nullptr if not found.
   Synthon *getSynthonFromPool(const std::string &smiles) const;
 
+  void orderSynthonsForSearch();
```
greglandrum: Docs?

DavidACosgrove:
The failing test passes on both my Macs (M1 and M4).  Could it be an Intel thing?  If so, I have an old Intel macbook I can try and get this building on.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.h
```diff
@@ -84,12 +103,32 @@ class SynthonSpaceSearcher {
...
+  // sets are uniquified.
+  FragSetUniquifyMode d_fragSetUniquifyMode{FragSetUniquifyMode::BySmiles};
+
  private:
   std::unique_ptr<std::mt19937> d_randGen;
 
-  const ROMol &d_query;
-  const SynthonSpaceSearchParams &d_params;
-  SynthonSpace &d_space;
+  ROMol d_query;
```
greglandrum: It feels very strange for the query and search parameters to not be `const`.

When I change them back to be `const`, everything still compiles, so I think this isn't even necessary

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.h
```diff
@@ -84,12 +103,32 @@ class SynthonSpaceSearcher {
...
+  SynthonSpaceSearchParams d_params;
+  SynthonSpace *d_space;
+  std::unique_ptr<ROMol> d_bestHitFound;
+  double d_bestSimilarity{0.0};
+
+  std::mutex d_myMutex;
```
greglandrum: Maybe a better name for this that indicates what its for?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+  std::int64_t maxHits{1000};  // The maximum number of hits to return.
+                               // Use -1 for no maximum.
+  std::uint64_t maxNumFragSets{
+      100000};  // The maximum number of fragment sets the query can
+                // be broken into.  Big molecules will create huge
+                // numbers of fragment sets that may cause excessive
+                // memory use.  If the number of fragment sets hits this
+                // number, fragmentation stops and the search results
+                // will likely be incomplete.
+  std::int64_t toTryChunkSize{
```
greglandrum: Why isn't this a uint64_t?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+      2500000};              // For similarity searching, especially
+                             // fingerprint similarity, there can be a
+                             // very large number of possible hits to
+                             // screen which can use a lot of memory and
+                             // crash the program.  It will also be very
+                             // slow.  To alleviate the memory use, the
+                             // possible hits are processed in chunks.
+                             // This parameter sets the chunk size.
+  std::int64_t hitStart{0};  // Sequence number of hit to start from.  So that
```
greglandrum, line 56: Why isn't this a uint64_t?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+             // before building a product and doing a final check.  The
+             // similarityCutoff is reduced by this value for the approximate
+             // check.  A lower value will give faster run times at the
+             // risk of missing some hits.  The value you use should have a
+             // positive correlation with your FOMO.
+  unsigned int minHitHeavyAtoms{0};  // Minimum number of heavy atoms in a hit.
+  int maxHitHeavyAtoms{-1};          // Maximum number of heavy atoms in a hit.
+  // -1 means no maximum.
+  double minHitMolWt{0};  // Minimum molecular weight for a hit.
+  double maxHitMolWt{0};  // Maximum molecular weight for a hit.  0.0 means
```
greglandrum: To be consistent this could be negative

Code/GraphMol/SynthonSpaceSearch/shape_search_catch_tests.cpp
greglandrum, line 1: These tests take much too long to run: ~5 minutes on my laptop, >8min on the CI machines. Together with pySynthonSpaceSearch, this increases the runtime of the test on the CI machines by more than 50%.

The longer running tests should be moved to LONG_TEST_catch_tests.cpp. Please reduce the size of the tests here and in pySynthonSpaceSearch so that they still test as much of the API as possible, but take a lot less time to run. I realize these won't be good regression tests; we'll have to rely upon what's in LONG_TEST_catch_tests.cpp for that.

greglandrum (changes requested):
As usual, I just went through the header files on this one. There are a few comments on those, but the big thing to address is the runtime of the tests - see my comment on shape_search_catch_tests.cpp

greglandrum (changes requested):
As usual, I just went through the header files on this one. There are a few comments on those, but the big thing to address is the runtime of the tests - see my comment on

greglandrum:
> As usual, I just went through the header files on this one. There are a few comments on those, but the big thing to address is the runtime of the tests - see my comment on

Apologies that I somehow managed to submit the review 1.5 times. The previous review is the real one

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
DavidACosgrove: No need - it's never used.  I've taken it out.  The synthon search orders are initialised on reading.

greglandrum:
@DavidACosgrove since this is a big one, I'm having copilot do a review. This is just to see if it catches anything obvious.

Let's see what it spits out...

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -39,6 +40,8 @@
 #include <GraphMol/MolStandardize/Tautomer.h>
 #include <GraphMol/SynthonSpaceSearch/SynthonSet.h>
 #include <GraphMol/SynthonSpaceSearch/SearchResults.h>
+#include <GraphMol/SynthonSpaceSearch/SynthonSet.h>
+#include <GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h>
```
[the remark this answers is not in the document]
greglandrum: @DavidACosgrove I think this one should be addressed

greglandrum (changes requested):
one very minor request

DavidACosgrove:
> one very minor request

If that's removing the duplicate inclusion of SynthonSet.h, I think it's already done.

Code/GraphMol/SynthonSpaceSearch/LONG_TEST_catch_tests.cpp
```diff
@@ -401,4 +410,350 @@ TEST_CASE("Rascal Biggy") {
...
+      CHECK(std::ranges::find(expNames[i], mol->getProp<std::string>(
+                                               common_properties::_Name)) !=
+            expNames[i].end());
+      auto scores = GaussianShape::ScoreMolecule(*queryMol, *mol);
+      CHECK_THAT(mol->getProp<double>("Similarity"),
+                 Catch::Matchers::WithinAbs(scores[0], 0.001));
+      ++j;
+    }
+#if 0
+    // Leave this in for now, in case we need to check brute force search
```
d-b-w, line 514: thanks for these comments about the hidden code

Code/GraphMol/SynthonSpaceSearch/LONG_TEST_catch_tests.cpp
```diff
@@ -15,8 +15,10 @@
 
 #include <GraphMol/SubstructLibrary/SubstructLibrary.h>
 #include <GraphMol/FileParsers/MolSupplier.h>
+#include <GraphMol/FileParsers/MolWriters.h>
```
d-b-w, line 18: so... How slow are the long tests? Is there a situation in which they run?

It would be a bummer if they stopped passing - it can be tricky to figure out when a test stopped passing if it isn't run regularly.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -80,73 +108,226 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
...
     while (stepper.d_currState[0] != numSynthons[0]) {
       toTry.emplace_back(hitset.get(), stepper.d_currState);
-      if (toTry.size() == static_cast<size_t>(d_params.toTryChunkSize)) {
+      if (toTry.size() == d_params.toTryChunkSize) {
         std::vector<std::unique_ptr<ROMol>> partResults;
-        processToTrySet(toTry, endTime, partResults, numHitsFound);
-        hitCount += partResults.size();
+        processToTrySet(toTry, endTime, partResults, numHitsFound,
+                        numPossHitsFound);
+        numHitsFound += partResults.size();
```
d-b-w: This is already incremented in `processToTrySet()`.

I guess this means there is a missing test - Maybe if the maxHits is smaller than the number of total hits & toTryChunkSize is set even smaller?


```suggestion
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -80,73 +108,226 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
...
+      break;
+    }
+    auto &[smiles, name] = checkLines[nextLine];
+    auto mol = v2::SmilesParse::MolFromSmiles(smiles);
+    std::unique_ptr<ROMol> prod;
+    try {
+      prod = molzip(*mol, mzparams);
+      MolOps::sanitizeMol(*dynamic_cast<RWMol *>(prod.get()));
+    } catch (std::exception &e) {
+      std::cout << "Failed to zip " << MolToSmiles(*prod) << " for " << name
```
d-b-w: I think this message is wrong - if `molzip()` fails prod will be empty.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -80,73 +108,226 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
...
+  }
+  std::string smiles, name, wholeLine;
+  for (std::uint64_t i = 0; i < startLine; ++i) {
+    std::getline(ifs, wholeLine);
+  }
+  for (std::uint64_t i = startLine; i < finishLine; ++i) {
+    std::getline(ifs, wholeLine);
+    if (ifs.eof()) {
+      break;
+    }
```
d-b-w: If the file doesn't have a terminating newline, this ignores the last line.

A file with a valid last line (but no terminating newline) would be a good test. 


```suggestion
    if (!std::getline(ifs, wholeLine)) {
        break;
    }
```

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -341,68 +568,52 @@ bool SynthonSpaceSearcher::quickVerify(
...
       return false;
     }
   }
   return true;
 }
 
+void SynthonSpaceSearcher::updateBestHitSoFar(const ROMol &possBest,
+                                              const double sim) {
+  if (sim > d_bestSimilarity) {
+    std::unique_lock lock1{d_bestHitsMutex};
```
d-b-w, line 580: I think that _technically_ the lock needs to be before you check inequality. It's possible for another thread to reset d_bestSimilarity between these. If that's a performance hit, you could check inequality twice (once before and once after the mutex).

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.cpp
```diff
@@ -841,17 +905,25 @@ std::unique_ptr<ROMol> buildProduct(
...
-  MolOps::sanitizeMol(*dynamic_cast<RWMol *>(zipProdMol.get()));
+  std::unique_ptr<ROMol> zipProdMol;
+  try {
+    zipProdMol = molzip(*prodMol, mzparams);
+    MolOps::sanitizeMol(*dynamic_cast<RWMol *>(zipProdMol.get()));
+  } catch (std::exception &e) {
+    std::cout << "AWOOGA :: Failed to zip " << MolToSmiles(*prodMol)
+              << " because " << std::endl
+              << e.what() << std::endl;
+  }
```
d-b-w, line 922: Does every caller of this check for null? Should we retain the exception?

Code/GraphMol/SynthonSpaceSearch/SynthonSet.cpp
```diff
@@ -554,6 +611,187 @@ std::unique_ptr<ROMol> SynthonSet::buildProduct(
...
+    std::string msg("Error:: in reaction " + getId() +
+                    " :: building molecule from synthons :");
+    for (size_t i = 0; i < synthonNums.size(); ++i) {
+      std::string sep = i ? " and " : " ";
+      msg += sep + d_synthons[i][synthonNums[i]].first + " (" +
+             d_synthons[i][synthonNums[i]].second->getSmiles() + ")";
+    }
+    msg += "\n" + std::string(e.what()) + "\n";
+    BOOST_LOG(rdErrorLog) << msg;
+    throw(e);
```
d-b-w: prefer bare throw

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.cpp
```diff
@@ -903,4 +975,540 @@ unsigned int countChiralAtoms(ROMol &mol, unsigned int *numExcDummies) {
...
+        BOOST_LOG(rdWarningLog) << e.what() << std::endl;
+        BOOST_LOG(rdWarningLog)
+            << "No conformers generated for sample molecule "
+            << sampleMols[molNum]->d_mol->getProp<std::string>(
+                   common_properties::_Name)
+            << " : " << MolToSmiles(*sampleMols[molNum]->d_mol)
+            << " when generating conformers for synthon "
+            << sampleMols[molNum]->d_synthon->getSmiles()
+            << " with isomer : " << MolToSmiles(*isomer) << std::endl;
+        exit(1);
```
d-b-w: woah - do we really want this to terminate the process!!? I'd prefer if we just rethrow. Or not catch! it feels odd that we are catching an Invariant.

Code/GraphMol/SynthonSpaceSearch/SynthonSpace.h
```diff
@@ -448,11 +582,12 @@ class RDKIT_SYNTHONSPACESEARCH_EXPORT SynthonSpace {
...
  * @param cancelled whether it received a SIGINT
  * @param fpGen optional fingerprint generator
+ * @param shapeParams optional parameters for conformer generation
  */
 RDKIT_SYNTHONSPACESEARCH_EXPORT void convertTextToDBFile(
     const std::string &inFilename, const std::string &outFilename,
-    bool &cancelled,
-    const FingerprintGenerator<std::uint64_t> *fpGen = nullptr);
+    bool &cancelled, const FingerprintGenerator<std::uint64_t> *fpGen = nullptr,
+    ShapeBuildParams *shapeParams = nullptr);
```
d-b-w, line 590: I haven't found a test that hits this. Please make sure it is tested

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+                                    // about another 35 characters or so
+                                    // depending on the size of the job.  0
+                                    // means no bar.
+  unsigned int maxSynthonAtoms{0};  // If > 0, sets a maximum number of heavy
+                                    // atoms, excluding dummies, for a synthon
+                                    // to have a shape made.
+  unsigned int maxEmbedAttempts{10};  // Maximum attempts for an embedding.
+  unsigned int timeOut{60};  // Maximum time in seconds to spend on each synthon
+                             // when generating conformers.
+  std::string
```
d-b-w, line 199: I haven't found a test that uses this. Because it's intended for failure recovery, we definitely need a test

greglandrum:
> > one very minor request
> 
> If that's removing the duplicate inclusion of SynthonSet.h, I think it's already done.

Ah, right you are. The github web interface confused me.... apologies!

greglandrum (approved):
LGTM

d-b-w (changes requested):
This looks great!

I tried to pick through looking for local logical errors (with the help of my robot friends, codex and claude).

I also paid special attention to:

* catch blocks. I expect them to be rarely hit in prod or testing, so review is particularly important.
* Unit tests. of course!

Code/GraphMol/SynthonSpaceSearch/LONG_TEST_catch_tests.cpp
DavidACosgrove: They take about 5 minutes on my Mac and frequently much longer on the CI machines.  @greglandrum was very firm, correctly I think, about taking them out of the CI tests.  But it does raise the issue you mention.  As a separate PR I am going to change the installation and contribution docs to say that if you're developing you should run CMake with -DRDK_BUILD_LONG_RUNNING_TESTS=ON and make sure the tests are run.  And I guess we'll have to play it by ear for a while to see how it goes.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
DavidACosgrove: I've put the test in testSynthonSpaceSubstructureSearch as that runs quickest.  And yes, it was a bug.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.cpp
DavidACosgrove: Several callers weren't checking for null, but they do now.  I think leaving the exception in is useful because it gives an indication of why the zip failed.  It's only ever zipping synthons so failures should be very rare, in which case there should be next to no overhead and it will help track down the offending synthons in what is likely to be a very large file.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.cpp
DavidACosgrove: Changed to throw an exception with more info.  The ShapeInput c'tor can throw an Invar in rare circumstances, one such being the subject of a recent bug report in GaussianShape.  I forget what triggered it to make me put the catch in, but I assume something must have done.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
DavidACosgrove: Do you mean the timeOut or maxEmbedAttempts?  In this instance they are both passed into the conformation generator. If that times out or fails the embedding then no conformations are generated for the molecule containing a synthon (in details::generateIsomerConformers) in which case it tries another molecule with the same synthon or ultimately no shapes are generated for that synthon.  I'm not sure they need testing separately here.  But perhaps I have missed the point.

Code/GraphMol/SynthonSpaceSearch/LONG_TEST_catch_tests.cpp
d-b-w: That seems fair. I think that as long as they always _compile_ as part of CI, the worst disasters are mitigated

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
d-b-w: crap - I meant the `interimFile`! I sneakily missed it in my comment span

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
```diff
@@ -0,0 +1,235 @@
...
+      interimFile;  // Interim file to write SynthonSpace to.  In the event of
+                    // a failure, a restart from this file will be possible.
```
d-b-w, line 201: This file - I think we should make sure that we can read and write it

d-b-w (approved):
Thanks for this work! I don't wanna hold it up anymore than I already have done.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearchHelpers.h
DavidACosgrove: Good point.  I will put a test in.

DavidACosgrove: The test is in the python wrapper as it was a lot easier to kill the shape generation there and show that the restart worked than in the C++ catch tests.

Issue #9069, opened 2026-01-29
In assignStereochemistry legacy mode, "force" flag does not work


JLVarjo:
In the legacy mode, assignStereochemistry does not respect the "force" flag since the check for `flagPossibleStereoCenters` argument in `legacyStereoPerception` has been commented out (Chirality.cpp line 2285). This has been done in PR #8875, but I wonder if it is really necessary?

greglandrum:
I'm not able to reproduce this. Here's a demonstration with the current release:
```
In [91]: m = Chem.MolFromSmiles('C[C@H]([13CH3])Cl')

In [92]: m.GetAtomWithIdx(1).GetChiralTag()
Out[92]: rdkit.Chem.rdchem.ChiralType.CHI_TETRAHEDRAL_CCW

In [93]: m.GetAtomWithIdx(2).SetIsotope(0)   #< remove the isotope so that it's no longer a potential chiral center

In [94]: Chem.AssignStereochemistry(m,force=False,cleanIt=True)

In [95]: m.GetAtomWithIdx(1).GetChiralTag()  #< we didn't force the assignment, so nothing happened
Out[95]: rdkit.Chem.rdchem.ChiralType.CHI_TETRAHEDRAL_CCW

In [96]: Chem.AssignStereochemistry(m,force=True,cleanIt=True)

In [97]: m.GetAtomWithIdx(1).GetChiralTag()  #< we did force the assignment, so chirality was removed
Out[97]: rdkit.Chem.rdchem.ChiralType.CHI_UNSPECIFIED


```
@JLVarjo am I misunderstanding something?

JLVarjo:
No, it was me who confused the `force` flag with the `flagPossibleStereoCenters` flag, and the actual issue was that the behavior of the latter was changed, at least what comes to the CIP rank calculation. This was discussed in more detail in #9286 and no further actions are required.

PR #9286, opened 2026-05-15
Fix force flag on legacyStereoPerception #9069 

JLVarjo:
This is a fix for https://github.com/rdkit/rdkit/issues/9069. My guess is that the removal of the assignStereoChemistry force flag has been used for debug purposes in development, and afterwards forgotten to re-enable. But if I'm wrong, please forget this PR.

### DIFF
--- a/Code/GraphMol/Chirality.cpp
+++ b/Code/GraphMol/Chirality.cpp
@@ -2283,7 +2283,7 @@ void legacyStereoPerception(ROMol &mol, bool cleanIt,
   // are neither stereocenters nor bonds that we need to consider.
   // The exception to this is when flagPossibleStereoCenters is
   // true; then we always need to do the work
-  bool hasStereoAtoms = false;  // flagPossibleStereoCenters;
+  bool hasStereoAtoms = flagPossibleStereoCenters;
   bool hasPotentialStereoAtoms = false;
   for (auto atom : mol.atoms()) {
     if (cleanIt) {


greglandrum:
@JLVarjo I don't see how #9069 manifests. Please add a test to this PR (I'd suggest putting it in `catch_chirality.cpp`) that demonstrates the problem that this PR is fixing (and shows that it's fixed)

JLVarjo:
Sorry for being unclear! My issue is that I need to get CIPranks for a molecule regardless of if there is stereo centers present in the molecule or not. This does not happen if `flagPossibleStereoCenters` argument does not set `hasStereoAtoms=true` on line 2286, as it used to do earlier.

I'll try to craft a test for this later...

greglandrum:
> Sorry for being unclear! My issue is that I need to get CIPranks for a molecule regardless of if there is stereo centers present in the molecule or not. This does not happen if `flagPossibleStereoCenters` argument does not set `hasStereoAtoms=true` on line 2286, as it used to do earlier.
> 

In that case, I think you're probably better off just calling `Chem.ComputeAtomCIPRanks()`; that will return the ranks of each atom.

By the way: if you just want a canonical ranking of the atoms that takes symmetry into account (which is what the CIP ranks are), you can call `Chem.CanonicalRankAtoms(m,breakTies=False)`.

JLVarjo:
Thanks Greg, I solved this with a wrapper in my own code which first calls `AssignStereoChemistry` and subsequently `ComputeAtomCIPRanks` in the case CIP ranks have not been computed.

`CanonicalRankAtoms` does not work for me since I need different canonical ranks for prochiral atoms, for which I have to make some CIP rank manipulations (long story, 1H NMR reasons).

So, on my behalf, feel free to close this PR, but I would consider cleaning the comments above line 2286, which still state that `flagPossibleStereoCenters=true` forces stereo assignment.

PR #9289, opened 2026-05-18
MacroMol general macro atom backend pls SCSRWRiter

tadhurst-cdd:
#### Reference Issue
Backend for MACROmol to suppport Schrodinger's MonomerMol and CDDs SCSR functionality
Full atomistic generator from MACROMols

Newly revised backend macro atom system that supports MonomerMol (HELM like) and SCSR mols

Also contains the full atomistic generator from a MACROMol

#### Any other comments?
Every MACROMol has a reference to one or more MACROMol libraries, which can be internally or externally stored.   This allows the MonomerMol use of an external Global Library or an external Custom library shared among several macro molecules.   It also allows the internal library needed by SCSR molecules.

The templates are stored in a vector, which has O(1) lookup for an index.  There is an unordered_map (d_key_to_index) which allows an O(log n) lookup of the index given a Template name and class.  This also allows multiple names to be assigned to a Template, and this allows single letter codes and 3 letter (PDB) names to be accessible for HELM-type macroatoms.

The names of the connection points to each monomer (Template) can be any string - MACROMol does not care what they are called - this allows SCSR's naming convention like Al r, Cx, HB1 etc. and HELMs R1, R1, Rx convention.


greglandrum:
@tadhurst-cdd it looks like the CI failures here are real, can you please take a look?

tadhurst-cdd:
I have addressed the CI issues, and all is clean now.  Please take a look
when you have a chance.
https://github.com/rdkit/rdkit/pull/9289

Thanks,
tad

External/ChemDraw/test.cpp
```diff
@@ -1303,6 +1303,9 @@ TEST_CASE("Round TRIP") {
       if (entry.path().string().find("ChemDraw") != std::string::npos) {
         continue;  // Skip ChemDraw directory
       }
+      if (entry.path().string().find("macromols") != std::string::npos) {
+        continue;  // Skip SCSR-macromols directory directory
```
bp-kelley: Typo  in comment "directory directory".

Out of curiosity, what was failing here?

tadhurst-cdd: I am glad you ask these questions.  The Chemdraw test looks for mol files anywhere is the RDKit code tree and tries to round-trip them.   I thought the SCSR files in the macromols were screwing that up, and filtering them out caused the test to not fail.    But it was really failing on some other test's temporary files in that directory, So rather than filter out the directory, I changed the other test to NOT put the temp files in the test_data directory - there are now created in a temp directory.  And the change to ChemDraw/test.cpp were not needed and were removed.

DavidACosgrove:
@tadhurst-cdd If the only thing holding the PR up is the GaussianShape tests and you're confident it's not part of your changes (which seems highly likely) then I suggest you put something really slack in to get it through.  There's a new PR in with significant changes that might make it moot in any case.  Obviously @greglandrum would have to agree.

Code/GraphMol/FileParsers/MacroMolUtils.h
```diff
@@ -0,0 +1,77 @@
...
+  bool outputSgroups = true;
+  MACROTemplateNames macroTemplateNames = MACROTemplateNames::All;
+};
+
+enum class MACROUseTemplateName {
+  UseFirstName,   //<!Use the first name in the template
+                  // def (For AA, the 3 letter code
+  UseSecondName,  //<!use the second name in the tempate def (
+                  // For AA, the 1 letter code)
+};
```
rachelnwalker, line 51: I would be in favor of always writing SCSR as if `MACROTemplateNames::All` was chosen, and removing the `MACROUseTemplateName` option altogether. I think we should be strict that on a given `MacroMol`, the identifier is a single `className` + `templateName` pair. A 3 letter code could map to that as an 'alias', but I worry that the option to write only some of these to sgroups could lead to strange roundtrip behavior. 

Would you be opposed to storing `templateName` and `3lettercode` separately on the template instead of in a shared vector?

rachelnwalker (commented):
I've talked through a few of these points with @tadhurst-cdd offline already, reiterating here in case anyone else wants to weigh in (also thank you to Tad for taking the time to understand some of our similar work along the way). A few points:

1. I vote we call this new data structure `MacroMol` instead of `MACROMol`, and standardize on that over `MonomerMol`. From our offline chat I think Tad's on board, the all-caps was just to match RDKit's `ROMol`/`RWMol`, but those are actual acronyms (Read-Only / Read-Write) whereas "MACRO" isn't.

2. Template library and MacroMol class - I made some suggestions against Tad's branch here https://github.com/cdd/rdkit/pull/23

3. Attachment point naming

>The names of the connection points to each monomer (Template) can be any string - MACROMol does not care what they are called - this allows SCSR's naming convention like Al r, Cx, HB1 etc. and HELMs R1, R1, Rx convention

I'd like to push for MacroMol using a single canonical attachment-point scheme internally, regardless of input format. MacroMol is an in-memory data model, so readers (SCSR, HELM, etc.) should translate *into* the canonical form and writers translate *out* of it. As-is the attachment semantics of a MacroMol would depend on which reader produced it; and without a canonical form, every writer and algorithm has to understand every reader's convention. It also means the same molecule built from different sources (SCSR, HELM, a sketcher, FASTA, a PDB file) won't necessarily come out equal, so anything cross-format like a subsequence search wouldn't be reliable.

Overall I think this is looking good and I'm excited to have roundtrip SCSR support in rdkit!

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,194 @@
...
+  return this->addAtom(atom, false, true);
+}
+
+void RDKit::MACROMol::addMacroBond(unsigned int fromAtomIdx,
+                                   unsigned int toAtomIdx,
+                                   Bond::BondType bondType,
+                                   std::string fromConnectionPoint,
+                                   std::string toConnectionPoint) {
+  auto bondIdx = this->addBond(fromAtomIdx, toAtomIdx, bondType);
+  auto bond = this->getBondWithIdx(bondIdx);
```
rachelnwalker: Noticed this while pairing with @emilyrrhodes on some follow up work - I believe `addBond` returns total number of bonds and not the idx, so I think this needs to be 

```
auto bondIdx = this->addBond(fromAtomIdx, toAtomIdx, bondType) - 1;
```

emilyrrhodes:
Are we only allowing specific classes for SCSR files?
https://github.com/cdd/rdkit/blob/ScsrWriter/Code/GraphMol/FileParsers/MolSGroupParsing.cpp#L1181
```
} else if (label == "CLASS" &&
                 !SubstanceGroupChecks::isValidClass(strValue)) {
        std::ostringstream errout;
        errout << "Unsupported SGroup template class '" << strValue
               << "' on line " << line;
```

"According to the Biovia doc, the CLASS attribute must have a value that is one of: AA, dAA, DNA, RNA, SUGAR, BASE, PHOSPHATE, LINKER, CHEM, LGRP, MODAA, MODdAA, MODDNA, MODRNA, XLINKAA, XLINKdAA, XLINKDNA, XLINKRNA. For an SCSR mol block, (and for any SGROUP), RDKit requires that the CLASS attribute be one of these values." Source: https://www.rdkit.org/docs/RDKit_Book.html

Does this mean we will do mapping between helm "classes" and SCSR "classes" for example? Will there be one parent class and helm class and SCSR class are just a view on this? Something else?

tadhurst-cdd:
> Are we only allowing specific classes for SCSR files? https://github.com/cdd/rdkit/blob/ScsrWriter/Code/GraphMol/FileParsers/MolSGroupParsing.cpp#L1181
> 
> ```
> } else if (label == "CLASS" &&
>                  !SubstanceGroupChecks::isValidClass(strValue)) {
>         std::ostringstream errout;
>         errout << "Unsupported SGroup template class '" << strValue
>                << "' on line " << line;
>         throw FileParseException(errout.str());
> ```
> 
> "According to the Biovia doc, the CLASS attribute must have a value that is one of: AA, dAA, DNA, RNA, SUGAR, BASE, PHOSPHATE, LINKER, CHEM, LGRP, MODAA, MODdAA, MODDNA, MODRNA, XLINKAA, XLINKdAA, XLINKDNA, XLINKRNA. For an SCSR mol block, (and for any SGROUP), RDKit requires that the CLASS attribute be one of these values." Source: https://www.rdkit.org/docs/RDKit_Book.html
> 
> Does this mean we will do mapping between helm "classes" and SCSR "classes" for example? Will there be one parent class and helm class and SCSR class are just a view on this? Something else?

The allowed classes are currently the SCSR allowed classes.   This controlls what classes of templates can be read in from a mol file. It is actuallly only applied to the SAP class definition,  so the class of LGRP would never be expected as a template class (just an SAP class).

MacroMol does not care what you call the class - the current MonomerMol tests that I have only load the class "PEPTIDE" and that works fine.

In the document https://pubs.acs.org/doi/10.1021/ci3001925#sec4_1. they list only PEPTIDE, RNA, and CHEM. They mention a possible furture SAC for saccarides.

Are they more?  Is there a definitive definition of HELM?

Do we expect to read in template defs from mol files for HELM?  

If we are to convert from HEML to SCSR and back, we need a conversion between the lists.  

*********
The same is true for the connection poinnts.  We talked in our recent group call about using a coded integer for the connection types.  MonomerMol is using R1, R2, R3, ATOM and pair, while SCSR is using Al Br, Cx, Hb, Hb1, Hb2, Hb3.
We really cannot use an integer fo the connection type - these are stored as SAPs in the internal template.

I REALLY DO NOT WANT MacoMol to know anything about HELM or SCSR, and allow it to keep unrestricted connection point names and template names.  The SCSR and MonomerMol interfaces should implement the expected template names and attachment point names.  

There is not a direct translation from the HELM list to the SCSR list - the SCSR list does not contain an attachment name of CHEM.  HELM "pair" maps to SCSR "Hb", but there is not noted mapping for Hb1, Hb2, Hb3.

The easient thing for me to do would be to change SCSR utils to use the internal attachment point names from the list:
R1,R2,R3,ATOM, pair, pair1, pair2, pair3.   

I think it is going to be up to the MonomerMol interface to translate the long list of Template types from SCSR to one of PEPTIDE, RNA, CHEM.

Thoughts?

Code/GraphMol/MacroMol.cpp
tadhurst-cdd: You are correct that this is an error.  But in the most recent version that code has been removed.

PR #9290, opened 2026-05-19
add checked atom and bond iterators

greglandrum:
This change introduces atom and bond iterators that will throw if the molecule structure is modified while the iterators are active.

These are maybe less useful in C++, but will be quite helpful in the nanobind wrappers.
The current wrappers are already doing this with the atom and bond iterators that we implemented in Python:
https://github.com/rdkit/rdkit/blob/master/rdkit/Chem/__init__.py#L60

### DIFF
--- a/Code/GraphMol/ROMol.h
+++ b/Code/GraphMol/ROMol.h
@@ -114,7 +114,7 @@ RDKIT_GRAPHMOL_EXPORT extern const int ci_ATOM_HOLDER;
 
 template <class Graph, class Vertex,
           class Iterator = typename Graph::vertex_iterator,
-          bool Checked = false>
+          bool CheckedAtoms = false, bool CheckedBonds = false>
 struct CXXAtomIterator {
   Graph *graph;
   Iterator vstart, vend;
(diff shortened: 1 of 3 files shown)


bp-kelley:
The failure looks annoying:

OSX:
```
Unable to open file /Users/runner/work/1/s/Code/JavaWrappers/gmwrapper/src/org/RDKit/SWIGTYPE_p_RDKit__CXXBondIteratorT_boost__adjacency_listT_boost__vecS_boost__vecS_boost__undirectedS_RDKit__Atom_p_RDKit__Bond_p_t_const_RDKit__Bond_pconst_boost__adjacency_listT_boost__vecS_boost__vecS_boost__undirectedS_RDKit__Atom_p_RDKit__Bond_p_t__out_edge_iterator_true_t.java: File name too long
```

Might need a rename here...

bp-kelley:
Same error on all 3 platforms...

ricrogz:
@greglandrum I created a PR against your branch that ~fixes~ avoids wrapping the CXX iterators.

ricrogz (approved):
LGTM

bp-kelley (approved):
LGTM

PR #9291, opened 2026-05-21
Adds MolToCDXMLBlock to FileParsers

bp-kelley:
Need to test C# swig wrapper as well before committing.

Adds 
```
RDKIT_FILEPARSERS_EXPORT std::string MolToCDXMLBlock(
    const RWMol &mol,
    CDXMLFormat format = CDXMLFormat::CDXML);
```

Throws a FileParserException if ChemDraw support is not built.

the Python api returns a str (for cdxml) or bytes (for cdx)

For SWIG, this is split into

```
MolToCDX
MolToCDXML
```

Because of differing return types.

### DIFF
--- a/Code/GraphMol/Wrap/rdmolfiles.cpp
+++ b/Code/GraphMol/Wrap/rdmolfiles.cpp
@@ -797,6 +797,21 @@ python::tuple MolsFromCDXML(python::object cdxml, bool sanitize,
   }
   return python::tuple(res);
 }
+
+python::object MolToCDXMLBlockHelper(const RDKit::ROMol &mol,
+				     RDKit::v2::CDXMLParser::CDXMLFormat format) {
+  auto block = RDKit::v2::CDXMLParser::MolToCDXMLBlock(mol, format);
+  // if CDXML return string
+  // if CDX return byteszo
+
+  if(format == RDKit::v2::CDXMLParser::CDXMLFormat::CDX) {
+    PyObject* py_bytes = PyBytes_FromStringAndSize(block.data(), block.size());
+    return python::object(python::handle<>(py_bytes));
+  } else {
+    return python::object(python::handle<>(PyUnicode_FromString(block.c_str())));
+  }
+}
+ 
 namespace {
 python::object translateMetadataToList(
     const std::vector<std::pair<std::string, std::string>> &metadata) {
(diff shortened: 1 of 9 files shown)


bp-kelley:
@greglandrum this is ready to look at when you have time.

Code/JavaWrappers/ROMol.i
```diff
@@ -116,6 +115,17 @@
   $result = JCALL1(NewByteArray, jenv, $1.size());
   JCALL4(SetByteArrayRegion, jenv, $result, 0, $1.size(), (const jbyte*)$1.c_str());
 }
+%typemap(jni) std::string RDKit::ROMol::MolToCDX "jbyteArray"
```
greglandrum, line 118: What you can do with SWIG is amazing, but the code to make it happen is just eye burning.

Code/GraphMol/Wrap/rdmolfiles.cpp
```diff
@@ -2746,6 +2761,34 @@ BOOST_PYTHON_MODULE(rdmolfiles) {
...
+
+        - mol: the molecule to write
+
+        - format: CDXMLFormat [default CDXML]
+
+      RETURNS:
+        a tuple of parsed Mol objects.)DOC";
+  
+  python::def("MolToCDXMLBlock", MolToCDXMLBlockHelper,
+              (python::arg("mol"), python::arg("format")), docString.c_str());
```
greglandrum: Given the name of the function, it seems like it would make sense to default the format to CDXML like it is in the C++ call:
```suggestion
              (python::arg("mol"), python::arg("format")=RDKit::v2::CDXMLParser::CDXMLFormat::CDXML), docString.c_str());
```

greglandrum (changes requested):
Small suggested change in the python wrapper

Code/JavaWrappers/ROMol.i
bp-kelley: And I always forget how to do it when I go back...

Code/GraphMol/Wrap/rdmolfiles.cpp
bp-kelley: Even the docstring says that should be the default.  I'm also missing python tests, so I'll add those as well.

bp-kelley:
@greglandrum default format updated and tests added.

Code/GraphMol/ChemReactions/Enumerate/EnumerateSynthons.cpp
greglandrum, line 1: Was this accidentally included? It isn't part of this PR

bp-kelley: It is not.

bp-kelley: Error caused by the user side of the keyboard ;). It's been resolved.

greglandrum (approved):
LGTM

greglandrum:
Will merge after the CI builds go through

PR #9292, opened 2026-05-21
All-In-One coordinate refinement for ETKDG

nmaeder:
This mainly implements the all-in-one conformer refinement of the conformer generator as introduced at the [last RDKit UGM](https://github.com/rdkit/UGM_2025/blob/main/presentations/Wed/Afternoon/02_niels-maeder.pdf).

This introduces an adapted coordinate refinement procedure for conformer generation.

It improves coherence with the bounds matrix and reduces occurences of unphysical geometries in embedded conformers.
There is a general speedup for KDG, for ETKDG speeds can vary on a per-molecule basis. Since the refinement process differs from the sequential approach, results for different molecules can differ.

The new code is not enabled by default, but can be activated by setting `useLegacyImplementation` to false in the `EmbedParameters` object.

Additionally, this includes some cleanup in `DistGeomUtils.cpp` and `Embedder.cpp.`

This also implements the use of designated initializers for the EmbedParameters object.

#### Any other comments?

Wherever possible i used the catch `GENERATE` functionality to let unit test run with the legacy as well as the new implementation.

I altered the `testDistGeomHelpers` test called `testProvideBoundsMatrix`. It was trying to set inconsistent bounds, this is now fixed.


ReleaseNotes.md
```diff
@@ -77,6 +77,10 @@ wszqkzqk, Diogo, dehaenw
 - `RWMol::addBond()` no longer removes RingInfo from the molecule, but it does
   clear the property cache of the begin and end atoms. This does not affect the
   general rule that molecules should be sanitized after adding/removing bonds.
+- The `DistGeomHelpers::EmbedParameters` struct no longer has a constructor that
```
greglandrum: this block should be in the 2026.09.1 section of the release notes

Code/GraphMol/ForceFieldHelpers/CrystalFF/TorsionPreferences.h
```diff
@@ -32,6 +32,29 @@ struct RDKIT_FORCEFIELDHELPERS_EXPORT ExpTorsionAngle {
   unsigned int idx[4];
 };
 
+namespace FC {
```
greglandrum: Please pick a more descriptive name than "FC" for this

Code/GraphMol/ForceFieldHelpers/CrystalFF/PlanarityContribs.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace ForceFields {
+namespace CrystalFF {
+
+struct RDKIT_FORCEFIELDHELPERS_EXPORT PlanarityContribsParams {
+  std::size_t idx1{0};        //!< index of atom1 in the ForceField's positions
+  std::size_t idx2{0};        //!< index of atom2 in the ForceField's positions
+  std::size_t idx3{0};        //!< index of atom3 in the ForceField's positions
+  std::size_t idx4{0};        //!< index of atom4 in the ForceField's positions
+  double forceConstant{1.0};  //!< force constant
```
greglandrum: It doesn't really hurt, but for future reference: this isn't the most useful type of comment and should just be skipped.

Code/GraphMol/ForceFieldHelpers/CrystalFF/PlanarityContribs.cpp
```diff
@@ -0,0 +1,161 @@
...
+                             pos[dimension * contrib.idx4 + 1],
+                             pos[dimension * contrib.idx4 + 2]);
+
+    totalEnergy +=
+        calcOopBendEnergy(calcOopChi(p1, p2, p3, p4), contrib.forceConstant);
+  }
+  return totalEnergy;
+}
+
+void PlanarityContribs::getGrad(double *pos, double *grad) const {
```
greglandrum, line 84: Rather than checking your math: have you tested this with a finite difference check like what's in https://github.com/rdkit/rdkit/pull/9249?

Code/GraphMol/DistGeomHelpers/Wrap/rdDistGeom.cpp
```diff
@@ -487,8 +510,50 @@ BOOST_PYTHON_MODULE(rdDistGeom) {
...
       .value("EXCEEDED_TIMEOUT",
              RDKit::DGeomHelpers::EmbedFailureCauses::EXCEEDED_TIMEOUT)
+      .value("MINIMIZATION",
+             RDKit::DGeomHelpers::EmbedFailureCauses::MINIMIZATION)
+      .value("KTERM_VIOLATION",
+             RDKit::DGeomHelpers::EmbedFailureCauses::KTERM_VIOLATION)
+      .value("CLASH", RDKit::DGeomHelpers::EmbedFailureCauses::CLASH)
       .export_values();
 
+  python::def(
```
greglandrum, line 520: very useful

Code/GraphMol/DistGeomHelpers/Wrap/rdDistGeom.cpp
```diff
@@ -646,6 +714,9 @@ BOOST_PYTHON_MODULE(rdDistGeom) {
   python::def("KDG", RDKit::getKDG,
               "Returns an EmbedParameters object for the KDG method.",
               python::return_value_policy<python::manage_new_object>());
+  python::def("DG", RDKit::getDG,
+              "REturns an EmbedParameters object for plain distance geometry.",
```
greglandrum:
```suggestion
              "Returns an EmbedParameters object for plain distance geometry.",
```

Code/DistGeom/Wrap/DistGeom.cpp
```diff
@@ -1,6 +1,6 @@
 // $Id$
```
greglandrum: Please remove this leftover from the SVN days

Code/DistGeom/DistGeomUtils.cpp
```diff
@@ -604,4 +617,174 @@ ForceFields::ForceField *construct3DImproperForceField(
...
+      std::make_unique<ForceFields::DistanceConstraintContribs>(ff);
+  for (std::size_t i = 1; i < numAtoms; ++i) {
+    for (std::size_t j = 0; j < i; ++j) {
+      if (fixedPts != nullptr && (*fixedPts)[i] && (*fixedPts)[j]) {
+        continue;
+      }
+      const double l = mmat.getLowerBound(i, j);
+      const double u = mmat.getUpperBound(i, j);
+      const auto dist = distMat[i * numAtoms + j];
+      const bool is1213 = (dist == 1.0 || dist == 2.0);
```
greglandrum: testing floating point equality is risky. It's probably ok here because it's likely that the distance is explicitly set to 1.0 or 2.0, but it would be safer to do a test with a tolerance like:

```suggestion
      const bool is1213 = (fabs(dist-1.0)<1e-4 || fabs(dist-2.0)<1e-4);
```

Code/DistGeom/DistGeomUtils.cpp
```diff
@@ -604,4 +617,174 @@ ForceFields::ForceField *construct3DImproperForceField(
...
+}
+
+RDKIT_DISTGEOMETRY_EXPORT ForceFields::ForceField *constructAllInOneForceField(
+    const BoundsMatrix &mmat, RDGeom::PointPtrVect &positions,
+    const ForceFields::CrystalFF::CrystalFFDetails &etkdgDetails,
+    const VECT_CHIRALSET *csets,
+    const std::map<std::pair<unsigned int, unsigned int>, double> *extraWeights,
+    const boost::dynamic_bitset<> *fixedPts) {
+  const std::size_t N = mmat.numRows();
+  CHECK_INVARIANT(N == positions.size(), "");
```
greglandrum: This and the next one are PRECONDITIONS:
```suggestion
  PRECONDITION(N == positions.size(), "");
```

Code/DistGeom/DistGeomUtils.cpp
```diff
@@ -604,4 +617,174 @@ ForceFields::ForceField *construct3DImproperForceField(
...
+  CHECK_INVARIANT(etkdgDetails.expTorsionAtoms.size() ==
```
greglandrum:
```suggestion
  PRECONDITION(etkdgDetails.expTorsionAtoms.size() ==
```

Code/DistGeom/DistGeomUtils.cpp
```diff
@@ -604,4 +617,174 @@ ForceFields::ForceField *construct3DImproperForceField(
...
+    const ForceFields::CrystalFF::CrystalFFDetails &etkdgDetails,
+    const VECT_CHIRALSET *csets,
+    const std::map<std::pair<unsigned int, unsigned int>, double> &CPCI,
+    const std::map<std::pair<unsigned int, unsigned int>, double> *extraWeights,
+    const boost::dynamic_bitset<> *fixedPts) {
+  auto *field = constructAllInOneForceField(mmat, positions, etkdgDetails,
+                                            csets, extraWeights, fixedPts);
+
+  bool is1_4 = false;
+  // double dielConst = 1.0;
```
greglandrum: Please remove

Code/ForceField/AngleConstraints.cpp
```diff
@@ -102,17 +106,46 @@ double AngleConstraintContribs::getEnergy(double *pos) const {
   return accum;
 }
 
+double AngleConstraintContribs::getEnergy(
```
greglandrum, line 116: It's too bad that here (and in a couple of other files in this PR) we have to duplicate code.
It might be worth figuring out if we can remove that duplication

Code/GraphMol/DistGeomHelpers/Wrap/testDistGeom.py
```diff
@@ -159,6 +170,51 @@ def test1Small(self):
     d6 = computeDist(conf.GetAtomPosition(2), conf.GetAtomPosition(3))
     self.assertTrue(feq(d6, 1.31, 0.01))
 
+  def test1Small(self):
+    # single double and tripple atoms cases should not fail
```
greglandrum, line 174:
```suggestion
    # single double and triple atoms cases should not fail
```

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -97,216 +100,56 @@ std::mutex &GetFailMutex() {
...
-                                true,     // useSmallRingTorsions
-                                false,    // useMacrocycleTorsions
-                                false,    // useMacrocycle14config
-                                0,        // timeout
-                                nullptr,  // CPCI
-                                nullptr   // callback
-);
+const EmbedParameters DG{};
+
+const EmbedParameters KDG{.useExpTorsionAnglePrefs = false,
```
greglandrum, line 105: This is soooo much more compact and readable. Nice!

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -477,6 +322,73 @@ bool _boundsFulfilled(const std::vector<int> &atoms,
...
+                const double threshold2 = 0.8) {
+  for (std::size_t i = 1; i < positions->size(); ++i) {
+    for (std::size_t j = 0; j < i; ++j) {
+      double dist2 = 0.0;
+      const RDGeom::PointND diff =
+          (*static_cast<RDGeom::PointND *>((*positions)[j]) -
+           *static_cast<RDGeom::PointND *>((*positions)[i]));
+      for (std::size_t d = 0; d < 3; ++d) {
+        dist2 += diff[d] * diff[d];
+      }
```
greglandrum, line 337: Any particular reason to not just do:
```suggestion
      dist2 = diff.lengthSq();
```

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -477,6 +322,73 @@ bool _boundsFulfilled(const std::vector<int> &atoms,
...
+      const double lb = boundsMat->getLowerBound(i, j);
+      if (dist2 < threshold2 and dist2 < lb * lb) {
+        return false;
+      }
+    }
```
greglandrum, line 345: Since this code is going to be called a lot, I think it's worth speeding it up a bit:

```suggestion
      if(dist2 >= threshold2) {
        continue;
      }
      const double lb = boundsMat->getLowerBound(i, j);
      if (dist2 < lb * lb) {
        return false;
      }
    }
```

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -477,6 +322,73 @@ bool _boundsFulfilled(const std::vector<int> &atoms,
...
+      std::cerr << e << ", ";
+    }
+    std::cerr << std::endl;
+#endif
+    return std::ranges::any_of(energies,
+                               [threshold](double e) { return e > threshold; });
+  };
+
+  for (const auto &contrib : field->contribs()) {
+    if (auto c = dynamic_cast<const ForceFields::AngleConstraintContribs *>(
```
greglandrum, line 381: dynamic_casts make me a bit nervous, but I don't really see a reasonable alternative here.

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -477,6 +322,73 @@ bool _boundsFulfilled(const std::vector<int> &atoms,
...
+        return false;
+      }
+    }
+  }
+  const double planarityTolerance = 0.7;
+  const std::size_t nCenters =
+      std::ranges::count_if(eargs.etkdgDetails->angles,
+                            [](const auto &angle) { return angle[3]; }) +
+      eargs.etkdgDetails->improperAtoms.size();
+  return totalEnergy <= (nCenters * planarityTolerance);
```
greglandrum: This seems like a pretty cheap test, can it be moved above the loop over contribs?

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -746,11 +706,43 @@ bool doubleBondGeometryChecks(const RDGeom::PointPtrVect &positions,
       }
     }
   }
+  if (doSP2Centers and !eargs.etkdgDetails->improperAtoms.empty()) {
+    // this is the arrangement:
+    //     a0
+    //       \       [ilb]
```
greglandrum, line 717: It took me *much* too long to figure out what "ilb" means.
:-)

greglandrum (changes requested):
Some small suggestions, but this mostly looks great.

Code/GraphMol/DistGeomHelpers/Embedder.cpp
nmaeder: Yes, if we have two points that are 4D, we still want to check their distance in 3D.
`diff.lengthSq()` would take the dimension the points have.

Code/DistGeom/DistGeomUtils.cpp
nmaeder: I can see your point, will change that accordingly.
Maybe for the future, we could add a `MolOps::getDistanceMat` that returns `* unsigned int` instead of `* double` since we ask for topological distance here, no?

Code/DistGeom/DistGeomUtils.cpp
nmaeder: they are `CHECK_INVARIANTS` in most other cases in said file to from earlier days, thats why i went for that. Will also quickly adapt the old code, as I am at it

Code/ForceField/AngleConstraints.cpp
nmaeder: We cannot remove it completely, as the singatures of the functions are different, but i can certainly extract the math into a helper function in an anonymous namespace above

Code/GraphMol/DistGeomHelpers/Embedder.cpp
nmaeder: we could add this element wise check to the base contrib, but that might be too much work for it only being used here at the moment.

Code/GraphMol/DistGeomHelpers/Embedder.cpp
nmaeder: Was the same for me, when i saw this for the first time in your code ;-)

Code/GraphMol/ForceFieldHelpers/CrystalFF/PlanarityContribs.h
nmaeder: will just remove them again, no need to add unecessary stuff to the code base

Code/GraphMol/ForceFieldHelpers/CrystalFF/PlanarityContribs.cpp
nmaeder: This is just a simplified version of the MMFF improper torsion implementation, so i was relying on it being fine.
Added a finite difference check to the unit tests to be safe

Code/GraphMol/DistGeomHelpers/Embedder.cpp
nmaeder: yes it can

nmaeder:
Adressed your comments in the latest commit

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -366,7 +369,14 @@ bool _checkKTerms(RDGeom::Point3DPtrVect &positions,
     return std::ranges::any_of(energies,
                                [threshold](double e) { return e > threshold; });
   };
-
+  const double planarityTolerance = 0.7;
```
greglandrum: Should be constexpr, no?

greglandrum (changes requested):
one tiny change, then we're done! :-)

Code/GraphMol/DistGeomHelpers/Embedder.cpp
nmaeder: absolutely!

nmaeder: Done

greglandrum (approved):
LGTM

greglandrum:

ricrogz:
The changes in this PR caused `testDistGeomHelpers` to almost duplicate its run time: on my computer it went from taking about 110 s  to approximately 210. This can be seen in CI too: in commit a31fcbf1abb323d56d2638c0e39d631fcb318aee in master, the test ran in 157 seconds, while when this PR was merged (c867fb121af1acbad0c4d48f0b94a1bf0e9d2a7f), it took 309 s. 

The test was already a slow one, often dominating the run time when running tests in parallel with `ctest -j8`. Now, I see `testDistGeomHelpers` running for an extra couple of minutes after all other tests have finished.

Should we split this test file in two (or three?) separate test files so that they can benefit more from running tests in parallel ?

nmaeder:
Yes, also noticed the significant increase - takes longer to run everything twice.
Happy to split it into different files, once we come up with a smart strategy to split them.
Can think of splitting them into LEGACY / AIO or more 'topic' specific, e.g. into I/O, BoundsMatrix related things and the actual embedding.
What do you think @greglandrum?

ricrogz:
> Yes, also noticed the significant increase - takes longer to run everything twice. Happy to split it into different files, once we come up with a smart strategy to split them. Can think of splitting them into LEGACY / AIO or more 'topic' specific, e.g. into I/O, BoundsMatrix related things and the actual embedding. What do you think @greglandrum?

I don't know how feasible it is to meaningfully split the test file by topic, or whether such a split  would be useful (imagine we do, and we end up with packing all tests except the ones about one specific topic into `testDistGeomHelpers #1` and the rest into `testDistGeomHelpers #2`, and  #then 1 takes 5 seconds and #2 takes 300s.

I'd rather do a "practical" split, splitting the tests so that each #1 and #2 take about half of the current run time (or into #1, #2 and #3, each taking 1/3 of the time)  and without caring too much about the topics.

nmaeder:
Am working on a solution I discussed this morning with @greglandrum, will open a PR as soon as it's ready, happy to get feedback there!

PR #9295, opened 2026-05-22
Feature BulkSimilarity: bulk Tanimoto matrix with runtime AVX-512 dispatch

mooreneural:
Adds `BulkSimilarity::tanimotoMatrix` (C++) and `rdkit.DataStructs.BulkTanimotoMatrix`
(Python) for computing an M×N Tanimoto matrix between two sets of
`ExplicitBitVect` fingerprints.

The kernel:

- **dispatches scalar vs AVX-512 at runtime.** The AVX-512 `VPOPCNTQ`
  (`_mm512_popcnt_epi64`) translation unit is only built when the host
  compiler accepts the flags, and selection happens at first call via
  CPUID + XGETBV (checking AVX512F + `AVX512_VPOPCNTDQ` and OS state),
  so prebuilt wheels (conda-forge, manylinux) run everywhere and still
  take the fast path on hardware that has it (Intel Ice Lake+, AMD Zen 4+).
- **is work-aware multi-threaded.** It parallelises across probe rows only
  once the matrix is large enough to amortize thread-creation cost, and
  gives each worker a coarse contiguous block of rows; small matrices run
  single-threaded.
- **cache-blocks the targets** into L2-resident tiles reused across each
  worker's probe rows, so wide matrices stay compute-bound instead of
  becoming memory-bandwidth bound.

This is intended to land before a follow-up PR that adds a CUDA backend; the
new module's API leaves room for that without breaking changes.

## What's in the box

- `Code/DataStructs/BulkSimilarity/` - new always-built library
  - `BulkSimilarity.h` - `tanimotoMatrix`, `packFingerprints`, `wordsForBits`,
    `activeKernel`
  - `BulkSimilarityCpu.cpp` - scalar kernel, runtime dispatcher, CPUID/XGETBV
    check, work-aware threading, target cache-blocking
  - `BulkSimilarityCpuAvx512.cpp` - AVX-512 VPOPCNTQ kernel (built only when
    compiler accepts the flags; uses `_mm512_maskz_loadu_epi64` for the tail)
  - `catch_tests.cpp` - 10 cases, including a 320-bit case that exercises the
    AVX-512 tail mask and a 130×4096 case that exercises the multi-threaded path
- `Code/DataStructs/BulkSimilarity/Wrap/` - `cBulkSimilarity` Python extension
  returning a 2-D `np.float64` NumPy array; imported defensively from
  `rdkit/DataStructs/__init__.py`, with numpy loaded lazily on first call
- `Code/Bench/similarity_matrix.cpp` - Catch2 benchmarks under
  `[similarity-matrix]`: naive nested `TanimotoSimilarity` vs
  `BulkSimilarity::tanimotoMatrix`, on a small single-threaded (64×1024) and a
  large multi-threaded (256×4096) perturbed-Morgan fingerprint matrix; also
  prints the active popcount kernel

## Performance

See the [benchmark comment](https://github.com/rdkit/rdkit/pull/9295#issuecomment-4748208040)
for numbers. In short, on AMD EPYC 9V45 (Zen 4, `avx512vpopcntdq`):
2.40× over the naive loop on a single-threaded 64×1024 matrix (isolating the
SIMD effect) and 2.68× on a multi-threaded 256×4096 matrix. The speedup tracks
AVX-512 availability across hosts (~1.6× on AVX-512-throttling Intel server
parts, ~1.04× on non-AVX-512 scalar hardware).

## Behavioural guarantees

- No new dependencies, no new languages, no new build options enabled by
  default. The AVX-512 path is auto-detected at configure time via
  `check_cxx_source_compiles`.
- The default scalar path uses the same `__builtin_popcountll` / `__popcnt64`
  intrinsics already in `BitOps.cpp`, so results are bit-identical to the
  existing `TanimotoSimilarity` reference (verified in tests).
- All fingerprints in a single call must share the same bit length and that
  length must be a positive multiple of 64. Mismatched / null inputs raise.

## Test plan

- [ ] `cmake --build . --target testBulkSimilarity && ctest -R testBulkSimilarity`
- [ ] `cmake --build . --target cBulkSimilarity && ctest -R pyBulkSimilarity`
- [ ] `cmake --build . --target bench && ./Code/Bench/bench "[similarity-matrix]"`
- [ ] Verify on a non-AVX-512 host that `BulkSimilarityActiveKernel()`
      reports `"scalar"`, and on Ice Lake / Zen 4+ that it reports
      `"avx512vpopcntdq"`
- [ ] Existing test suites (especially `testDataStructs`, `pyBV`) still pass

## Follow-ups (separate PRs)

- CUDA `tanimotoMatrixCuda` backend (prototype branch already exists)
- AVX2 Harley–Seal fallback for the Skylake → Tiger Lake gap
- `SparseBitVect` and `SparseIntVect` (count) variants


mooreneural:
## Benchmarks

Added a `[similarity-matrix]` group in `Code/Bench/similarity_matrix.cpp`
(Catch2, 30 samples, Morgan-r2 / 2048-bit fingerprints) comparing a naive
nested `TanimotoSimilarity` loop against `BulkSimilarity::tanimotoMatrix`,
built on a GitHub-hosted `ubuntu-24.04` runner.

**AMD EPYC 9V45 (Zen 4); active kernel `avx512vpopcntdq`:**

| matrix | naive nested `TanimotoSimilarity` | `BulkSimilarity::tanimotoMatrix` | speedup |
|---|---|---|---|
| 64 Ã 1024 (single-threaded â€” isolates SIMD) | 891 Âµs | 371 Âµs | **2.40Ã** |
| 256 Ã 4096 (multi-threaded) | 15.24 ms | 5.69 ms | **2.68Ã** |

The kernel reports the popcount path it actually selected (CPUID-based) via
`BulkSimilarity::activeKernel()`. The magnitude tracks AVX-512 availability
across runners (same benchmark, all `ubuntu-24.04`):

| host | active kernel | 64Ã 1024 speedup |
|---|---|---|
| AMD EPYC 9V45 (Zen 4) | `avx512vpopcntdq` | 2.4Ã |
| Intel Xeon Platinum 8370C (Ice Lake) | `avx512vpopcntdq` | 1.6Ã (AVX-512 frequency throttling) |
| AMD EPYC 7763 (Zen 3) | `scalar` | 1.04Ã (no AVX-512) |

Notes:

- The multi-threaded large case (2.68Ã) exceeds the single-threaded
  SIMD-only case (2.40Ã) because the targets are cache-blocked into
  L2-resident tiles. Without that, wide matrices become memory-bandwidth
  bound and threading does not help.
- On hardware without AVX-512 the scalar path is only marginally faster
  than the naive loop for small matrices, with threading still helping
  larger ones.

Reproduce: build the `bench` target and run
`./Code/Bench/bench '[similarity-matrix]'`.

mooreneural:
Hi @greglandrum thanks again for merging [#9298](https://github.com/rdkit/rdkit/pull/9298)!

Whenever you get a chance (no rush), this one's ready for a look. Quick
heads-up: I force-pushed the branch a little while ago, that was only to
re-run CI after an unrelated `Ubuntu_x64` failure (the runner couldn't find
`GL/gl.h` while configuring Qt5 for `MolDraw2D/Qt`); no code changed in that
push, and all checks are green now.

It's a feature-sized change (a new `BulkSimilarity` module + public API), so
There are also benchmark numbers in a comment
above.

Thanks!

mooreneural:
That was right: `cBulkSimilarity` was only registered in the Boost.Python tree, so a nanobind-only build would have dropped `BulkTanimotoMatrix` / `BulkSimilarityActiveKernel` from `rdkit.DataStructs`. Every other module ships both `Wrap/` and `nbWrap/`, so this was the odd one out.

Added `Code/DataStructs/BulkSimilarity/nbWrap/` with a nanobind implementation of the same two entry points, gated on `RDK_BUILD_NANOBIND_WRAPPERS` alongside the existing Boost.Python gate. The matrix is returned as a 2-D NumPy array with buffer ownership handed over via `nb::capsule`, following `Code/GraphMol/nbWrap/MolOps.cpp`. Per the convention in `Code/DataStructs/nbWrap/`, the nanobind build reuses the existing `Wrap/testBulkSimilarity.py` rather than duplicating it.

I also dropped the `try/except ImportError` around the import in `rdkit/DataStructs/__init__.py`. Now that the extension is built under both backends it should always be present, and swallowing the error would only hide a broken build.

Also worth noting for the AVX-512 part: #9216 landed the same shape for ARM SVE (compile-time availability gate, separate translation unit, runtime CPU feature check), so the CPUID/XGETBV dispatch here is consistent with what's already in the tree.

PR #9297, opened 2026-05-22
add ability to block atoms/bonds from participating in tautomer zones

greglandrum:
Atoms and/or bonds can be excluded from consideration as being part of the tautomeric zones by setting the property `_MolHashExcludeFromTautomerism` on them.

This allows very fine-grained customization of what is considered tautomerizable and what not.

### DIFF
--- a/Code/GraphMol/MolHash/nmmolhash.h
+++ b/Code/GraphMol/MolHash/nmmolhash.h
@@ -22,6 +22,8 @@
 namespace RDKit {
 class RWMol;
 namespace MolHash {
+constexpr const char *excludeFromTautomerismProp =
+    "_MolHashExcludeFromTautomerism";
 enum class HashFunction {
   AnonymousGraph = 1,
   ElementGraph = 2,
--- a/Code/GraphMol/MolHash/nmmolhash.h
+++ b/Code/GraphMol/MolHash/nmmolhash.h
@@ -22,6 +22,9 @@
 namespace RDKit {
 class RWMol;
 namespace MolHash {
+// atoms or bonds with this property set (value not important) will not be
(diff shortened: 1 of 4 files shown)


Code/GraphMol/MolHash/nmmolhash.h
```diff
@@ -22,6 +22,8 @@
 namespace RDKit {
 class RWMol;
 namespace MolHash {
+constexpr const char *excludeFromTautomerismProp =
```
bp-kelley, line 28: Perhaps add documentation here:

```
//! Atoms with the _MolHasExcludeFromTautomerism property set (with any value)
//!  will be preserved during tautomer hashing
```

bp-kelley (changes requested):

bp-kelley:
Neat, is this just part of MolHash?  I might be nice (another pr) to have the same capability for normal tautomer expansion.  I.e. keep this particular section but change the rest.

I gave a stab at documenting the header file, but this probably needs to be in the Python wrapper/RDKit Book as well.

Code/GraphMol/MolHash/nmmolhash.h
greglandrum: > Perhaps add documentation here:
> 
> ```
> //! Atoms with the _MolHasExcludeFromTautomerism property set (with any value)
> //!  will be preserved during tautomer hashing
> ```

Good point. I will add something similar to this and expose the constant to Python as well (unfortunately without a docstring since variables can't have docstrings), but I don't think we have any decent way of documenting this short of writing a new section in the RDKit Book explaining the molhash functionality. That's not a terrible idea, but is real work and won't be visible in IDEs

greglandrum:
@bp-kelley you ok with this updated version?

PR #9298, opened 2026-05-22
Fix BFGS gradient-convergence denominator for negative energies

mooreneural:
In `BFGSOpt::minimize`, the gradient-convergence denominator
`std::max(funcVal * gradScale, 1.0)` clamps to `1.0` whenever the current
energy is negative, which over-tightens the gradient test by a factor of
`|funcVal * gradScale|`. Force fields with stabilizing electrostatic or
dispersion terms (MMFF94, UFF with charges) routinely produce negative
energies mid-minimization, so this affects realistic workloads — extra BFGS
iterations, or occasionally returning the "too many iterations" status when
convergence would otherwise have been reached.

The fix uses `fabs(funcVal)`, matching the magnitude-as-denominator pattern
used three lines below (`std::max(fabs(pos[i]), 1.0)`).

## Changes

- `Code/Numerics/Optimizer/BFGSOpt.h`: one-line change plus a comment
  explaining why the absolute value is required.
- `Code/Numerics/Optimizer/testOptimizer.cpp`: new
  `testBFGSOptimizationNegativeEnergy` case that minimizes a 2D quadratic
  whose value is always negative along the convergence path
  (`f(x,y) = (x-3)^2 + (y+1)^2 - 100`) and verifies the optimizer reaches
  the analytic minimum.

## Test plan

- [ ] `ctest -R testOptimizer` passes (existing + new case)
- [ ] No behavioural change for the existing `circ_0_0` / `func2` cases
      (those have non-negative energies on the convergence path)

### DIFF
--- a/Code/Numerics/Optimizer/BFGSOpt.h
+++ b/Code/Numerics/Optimizer/BFGSOpt.h
@@ -252,7 +252,12 @@ int minimize(unsigned int dim, double *pos, double gradTol,
 
     // is the gradient converged?
     test = 0.0;
-    double term = std::max(funcVal * gradScale, 1.0);
+    // Use |funcVal| so that negative energies (which arise routinely
+    // mid-minimization in force fields containing stabilizing
+    // electrostatic or dispersion terms) do not drive
+    // funcVal * gradScale below zero and clamp the denominator to 1.0,
+    // which would artificially tighten the gradient convergence test.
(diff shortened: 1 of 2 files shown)


greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @mooreneural !

PR #9300, opened 2026-05-23
Github9280

### LINKED ISSUE #9280: DrawText::setFontScale Atom Font Scaling Issue
**Description**

When `DrawText::base_font_size_` is increased above its default value of 0.6 to make atom labels larger, the `DrawText::max_font_size_` cap is applied prematurely.

Rather than clamping `DrawText::fontSize()` when it exceeds the `DrawText::max_font_size_` (such as the default 40 px), the condition fires early at `max_font_size_ × DEFAULT_FONT_SCALE / base_font_size_` px instead of `max_font_size_` px - partially or fully cancelling out the intended size increase.

This is as a result of these lines in `DrawText::setFontScale` in `Code/GraphMol/MolDraw2D/DrawText.cpp`:

```
if (max_font_size_ > 0.0 &&
    nfs * (baseFontSize() / DEFAULT_FONT_SCALE) > max_font_size_) {
  font_scale_ = max_font_size_ / baseFontSize();
  return false;
}
```

The result is that atom labels are snapped to the maximum size before they should be, making them disproportionately large relative to the bond lengths.

Since bonds scale correctly with the molecule geometry but atom labels don't, the structure looks visually unbalanced - heteroatom labels (N, O, NH etc.) appear too large for the molecule they're attached to.

The `DEFAULT_FONT_SCALE` used in`setFontScale` appears to have been introduced as part of this change:

`f93016a77 — "Refactor mol draw2 d (#4948)" (Feb 2022)`

**Steps to reproduce**

A Python script like this will generate two images (see screenshots) where the cap is exceeded prematurely with a `base_font_size_` of 1.0. The image generated with a value of 0.9 for comparison:

```
from rdkit import Chem

SMILES = "CC(C)Oc1ccc(N2CCc3nccc(C(=O)Nc4ccccn4)c3C2)nc1"

mol = Chem.MolFromSmiles(SMILES)
mol = rdMolDraw2D.PrepareMolForDrawing(mol)

for bfs in [0.9, 1.0]:                                                                                                                                                                                                                           
    drawer = rdMolDraw2D.MolDraw2DSVG(358, 290)
    drawer.drawOptions().baseFontSize = bfs
    drawer.DrawMolecule(mol)
    drawer.FinishDrawing()
    filename = f"output_bfs_{str(bfs).replace('.', '_')}.svg"
    with open(filename, "w") as f:
        f.write(drawer.GetDrawingText())
    print(f"baseFontSize={bfs}  fontSize={drawer.FontSize():.2f}px  -> {filename}")
```
this should output:
```
baseFontSize=0.9  fontSize=22.20px  -> output_bfs_0_9.svg
baseFontSize=1.0  fontSize=40.00px  -> output_bfs_1_0.svg
```

**Expected behavior**

The cap shouldn't be exceeded with a value of 1.0 i.e. the function should behave as below:

```
if (max_font_size_ > 0.0 &&
    nfs > max_font_size_) {
  font_scale_ = max_font_size_ / baseFontSize();
  return false;
}
```

and the `fontSize` from the script should be 24.69 px for a `baseFontSize` of 1.0, this would be derived from:

`scale_ × baseFontSize = 24.69 × 1.0`

where `scale_ = min(322.2 / 13.05, 261.0 / 9.67) = 24.69` (canvas inner dimensions divided by molecular bounding box extent)

The expected atom sizing result is shown in the screenshots.

**Screenshots**

Reference value of 0.9 `base_font_size_`
<img width="358" height="290" alt="Image" src="https://github.com/user-attachments/assets/23ad313c-8f44-460b-9fdf-03ad47a48a93" />

What a value of 1.0 `base_font_size_` currently looks like:
<img width="358" height="290" alt="Image" src="https://github.com/user-attachments/assets/71cd9cd5-6ece-4aef-8122-c489a99c17bf" />

What a value of 1.0 `base_font_size_` should look like:
<img width="358" height="290" alt="Image" src="https://github.com/user-attachments/assets/3d405f36-a149-46bf-b1dc-2c67209ceb53" />

**Configuration:**

 - RDKit version: 2026.3.1
 - OS: macOS 15.4
 - Python version: 3.14
 - Are you using conda? No
 - How did you install the RDKit? pip (via uv)

**Additional context**

As a workaround if no capping is required i.e. atom labels should always scale with the rest of the image then an excessively large `max_font_size_` can be set e.g. 999px. This then means that the atom sizing is never adjusted prematurely.

A similar issue will exist for `min_font_size_` scaling from inspecting the code.

This report was prepared with assistance from Claude (Anthropic).

DavidACosgrove:
#### Reference Issue
Fixes #9280

Implements the fix as suggested in the issue, and also the min_font_size_ equivalent.

#### Any other comments?
Some extra formatting has happened in catch_tests.cpp.
I've scanned the outputs for the other drawings against the current master branch and nothing else seems to be broken, at least by this PR.

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawText.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawText.cpp
@@ -78,13 +78,11 @@ bool DrawText::setFontScale(double new_scale, bool ignoreLimits) {
   }
   font_scale_ = new_scale;
   double nfs = fontSize();
-  if (max_font_size_ > 0.0 &&
-      nfs * (baseFontSize() / DEFAULT_FONT_SCALE) > max_font_size_) {
+  if (max_font_size_ > 0.0 && nfs > max_font_size_) {
     font_scale_ = max_font_size_ / baseFontSize();
     return false;
   }
-  if (min_font_size_ > 0.0 &&
-      nfs * (baseFontSize() / DEFAULT_FONT_SCALE) < min_font_size_) {
+  if (min_font_size_ > 0.0 && nfs < min_font_size_) {
     font_scale_ = min_font_size_ / baseFontSize();
     return false;
   }
--- a/Code/GraphMol/MolDraw2D/catch_tests.cpp
+++ b/Code/GraphMol/MolDraw2D/catch_tests.cpp
@@ -4868,8 +4868,9 @@ TEST_CASE("legend position Top Left Right and vertical text", "[drawing]") {
       y = std::stod(match[2].str());
       return true;
     }
-    std::regex pathRgx("class='legend' d='M (-?[0-9]+\\.?[0-9]*) "
-                       "(-?[0-9]+\\.?[0-9]*)");
+    std::regex pathRgx(
+        "class='legend' d='M (-?[0-9]+\\.?[0-9]*) "
+        "(-?[0-9]+\\.?[0-9]*)");
     if (std::regex_search(text, match, pathRgx) && match.size() == 3) {
       x = std::stod(match[1].str());
       y = std::stod(match[2].str());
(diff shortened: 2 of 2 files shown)


greglandrum (approved):
LGTM

PR #9303, opened 2026-05-25
Refactor BreakTies control flow to use explicit termination tracking (#9284)

rezabagher:
#### Reference Issue
Related to #9284

Fixes an infinite loop in `Canon::BreakTies` triggered by highly symmetric
disconnected structures (e.g. Ammonium iron(III) sulfate dodecahydrate).

**Root cause**

For molecules with identical disconnected fragments and many isolated atoms,
`RefinePartitions` never converged, causing the `i -= 1` backtracking to
fire every iteration and the outer loop to never advance.

**Fix**

Replaced `i -= 1` with a `boost::dynamic_bitset<> finishedClass` that marks
symmetry classes as done once resolved to singletons, guaranteeing the outer
loop terminates in exactly `nAtoms` steps.

#### Any other comments?

Regression test added to `catch_canon.cpp` using the exact CTAB from the
bug report, covering termination, determinism, and cross-instance invariance.

- [x] New unit tests added to `catch_canon.cpp`

### DIFF
--- a/Code/GraphMol/catch_canon.cpp
+++ b/Code/GraphMol/catch_canon.cpp
@@ -1767,3 +1767,106 @@ TEST_CASE("Canonicalization issues watch (see GitHub Issue #8775)") {
 
   checkSmilesRoundtrip(smiles, shouldMatch);
 }
+
+TEST_CASE(
+    "Issue 9284: Infinite loop in BreakTies for disconnected/symmetric components",
+    "[canon][bug]") {
+  // Exact CTfile from the bug report - Ammonium iron(III) sulfate dodecahydrate
+  // Two identical SO4 groups + 13 isolated O atoms + isolated N + Fe
+  // caused BreakTies() to loop forever due to oscillating i -= 1
+  constexpr const char *ctab = R"ctab(
+     RDKit          2D
+
+ 24  8  0  0  0  0  0  0  0  0999 V2000
+    0.0000    0.0000    0.0000 N   0  0  0  0  0  0  0  0  0  0  0  0
+    1.0000    0.0000    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
+    0.0000    1.0000    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
+    2.0000    0.0000    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
+    0.0000    2.0000    0.0000 O   0  0  0  0  0  0  0  0  0  0  0  0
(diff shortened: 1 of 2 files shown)


greglandrum:
@rezabagher Thanks for the contribution.

I haven't looked at your proposed code changes yet, but I did try running the catch test you added using master and everything works fine. I'm also able to generate canonical ranks and canonical SMILES for the CTAB you provide for testing with the 2026.03.1 release. 

Were you able to actually reproduce an infinite loop with this test file? If so, on which operating system and using which compiler?

rezabagher:
Thanks for checking @greglandrum.
To be transparent: I wrote the test first, saw it fail, then wrote the code changes — but the test had a typo that I only caught afterward, and I didn’t re-run it against master before submitting. That was my mistake.

The previous implementation used a manual loop index decrement (`i -= 1`) to revisit partitions after refinement shifted their positions. The existing comment (`“not sure if this works each time”`) suggested that logic was already somewhat fragile, which also required tracking `oldPart` to detect whether a shift had occurred.

The proposed change replaces that mechanism with a `boost::dynamic_bitset<> finishedClass` that tracks resolved symmetry classes directly. As a side effect, this removes the need for `oldPart`and simplifies the degree-zero guard. 

That said, you know this codebase far better than I do. If, after reviewing the changes, you feel they are unnecessary or too risky, I’m happy to close the PR.

rezabagher:
Also updated the issue reference from Fixes to Related to — I didn't want the issue to auto-close on merge.

greglandrum:
@rezabagher I'm sure that it's possible that there is a bug in the canonicalization code, but I'm reluctant to make changes to it without having a test case that actually demonstrates the bug, i.e. an example which fails (or ends up in an infinite loop) on master but passes with your fix applied.

rezabagher:
Completely understandable, especially since canonicalization is a sensitive part of the codebase. I'll try to find a molecule that reliably reproduces the issue on master over the next few days. If I can't, I'll close this PR rather than leave unnecessary noise in the queue.

rezabagher:
I ran RDKit master against several symmetric and disconnected structures, including C60, cubane, benzene dimer, naphthalene, stereocenter cases, multiple disconnected fragment sets, and neutral/ionic ferrocene variants. I also ran a dedicated Catch2 stress test that subjects BreakTies to extreme permutation pressure using 200 identical disconnected ethane fragments over 500 randomized rounds.

In all cases, canonicalization completed without issues on master. I am closing this PR due to the lack of a reproducible failure on the current master branch. I actually learned some important things here. 
Thanks Greg

PR #9305, opened 2026-05-27
Improve synthon substructure search performance ~20%

d-b-w:

Improves synthon substructure search performance by almost 20%, especially for high-hitrate molecules. There are three basic optimizations here:

* Use a map instead of a vector in sortAndUniquifyToTry for lookups
* Use boost::unordered_flat_set instead of std::unordered_set for faster lookups, insertions, and creation
* Key sets on a hash instead of storing millions of strings.

I searched a group of high-hitrate molecules (more than 2million hits), on the 140B-product Freedom space. The overall improvement was something like 20%. Some standouts:

* Benzene:       6.92s → 5.64s  (−19%),
* Ciprofloxacin: 6.82s → 6.09s  (−11%)
* Aspirin:       4.51s → 3.42s  (−24%)

#### Any other comments?

I have a few other small optimizations coming. I think they are independent, so I'll put up separate reviews

Co-authored by claude Sonnet 4.6 with heavy driving by me.

### DIFF
--- a/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.h
+++ b/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.h
@@ -147,6 +147,11 @@ RDKIT_SYNTHONSPACESEARCH_EXPORT std::string buildProductName(
 RDKIT_SYNTHONSPACESEARCH_EXPORT std::string buildProductName(
     const RDKit::SynthonSpaceSearch::SynthonSpaceHitSet *hitset,
     const std::vector<size_t> &fragNums);
+// Hash of the product identity — same byte sequence as buildProductName but
+// without allocating the concatenated string. Use as a dedup key.
+RDKIT_SYNTHONSPACESEARCH_EXPORT std::size_t buildProductHash(
+    const RDKit::SynthonSpaceSearch::SynthonSpaceHitSet *hitset,
+    const std::vector<size_t> &fragNums);
 // Zip the fragments together to make a molecule.  Assumes the connection
 // points are marking by isotope numbers on dummy atoms.
 RDKIT_SYNTHONSPACESEARCH_EXPORT std::unique_ptr<ROMol> buildProduct(
--- a/Code/GraphMol/SynthonSpaceSearch/SearchResults.h
+++ b/Code/GraphMol/SynthonSpaceSearch/SearchResults.h
@@ -12,6 +12,7 @@
 #define RDKIT_SYNTHONSPACE_SEARCHRESULTS_H
 
 #include <functional>
+#include <boost/unordered/unordered_flat_set.hpp>
 #include <RDGeneral/export.h>
 #include <GraphMol/ROMol.h>
 
@@ -77,7 +78,7 @@ class RDKIT_SYNTHONSPACESEARCH_EXPORT SearchResults {
   std::vector<std::unique_ptr<ROMol>> d_hitMolecules;
   // Only used when merging in another set, so will be
   // filled in then if needed, empty otherwise.
-  std::unordered_set<std::string> d_molNames;
+  boost::unordered_flat_set<std::string> d_molNames;
 
   std::uint64_t d_maxNumResults;
   bool d_timedOut{false};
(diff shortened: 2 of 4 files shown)


Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.h
```diff
@@ -147,6 +147,11 @@ RDKIT_SYNTHONSPACESEARCH_EXPORT std::string buildProductName(
 RDKIT_SYNTHONSPACESEARCH_EXPORT std::string buildProductName(
     const RDKit::SynthonSpaceSearch::SynthonSpaceHitSet *hitset,
     const std::vector<size_t> &fragNums);
+// Hash of the product identity — same byte sequence as buildProductName but
+// without allocating the concatenated string. Use as a dedup key.
+RDKIT_SYNTHONSPACESEARCH_EXPORT std::size_t buildProductHash(
```
DavidACosgrove, line 152: I guess collisions are possible this way, but the theoretical probability is hopefully small enough that it is out-weighed by the performance gain, in memory as well as speed.

DavidACosgrove (approved):
This all looks fine to me.  The CI failure is in rdFMCS.

bp-kelley:
Another optimization to try in SynthonSet.cpp, use insertMol instead of combineMols.  The later has exponential behavior as it copies combineMol every step and has better memory usage to boot.

```
  for (size_t i = 0; i < synthons[longVecNum].size(); ++i) {
    auto combMol = std::make_unique<ROMol>();
    for (size_t j = 0; j < synthons.size(); ++j) {
      if (j == longVecNum) {
        combMol.reset(combineMols(*combMol, *synthons[j][i]));
      } else {
        combMol.reset(combineMols(*combMol, *synthons[j].front()));
      }
    }
```

to

```
  for (size_t i = 0; i < synthons[longVecNum].size(); ++i) {
    auto combMol = std::make_unique<ROMol>();
    for (size_t j = 0; j < synthons.size(); ++j) {
      if (j == longVecNum) {
	combMol->insertMol(*synthons[j][i]);
      } else {
	combMol->insertMol(*synthons[j].front());
      }
    }
```

bp-kelley:
molzip has the same behavior btw.  I'll see if I can fix it there as well.

DavidACosgrove:
> molzip has the same behavior btw. I'll see if I can fix it there as well.

Making molzip faster would be great.  It's a significant part of the runtime for the final check of potential hits.

greglandrum:
> Another optimization to try in SynthonSet.cpp, use insertMol instead of combineMols. The later has exponential behavior as it copies combineMol every step and has better memory usage to boot.
> 

This one makes a lot of sense. Maybe worth doing in a separate PR @DavidACosgrove ?

greglandrum (approved):
LGTM

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearch_details.h
d-b-w: An earlier draft used ingt128, not sure how I dropped that.

DavidACosgrove:
> > Another optimization to try in SynthonSet.cpp, use insertMol instead of combineMols. The later has exponential behavior as it copies combineMol every step and has better memory usage to boot.
> 
> This one makes a lot of sense. Maybe worth doing in a separate PR @DavidACosgrove ?

Will do.

DavidACosgrove:
> Another optimization to try in SynthonSet.cpp, use insertMol instead of combineMols. The later has exponential behavior as it copies combineMol every step and has better memory usage to boot.
> 
> ```
>   for (size_t i = 0; i < synthons[longVecNum].size(); ++i) {
>     auto combMol = std::make_unique<ROMol>();
>     for (size_t j = 0; j < synthons.size(); ++j) {
>       if (j == longVecNum) {
>         combMol.reset(combineMols(*combMol, *synthons[j][i]));
>       } else {
>         combMol.reset(combineMols(*combMol, *synthons[j].front()));
>       }
>     }
> ```
> 
> to
> 
> ```
>   for (size_t i = 0; i < synthons[longVecNum].size(); ++i) {
>     auto combMol = std::make_unique<ROMol>();
>     for (size_t j = 0; j < synthons.size(); ++j) {
>       if (j == longVecNum) {
> 	combMol->insertMol(*synthons[j][i]);
>       } else {
> 	combMol->insertMol(*synthons[j].front());
>       }
>     }
> ```

I'm working on the PR for this.  There are a couple of other instances where it also makes sense.  However, this one is only used when building the fingerprint database, so won't impact search speed.  There's no harm in making the fingerprint database creation faster.

PR #9307, opened 2026-05-27
Synthon substructure search 2x performance

d-b-w:

processPartHitsFromDetails called haveEnoughHits after each verified hit, scanning every slot of the results vector to count non-null entries. This was a _nuts_ number of non-null checks for high-hitrate molecules:

Screenshot 2026-05-27 at 9 08 15 AM

Replace with a std::atomic<int64_t> numHitsFound counter in makeHitsFromToTry, incremented via fetch_add on each verified hit. The early-exit condition becomes a single atomic read, O(1) per hit regardless of vector size. The atomic is local to makeHitsFromToTry so it resets correctly per chunk and is safe for the multi-threaded path without added synchronization.

I searched a group of high-hitrate molecules (more than 2million hits), on the 140B-product Freedom space. The overall improvement relative to #9305 was something like 2X. Some standouts:

* benzene: 5.6s -> 3.0s
* acetaminophen: 3.6s -> 1.0s

#### Any other comments?

This is an independent optimization from #9305. They stack for an even larger net improvement. I think there are a few more more microoptimizations in here before I need to start rethinking the core logic.

### DIFF
--- a/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
+++ b/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
@@ -530,7 +530,7 @@ void processPartHitsFromDetails(
                                                 toTry[thisTry].second)) {
       results[thisTry] = std::move(prod);
       ++numHitsFound;
-      if (maxHits != -1 && numHitsFound + 1 >= maxHits + hitStart) {
+      if (maxHits != -1 && numHitsFound >= maxHits + hitStart) {
         break;
       }
     }
--- a/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
+++ b/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
@@ -65,6 +65,7 @@ void SynthonSpaceSearcher::search(const SearchResultCallback &cb) {
 
   // from buildAllhits
   std::vector<std::pair<const SynthonSpaceHitSet *, std::vector<size_t>>> toTry;
+  std::atomic<std::int64_t> numHitsFound = 0;
   std::int64_t hitCount = 0;
   bool stop = false;  // set by callback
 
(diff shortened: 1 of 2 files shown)


Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -543,8 +529,8 @@ void processPartHitsFromDetails(
     if (auto prod = searcher->buildAndVerifyHit(toTry[thisTry].first,
                                                 toTry[thisTry].second)) {
       results[thisTry] = std::move(prod);
-      if (haveEnoughHits(results, searcher->getParams().maxHits,
-                         searcher->getParams().hitStart)) {
+      if (maxHits != -1 &&
+          numHitsFound.fetch_add(1) + 1 >= maxHits + hitStart) {
```
DavidACosgrove: This function now has 2 different ways of incrementing an atomic variable - mostRecentTry uses the plain increment operator, numHitsFound uses fetch_add.  Mistral.ai's Vibe suggests they are both perfectly valid but fetch_add() is more explicit about dealing with an atomic variable so is probably preferable.  Maybe it would be better to make mostRecentTry conform.

DavidACosgrove (approved):
This makes sense to me.

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
greglandrum: my two cents: I find the `++` form a lot more readable, so I'd prefer that we use that form if we're going to pick one

d-b-w: swapped to ++ for both

greglandrum:
@d-b-w it looks like the CI failures here are real

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -543,8 +529,8 @@ void processPartHitsFromDetails(
     if (auto prod = searcher->buildAndVerifyHit(toTry[thisTry].first,
                                                 toTry[thisTry].second)) {
       results[thisTry] = std::move(prod);
-      if (haveEnoughHits(results, searcher->getParams().maxHits,
-                         searcher->getParams().hitStart)) {
+      ++numHitsFound;
+      if (maxHits != -1 && numHitsFound + 1 >= maxHits + hitStart) {
```
greglandrum: This shouldn't be `+1`
```suggestion
      if (maxHits != -1 && numHitsFound >= maxHits + hitStart) {
```

d-b-w:
> @d-b-w it looks like the CI failures here are real

Addressed

greglandrum (approved):
LGTM

PR #9308, opened 2026-05-28
switch to specific versions of other actions in the templates action

greglandrum:
I think we're safer using specific versions of the other actions we use


ricrogz (approved):
LGTM

PR #9219, opened 2026-04-01
Fix STEREOANY (wavy bond) loss during InChI roundtrip

rodyarantes:
## Summary

Wavy bonds (`Bond::STEREOANY` on double bonds) are silently dropped during InChI roundtrip and CXSMILES serialization. This PR fixes the full pipeline so that `CXSMILES(|w:|) → mol → InChI(-SUU) → mol → CXSMILES(|w:|)` preserves wavy bond information.

**Root cause:** Three bugs across two files — all in RDKit's conversion layers, not in the InChI C library.

### 1. InChI reverse path (`InchiToMol`) — `External/INCHI-API/inchi.cpp`

The stereo0D processing loop skipped `INCHI_PARITY_UNDEFINED` entries before they could reach the double bond handler, which already had an `else` clause that correctly sets `Bond::STEREOANY`:

```cpp
// Before: UNDEFINED skipped, never reaches STEREOANY handler
if (stereo0DPtr->parity == INCHI_PARITY_NONE ||
    stereo0DPtr->parity == INCHI_PARITY_UNDEFINED) {
  continue;
}
```

**Fix:** Only skip `INCHI_PARITY_NONE`. Add a guard inside the `INCHI_StereoType_Tetrahedral` case to prevent `UNDEFINED`/`UNKNOWN` from incorrectly assigning chirality.

### 2. InChI forward path (`MolToInchi`) — `External/INCHI-API/inchi.cpp`

`STEREOANY` bonds only collapsed coordinates to prevent InChI from inferring stereo from geometry, but never sent an explicit `inchi_Stereo0D` entry. With `-SUU`, InChI needs this entry to produce the correct unknown stereo annotation (`/b..?`).

**Fix:** Keep the coordinate collapse **and** send a `stereo0D` entry with `INCHI_PARITY_UNKNOWN`. Since RDKit clears `getStereoAtoms()` for STEREOANY bonds, neighbors are discovered by iterating atom bonds.

### 3. CXSMILES writer — `Code/GraphMol/SmilesParse/CXSmilesOps.cpp`

`MolToCXSmiles` did not write `|w:|` for double bonds with `Bond::STEREOANY` unless `BondDir::UNKNOWN` was already set on an adjacent single bond. After InChI roundtrip, `STEREOANY` is set on the double bond but no `BondDir` is set on neighbors, so `|w:|` was silently dropped.

[...]

PR #9315, opened 2026-05-29
Fix STEREOANY (wavy bond) loss during InChI roundtrip

rodyarantes:
Replaces #9219, split per @greglandrum's review request to separate the CXSMILES writer changes into their own PR.

## Summary

When converting molecules with wavy bonds (`Bond::STEREOANY` on double bonds) through InChI and back, the stereo information was silently dropped. This PR fixes two bugs in `External/INCHI-API/inchi.cpp`:

**Reverse path (`InchiToMol`):** The stereo0D processing loop skipped `INCHI_PARITY_UNDEFINED` entries before they could reach the double bond handler. The handler already had an `else` clause that correctly sets `Bond::STEREOANY`, but it was never reached. Fix: only skip `INCHI_PARITY_NONE` at the top level, and add a guard in the `Tetrahedral` case to prevent `UNDEFINED`/`UNKNOWN` from incorrectly setting chirality.

**Forward path (`MolToInchi`):** `STEREOANY` double bonds were only handled by collapsing the coordinates — InChI then produced no stereo annotation under `-SUU`. Fix: also emit a stereo0D entry with `INCHI_PARITY_UNKNOWN` so InChI's `-SUU` output correctly carries the "stereo unknown" designation. `StereoAtoms` may be cleared for `STEREOANY`, so we locate the two outer neighbors by iterating bonds.

## Tests

New `testStereoAnyRoundtrip` in `External/INCHI-API/test.cpp` covers 9 representative cases (Schiff base, oxime, cinnamic acid, chalcone, crotonaldehyde, tamoxifen-like, retinal-like, plus two molecules with a chiral center adjacent to the wavy bond).

`rdkit/Chem/UnitTestInchi.py` counters shift by 1 (689 same, 492 reasonable) because the new STEREOANY emission produces a more accurate roundtrip for one entry in the test inventory.

## Note

The CXSMILES writer changes that were originally bundled in #9219 are being moved to a separate PR (forthcoming) with dedicated tests, per the review feedback.

### DIFF
--- a/External/INCHI-API/inchi.cpp
+++ b/External/INCHI-API/inchi.cpp
@@ -1439,8 +1439,7 @@ RWMol *InchiToMol(const std::string &inchi, ExtraInchiReturnValues &rv,
         Chirality::assignAtomCIPRanks(*m, ranks);
         for (unsigned int i = 0; i < numStereo0D; i++) {
           inchi_Stereo0D *stereo0DPtr = inchiOutput.stereo0D + i;
-          if (stereo0DPtr->parity == INCHI_PARITY_NONE ||
-              stereo0DPtr->parity == INCHI_PARITY_UNDEFINED) {
+          if (stereo0DPtr->parity == INCHI_PARITY_NONE) {
             continue;
           }
           switch (stereo0DPtr->type) {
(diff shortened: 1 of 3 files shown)


rodyarantes:
Hi @greglandrum — thanks again for the careful review on #9219! You mentioned there that the InChI wrapper changes and tests looked good to you, so this PR is just that half, rebased onto current master with no other changes from what you already reviewed. Whenever you have a moment to take another look, that would be much appreciated. Happy to address any further feedback.

greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @rodyarantes !

PR #9316, opened 2026-05-30
Fix CXSMILES writer to emit |w:| for STEREOANY double bonds

rodyarantes:
Split off from #9219 per @greglandrum's review request, with dedicated tests demonstrating each fix.

## Summary

When a `STEREOANY` double bond has its stereo information set (for example, after an InChI roundtrip restores it from an `UNKNOWN`-parity `stereo0D`, but also any programmatic `SetStereo(STEREOANY)` with stereo atoms), the CXSMILES writer previously dropped the marker silently. The CXSMILES parser handles `|w:|` by setting `BondDir::UNKNOWN` on the adjacent single bond, so the writer needs to mirror that — emit a wavy marker on an adjacent single bond.

## Change

The new code path in `get_bond_config_block()` (`Code/GraphMol/SmilesParse/CXSmilesOps.cpp`) runs after the main wedging loop and only fires when nothing else has handled the `STEREOANY` double bond. It carries four guards so it does not duplicate or contradict existing markers:

- **Ring double bonds** are handled by `get_ringbond_cistrans_block()`, which emits `|ctu:|`; skip them here.
- **Adjacent `BondDir::UNKNOWN`**: the main loop already emitted `|w:|` (parser path).
- **Adjacent `_MolFileBondCfg`**: mol-file wedging will produce the wavy via `wU`/`wD`; do not double-emit.
- **Bond carries `_MolFileBondStereo`** (crossed double bond from a mol file) or has no `stereoAtoms` set: defer to existing paths.

## Tests

New `TEST_CASE("CXSMILES writer emits |w:| for STEREOANY double bonds")` in `Code/GraphMol/SmilesParse/cxsmiles_test.cpp` covers each guard:

| sub-test | demonstrates |
|---|---|
| acyclic STEREOANY set programmatically produces `\|w:\|` | core fix |
| ring STEREOANY uses `\|ctu:\|`, never `\|w:\|` | ring-bond skip |
| `\|w:\|` roundtrip is idempotent and does not duplicate | adjacent `BondDir::UNKNOWN` guard |
| STEREOANY with empty stereo atoms does not emit `\|w:\|` | empty-stereo-atoms guard |
| STEREOANY with `_MolFileBondStereo` defers to mol-file wedging | crossed-double-bond guard |
| adjacent `_MolFileBondCfg=2` suppresses duplicate `\|w:\|` | mol-file-cfg guard |

Verified locally — `cxsmilesTest`, `testExtendedStereoParsing`, `fileParsersCatchTest`, `molfileStereoCatchTest`, and the InChI test (`testInchi`) all pass.

## Related

This is the CXSMILES half of the original bundled PR #9219. The InChI half is now #9315.

### DIFF
--- a/Code/GraphMol/SmilesParse/CXSmilesOps.cpp
+++ b/Code/GraphMol/SmilesParse/CXSmilesOps.cpp
@@ -2265,6 +2265,74 @@ std::string get_bond_config_block(
           boost::str(boost::format("%d.%d") % begAtomOrder % i));
     }
   }
+  // For double bonds with STEREOANY, emit a wavy marker on an adjacent single
+  // bond so that |w:| is written. This mirrors what the CXSMILES parser does
+  // when reading |w:| (it sets BondDir::UNKNOWN on the single bond neighbor).
+  for (unsigned int i = 0; i < bondOrder.size(); ++i) {
+    auto idx = bondOrder[i];
+    const auto bond = mol.getBondWithIdx(idx);
+    if (bond->getBondType() != Bond::BondType::DOUBLE ||
+        bond->getStereo() != Bond::BondStereo::STEREOANY) {
+      continue;
+    }
+    // Skip ring double bonds — they are handled by
+    // get_ringbond_cistrans_block() which emits |ctu:| instead.
+    if (mol.getRingInfo()->isInitialized() &&
+        mol.getRingInfo()->numBondRings(idx)) {
+      continue;
+    }
+    // Check if an adjacent single bond already has BondDir::UNKNOWN
+    // (set by the CXSMILES parser) or _MolFileBondCfg (set by the
+    // mol file parser). If so, skip — the wavy bond is already handled
+    // by the main loop above, or by wU/wD markers from mol file wedging.
+    bool alreadyHandled = false;
+    for (const auto &nbond : mol.atomBonds(bond->getBeginAtom())) {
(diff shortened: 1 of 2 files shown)


rodyarantes:
Hi @greglandrum — thanks again for the review feedback on #9219! You suggested moving the CXSMILES changes into a separate PR "along with some tests that demonstrate what they are trying to fix," so here it is. I added a dedicated `TEST_CASE` in `cxsmiles_test.cpp` with one sub-test per guard in the new code path (acyclic STEREOANY, ring STEREOANY, idempotent roundtrip, empty stereo atoms, `_MolFileBondStereo`, and `_MolFileBondCfg`), so each part of the change has a small concrete demonstration of what it fixes. Whenever you have time to take a look, that would be much appreciated — and happy to iterate on anything.

rodyarantes:
Hey @greglandrum - This is the other PR. Let me know if you need any additional information or context for this.

greglandrum (changes requested):
Sorry for the slow review of this one.

The idea of the change makes sense to me, but I think that it would be better to move the logic for picking the bond that is going to be marked as wiggly to a standalone function in the file WedgeBonds.cpp (with a declaration in Chirality.h).

This would allow us to implement a function that applies the IUPAC recommendation to use wiggly bonds to mark unspecified double bonds. (Note: we don't currently do this because some of us think that it's a terrible recommendation, but I could still imagine wanting to allow it at some point in the future)

PR #9318, opened 2026-05-30
Build: forward-slash Python3_EXECUTABLE when generating rdkit-stubs

skearnes:
`rdkit-stubs/CMakeLists.txt` assembles a cmake script (`RUN_GEN_RDKIT_STUBS_PY`) that is later executed via `cmake -P`. Every path embedded into that generated script is first converted to forward slashes — `CMAKE_SOURCE_DIR`, `CMAKE_CURRENT_BINARY_DIR`, and `PYTHON_INSTDIR` all go through `string(REGEX REPLACE "\\\\" "/" ...)` — so that the backslashes in native Windows paths are not treated as escape sequences when the generated script is re-parsed.

`Python3_EXECUTABLE` was the one path that was left unconverted, so on Windows the generated `COMMAND` line still contained backslashes and failed to parse. This applies the same forward-slash conversion already used for the neighboring paths.

#### Any other comments?
This has been carried as a patch in [conda-forge/rdkit-feedstock](https://github.com/conda-forge/rdkit-feedstock); upstreaming it so the feedstock can drop the patch.

### DIFF
--- a/rdkit-stubs/CMakeLists.txt
+++ b/rdkit-stubs/CMakeLists.txt
@@ -14,13 +14,14 @@ set(CONCURRENCY "")
 if (DEFINED ENV{CMAKE_BUILD_PARALLEL_LEVEL})
   set(CONCURRENCY "--concurrency $ENV{CMAKE_BUILD_PARALLEL_LEVEL} ")
 endif()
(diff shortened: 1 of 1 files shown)


greglandrum (approved):
LGTM

greglandrum:
The CI failure is not related to this PR

PR #9321, opened 2026-06-03
Some minor changes to GaussianShape

DavidACosgrove:
#### Reference Issue
No issue.

Whilst merging master after PR #9265 into PR #9279, a few things cropped up.
Firstly, in his review of #9279 @d-b-w requested a change to the signature of ShapeInput::merge() function to show that other is altered by the merge.
Also in that function, the PRECONDITION that the SMILES of this and other matched turned out to be too strict.  The SynthonSpace shape searching puts all stereoisomers of a molecule into the same shape, and the PRECONDITION broke that.   The new test on the types of the Gaussians is not very strict, since all atoms have type 0 but better than nothing.  I thought about checking the d_alphas but that would be prone to floating point issues.
There were a couple of commented out debugging writes.


greglandrum (approved):
LGTM

PR #9323, opened 2026-06-03
Support generating query molecules from CDXML

pechersky:
Chemists can use ChemDraw (native or web applet) to specify queries. But previously, one would have needed to rely on conversion to MDL or out of band parsing to make correct queries from those files.

Instead, use the ChemDraw library to natively parse these query configurations. The primary target for these query mols is SMARTS usage inside a pg cartridge.

#### Any other comments?

The precise semantics of how one handles things like explicit Hs was based off of chemist feedback. I'm fine to start a bikeshed conversation on how we want to handle some semantics.

GPT 5.4 xhigh via `copilot` harness assisted in writing the code and re-running the tests. User facilitated in restructuring git history to make it easier to review.

cc @bp-kelley


bp-kelley:
@pechersky I'll have time to review this over the next week.

Code/GraphMol/FileParsers/CDXMLParser.cpp
```diff
@@ -114,6 +127,335 @@ Atom *addquery(Q *qry, std::string symbol, RWMol &mol, unsigned int idx) {
   return res;
 }
 
+bool isRQueryLabel(const std::string &label) {
```
bp-kelley, line 130: This looked a bit of effort, but I'm not sure we really want to support the legacy CDXML parser anymore.  I'm ok having to use the full ChemDraw parser to read queries.

bp-kelley: I might be misunderstanding this code though...

bp-kelley:
My first pass seems positive.  However, since this is a fundamental piece of kit, I'm going to spend some time comparing the query mols generated here versus the ones we generate from V3K files.  At the minimum, we should know the differences.

greglandrum:
@bp-kelley let me know when this one is ready and I will do a final quick check and merge

bp-kelley:
@greglandrum Gareth and I are going to go over this as well, he has a lot more experience converting chemdraw queries to RDKit, so this will be delayed until he finishes his climbing vacation.

bp-kelley:
@pechersky I'm back on this one, will let you know as it goes.

greglandrum:
@pechersky this now has merge conflicts. It will no doubt help @bp-kelley's review if you can resolve those.

Code/GraphMol/FileParsers/CDXMLParser.cpp
bp-kelley: @pechersky any comment here?  I think we should revert the changes in the legacy code and require use of the ChemDraw library

bp-kelley:
@pechersky do you want me to resolve the merge conflicts?  I'd like to finish this up in the next week or so.

pechersky:
Feel free to contribute changes here, I'm on leave for at least the next month and won't be able to write code during that time.

bp-kelley:
@greglandrum I finally reviewed this, and fixed the issues greptile found.  I'm not sure I can bush directly here, but I'll add a second PR from my fixed branch until we can sort it out.  Thanks for fixing the merge conflict.

bp-kelley:
@greglandrum I moved the PR to here:  https://github.com/rdkit/rdkit/pull/9567

PR #9328, opened 2026-06-04
Fix DistGeomHelpers tests on arm64

tadhurst-cdd:
#### Reference Issue
Gaussian catch_test.cpp was giving round off errors depending on the platform.
Same this for DistGeometryHelper tests

For ARM64, the expected values for one test  in Gaussian catch_test.spp were adjusted slightly.
The expected values for one test in DistGeometryHelperd testDGeomHelpers were adjusted slightly.

#### Any other comments?
Brian Kelly has indicated that a larger PR is in the works that may be changing the Gaussian catch stuff (I think).


bp-kelley:
@tadhurst-cdd I can't trigger the failures on my ARM docker, but I'll give another try this week to see if I can.

tadhurst-cdd:
I re-pulled master and merged with my branches, and the gaussian shape
issue may be disappeared 9been fixed) or something. I am trying it right
now with the gaussian fix removed.

The Dist Geo issue works find for me locally when I build in debug and
fails when I build is release mode.  I am leaving the change to dist geom
to be 12.5 rather then 12.
tad

greglandrum (approved):
LGTM

PR #9332, opened 2026-06-06
Catch zero-length vector in getAtomSymbolAndOrientation.

### LINKED ISSUE #9329: MolDraw2D: "Cannot normalize a zero length vector" — wedge/atom-note placement not guarded against degenerate 2D coordinates
### Description

`rdMolDraw2D.PrepareAndDrawMolecule` / `DrawMolecule` raise
`RuntimeError: Cannot normalize a zero length vector` when the 2D layout contains a
zero-length (or near-zero) direction vector. The drawing code normalizes bond-wedging and
atom-note direction vectors without guarding against zero length, so it hard-crashes
instead of falling back to a default direction. In a batch rendering pipeline a single
such molecule aborts the whole run.

It bites valid molecules in two ways:

1. **Coincident atoms.** `Compute2DCoords` (legacy) occasionally places two bonded atoms at
   the same position for valid molecules, producing a zero-length bond.
2. **Atom-note placement.** Setting `atomNote` on certain valid molecules crashes note-label
   placement, *even when the same molecule draws fine without notes*.

### Minimal, self-contained reproduction

The drawing path normalizes a bond vector that is zero whenever two bonded atoms share a
position (the degenerate state `Compute2DCoords` can produce):

```python
from rdkit import Chem
from rdkit.Geometry import Point3D

conf = Chem.Conformer(mol.GetNumAtoms())
conf.SetAtomPosition(0, Point3D(0.0, 0.0, 0.0))
conf.SetAtomPosition(1, Point3D(0.0, 0.0, 0.0))   # atoms 0 and 1 coincide -> zero-length bond
conf.SetAtomPosition(2, Point3D(1.0, 0.0, 0.0))
mol.AddConformer(conf)

d = rdMolDraw2D.MolDraw2DSVG(200, 200)
rdMolDraw2D.PrepareAndDrawMolecule(d, mol)
# RuntimeError: Cannot normalize a zero length vector
```

### Occurs organically via the standard API (no manual coordinates)

For some valid molecules, `MolFromSmiles` -> `AllChem.Compute2DCoords` ->
`atom.SetProp("atomNote", ...)` -> `PrepareAndDrawMolecule` raises the same error, while the
identical molecule renders fine **without** the atom notes. In our data the affected molecules
have substituted-cyclopropane / fused-small-ring motifs (~4 in 245k drug-like molecules).

Related and closed: #8637 is the bond-wedging variant for `C12O/C=C\OC2C1CCC`, which no longer
reproduces in 2025.09.6, but the atom-note variant above still does.

### Expected

When a wedge/note direction vector has zero length, the drawing code should fall back to a
default direction (or skip the offset) rather than raising — so a single degenerate layout
cannot abort rendering of an otherwise-valid molecule.

### Workaround

`rdDepictor.SetPreferCoordGen(True)` before `Compute2DCoords` yields a non-degenerate layout
that both draws and places notes cleanly.

### Versions
- RDKit 2025.09.6
- Python 3.12, Linux

DavidACosgrove:
#### Reference Issue
Fixes #9329

Catches the exception thrown by directionVector when trying to work out the orientation of atom labels.

#### Any other comments?
I believe that the try has no overhead, so this should be faster than testing the length before calling thus giving 2 length tests. Unwinding the exception will be slow, but that should be rare.  Indeed, this code hasn't been changed since 2022 and it's the first time someone has noticed the problem :-).

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawMol.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawMol.cpp
@@ -4082,10 +4082,14 @@ bool isLinearAtom(const Atom &atom, const std::vector<Point2D> &atCds) {
     ROMol const &mol = atom.getOwningMol();
     int i = 0;
     for (auto nbr : make_iterator_range(mol.getAtomNeighbors(&atom))) {
-      Point2D bond_vec = at1_cds.directionVector(atCds[nbr]);
-      bond_vec.normalize();
-      bond_vecs[i] = bond_vec;
-      bts[i] = mol.getBondBetweenAtoms(atom.getIdx(), nbr)->getBondType();
+      try {
+        Point2D bond_vec = at1_cds.directionVector(atCds[nbr]);
+        bond_vecs[i] = bond_vec;
+        bts[i] = mol.getBondBetweenAtoms(atom.getIdx(), nbr)->getBondType();
+      } catch (std::runtime_error &e) {
+        // A zero-length vector throws and can be ignored.
+        continue;
+      }
       ++i;
     }
     return (bts[0] == bts[1] && bond_vecs[0].dotProduct(bond_vecs[1]) < -0.95);
--- a/Code/GraphMol/MolDraw2D/catch_tests.cpp
+++ b/Code/GraphMol/MolDraw2D/catch_tests.cpp
@@ -379,7 +379,8 @@ const std::map<std::string, std::hash_result_t> SVG_HASHES = {
     {"test_Github9280_1.0.svg", 1658116840U},
     {"test_Github9280_2.0.svg", 1805554327U},
     {"test_Github9280_0.3.svg", 893100468U},
-    {"test_Github9280_0.2.svg", 770838895U}};
+    {"test_Github9280_0.2.svg", 770838895U},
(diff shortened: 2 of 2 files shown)


Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -4082,10 +4082,14 @@ bool isLinearAtom(const Atom &atom, const std::vector<Point2D> &atCds) {
     ROMol const &mol = atom.getOwningMol();
     int i = 0;
     for (auto nbr : make_iterator_range(mol.getAtomNeighbors(&atom))) {
-      Point2D bond_vec = at1_cds.directionVector(atCds[nbr]);
-      bond_vec.normalize();
```
greglandrum, line 4086: You lost the normalize, which probably removes the need for the try...catch in the first place, but it certainly means that the dotProduct threshold check that is done further below now means something different.

greglandrum: Never mind! `directionVector()` does the normalization

greglandrum (approved):
LGTM

PR #9333, opened 2026-06-06
Catch a few more potential zero length vectors.

DavidACosgrove:
#### Reference Issue
No issue, but follows on from #9329

3 more cases where the code could potentially throw zero length vector errors.
2 of the changes are just tidying the existing detection which involved duplicate length calculations and already have tests.  The other is drawing brackets around polymer units for which a test is added.

#### Any other comments?
I have checked that not of the drawings change compared with master.

### DIFF
--- a/Code/GraphMol/MolDraw2D/DrawMol.cpp
+++ b/Code/GraphMol/MolDraw2D/DrawMol.cpp
@@ -792,62 +792,66 @@ void DrawMol::extractBrackets() {
       // bracket is largely horizontal or largely vertical.
       const auto &brkShp = *postShapes_.back();
       Point2D longline = brkShp.points_[1] - brkShp.points_[2];
-      longline.normalize();
-      static const double cos45 = 1.0 / sqrt(2.0);
-      bool horizontal = fabs(longline.x) > cos45;
-      size_t labelBrk = postShapes_.size() - 1;
-      for (int i = 1; i < numBrackets; ++i) {
-        const auto &brkShp = *postShapes_[postShapes_.size() - i - 1];
-        if (horizontal) {
-          if (brkShp.points_[2].y > postShapes_[labelBrk]->points_[2].y) {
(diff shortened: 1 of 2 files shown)


Code/GraphMol/MolDraw2D/DrawMol.cpp
```diff
@@ -792,62 +792,66 @@ void DrawMol::extractBrackets() {
...
-      for (int i = 1; i < numBrackets; ++i) {
-        const auto &brkShp = *postShapes_[postShapes_.size() - i - 1];
-        if (horizontal) {
-          if (brkShp.points_[2].y > postShapes_[labelBrk]->points_[2].y) {
-            labelBrk = postShapes_.size() - i - 1;
-          }
-        } else {
-          if (brkShp.points_[2].x > postShapes_[labelBrk]->points_[2].x) {
-            labelBrk = postShapes_.size() - i - 1;
+      try {
```
greglandrum, line 795: I really don't like these huge try..catch blocks. If you're going to use this approach (and I wouldn't mind seeing some proof that it doesn't actually affect performance), please limit the amount of code in the try to be the code that could actually throw

DavidACosgrove: According to some people on the interweb, the try bit is essentially zero cost. Unwinding the exception is expensive, but that should be rare. On the other side of the equation, without the try/catch I would have to check the length of the vector every time before proceeding, which clearly does have a cost.

DavidACosgrove: This is what chat.mistral.ai has to say on the matter.  Apologies, it had to be a screenshot:
image

greglandrum:
Changes look fine so far. Pleas fix the merge conflict and then I can merge

greglandrum (approved):
LGTM

PR #9337, opened 2026-06-09
Refactor BoundsMatrixBuilder

kabu00002:
Refactoring of the `BoundsMatrixBuilder.cpp` code. The focus was to restructure the set14Bounds code to avoid code duplication; only minor changes were done on the remaining code in the specified file.

### DIFF
--- a/Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
+++ b/Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
@@ -273,6 +273,9 @@ inline std::size_t getUnifiedId(const unsigned int id1, const unsigned int id2,
 }
 
 inline bool squishBond(const ROMol &mol, const Bond *bond) {
+  // determines whether the corresponding atoms are larger heteroatoms (at least
+  // of of them) in conjugated 5 rings, because we need to add a bit of extra
+  // flex for them
   return bond->getIsConjugated() &&
          (bond->getBeginAtom()->getAtomicNum() > 10 ||
           bond->getEndAtom()->getAtomicNum() > 10) &&
--- a/Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
+++ b/Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
@@ -258,7 +258,7 @@ void _checkAndSetBounds(unsigned int i, unsigned int j, double lb, double ub,
 
 inline std::size_t getUnifiedId(const unsigned int id1, const unsigned int id2,
                                 const unsigned int n) {
-  // returns an id for (id1, id2) independed of order within range (0, 2*n - 1)
+  // returns an id for (id1, id2) independent of order within range (0, 2*n - 1)
   // assuming id1 < n and id2 < n
   return id1 < id2 ? (static_cast<std::size_t>(id1) * n + id2)
                    : (static_cast<std::size_t>(id2) * n + id1);
@@ -266,7 +266,7 @@ inline std::size_t getUnifiedId(const unsigned int id1, const unsigned int id2,
 
 inline std::size_t getUnifiedId(const unsigned int id1, const unsigned int id2,
                                 const unsigned int id3, const unsigned int n) {
-  // returns an id for (id1, id2, id3) independed of order of id1, id3 within
+  // returns an id for (id1, id2, id3) independent of order of id1, id3 within
   // range (0, 3*(n) - 1) assuming id1 < n, id2 < n and id3 < n
   return id1 < id3 ? (static_cast<std::size_t>(id1) * n * n + id2 * n + id3)
                    : (static_cast<std::size_t>(id3) * n * n + id2 * n + id1);
--- a/Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
+++ b/Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
@@ -36,7 +36,6 @@ const double VDW_SCALE_15 = 0.7;
 constexpr double H_BOND_LENGTH = 1.8;
 const double MAX_UPPER = 1000.0;
 static const double minMacrocycleRingSize = 9;
-#include <map>
 
 namespace RDKit {
 namespace DGeomHelpers {
@@ -47,9 +46,6 @@ typedef boost::shared_ptr<RDNumeric::DoubleSymmMatrix> SymmDoubleMatPtr;
 typedef boost::dynamic_bitset<> BIT_SET;
 
 //! Bunch of functions to set distance bound based on topology
-
-typedef std::map<int, double> INT_DOUBLE_MAP;
-typedef INT_DOUBLE_MAP::const_iterator INT_DOUBLE_MAP_CI;
 
 typedef std::vector<long int> LINT_VECT;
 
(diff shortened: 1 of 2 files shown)


Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -235,6 +256,29 @@ void _checkAndSetBounds(unsigned int i, unsigned int j, double lb, double ub,
   }
 }
 
+inline std::size_t getUnifiedId(const unsigned int id1, const unsigned int id2,
+                                const unsigned int n) {
+  // returns an id for (id1, id2) independed of order within range (0, 2*n - 1)
```
greglandrum:
```suggestion
  // returns an id for (id1, id2) independent of order within range (0, 2*n - 1)
```

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -235,6 +256,29 @@ void _checkAndSetBounds(unsigned int i, unsigned int j, double lb, double ub,
...
+  // assuming id1 < n and id2 < n
+  return id1 < id2 ? (static_cast<std::size_t>(id1) * n + id2)
+                   : (static_cast<std::size_t>(id2) * n + id1);
+}
+
+inline std::size_t getUnifiedId(const unsigned int id1, const unsigned int id2,
+                                const unsigned int id3, const unsigned int n) {
+  // returns an id for (id1, id2, id3) independed of order of id1, id3 within
```
greglandrum:
```suggestion
  // returns an id for (id1, id2, id3) independent of order of id1, id3 within
```

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -235,6 +256,29 @@ void _checkAndSetBounds(unsigned int i, unsigned int j, double lb, double ub,
...
+  // range (0, 3*(n) - 1) assuming id1 < n, id2 < n and id3 < n
+  return id1 < id3 ? (static_cast<std::size_t>(id1) * n * n + id2 * n + id3)
+                   : (static_cast<std::size_t>(id3) * n * n + id2 * n + id1);
+}
+
+inline bool squishBond(const ROMol &mol, const Bond *bond) {
```
greglandrum, line 275: I think it would be good to add the comment from below about what this function is actually for

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -361,57 +406,57 @@ void setLowerBoundVDW(const ROMol &mol, DistGeom::BoundsMatPtr mmat, bool,
...
-  if (isLargerSP2Atom(mol.getAtomWithIdx(aid3))) {
-    distTol *= 2;
-  }
-  auto du = dl + distTol;
-  dl -= distTol;
+
+  const auto distTol =
+      DIST13_TOL * (1u << isLargerSP2Atom(mol.getAtomWithIdx(aid1))
+                       << isLargerSP2Atom(mol.getAtomWithIdx(aid))
+                       << isLargerSP2Atom(mol.getAtomWithIdx(aid3)));
```
greglandrum: We can leave it this way, but I'm not convinced that this is easier to read (and I really doubt that it's faster).

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -1369,296 +1119,215 @@ bool _checkMacrocycleTwoInSameRingAmideEster14(
...
+  PRECONDITION(bnd2, "");
+  PRECONDITION(bnd3, "");
 
-      path14.type = Path14Configuration::OTHER;
+  unsigned int bid1 = bnd1->getIdx();
+  unsigned int bid2 = bnd2->getIdx();
+  unsigned int bid3 = bnd3->getIdx();
+
+  const Atom *atm2 = mol.getAtomWithIdx(accumData.bondAdj->getVal(bid1, bid2));
+  PRECONDITION(atm2, "");
```
greglandrum: This isn't a PRECONDITION.
If you wanted to do a test here, it would be `TEST_ASSERT()` or `CHECK_INVARIANT()`, but I don't think the check is necessary at all: `getAtomWithIdx()` has a `POSTCONDITION()` on it that the result is non-null.

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -1369,296 +1119,215 @@ bool _checkMacrocycleTwoInSameRingAmideEster14(
...
+  const Atom *atm3 = mol.getAtomWithIdx(accumData.bondAdj->getVal(bid2, bid3));
+  PRECONDITION(atm3, "");
```
greglandrum: same comment here

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -1369,296 +1119,215 @@ bool _checkMacrocycleTwoInSameRingAmideEster14(
...
 
-    // we only overwrite bounds if they are not 1-2 nor 1-3 distances
-    _checkAndSetBounds(aid1, aid4, dl, du, mmat);
-    accumData.paths14.push_back(path14);
-    accumData.visited14Bounds.set(pid);
+  CHECK_INVARIANT(ba12 > 0.0, "");
+  CHECK_INVARIANT(ba23 > 0.0, "");
+
+  const Atom *atm1 = mol.getAtomWithIdx(aid1);
+  const Atom *atm4 = mol.getAtomWithIdx(aid4);
```
greglandrum, line 1254: This is better: no checks on the return value of `getAtomWithIdc()`

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -1730,30 +1398,25 @@ void set14Bounds(const ROMol &mol, DistGeom::BoundsMatPtr mmat,
         for (const auto bnd3 : mol.atomBonds(mol.getAtomWithIdx(aid3))) {
           auto bid3 = bnd3->getIdx();
           if (bid3 != bid2) {
-            auto id1 = bid1 * nb * nb + bid2 * nb + bid3;
-            auto id2 = bid3 * nb * nb + bid2 * nb + bid1;
-            if (donePaths.find(id1) == donePaths.end() &&
-                donePaths.find(id2) == donePaths.end()) {
+            auto id = getUnifiedId(bid1, bid2, bid3, nb);
+            if (!donePaths[id]) {  // == donePaths.end()) {
```
greglandrum: no need to leave the old code in there:
```suggestion
            if (!donePaths[id]) {
```

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -1369,296 +1119,215 @@ bool _checkMacrocycleTwoInSameRingAmideEster14(
...
+                                   *torsionValue.value) -
+           GEN_DIST_TOL;
+      du = dl + 2 * GEN_DIST_TOL;
+      break;
+    default:  // NONE  => do not set the bounds => nothing more to do
+      return;
+  }
+
+  Path14Configuration path14 = {
+      .bid1 = bid1, .bid2 = bid2, .bid3 = bid3, .type = torsionValue.type};
```
greglandrum: Since you're setting all the values, I think it's clearer to leave out the designators:
```suggestion
  Path14Configuration path14 = { bid1, bid2, bid3, torsionValue.type};
```
Otherwise when reading the code you need to check to see if one of the values is being left out.

greglandrum (changes requested):
I like it!
I've suggested a few changes.

Code/GraphMol/DistGeomHelpers/BoundsMatrixBuilder.cpp
```diff
@@ -235,6 +256,32 @@ void _checkAndSetBounds(unsigned int i, unsigned int j, double lb, double ub,
...
+                                const unsigned int id3, const unsigned int n) {
+  // returns an id for (id1, id2, id3) independent of order of id1, id3 within
+  // range (0, 3*(n) - 1) assuming id1 < n, id2 < n and id3 < n
+  return id1 < id3 ? (static_cast<std::size_t>(id1) * n * n + id2 * n + id3)
+                   : (static_cast<std::size_t>(id3) * n * n + id2 * n + id1);
+}
+
+inline bool squishBond(const ROMol &mol, const Bond *bond) {
+  // determines whether the corresponding atoms are larger heteroatoms (at least
+  // of of them) in conjugated 5 rings, because we need to add a bit of extra
```
greglandrum:
```suggestion
  // one of them) in conjugated 5 rings, because we need to add a bit of extra
```

greglandrum (changes requested):
one further comment typo suggestion here, then this is ready to go

greglandrum (approved):
LGTM

PR #9339, opened 2026-06-09
Fix #9338 by canonicalizing double bond stereo limited to STEREO E/Z/CIS/TRANS

### LINKED ISSUE #9338: TautomerEnumerator.Canonicalize() raises pre-condition violation when `SetRemoveBondStereo(False)` on a molecule with atropisomer bond
**Describe the bug**
Canonicalization aborts with
```
    bad bond stereo
    Violation occurred on line 362 in file Code/GraphMol/Canon.cpp
    Failed Expression: dblBond->getStereo() > Bond::STEREOANY
```

**To Reproduce**
```
from rdkit import Chem

molblock = """
     RDKit          2D

M  V30 BEGIN CTAB
M  V30 COUNTS 7 6 0 0 0
M  V30 BEGIN ATOM
M  V30 1 C 66.546902 -10.607295 0.000000 0
M  V30 2 C 67.407319 -9.379170 0.000000 0
M  V30 3 N 68.901071 -9.509378 0.000000 0
M  V30 4 C 69.532321 -10.864587 0.000000 0
M  V30 5 O 68.675029 -12.088546 0.000000 0
M  V30 6 C 71.021905 -10.994795 0.000000 0
M  V30 7 C 67.180236 -11.966671 0.000000 0
M  V30 END ATOM
M  V30 BEGIN BOND
M  V30 1 2 4 5
M  V30 2 1 4 6 CFG=1
M  V30 3 1 3 4
M  V30 4 2 2 1
M  V30 5 1 1 7
M  V30 6 1 3 2
M  V30 END BOND
M  V30 END CTAB
M  END
"""

mol = Chem.MolFromMolBlock(molblock)
assert mol is not None

enum = rdMolStandardize.TautomerEnumerator()
enum.SetRemoveBondStereo(False)
enum.Canonicalize(mol)
```

**Expected behavior**
No crash. If bond stereo can't be retained, clear the stereo in a valid way.

**Screenshots**
Image

 - RDKit version:  2026.3.3
 - OS: AL2023
 - Python version (if relevant): 3.11.15
 - Are you using conda? no
 - If you are not using conda: how did you install the RDKit? built from source

**Additional context**
Fragment is unusual -- came from reading a CDXML with a WedgeEnd in a `Fragment`.

I bisected this locally:

 - good: 872b054d5caa868da0ab7e8a616760ca306e95ff
 - first bad: d5aa90e18f2877e7c29a9a732df98ab9a9c5faf3 (Fix for issue #8965 (#8968))

pechersky:
#### Reference Issue
Fixes #9338

Canonicalization could have copied non-double-bond stereo, like atrop bond stereo, to a bond whose order became double. This happened in our carbonyl-wedge situation. Then you'd have an invalid bond -- double yet atrop. Now we only preserve within the limited allowed bond stereo for a double bond, and clear it otherwise.

#### Any other comments?
Bisected and implemented with GPT 5.4, from MWE identified manually.
cc @ricrogz

### DIFF
--- a/Code/GraphMol/Canon.cpp
+++ b/Code/GraphMol/Canon.cpp
@@ -29,7 +29,13 @@
 namespace RDKit {
 namespace Canon {
 namespace {
-constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
+static bool isCanonicalizableStereoDoubleBond(const Bond &bond) {
+  const auto stereo = bond.getStereo();
+  return bond.getBondType() == Bond::DOUBLE && stereo >= Bond::STEREOZ &&
+         stereo <= Bond::STEREOTRANS && bond.getStereoAtoms().size() >= 2;
+}
+
+static constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
   switch (bondDir) {
     case Bond::ENDUPRIGHT:
       return Bond::ENDDOWNRIGHT;
@@ -724,8 +730,7 @@ void canonicalizeDoubleBonds(ROMol &mol, const UINT_VECT &bondVisitOrders,
                                          const Bond *nbrBnd) -> Bond * {
     auto otherAtom = nbrBnd->getOtherAtom(dblBndAtom);
     for (const auto bond : mol.atomBonds(otherAtom)) {
-      if (bond != nbrBnd && bond->getBondType() == Bond::DOUBLE &&
-          bond->getStereo() > Bond::STEREOANY) {
+      if (bond != nbrBnd && isCanonicalizableStereoDoubleBond(*bond)) {
         return bond;
       }
     }
@@ -765,11 +770,10 @@ void canonicalizeDoubleBonds(ROMol &mol, const UINT_VECT &bondVisitOrders,
       bond->setBondDir(Bond::NONE);
     }
 
-    if (bond->getBondType() != Bond::DOUBLE ||
-        bond->getStereo() <= Bond::STEREOANY ||
-        bond->getStereoAtoms().size() < 2) {
+    if (!isCanonicalizableStereoDoubleBond(*bond)) {
       // not a bond that can have stereo or that needs canonicalization
       bond->setStereo(Bond::STEREONONE);
+      bond->getStereoAtoms().clear();
       continue;
     }
 
@@ -817,6 +821,12 @@ void canonicalizeDoubleBonds(ROMol &mol, const UINT_VECT &bondVisitOrders,
       const auto currentBond = connectedBondsQ.front();
       connectedBondsQ.pop();
       if (seen_bonds[currentBond->getIdx()]) {
+        continue;
+      }
+      if (!isCanonicalizableStereoDoubleBond(*currentBond)) {
+        currentBond->setStereo(Bond::STEREONONE);
+        currentBond->getStereoAtoms().clear();
+        seen_bonds[currentBond->getIdx()] = true;
         continue;
       }
 
--- a/Code/GraphMol/Canon.cpp
+++ b/Code/GraphMol/Canon.cpp
@@ -32,7 +32,7 @@ namespace {
 static bool isCanonicalizableStereoDoubleBond(const Bond &bond) {
   const auto stereo = bond.getStereo();
   return bond.getBondType() == Bond::DOUBLE && stereo >= Bond::STEREOZ &&
-         stereo <= Bond::STEREOTRANS && bond.getStereoAtoms().size() >= 2;
+         stereo <= Bond::STEREOTRANS && bond.getStereoAtoms().size() == 2;
 }
 
 static constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
--- a/Code/GraphMol/MolStandardize/Tautomer.cpp
+++ b/Code/GraphMol/MolStandardize/Tautomer.cpp
@@ -347,9 +347,8 @@ const RingInfo *getFastRingInfo(ROMol &mol) {
   // SymmSSSR here since that has a performance cost.
   // this might be not precise enough for large rings like 9-or-more macrocycles
   auto ringInfo = mol.getRingInfo();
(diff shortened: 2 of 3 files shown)


Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -341,6 +341,40 @@ int scoreHeteroHs(const ROMol &mol) {
...
+const RingInfo *getFastRingInfo(ROMol &mol) {
+  auto ringInfo = mol.getRingInfo();
+  if (!ringInfo || !ringInfo->isFindFastOrBetter()) {
+    MolOps::fastFindRings(mol);
+    ringInfo = mol.getRingInfo();
+  }
+  return ringInfo;
+}
+
+bool isTautomerDoubleBondInRing(const RingInfo *ringInfo, const Bond &bond) {
```
ricrogz: This doesn't check whether the bond is a double or not, so I wouldn't add that to the name of the function.

Also, it doesn't just return ring bonds, but it may also return bonds that aren't necessarily part of a ring but join two rings  (`beginInRing && endInRing`)

Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -426,39 +460,18 @@ bool TautomerEnumerator::setTautomerStereoAndIsoHs(
       }
     }
     auto tautBond = tautBonds[bondIdx];
-    if (tautBond->getBondType() != Bond::DOUBLE || d_removeBondStereo) {
+    if (tautBond->getBondType() != Bond::DOUBLE || d_removeBondStereo ||
+        !hasValidSpecifiedDoubleBondStereo(*bond)) {
       // When bond stereo is being removed for bonds involved in tautomerism,
       // use STEREOANY (for *non-ring* double bonds) instead of STEREONONE.
       // This prevents downstream tools (notably InChI) from inferring a specific E/Z
       // assignment from 2D coordinates after bond orders have been changed.
```
ricrogz, line 472: Please update this comment: as mentioned above, we also consider bonds between rings which aren't actually in a ring as "ring bonds".

Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -426,39 +460,18 @@ bool TautomerEnumerator::setTautomerStereoAndIsoHs(
...
-                                 !isRingBond)
-                                    ? Bond::STEREOANY
-                                    : Bond::STEREONONE;
-      modified |= (tautBond->getStereo() != targetStereo);
+      const RingInfo *ringInfo =
+          tautBond->getBondType() == Bond::DOUBLE ? getFastRingInfo(taut)
+                                                  : nullptr;
+      const auto targetStereo = getClearedTautomerBondStereo(ringInfo, *tautBond);
+      modified |= (tautBond->getStereo() != targetStereo ||
+                   !tautBond->getStereoAtoms().empty());
```
ricrogz: I'm not sure this condition makes sense. Stereo atoms are only relevant in bonds do have stereo (so removing them from STEREOANY/STEREONONE bonds isn't really a relevant change), and if any of those is changed into STEREOANY/STEREONONE, it will get caught by the other condition.

Code/GraphMol/Canon.cpp
```diff
@@ -29,7 +29,13 @@
 namespace RDKit {
 namespace Canon {
 namespace {
-constexpr Bond::BondDir flipStereoBondDir(Bond::BondDir bondDir) {
+static bool isCanonicalizableStereoDoubleBond(const Bond &bond) {
+  const auto stereo = bond.getStereo();
+  return bond.getBondType() == Bond::DOUBLE && stereo >= Bond::STEREOZ &&
+         stereo <= Bond::STEREOTRANS && bond.getStereoAtoms().size() >= 2;
```
ricrogz: Hm, I don't think `bond.getStereoAtoms().size() > 2` can actually happen...

ricrogz (approved):
A few nits below, but looks good otherwise.

Code/GraphMol/Canon.cpp
pechersky: It could happen if somehow a molecule was edited/malformed to begin with -- but no, we don't have direct code that sets above 2 right now. Will edit.

Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -341,6 +341,44 @@ int scoreHeteroHs(const ROMol &mol) {
...
 
+namespace {
+const RingInfo *getFastRingInfo(ROMol &mol) {
+  // We only need to know whether the bond is in a ring; avoid forcing
+  // SymmSSSR here since that has a performance cost.
+  // this might be not precise enough for large rings like 9-or-more macrocycles
+  auto ringInfo = mol.getRingInfo();
+  if (!ringInfo || !ringInfo->isFindFastOrBetter()) {
+    MolOps::fastFindRings(mol);
+    ringInfo = mol.getRingInfo();
```
greglandrum: This call isn't necessary: `ringInfo` is a pointer to the molecule's ringInfo

Code/GraphMol/MolStandardize/Tautomer.cpp
```diff
@@ -341,6 +341,44 @@ int scoreHeteroHs(const ROMol &mol) {
 }
 }  // namespace TautomerScoringFunctions
 
+namespace {
+const RingInfo *getFastRingInfo(ROMol &mol) {
+  // We only need to know whether the bond is in a ring; avoid forcing
+  // SymmSSSR here since that has a performance cost.
+  // this might be not precise enough for large rings like 9-or-more macrocycles
+  auto ringInfo = mol.getRingInfo();
+  if (!ringInfo || !ringInfo->isFindFastOrBetter()) {
```
greglandrum: The molecule should always have a valid ringInfo pointer, so you don't need the first part of this:

```suggestion
  if (!ringInfo->isFindFastOrBetter()) {
```

greglandrum (changes requested):
Two small suggestions

greglandrum (approved):
LGTM

PR #9342, opened 2026-06-10
Add opt-in unity (jumbo) builds + fix latent missing include guards

skearnes:
## Summary

Adds an **opt-in** `RDK_USE_UNITY_BUILDS` CMake option (default **OFF**) that builds the RDKit libraries using CMake [unity (jumbo) builds](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html): each library's sources are batched into a few translation units so the shared headers are parsed once per batch instead of once per file. This is the GCC-friendly complement to the precompiled-headers work in #9236 — it reduces real parse work for every compiler (it doesn't rely on a PCH), and helps incremental rebuilds too.

Along the way it fixes a handful of **latent bugs** that unity builds expose (headers with no include guard; a leaking file-scope constant; a linkage mismatch that risks an ODR violation; a test that `#include`s a `.cpp`).

### Measured win
Rebuilding the `GraphMol` library (Apple clang, 8 cores): **~14.5s → ~6.6s** (36 translation units collapse to 5). The win applies to every library not excluded.

## What's here

**1. Missing include guards (5 headers)** — latent everywhere, exposed by unity builds (multiple `.cpp` in one batch each include the header → redefinition):
`GraphMol/new_canon.h`, `Numerics/Optimizer/BFGSOpt.h`, `GraphMol/FileParsers/MaestroProperties.h`, `GraphMol/Depictor/Templates.h`, `GraphMol/MolEnumerator/LinkNode.h`. Correctness fixes independent of unity — a header should be idempotent.

  A 6th candidate, `GraphMol/Depictor/TemplateSmarts.h`, is intentionally left alone: it's automatically generated and imported from [rdkit/molecular_templates](https://github.com/rdkit/molecular_templates), so a guard added here would be overwritten on the next template update ([thanks @ricrogz](https://github.com/rdkit/rdkit/pull/9342#discussion_r3388745929)). Upstream is adding a `#pragma once` to the generator instead, and it's a non-issue for unity anyway since it's only ever included via `Templates.h`, which already has its own guard.

**2. `UniformGrid3D.cpp` tolerance constants** — `OFFSET_TOL`/`SPACING_TOL` were at namespace scope (value `1e-8`) while `UniformRealValueGrid3D.cpp` has same-named anonymous-namespace constants with a *different* value (`1e-4`); they become ambiguous when batched. Made file-local and uniquely named.

**3. `FMCS.cpp` `ringFusionCheck` linkage** — the helper was marked `inline` (external linkage), but its lambda captures `RingBondCountVect` locals, and `RingBondCountVect` lives in the file's anonymous namespace (internal linkage). An external-linkage closure type with an internal-linkage member is an ODR hazard that GCC rejects under `-Werror=subobject-linkage`. The helper is only called within the same TU, so the `inline` was unnecessary — moved it into the anonymous namespace alongside `RingBondCountVect`. (Caught by the Linux/GCC CI job below.)

**4. Vestigial `#include <AddHs.cpp>` in the DetermineBonds test** — the test only uses the public `MolOps::removeAllHs` (via headers + the linked library) and none of `AddHs.cpp`'s file-local helpers, but pulled in the whole `.cpp`, duplicating its symbols at link time under unity. Removed the stale include.

**5. The `RDK_USE_UNITY_BUILDS` option**, applied via a `rdkit_maybe_enable_unity_build()` helper in `rdkit_library`. Skipped for:
  - **Vendored third-party code under `External/`** — excluded automatically by path (we don't reformat/batch upstream code, and refreshes could introduce collisions).
  - **Internal libraries listed in `RDK_UNITY_BUILD_EXCLUDED_LIBRARIES`** that still have file-local symbol collisions to resolve.

## Excluded internal libraries (follow-up work)

These have file-local symbols (`moldata3D`, `RAD2DEG`, `roundn`, `getEigenVect`, …) with colliding names across files in the same library — each a mechanical fix (unique names / anonymous-namespace scope) that can shrink the list incrementally:

`CIPLabeler`, `ChemTransforms`, `Descriptors`, `FileParsers`, `ForceField`, `ForceFieldHelpers`, `MolDraw2D`, `MolStandardize`, `RGroupDecomposition`, `SmilesParse`, `hc`.

## Verification
- `RDK_USE_UNITY_BUILDS=ON` full build (default features): **clean (0 errors)**.
- **Off-default libraries tested**: rebuilt with `RDK_BUILD_INCHI_SUPPORT`, `RDK_BUILD_XYZ2MOL_SUPPORT`, `RDK_BUILD_STRUCTCHECKER_SUPPORT` enabled — the off-default internal libraries (DetermineBonds, StructChecker, MolProcessing) are **unity-clean**, and `External/` InChI is correctly auto-excluded. So the exclude list isn't silently incomplete for these configs. (MinimalLib/CFFI couldn't be exercised here — they need emscripten / a static-libs config — but their sources compiled clean under unity.)
- Default (`OFF`) is unchanged behavior; the include-guard, Geometry, and test fixes also build fine non-unity.

**CI coverage:** flips `-DRDK_USE_UNITY_BUILDS=ON` on the existing `Ubuntu_x64` job (Linux/GCC, full feature set, `-Werror`), mirroring how #9236 enabled precompiled headers via a flag on existing jobs rather than a dedicated one. This keeps the option and its exclude list from silently rotting as new sources are added, and exercises GCC (where the PCH path is disabled), with no duplicated job to maintain. The default (non-unity) build stays covered by the other Linux jobs (py312, cartridge, java, limitexternal, minimallib).

My local validation was clang without `-Werror`, so this GCC CI job is doing exactly its job: its first run surfaced the `FMCS` `-Wsubobject-linkage` issue above (clang doesn't diagnose it), now fixed. Further runs may turn up another such case to fix or one more library to exclude — that's the point of the job.

## Note: expected docs conflict with #9353

This PR and #9353 are independent branches off `master`, and both add a "Speeding up the build" section to `Docs/Book/Install.md` at the same spot (one documenting `RDK_USE_UNITY_BUILDS`, the other `RDK_USE_PRECOMPILED_HEADERS`; both share the same `(s)ccache` bullet). Whichever merges second will hit a small add/add conflict there — resolve by keeping both option bullets and a single copy of the shared intro + ccache bullet.


Code/GraphMol/Depictor/TemplateSmarts.h
```diff
@@ -14,6 +14,9 @@
 //
 //
 
+#ifndef RD_DEPICTOR_TEMPLATESMARTS_H
+#define RD_DEPICTOR_TEMPLATESMARTS_H
```
ricrogz: This file is automatically generated and imported from https://github.com/rdkit/molecular_templates, so this change would be overwritten in the next update.
I just added a PR there to add a `#pragma once` (no real reason, but I prefer that to `#ifndef`/`#define`) at the top when this file is generated.

skearnes: Good catch, thanks — reverted in fb4a4e037. I'd missed the "automatically generated" banner. It's moot for unity builds anyway since `TemplateSmarts.h` is only ever included via `Templates.h`, which has its own guard, so it can't be processed twice in one TU. I'll let the upstream `#pragma once` handle it.

greglandrum (approved):
LGTM

greglandrum:
@ricrogz are you 👍 on this one now?

ricrogz (approved):
LGTM 👍

ricrogz:
> @ricrogz are you 👍 on this one now?

Yes, looks good to me.

PR #9344, opened 2026-06-10
Patch GetProp default typing in rdchem stubs

### LINKED ISSUE #9335: `rdchem.pyi` typing issue: non-default parameter following default parameter
**Describe the bug**
There is a typing issue in `rdchem.pyi` line 347 due to a non-default parameter following a default parameter, which is invalid in python. This causes `mypy` to error on it too.

**To Reproduce**
No real reproduction steps, its just invalid python.

Found this issue by running `mypy` over a codebase that uses `rdkit`

**Expected behavior**
No typing issue. Likely the typing should have been:

```
    def GetProp(self, key: str, autoConvert: bool = False, default: typing.Any = None) -> typing.Any:
```

**Screenshots**
Image

Mypy output:
```
.nox/typecheck/lib/python3.14/site-packages/rdkit-stubs/Chem/rdchem.pyi:347: error: Parameter without a default follows parameter with a default  [syntax]
```

 - RDKit version: 2026.3.3
 - OS: N/A
 - Are you using conda? No
 - If you are not using conda: uv

**Additional context**
None

pechersky:
#### Reference Issue
Fixes #9335

include `default = ...` in typing patch so that downstream checker doesn't complain about non-default kwarg after default kwarg. Also tighten the chunks to not have to carry as much patch context.

### DIFF
--- a/Scripts/gen_rdkit_stubs/patch/rdkit/Chem/rdchem.pyi.diff
+++ b/Scripts/gen_rdkit_stubs/patch/rdkit/Chem/rdchem.pyi.diff
@@ -1,20 +1,37 @@
 --- a/rdkit/Chem/rdchem.pyi
 +++ b/rdkit/Chem/rdchem.pyi
-@@ -1988,7 +1988,7 @@
-             C++ signature :
-                 RDKit::Atom* GetAtomWithIdx(RDKit::ROMol {lvalue},unsigned int)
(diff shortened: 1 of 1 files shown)


ijpulidos:
Related to #9346 FYI

greglandrum (approved):
LGTM

PR #9346, opened 2026-06-11
Typing fix: Reorder arguments to `GetPyPropOrDefault<T>` bindings

### LINKED ISSUE #9335: `rdchem.pyi` typing issue: non-default parameter following default parameter

ianmkenney:
The generated Python stubs represent invalid Python syntax.

#### Reference Issue

Fixes #9335 

~~This PR sets the `GetProp` python "default" argument to python::object(), which _should_ represent `None` and is compatible with `GetPyPropOrDefault`.~~

This PR switches the positions of `default` and `autoconvert` defined in the Python wrappers for `GetPyPropOrDefault<T>` templates. This should preserve still support overloading.

### DIFF
--- a/Code/GraphMol/Wrap/Atom.cpp
+++ b/Code/GraphMol/Wrap/Atom.cpp
@@ -348,7 +348,7 @@ struct atom_wrapper {
             "GetProp", GetPyPropOrDefault<Atom>,
             (python::arg("self"), python::arg("key"),
              python::arg("autoConvert") = false,
-             python::arg("default")),
+             python::arg("default") = python::object()),
             "Returns the value of the property.\n\n"
             "  ARGUMENTS:\n"
             "    - key: the name of the property to return (a string).\n\n"
--- a/Code/GraphMol/Wrap/Atom.cpp
+++ b/Code/GraphMol/Wrap/Atom.cpp
@@ -347,13 +347,13 @@ struct atom_wrapper {
         .def(
             "GetProp", GetPyPropOrDefault<Atom>,
             (python::arg("self"), python::arg("key"),
-             python::arg("autoConvert") = false,
-             python::arg("default") = python::object()),
+             python::arg("default"),
+             python::arg("autoConvert") = false),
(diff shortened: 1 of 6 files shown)


ianmkenney:
This seems to have corrected the first instance of this bug (`rdchem.pyi:347`), but there are other methods that have the same problem, addressing those now. `mypy` fully stops checking each time is encounters the bad signature.

ianmkenney:
The generated and installed stubs no longer cause `mypy` to crash when checking a module importing `rdkit`.

greglandrum:
@ianmkenney we thought about making a similar change when we added the default argument. Unfortunately it would silently break existing code because it would change the meaning of `foo.GetProp('bar',True)`. So we had to do things this way. If you're interested, there's a longish discussion in #9242 

In any case, I think the underlying problem in the .pyi files was addressed by #9344.
Unless you disagree, we can close this.

ianmkenney:
@greglandrum Yeah, I don't immediately see a clean way to fix this at this layer without breaking the API. As soon as I saw the overloaded methods, I figured it was going to get tricky in Python-land. Thank you for taking a look and providing some context!

Closing.

PR #9349, opened 2026-06-11
Make substructure search Ctrl+c interruptible

ricrogz:
Substructure searching is another of those calculations that can run for a long, long time if the query being searched for has the right size and symmetry conditions. And it currently doesn't respond to attempts to interrupt it via Ctrl+c .

This patch allows it to be interrupted in the same way we do with other potentially long running calculations like CIP labeling or embedding.

### DIFF
--- a/Code/GraphMol/Substruct/catch_tests.cpp
+++ b/Code/GraphMol/Substruct/catch_tests.cpp
@@ -15,6 +15,12 @@
 #include <tuple>
 #include <utility>
 
+#ifdef RDK_TEST_MULTITHREADED
+#include <csignal>
+#include <thread>
+#include <chrono>
+#endif
+
 #include <GraphMol/RDKitBase.h>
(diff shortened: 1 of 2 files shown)


Code/GraphMol/Substruct/catch_tests.cpp
```diff
@@ -1122,4 +1128,119 @@ TEST_CASE("quick return when the query has more atoms than the molecule") {
...
+M  V30 34 1 35 30
+M  V30 35 1 36 32
+M  V30 36 1 37 32
+M  V30 37 1 38 32
+M  V30 END BOND
+M  V30 END CTAB
+M  END
+)";
+
+  v2::FileParsers::MolFileParserParams params{.sanitize = false};
```
greglandrum, line 1224: I need to remember to use this pattern...

greglandrum (approved):
That required surprisingly little code.
Nice!

LGTM

PR #9353, opened 2026-06-12
Add opt-in precompiled-header support (reworked #9236)

skearnes:
## Summary

Adds an **opt-in** `RDK_USE_PRECOMPILED_HEADERS` CMake option (default **OFF**) that precompiles `Code/stdafx.h` — a bundle of the STL/boost headers that dominate parse time plus the core RDKit headers pulled in by nearly every TU — via `target_precompile_headers` on `rdkit_base`.

The PCH is **scoped to library targets**. The `INTERFACE` PCH on `rdkit_base` would otherwise propagate to every consumer (~410 targets: ~80 libraries, ~200 C++ tests, ~65 Python extensions, plus a few executables), and MSVC writes a separate multi-hundred-MB PCH per target — that exhausts the Windows CI disk and collides with `WINDOWS_EXPORT_ALL_SYMBOLS`. Libraries are where the parse time is, so the PCH stays there and everything else opts out (see exclusions below).

Enabled in CI for the **macOS and Windows (VS + DLL)** jobs; left **off for GCC/Linux**, where PCH regresses (GCC's PCH is a whole-heap memory image with no lazy deserialization, so it thrashes under high `-j`). This is the precompiled-header counterpart to the unity-build work in #9342.

## What gets the PCH, and what opts out

**Libraries** (`rdkit_library`) get the PCH. The following opt out:

- **C libraries** — `hc`, `avalon_clib`, `freesasa_clib`, `Inchi`, `RingDecomposerLib` (via a `NO_PCH` argument to `rdkit_library`) and `yaehmop_eht` (via `DISABLE_PRECOMPILE_HEADERS`). A C TU can't consume the C++ `stdafx.h` (`'algorithm' file not found`).
- **Tests** (`rdkit_test`, `rdkit_catch_test`) and **Python extensions** (`rdkit_python_extension`) — disabled in the macros. Hundreds of small targets, little benefit, and the source of the Windows disk blow-up.
- **`RDKit_minimal`, `rdkitcffi`, `cffi_test`** — executables / an export-all shared lib (details below).
- **Five library TUs** keep the PCH for the rest of their library but skip it themselves via `SKIP_PRECOMPILE_HEADERS`: the four `#define USE_BETTER_ENUMS` JSON parsers (`SmilesJSONParsers.cpp`, `MolFragmenterJSONParser.cpp`, `MolDrawOptionsJSONParsers.cpp`, `RGroupDecompJSONParsers.cpp`) and `AtomIterators.cpp` (exported explicit template instantiations).

## Notable interactions resolved

- **`USE_BETTER_ENUMS` + PCH.** `BetterEnums.h` is intentionally re-includable (no include guard) so `RDKitBase.h` can be baked into the PCH in plain-enum mode while a TU that later `#define`s `USE_BETTER_ENUMS` still upgrades to the full implementation; the undef/re-include is gated on `enum.h`'s own guard (`BETTER_ENUMS_ENUM_H`) to avoid a macro-clobber that previously left `BETTER_ENUM` undefined. **However**, re-includability only un-pins `BetterEnums.h` itself — the headers that *declare* the enums (e.g. `MolOps.h`'s `SanitizeFlags`) are `#pragma once`, so once baked into the plain-enum PCH their enums stay plain. TUs that need the real better-enums therefore skip the PCH (the four JSON parsers above). This surfaced on macOS as `JSONParsers.cpp: no matching function for call to 'flagsFromJson'` (`no type named '_integral' in 'MolOps::SanitizeFlags'`).
- **Eigen `make_unsigned` ambiguity (MSVC).** `MolDraw2D.cpp` pulls in Eigen, whose internals use unqualified `make_unsigned`. With the PCH pre-including boost's `type_traits`, the file's global `using namespace boost` made the name ambiguous with `std::make_unsigned` under MSVC's instantiation-time lookup (`error C2872`). Removed the unused `using namespace boost`.
- **Exported explicit template instantiation in a DLL.** `AtomIterators.cpp` does `template class RDKIT_GRAPHMOL_EXPORT AtomIterator_<...>`. The PCH is built on `rdkit_base` without `RDKIT_GRAPHMOL_BUILD`, so it bakes the header in as `dllimport`; MSVC then omits the instantiations from `GraphMol.dll` (ordinary members of `dllexport` classes are still exported, which is why nothing else broke), giving `LNK2019` in consumers. Fixed by compiling that TU without the PCH.
- **`WINDOWS_EXPORT_ALL_SYMBOLS` + PCH.** `rdkitcffi` statically bundles other libraries whose `cmake_pch.obj` carry `_PchSym_` symbols, corrupting the auto-generated export `.def` (`LNK4022`/`LNK2001` unresolved external symbol `'__'`). Excluded from the PCH.

## Relationship to #9236

This is a reworked version of @greglandrum's #9236, opened so CI can run on a cleaned-up, rebased branch — happy to fold it back however Greg prefers. Notable differences from the original:

- **Split out** the unrelated `-Wdeprecated-copy` fixes (`Seed`, `SubstructTerm`) into #9352, so the PCH change stays purely additive.
- **Dropped** the `SmilesJSONParsers.h` `#pragma once` → include-guard conversion — `#pragma once` is already PCH-safe and the change ran counter to RDKit's direction.
- **Scoped the PCH to libraries** rather than letting the `INTERFACE` property reach every target (the original would not build on Windows CI — disk exhaustion + the link issues above).
- **Documented** build-speed options in `Docs/Book/Install.md` (a new "Speeding up the build" section covering `RDK_USE_PRECOMPILED_HEADERS` and `(s)ccache`).

## Verification

- CI is **green with PCH enabled** on **macOS_x64**, **Windows_VS2022_x64**, **Windows_VS2022_x64_dll**, and **Windows_VS2022_x64_swig** (and on the Linux/Java/cartridge/minimallib jobs, where PCH is off).
- Default (`OFF`) is unchanged behavior.

> ℹ️ **Known-unrelated CI red:** the `Ubuntu_x64` job fails at cmake-configure with `Qt5Gui ... Failed to find "GL/gl.h"`. This is a CI-image flake (PCH is **off** on that job; the same configure error hits unrelated PRs, e.g. #9366), not a change in this branch — it just needs that one job re-run.

Refs #9236.


pechersky:
How about suggesting to users to use `(s)ccache`?

skearnes:
Good call — added a "Speeding up the build" section to `Docs/Book/Install.md` (commit a3ae01c84) that recommends `ccache`/`sccache` alongside `RDK_USE_PRECOMPILED_HEADERS`, with the `CMAKE_C(XX)_COMPILER_LAUNCHER` invocation and a note that PCH and the compiler cache interact (`ccache`'s `pch_defines`/`time_macros` sloppiness).

I left ccache out of the CI builds themselves: the Microsoft-hosted agents are ephemeral, so it'd need a persistent pipeline cache to help at all, and enabling it in the same PR that introduces PCH would muddy the build signal. Better as a separate infra PR if we want it — and it'd benefit all jobs, not just the PCH ones.

greglandrum:
> 
> I left ccache out of the CI builds themselves: the Microsoft-hosted agents are ephemeral, so it'd need a persistent pipeline cache to help at all,

Yeah @whosayn had looked at ccache and CI before and encountered this problem. It's too bad, because this would *really* help our CI go faster.

pechersky:
For Azure -- how about using a Cache task? https://learn.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops&tabs=bundler#cache-task-how-it-works

Our github rdkit builds utilize https://github.com/mozilla-actions/sccache-action

So for Azure (proposed by Gemini):
```
variables:
  # Route standard C/C++ compilers through the sccache executable wrapper
  CC: sccache gcc
  CXX: sccache g++
  # Local folder where sccache writes its compiled .o / .obj object fragments
  SCCACHE_DIR: $(Pipeline.Workspace)/.sccache

steps:
  # 1. Fetch matching compiled object files from prior runs
  - task: Cache@2
    displayName: 'Fetch sccache C++ artifacts'
    inputs:
      # Key changes if your build configuration or agent OS changes
      key: 'sccache-cpp | "$(Agent.OS)" | **/CMakeLists.txt' 
      path: $(SCCACHE_DIR)
      restoreKeys: |
        sccache-cpp | "$(Agent.OS)"

  # 2. Download and install sccache binary
  - bash: |
      if ! command -v sccache &> /dev/null; then
        echo "Installing sccache..."
        curl -sSL https://github.com | tar xz
        sudo mv sccache-v0.8.2-x86_64-unknown-linux-musl/sccache /usr/local/bin/
        chmod +x /usr/local/bin/sccache
      fi
      
      # Start background daemon to intercept compilation traffic
      sccache --start-server
    displayName: 'Install and start sccache'

  # 3. Configure and run your C++ build system
  - bash: |
      mkdir build && cd build
      cmake .. -DCMAKE_BUILD_TYPE=Release
      cmake --build . --parallel $(nproc)
    displayName: 'Compile C++ Project'

  # 4. Print optimization hit/miss metrics and safely shut down the daemon
  - bash: |
      echo "=== SCCACHE HIT/MISS STATS ==="
      sccache --show-stats
      sccache --stop-server
    displayName: 'Show C++ cache performance'
    condition: always() # Ensures stats display and server closes even if compilation fails
```

skearnes:
@pechersky ok if we explore caching in a separate PR?

skearnes:
Merged current `master` and worked through the failures that enabling PCH surfaced on CI. Summary of the latest set of changes (commits `b3baeff7…889c4668`):

**Scope the PCH to library targets** (`889c4668`) — the headline change. The `INTERFACE` PCH on `rdkit_base` was reaching all ~410 targets; MSVC writes a separate multi-hundred-MB PCH per target, which exhausted the Windows CI disk (`C1088 … cmake_pch.pch: No space left on device` across dozens of tests). Now `rdkit_test`, `rdkit_catch_test`, and `rdkit_python_extension` opt out in the macros, and `RDKit_minimal` / `rdkitcffi` / `cffi_test` opt out directly. This also fixes:
- `rdkitcffi` — `WINDOWS_EXPORT_ALL_SYMBOLS` + statically-bundled libs pulled `_PchSym_` symbols into the auto-generated export `.def` → `LNK4022`/`LNK2001` unresolved `'__'`.
- `cffi_test` — a C executable can't precompile the C++ `stdafx.h` (`'algorithm' file not found`, the macOS failure).

**Per-platform fixes uncovered along the way:**
- `USE_BETTER_ENUMS` TUs vs. the plain-enum PCH (`b3baeff7`) — `MolOps.h`'s `SanitizeFlags` is `#pragma once`, so once baked into the PCH it stays a plain enum and `flagsFromJson<SanitizeFlags>` failed to compile on macOS. The four library JSON parsers now `SKIP_PRECOMPILE_HEADERS`.
- `MolDraw2D.cpp` (`85b6cce5`) — dropped an unused global `using namespace boost` that made Eigen's unqualified `make_unsigned` ambiguous with boost's under MSVC once the PCH pre-included boost `type_traits` (`C2872`).
- `AtomIterators.cpp` (`5ff96ae2`) — the exported explicit class-template instantiations were suppressed to `dllimport` (the PCH is built without `RDKIT_GRAPHMOL_BUILD`), so they were missing from `GraphMol.dll` (`LNK2019`). Compiled without the PCH.

Plus the `master` merge (kept both the PCH and unity-build `Install.md` sections) and some comment cleanup.

**Status:** the previous run was green with PCH on across **macOS_x64**, **Windows_VS2022_x64**, **Windows_VS2022_x64_dll**, and **Windows_VS2022_x64_swig**. Marking the PR ready re-triggered the pipeline, so CI is running again now.

The one remaining red is `Ubuntu_x64`, which fails at cmake-configure with `Qt5Gui … Failed to find "GL/gl.h"` — a CI-image flake (PCH is **off** on that job; the same configure error hits unrelated PRs, e.g. #9366). It just needs that job re-run.

PR #9361, opened 2026-06-15
Check OOB causing OSV-2022-1288

thomp-j:
#### Reference Issue
https://issues.oss-fuzz.com/issues/376787368

The "first" iterator ended up past the "last" iterator, causing memory issues. This checks for that case before accessing "first" to avoid the error.


greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @thomp-j!

PR #9364, opened 2026-06-16
Bugfix/rgd multicore targetmol crash

### LINKED ISSUE #9363: rdRGroupDecomposition: error when using both allowMultipleCoresInSameMol and includeTargetMolInResults
**Describe the bug**

When both `allowMultipleCoresInSameMol=True` and `includeTargetMolInResults=True` are used in `rdRGroupDecomposition`, decomposition fails with a runtime `Range Error` during result generation.

Each option works independently:

- `allowMultipleCoresInSameMol=True`, `includeTargetMolInResults=False` works
- `allowMultipleCoresInSameMol=False`, `includeTargetMolInResults=True` works

The crash appears only when both are enabled together.

**To Reproduce**

```python
from rdkit import Chem
from rdkit.Chem import rdRGroupDecomposition as rdRGD

smiles_list = ["c1ccc(C)c(N)c1", "c1ccc(F)c(O)c1", "c1ccc(c2ccccc2)c(N)c1"]
mols = [Chem.MolFromSmiles(s) for s in smiles_list]
scaffold = Chem.MolFromSmarts("[*:1]c1c([*:2])cccc1")

options = rdRGD.RGroupDecompositionParameters()
options.allowMultipleCoresInSameMol = True
options.includeTargetMolInResults = True

decomp = rdRGD.RGroupDecomposition(scaffold, options)
for mol in mols:
    decomp.Add(mol)
decomp.Process()
decomp.GetRGroupsAsColumns(asSmiles=True)
```

**Expected behavior**

Decomposition should succeed and return `Core`, `Mol`, and `R*` outputs without error.

**Observed behavior**

`RuntimeError: Range Error` is thrown from `Code/GraphMol/ROMol.cpp` during result generation when both flags are set.

**Screenshots**

- RDKit version: `2026.09.1pre` (also reproduced with `2026.03.2`)
- OS: macOS (arm64)
- Python version: 3.12
- Are you using conda? Yes
- If you are using conda, which channel did you install the rdkit from? conda-forge (also reproduced in a local source build)
- If you are not using conda: N/A

**Additional context**

Likely cause is shared mutable target-molecule highlight state across multiple matches when a single molecule produces multiple core hits. This can lead to stale index remapping and out-of-range access.

jonswain:
#### Reference Issue

Fixes #9363

#### What does this implement/fix? Explain your changes,

This PR fixes a crash in RGroup decomposition when both of the following options are enabled together:

- `includeTargetMolInResults = true`
- `allowMultipleCoresInSameMol = true`

Root cause:

- The target molecule used for highlight bookkeeping was shared across multiple decomposition matches from the same input molecule.
- That shared mutable state caused invalid/stale atom and bond index remapping during highlight processing and could trigger a `Range Error`.

Code change:

- Updated RGroup decomposition match construction so each match stores its own copy of the target molecule for highlighting instead of sharing one mutable instance.
- This isolates per-match highlight trimming/index remapping and prevents cross-row aliasing side effects.

Tests added:

- C++ regression test in `Code/GraphMol/RGroupDecomposition/catch_rgd.cpp`
- Python wrapper regression test in `Code/GraphMol/RGroupDecomposition/Wrap/test_rgroups.py`
- Both tests cover the combined-options scenario and assert successful decomposition plus consistent `Core` and `Mol` output sizing.

#### Any other comments?

Validation run locally:

- `testRGroupDecomp`
- `testRGroupDecompInternals`
- `rgroupCatchTests`
- `pyRGroupDecomposition`

AI usage acknowledgment:

- I used GitHub Copilot (GPT-5.3-Codex) to assist with implementation drafting and test authoring, then reviewed and validated the final changes and behavior locally before preparing this PR.

### DIFF
--- a/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
+++ b/Code/GraphMol/RGroupDecomposition/RGroupDecomp.cpp
@@ -586,7 +586,8 @@ int RGroupDecomposition::add(const ROMol &inmol) {
           potentialMatches.emplace_back(core_idx, numberMissingUserGroups,
                                         match, extractedCore);
           if (data->params.includeTargetMolInResults) {
-            potentialMatches.back().setTargetMoleculeForHighlights(mol);
+            potentialMatches.back().setTargetMoleculeForHighlights(
+                RWMOL_SPTR(new RWMol(*mol)));
           }
         }
       }
--- a/Code/GraphMol/RGroupDecomposition/Wrap/test_rgroups.py
+++ b/Code/GraphMol/RGroupDecomposition/Wrap/test_rgroups.py
@@ -956,6 +956,37 @@ M  END
     for row in rows:
       checkRow(row)
 
+  def testIncludeTargetMolInResultsWithMultipleCoresInSameMol(self):
+    core = Chem.MolFromSmarts("[*:1]c1c([*:2])cccc1")
+    self.assertIsNotNone(core)
+    mols = [
+      Chem.MolFromSmiles(smi) for smi in [
+        "c1ccc(C)c(N)c1",
+        "c1ccc(F)c(O)c1",
+        "c1ccc(c2ccccc2)c(N)c1",
+      ]
+    ]
+    self.assertTrue(all(mols))
+    ps = RGroupDecompositionParameters()
+    ps.allowMultipleCoresInSameMol = True
+    ps.includeTargetMolInResults = True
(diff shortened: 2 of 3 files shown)


greglandrum (approved):
LGTM

greglandrum:
Thanks for the contribution @jonswain !

bp-kelley:
yes @jonswain this was a nice fix in a complicated bit of code.

PR #9365, opened 2026-06-16
Perf: Caches for faster reaction enumeration

jpfeuffer:
## Enumeration performance improvements: match caching, graft caching, and symmetric deduplication

This PR adds three opt-in performance features to the RDKit reaction library enumeration engine. All changes are backward-compatible — default behavior is unchanged — and the new options are controlled via `EnumerationParams`.

## Summary
- Reactant match cache (cacheMode = ReactantCacheMode::MatchOnly)
New in this PR. Previously, every Cartesian-product combination re-matched each reagent against its reaction template from scratch (EnumerateLibrary::next() called rxn.runReactants(reactants) with no caching). With this option, each (reagent, template) pair is matched once and the result is replayed across all combinations. The m_matchCache is owned by EnumerateLibrary but only consulted when caching is enabled.

- Graft/fragment cache (cacheMode = ReactantCacheMode::Full)
The reagent subgraph contributed to each product is extracted once per (reagent, match) pair and replayed for every downstream combination, eliminating redundant bond-by-bond assembly. Implies MatchOnly. ~2x speedup in isolation on assembly-dominated reactions.

- Symmetric match deduplication (dedupeSymmetricMatches = true)
Substructure matches that map onto symmetry-equivalent atom sets are collapsed before the Cartesian product, removing duplicate products for symmetric reagents and halving the work for fully symmetric inputs. Requires cacheMode >= MatchOnly.

### Defaults / compatibility
cacheMode defaults to ReactantCacheMode::None and dedupeSymmetricMatches defaults to false, so default behavior is byte-for-byte identical to the previous baseline.
All three features are opt-in via EnumerationParams and exposed in the Python wrapper.

### Benchmarks
On a realistic three-component library with complex SMARTS, combining all three features yields ~22x end-to-end speedup (24.0 s → 1.09 s).

### API

The two previous boolean flags (`cacheReactantGrafts`, and the always-on match cache) are replaced by a single `ReactantCacheMode` enum that makes the hierarchy explicit:

```
ReactantCacheMode.NoCache    — no caching (default for backwards-compat)
ReactantCacheMode.MatchOnly  — cache reactant-template matches
ReactantCacheMode.Full       — cache matches + per-reagent product grafts
```

`EnumerationParams.cacheMode` and `EnumerationParams.dedupeSymmetricMatches` (default `false`) are the two knobs. Both are serialized in a single new `EnumerateLibrary` serialization version (v0 → v1); old pickles at version 0 deserialize cleanly with defaults applied. I am not 1000% sure if the rank-based deduplication covers all cases for global symmetry also, but in simple cases like symmetric functional groups that are sketched weirdly for various reasons like an (NH of an NH2 group) it works wonders.

<details>
<summary>New accessors and types</summary>

- `EnumerateLibrary.GetCacheMode()` — returns current `ReactantCacheMode`
- `EnumerateLibrary.GetMatchCacheSize()` / `GetGraftCacheSize()` — cache introspection
- `ReactionRunnerUtils::ReactantMatchCache`, `ReactantGraftCache`, `dedupeMatchesBySymmetry`
- `run_Reactants` overloads wiring match cache only, or both caches + dedup flag
- Benchmark target `reaction_enumerate` in Bench

Graft caches are cleared on deserialization and preserved on copy. Match caches are always live.
</details>

<details>
<summary>Benchmark results (macOS M1, three-component library)</summary>

| Mode | Time | Speedup |
|---|---|---|
| No caching | 24.0 s | 1x |
| Dedup + MatchOnly | 2.6 s | ~9x |
| Dedup + Full | 1.09 s | **~22x** |

Correctness assertions (identical unique product sets) pass in all cases.
</details>

<details>
<summary>Test coverage</summary>

- Catch2 unit tests: cached vs. uncached output equality, cache key correctness, graft reuse across combinations, dedup on symmetric/asymmetric inputs, serialization round-trips, `ReactantCacheMode` enum values
- Python tests: defaults, flag propagation, cache sizes after enumeration, dedup collapses symmetric reagents, `GetCacheMode()` accessor
- Full CTest suite (testReaction, testReactionFingerprints, testEnumeration, rxnTestCatch) passing
</details>

<details>
<summary>Commits</summary>

- **feat: cache reactant template matches in run_Reactants**
- **feat: use reactant match cache in EnumerateLibrary**
- **erf: share reactant match cache with enumeration prefilter**
- **test: add reaction enumeration matching benchmark**
- **feat: add symmetric-match dedup helper for enumeration**
- **feat: apply symmetric-match dedup in cached run_Reactants**
- **feat: expose symmetric-match dedup on EnumerateLibrary**
- **test: add symmetric-match dedup enumeration benchmark**
- **test: add product-assembly golden oracle and baseline bench**
- **refactor: split product assembly into extract/apply graft**
- **feat: add opt-in reactant graft cache to run_Reactants**
- **feat: cache reactant grafts across EnumerateLibrary runs**
- **test: benchmark reactant graft cache on vs off**
- **test: add three-component symmetric-monomer benchmark**
- **feat: expose graft cache and symmetric dedup in Python**
- **api: collapse cacheModes into one param**

</details>

#### Reference Issue
https://github.com/rdkit/rdkit/discussions/7563

### DIFF
--- a/Code/GraphMol/ChemReactions/Enumerate/Enumerate.cpp
+++ b/Code/GraphMol/ChemReactions/Enumerate/Enumerate.cpp
@@ -150,13 +150,17 @@ BBS removeNonmatchingReagents(const ChemicalReaction &rxn, BBS bbs,
           cache && params.reagentMaxMatchCount == INT_MAX &&
           !hasProtectedAtoms(*mol);
       if (canPrimeCache) {
-        const auto cacheKey =
-            std::make_tuple(static_cast<unsigned int>(reactant_idx), mol.get(),
-                  1000u, false);
-        const VectMatchVectType reactantMatches =
+        VectMatchVectType reactantMatches =
             ReactionRunnerUtils::getReactantMatchesToTemplate(
                 *mol.get(), *reactantTemplate.get(), 1000u,
                 rxn.getSubstructParams());
+        if (params.dedupeSymmetricMatches) {
+          reactantMatches = ReactionRunnerUtils::dedupeMatchesBySymmetry(
+              *mol.get(), reactantMatches);
+        }
+        const auto cacheKey = std::make_tuple(
+            static_cast<unsigned int>(reactant_idx), mol.get(), 1000u,
+            params.dedupeSymmetricMatches);
         matches = reactantMatches.size();
         if (matches) {
           cache->emplace(cacheKey, reactantMatches);
@@ -213,6 +217,7 @@ BBS removeNonmatchingReagents(const ChemicalReaction &rxn, BBS bbs,
 EnumerateLibrary::EnumerateLibrary(const ChemicalReaction &rxn, const BBS &bbs,
                                    const EnumerationParams &params)
     : EnumerateLibraryBase(rxn, new CartesianProductStrategy),
+      m_dedupeSymmetricMatches(params.dedupeSymmetricMatches),
       m_matchCache(),
       m_bbs(removeNonmatchingReagents(m_rxn, bbs, params, &m_matchCache)) {
   m_enumerator->initialize(m_rxn, m_bbs);  // getSizesFromBBs(bbs));
@@ -223,6 +228,7 @@ EnumerateLibrary::EnumerateLibrary(const ChemicalReaction &rxn, const BBS &bbs,
                                    const EnumerationStrategyBase &enumerator,
                                    const EnumerationParams &params)
     : EnumerateLibraryBase(rxn),
+      m_dedupeSymmetricMatches(params.dedupeSymmetricMatches),
       m_matchCache(),
       m_bbs(removeNonmatchingReagents(m_rxn, bbs, params, &m_matchCache)) {
   m_enumerator.reset(enumerator.copy());
@@ -231,7 +237,10 @@ EnumerateLibrary::EnumerateLibrary(const ChemicalReaction &rxn, const BBS &bbs,
 }
 
 EnumerateLibrary::EnumerateLibrary(const EnumerateLibrary &rhs)
-    : EnumerateLibraryBase(rhs), m_matchCache(rhs.m_matchCache), m_bbs(rhs.m_bbs) {}
+    : EnumerateLibraryBase(rhs),
+      m_dedupeSymmetricMatches(rhs.m_dedupeSymmetricMatches),
+      m_matchCache(rhs.m_matchCache),
+      m_bbs(rhs.m_bbs) {}
 
 std::vector<MOL_SPTR_VECT> EnumerateLibrary::next() {
   PRECONDITION(static_cast<bool>(*this), "No more enumerations");
@@ -242,7 +251,7 @@ std::vector<MOL_SPTR_VECT> EnumerateLibrary::next() {
     reactants[i] = m_bbs[i][reactantIndices[i]];
   }
 
-  return run_Reactants(m_rxn, reactants, m_matchCache);
+  return run_Reactants(m_rxn, reactants, m_matchCache, m_dedupeSymmetricMatches);
 }
 
 void EnumerateLibrary::toStream(std::ostream &ss) const {
--- a/Code/GraphMol/ChemReactions/Enumerate/Enumerate.cpp
+++ b/Code/GraphMol/ChemReactions/Enumerate/Enumerate.cpp
@@ -219,10 +219,13 @@ EnumerateLibrary::EnumerateLibrary(const ChemicalReaction &rxn, const BBS &bbs,
     : EnumerateLibraryBase(rxn, new CartesianProductStrategy),
       m_dedupeSymmetricMatches(params.dedupeSymmetricMatches),
       m_matchCache(),
-      m_cacheReactantGrafts(params.cacheReactantGrafts),
+      m_cacheMode(params.cacheMode),
       m_graftCache(),
-      m_bbs(removeNonmatchingReagents(m_rxn, bbs, params, &m_matchCache)) {
-  m_enumerator->initialize(m_rxn, m_bbs);  // getSizesFromBBs(bbs));
+      m_bbs(removeNonmatchingReagents(
+          m_rxn, bbs, params,
+          params.cacheMode != ReactantCacheMode::None ? &m_matchCache
+                                                      : nullptr)) {
+  m_enumerator->initialize(m_rxn, m_bbs);  // getSizesFromBBs(bbs))
   m_initialEnumerator.reset(m_enumerator->copy());
 }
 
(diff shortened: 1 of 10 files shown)


bp-kelley:
@jpfeuffer this is a nice improvement that I've been thinking about now for > 10 years.  The failures, however, are real and effect results.

jpfeuffer:
Glad that you think it is a worthwhile improvement. I will investigate the failures. Didn't expect an effect outside of the few tests that I touched but will follow up on this!

jpfeuffer:
@bp-kelley Ready for further review!

bp-kelley:
Thanks, I'll get to it this week.

greglandrum:
@bp-kelley let me know when you're 👍 on this and I'll do a final review.

bp-kelley:
Will do.  Most likely next week.

Code/GraphMol/ChemReactions/Enumerate/Enumerate.cpp
```diff
@@ -129,8 +144,31 @@ BBS removeNonmatchingReagents(const ChemicalReaction &rxn, BBS bbs,
     for (size_t reagent_idx = 0; reagent_idx < bbs[reactant_idx].size();
          ++reagent_idx) {
       ROMOL_SPTR mol = bbs[reactant_idx][reagent_idx];
-      size_t matches =
-          countMatches(*mol.get(), *reactantTemplate.get(), maxMatches);
+      size_t matches = 0;
+
+      const bool canPrimeCache =
+          cache && params.reagentMaxMatchCount == INT_MAX &&
```
bp-kelley, line 150: Let's be explicit with cache != nullptr

Code/GraphMol/ChemReactions/Enumerate/Enumerate.cpp
```diff
@@ -179,23 +217,46 @@ BBS removeNonmatchingReagents(const ChemicalReaction &rxn, BBS bbs,
 EnumerateLibrary::EnumerateLibrary(const ChemicalReaction &rxn, const BBS &bbs,
                                    const EnumerationParams &params)
     : EnumerateLibraryBase(rxn, new CartesianProductStrategy),
-      m_bbs(removeNonmatchingReagents(m_rxn, bbs, params)) {
-  m_enumerator->initialize(m_rxn, m_bbs);  // getSizesFromBBs(bbs));
+      m_dedupeSymmetricMatches(params.dedupeSymmetricMatches),
```
bp-kelley, line 220: I

bp-kelley, line 220: Perhaps just save the whole EnumerationParams here as m_params

Code/GraphMol/ChemReactions/ReactionRunner.h
```diff
@@ -61,6 +78,18 @@ RDKIT_CHEMREACTIONS_EXPORT std::vector<MOL_SPTR_VECT> run_Reactants(
...
     unsigned int maxProducts = 1000);
 
+//! Runs a reaction on a set of reactants using a reactant match cache
+/*!
+  This overload reuses previously computed reactant-template matches for
+  repeated reagent/template combinations. The cache key is the reactant
+  template index, the reagent pointer identity, the effective
+  maxMatches/maxProducts value, and whether symmetry deduplication is on.
+*/
+RDKIT_CHEMREACTIONS_EXPORT std::vector<MOL_SPTR_VECT> run_Reactants(
```
bp-kelley, line 88: I'm half tempted to have a ReactantParams here as opposed to the multiple params here.

Code/GraphMol/ChemReactions/ReactionRunner.h
```diff
@@ -104,14 +133,86 @@ RDKIT_CHEMREACTIONS_EXPORT ROMol *reduceProductToSideChains(
...
+  a product and can be replayed onto any number of products. The two trailing
+  flags are part of the key because the match list a match index refers to
+  depends on dedupeSymmetricMatches, and whether a graft carries conformer
+  coordinates depends on doConfs; including them keeps a single cache correct
+  even when reused across calls with differing settings. The ROMol objects whose
+  raw pointers are used as keys must outlive this cache and must not be
+  structurally modified between calls, otherwise stale graft data may be
+  replayed. Not thread-safe.
+*/
+using ReactantGraftCache = std::map<
```
bp-kelley, line 158: Let's save us some future pain here and make const ROMol *  become: std::shared_ptr<const ROMol> so we don't have to worry about memory.   (for simplicity you can make this shared_ptr<ROMol> as well.

Code/GraphMol/ChemReactions/ReactionRunner.h
```diff
@@ -36,10 +36,27 @@
...
+using VectMatchVectType = std::vector<MatchVectType>;
+//! Caches reactant-template substructure matches across run_Reactants() calls.
+/*!
+  Keyed on (reactant template index, reagent pointer identity, effective
+  maxMatches, dedupeSymmetricMatches). The ROMol objects whose raw pointers
+  are used as keys must outlive this cache and must not be structurally
+  modified between calls, otherwise stale match data may be returned.
+  Not thread-safe.
+*/
+using ReactantMatchCache =
```
bp-kelley, line 56: See previous point about const ROMol *

Code/GraphMol/ChemReactions/ReactionRunner.cpp
```diff
@@ -213,9 +252,33 @@ bool getReactantMatches(const MOL_SPTR_VECT &reactants,
   for (auto iter = rxn.beginReactantTemplates();
        iter != rxn.endReactantTemplates(); ++iter, i++) {
     if (matchSingleReactant == MatchAll || matchSingleReactant == i) {
-      auto matches =
-          getReactantMatchesToTemplate(*reactants[i].get(), *iter->get(),
-                                       maxMatches, rxn.getSubstructParams());
+      const auto cacheKey =
+          std::make_tuple(i, reactants[i].get(), maxMatches,
```
bp-kelley, line 256: iter is already the template, you don't really need 'reactants[I].get()'  Here `*iter` is the reactant.

Code/GraphMol/ChemReactions/ReactionRunner.cpp
```diff
@@ -1319,6 +1395,247 @@ void generateProductConformers(Conformer *productConf, const ROMol &reactant,
...
+    return new Bond(srcBond);
+  }
+  return new QueryBond(dynamic_cast<const QueryBond &>(srcBond));
+}
+
+// Remap a copied bond's stereo-atom references (held in iso atom indices) to
+// the corresponding product atom indices. The bond already carries its stereo
+// and stereo-atom values from the copy constructor, so we only translate the
+// indices here, preserving their order (begin/end orientation is unchanged).
+void remapBondStereoAtoms(const Bond &isoBond, Bond *prodBond,
```
bp-kelley, line 1422: Stereo is notoriously trick, do we have enough tests for this?

bp-kelley:
This looks good, I only have minor comments.  I would like to take one of the synthon libraries (with smarts) and run them cached and uncached.  Perhaps @DavidACosgrove has an example or two with smarts.

My only real concern is stereo, we may need some more tests.

@greglandrum  I think this is worth looking at.

DavidACosgrove:
> This looks good, I only have minor comments. I would like to take one of the synthon libraries (with smarts) and run them cached and uncached. Perhaps @DavidACosgrove has an example or two with smarts.
> 
@bp-kelley If you have FreedomSpace 5.0 you should have the file 2026-01_Freedom_5_reactions_SMARTS.txt which has the SMARTS strings for building the products from the synthons.

> My only real concern is stereo, we may need some more tests.
> 
> @greglandrum I think this is worth looking at.

jpfeuffer:
@bp-kelley Your comments make sense and I can try to address them. At some point (not sure yet when I get to it), I will also have a look at FreedomSpace (which is very similar to my initial use-case).

greglandrum:
> @bp-kelley Your comments make sense and I can try to address them. 

@jpfeuffer  please @ me when you've made the changes you decide to do.

PR #9366, opened 2026-06-17
Adding github action npm package build and publish

charnley:

This initializes the RDKit MinimalLib WASM build and publishes it within RDKit-repo to keep everything aligned every time RDKit makes a new release.

- Minimal setup GitHub Action package compile, as the package only contains .js, .wasm, and .d.ts.
- Small fixes to multi-dockerfile setup, though renamed the files to `.dockerfile` to be consistent.
- Automatic TypeScript interface generation (.d.ts) and added variable hints in jswrapper (otherwise it will typehint with `_1` variable names)
- Fixed some small issues with tests.js (duplicate tests and variable init)
- Take npm version from CMakeLists.txt during GitHub Action.
- Export `.js` directly as module-friendly (`EXPORT_ES6=1`)
- Changed MinimalLib readme to be very minimal, moving all examples to rdkitjs.com. This is also because the README will be part of the NPM package and hosted on npmjs.com.
- Changed default behavior of `build_rdkitjs.sh` to take local source code, not clone. Clone now requires explicit `GET_SRC` to be set.
- `build_rdkitjs.sh` now also takes `NPM_VERSION` to set the version of the resulting packages.
- `build_rdkitjs.sh` now results in `build` folder to be consistent with NPM expectations. While still keeping resulting files in a `build/dist/` folder to be backward compatible. 

The resulting package will look like (package will be deleted after merge):

https://www.npmjs.com/package/@charnley/rdkit-minimal/

    npm i @charnley/rdkit-minimal

#### Any other comments?

These are changes that need to be made before merge

- [x] Change GitHub Actions to `@rdkit/rdkit`.
- [x] Set up trusted publisher / OIDC for `@rdkit/rdkit` on npmjs.com.

There is one open question

- [x] DRDK_BUILD_MINIMAL_LIB_RGROUPDECOMP seems to be currently disabled by default. Should it be?
- [x] There is currently a build (other than release) for minimal lib on GitHub Actions. Should we keep it, or only have the publish action?

ps, I used opencode/deepseek for some tedious tasks and to summarize commit msgs


damingerdai:
it looks nice for me. i am working for using nextjs and rdkit to build a fullstack application.

Code/MinimalLib/.gitignore
greglandrum, line 1: I don't think this belongs here

greglandrum:
@charnley the CI failures look real

charnley:
> @charnley the CI failures look real

I tried to be smart and only rely on debian `apt` packages, but this downgraded emscripten heavily and it seem to compensate I made it incompatible with newest emscripten. I will fix it.

greglandrum:
@charnley this is still marked as a draft. Is it ready for review/merging?

charnley:
@greglandrum @ptosco It is ready for Paolo to tell me his thoughts for the two Docker-file approaches : ) There are some open to-dos noted in the description before merge.

Code/MinimalLib/.gitignore
charnley: The .gitignore content is minimal (node.js etc) and folder-specific, and I didn't want to dirty the global .gitignore.

charnley:
@greglandrum  @ptosco , I had a short chat with Paolo and reverted my single-dockerfile setup to make it easier to review. I also cleaned up the last remaining tasks. It is ready for review. I updated my original PR statement with the scope of changes.

ptosco:
@charnley I have submitted a PR against your branch with some suggested changes.
Most importantly, `demo.html` fails to load the newly built `RDKit_minimal.js` and `RDKit_minimal.wasm` in my WSL `apache2` instance with the following error:
image
Am I missing something? Does `demo.html` load for you?

greglandrum:
@charnley if you fix the merge conflicts, I think we can merge this

charnley:
Changes from Paolo's comment

- I reverted `EXPORT_ES6` flag from emscripten, as it provided a version that required it to be loaded as `<script type="module">`, hence the error in the demo that Paolo found
- I enabled `DRDK_BUILD_MINIMAL_LIB_RGROUPDECOMP=ON`, for some reason that was never enabled as default for minimallib, and it looks like the demo/test is working fine.

The only thing holding me back was whether we need a test for the demo files. I have a local `playwright` Docker setup to run an HTTP server and test whether `demo.html` loads correctly. What do you think @greglandrum ?

Another question: currently there is a GitHub build (test) and publish. Should we keep both or only run on release publish? Personally, I would keep it. @greglandrum 

I want to (in another PR) split up the test file and also print out when a test is SKIPPED (as was the case with the rgroup flag).

greglandrum:
> Changes from Paolo's comment
> 
> * I reverted `EXPORT_ES6` flag from emscripten, as it provided a version that required it to be loaded as `<script type="module">`, hence the error in the demo that Paolo found
> * I enabled `DRDK_BUILD_MINIMAL_LIB_RGROUPDECOMP=ON`, for some reason that was never enabled as default for minimallib, and it looks like the demo/test is working fine.
> 
> The only thing holding me back was whether we need a test for the demo files. I have a local `playwright` Docker setup to run an HTTP server and test whether `demo.html` loads correctly. What do you think @greglandrum ?

What would be the mechanics of doing the testing?

> 
> Another question: currently there is a GitHub build (test) and publish. Should we keep both or only run on release publish? Personally, I would keep it. @greglandrum

I'm not sure I understand the question, but I will answer what I think you're asking:
I am reluctant to add anything new to the current CI stack. We already ensure that the JS stuff builds there and that the minimallib tests run.
Ideally:
 - we would run the npm related build (and tests) whenever a branch whose name starts with "Release" is updated.
 - we would do the build and push whenever a release is done

I'm not sure how much of this is possible.

charnley:
> What would be the mechanics of doing the testing?

The tldr is: JavaScript is used in three ways, but we are currently only testing one. 

- Via `var initRDKitModule = require('RDKit_minimal.js');` tested in `node tests/test.js`
- Via CDN `<script src="RDKit_minimal.js" />` tested by opening `demo.html` which had the error Paolo caught.
- Via `import { initRDKitModule } from "@rdkit/rdkit"` when installed into `node_modules` and used in some bundled application.

I don't like having to open `demo.html` to catch errors, and I'd prefer to set up a browser emulator (`playwright`) to test that it works. We're also not currently testing that the package works for bundlers. Will leave it for now though.

> I am reluctant to add anything new to the current CI stack. We already ensure that the JS stuff builds there and that the minimallib tests run.

Understood. I removed the Action that is executed per-commit, leaving only the publish github action (per-release). Instead I added a "package test" to the Docker workflow to test that the package is deployable, which is run as part Azure. This should catch if any changes break the npm package.

charnley:
I had to make some changes to the build because I ran into issues with emsdk and freetype that caused an error due to incompatibility with `WASM_LEGACY_EXCEPTIONS=True`. I got 

    emcc: error: attempt to set `WASM_LEGACY_EXCEPTIONS` to `True`; use 1/0 to set boolean settings

- Inconsistency with the default compiler flag. `build_rdkitjs.sh` was using `-fwasm-exceptions`, and Azure was using `-fexceptions -sNO_DISABLE_EXCEPTION_CATCHING`. Now the default is set to `-fwasm-exceptions` for both.
- Above showed a bug where emsdk was downloading a new version of FreeType (`USE_FREETYPE=1`), even though the version was locked in the dependency Dockerfile (above error). CMake files were updated to use the compiled locked version.

It was not failing on Azure Pipelines because the default flags were `-fexceptions` for the `docker compose` command, but `build_rdkitjs.sh` was failing on the master branch as well.

(My poor CPU had to recompile MinimalLib so many times over the weekend to isolate this issue)

greglandrum:
> > What would be the mechanics of doing the testing?
> 
> The tldr is: JavaScript is used in three ways, but we are currently only testing one.
> 
> * Via `var initRDKitModule = require('RDKit_minimal.js');` tested in `node tests/test.js`
> * Via CDN `<script src="RDKit_minimal.js" />` tested by opening `demo.html` which had the error Paolo caught.
> * Via `import { initRDKitModule } from "@rdkit/rdkit"` when installed into `node_modules` and used in some bundled application.
> 
> I don't like having to open `demo.html` to catch errors, and I'd prefer to set up a browser emulator (`playwright`) to test that it works. We're also not currently testing that the package works for bundlers. Will leave it for now though.

Ok. Since I don't do real work in the JS space and don't particularly feel like investing the time to learn all this stuff, I'm going to go with whatever you say is needed here.

> 
> > I am reluctant to add anything new to the current CI stack. We already ensure that the JS stuff builds there and that the minimallib tests run.
> 
> Understood. I removed the Action that is executed per-commit, leaving only the publish github action (per-release). Instead I added a "package test" to the Docker workflow to test that the package is deployable, which is run as part Azure. This should catch if any changes break the npm package.

Great, thanks.

charnley:
> Ok. Since I don't do real work in the JS space and don't particularly feel like investing the time to learn all this stuff, I'm going to go with whatever you say is needed here.

I will think about it, and maybe improve the test docker step to include these in the future, but should be fine for now.

Then I think we are ready to merge.

greglandrum (approved):
LGTM
approval based on @ptosco's feedback

PR #9369, opened 2026-06-20
Use the same libexpat as the current python when building

bp-kelley:
@greglandrum noticed an issue with a conflict between the version of expat ChemDraw was using to build and the one python uses that could cause segfaults at runtime.

This force ChemDraw to use the system expat, which, at least under conda builds, should be the same one python uses.

### DIFF
--- a/.azure-pipelines/vs_build.yml
+++ b/.azure-pipelines/vs_build.yml
@@ -13,7 +13,7 @@ steps:
         libboost-devel=$(boost_version) ^
         numpy matplotlib cairo pillow eigen ^
         numpy=2.4 pandas=3 ^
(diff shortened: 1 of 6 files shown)


bp-kelley:
@greglandrum hopefully this just works (TM) the fix was pretty simple and worked on my tests systems.

there might be necessary changes to rdkit-feedstock to ensure libexpat is installed during the C++ builds, it should be automatic when python is included in "condo install/create"

bp-kelley:
Looks like I'll have to dink around with ensuring libexpat is included in some of the builds.

greglandrum (approved):
LGTM

PR #9371, opened 2026-06-23
Replace trivial Boost usages with standard-library equivalents

skearnes:
## Summary

Rung 0 of a broader effort to reduce RDKit's dependence on Boost: mechanical, low-risk replacements of trivial Boost usages with their C++ standard-library equivalents, plus removal of dead Boost includes. No functional or ABI changes intended.

This is independent of (and complementary to) the boost::iostreams removal in #9370.

## What changed

- **Dead/unused includes removed:** `boost/graph` kruskal + johnson (`MolOps.cpp`, never used), `boost/math/special_functions/round` (`TorsionConstraint.cpp`, `MMFF/testMultiThread.cpp`), an unused `boost/noncopyable` (`catch_chirality.cpp`) and `boost/foreach` (`Atropisomers.cpp`).
- **`boost::optional` → `std::optional`** (`python_streambuf.h`).
- **`boost::variant` → `std::variant`** for `MolPickler`'s `QueryDetails`, and its consumers `MolPickler.cpp` and `MolInterchange/Writer.cpp` (`boost::get` → `std::get`, `.which()` → `.index()`).
- **`boost/cstdint` → `<cstdint>` / `std::*int*_t`** in `StreamOps.h`, plus the three files that had been relying on it transitively (`SparseIntVect.h`, `BitOps.cpp`, `FPBReader.cpp`), making them self-contained.
- **`boost::noncopyable` → explicitly deleted copy operations** (`MolWriter`, the test-fixture template, the UFF/MMFF `DefaultTorsionBondSmarts` helpers). `MolWriter`, a base class, also gets a defaulted default constructor.
- The unbuilt `UNSAFE_RDVALUE` path (`RDValue-doublemagic.h`) gets the same type-traits/`static_assert`/cstdint cleanup for consistency.

## Verification

- Builds clean locally (clang): GraphMol, FileParsers, ForceFieldHelpers, DataStructs, MolInterchange; `testMolWriter`, `graphmolTest1`, and `testPickler`'s query-serialization section pass.
- All changed files additionally pass a `g++-16` / libstdc++ `-fsyntax-only` check (to catch transitive-include gaps that clang/libc++ would hide). The two files that fail standalone (`python_streambuf.h`, `RDValue-doublemagic.h`) fail identically on master — pre-existing, context-dependent, and not introduced here.

Draft pending full CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9372, opened 2026-06-23
Make public C++ headers self-contained

skearnes:
## Why

This is groundwork for the ongoing effort to reduce RDKit's reliance on Boost (#9370 removes `boost::iostreams`; #9371 replaces trivial Boost usages with their standard-library equivalents). A recurring hazard in that work: removing a widely-included umbrella header (e.g. `boost/iostreams`) breaks the build in seemingly unrelated files, because those files were silently depending on the umbrella to transitively pull in `std::string`, `std::vector`, `ROMol`, the `PRECONDITION` macro, and so on. The breakage typically only shows up under libstdc++ (not libc++) and only on certain include orders, so it surfaces as confusing, platform-specific CI failures partway through a refactor.

Making the headers self-contained up front removes that hazard: each header carries its own dependencies, so dropping or reordering umbrella includes can no longer expose a hidden gap. It also benefits downstream consumers, who can include any public header without first discovering the right preamble by trial and error.

## What changed

The changes are additive (new `#include`s / forward declarations only) — no behavior or API change.

- **34 headers made self-contained** by adding the includes / forward declarations they were relying on transitively. Affected areas: `DataStructs`, `GraphMol/Descriptors`, `ForceFieldHelpers`, `FileParsers`, `ChemReactions`, `MolDraw2D`, `MolEnumerator`, `MolInterchange`, `SLNParse`, `SmilesParse`, `SynthonSpaceSearch`, `GaussianShape`, and the core `Canon`/`Resonance`/`StereoGroup`/`Subset` headers.
- **`Data/Fonts/font_dumper.py`** now emits `#include <string>` so regenerated `Fonts/*regular.h` stay self-contained (matching the checked-in change to the two existing dumps).

## Verification

Each affected header was confirmed to compile as a standalone translation unit (`#include` it and nothing else) under `g++` / libstdc++ — the stricter standard library, deliberately chosen to surface the gaps that libc++ tolerates. The fixes were validated across all in-scope public headers under `Code/`:

```
Checking 452 headers with g++ -std=gnu++20
All 452 headers are self-contained.
```


Code/GraphMol/MolEnumerator/LinkNode.h
```diff
@@ -9,6 +9,7 @@
 //
 #ifndef RD_MOLENUMERATOR_LINKNODE_H
 #define RD_MOLENUMERATOR_LINKNODE_H
+#include <GraphMol/ROMol.h>
```
greglandrum, line 12: Shouldn't we also explicitly include `sstream` in this file?
We only get `stringstream` because Invariant.h brings in RDLog.h and that brings in stringstream

Code/GraphMol/SmilesParse/CanonicalizeStereoGroups.h
```diff
@@ -10,6 +10,10 @@
 #include <RDGeneral/export.h>
 #ifndef RD_CANONICALIZESTEROGROUPS_H
 #define RD_CANONICALIZESTEROGROUPS_H
+#include <stdexcept>
```
greglandrum, line 13: elsewhere we bring in `exception`, not `stdexcept`; should be consistent

Code/GraphMol/SynthonSpaceSearch/Synthon.h
```diff
@@ -10,6 +10,8 @@
 
 #ifndef RDKIT_REAGENT_H
 #define RDKIT_REAGENT_H
+#include <DataStructs/ExplicitBitVect.h>
+#include <memory>
```
greglandrum, line 13: I think this one should probably have vector too

greglandrum: And if you do a forward declaration of ExplicitBitVect you can get rid of that header:
```
#include <memory>
#include <vector>

#include <string>

#include <RDGeneral/export.h>

class ExplicitBitVect;
```

Code/GraphMol/ChemReactions/ReactionParser.h
```diff
@@ -35,6 +35,7 @@
 #include <RDGeneral/export.h>
 #ifndef RD_REACTIONPARSER_H_21Aug2006
 #define RD_REACTIONPARSER_H_21Aug2006
+#include <GraphMol/ChemReactions/Reaction.h>
```
greglandrum, line 38: Since you've included this, you no longer need the forward declaration of `ChemicalReaction` below.

greglandrum (changes requested):
Small suggested changes

Code/GraphMol/MolEnumerator/LinkNode.h
skearnes: Done.

Code/GraphMol/SmilesParse/CanonicalizeStereoGroups.h
skearnes: I'd like to keep `<stdexcept>` here: `RigorousEnhancedStereoException` derives from `std::runtime_error`, which is declared in `<stdexcept>`. `<exception>` provides only `std::exception` (and friends) and does not declare `runtime_error` — g++/clang both report *"no type named 'runtime_error' in namespace 'std'"* if you swap it — so switching would actually break self-containment for this header.

This also matches the existing convention: `RDGeneral/Exceptions.h`, `RDGeneral/Invariant.h`, and `RDGeneral/BadFileException.h` all bring in `<stdexcept>` for the same reason. I think the `<exception>` cases are the classes that derive from `std::exception` directly — e.g. `ChemicalReactionParserException` in this same PR — which is a different base and correctly uses `<exception>`.

Code/GraphMol/SynthonSpaceSearch/Synthon.h
skearnes: Done — added `<vector>` and switched to a forward declaration of `ExplicitBitVect`, dropping the `DataStructs/ExplicitBitVect.h` include.

One subtlety worth flagging: `Synthon`'s destructor and move-assignment are inline-defaulted, so a translation unit that destroys or move-assigns a `Synthon` needs `ExplicitBitVect` complete. That's not a regression — `ROMol` was already forward-declared here in exactly the same way, so the requirement already existed (a consumer that odr-uses `~Synthon` gets incomplete-type errors for *both* `ROMol` and `ExplicitBitVect`, not just the latter). The header itself stays self-contained: it only parses, and those special members are instantiated at odr-use rather than on `#include`.

If we ever want the forward declarations to be fully robust for consumers, the clean fix is to move the destructor and move-assignment out-of-line as `= default` in `Synthon.cpp` (matching the copy ctor/assignment, which are already defined there). That's a structural change beyond this PR's additive scope, so I've left it out — happy to do it as a follow-up if you'd prefer.

Code/GraphMol/ChemReactions/ReactionParser.h
skearnes: Done.

greglandrum (approved):
LGTM

PR #9373, opened 2026-06-23
MacroMol core

emilyrrhodes:
#### Reference Issue
MacroMol/MonomerMol

This is a small PR to describe the root commonality amongst a variety of macro/monomeric representations.

#### Description
This is PR 1 of 6
[MacroMol core](https://github.com/rdkit/rdkit/pull/9373) ⬅️
[MacroMolTemplate](https://github.com/rdkit/rdkit/pull/9485)
[MacroMol Local Library](https://github.com/rdkit/rdkit/pull/9486)
[MacroMol Global Library](https://github.com/rdkit/rdkit/pull/9487)
[MolToMacroMol](https://github.com/rdkit/rdkit/pull/9488)
[MolFromMacroMol](https://github.com/rdkit/rdkit/pull/9489)


Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,120 @@
...
+  props.beginAttachPt = beginAttachPt;
+  props.endAttachPt = endAttachPt;
+  props.bondType = bondType;
+  props.isDirectional = isDirectional;
+
+  std::vector<MacroBondProps> propsList;
+  bond->getPropIfPresent(common_properties::macroBondProps, propsList);
+  propsList.push_back(props);
+  bond->setProp(common_properties::macroBondProps, propsList);
+}
```
rachelnwalker, line 66: Emily and I have talked offline a bit about this - since `MacroAtom`s can have any number of attachment points, two `MacroAtom`s can be attached to each other in more than one way. There is a clear HELM example of this where two cysteines are connected by both a backbone bond and a disulfide bond:

`PEPTIDE1{C.C}$PEPTIDE1,PEPTIDE1,1:R3-2:R3$$$V2.0`

Since `MacroMol` is a subclass of `RWMol`, we cannot store multiple actual bonds. We got around this in our original implementation by just storing multiple properties (LINKAGE and CUSTOM_LINKAGE)  - but that isn't easily expandable to >2 linkages between monomers.

I have two main concerns with storing this as a vector property though:
1. The way RDKit deals with vector<MacroBondProps> is different then how it deals with types like std::string, int, double, which RDValue has native tagged slots for. I think this means that the property access and setting will be much more expensive.
2. We loose the ability to use the bond type in writing SMILES

Since this is specifically for an edge case, I'm tempted to say that we should omit support for >1 bond between macroatoms for now, and later add support for it in a similar way to our implementation using a backup property https://github.com/schrodinger/sketcher/blob/93599daef0865272d929e7925a13bcba9f2e9ec7/src/schrodinger/rdkit_extensions/monomer_mol.cpp#L361

I realize that is a bit hacky - but it is one of the downsides of subclassing RWMol directly. I think that the benefits of subclassing RWMol outweigh the negatives here - we get 

- I/O: pickling, drawing
- Algorithms: substructure matching, canonicalization, ring perception
- Iteration / queries: mol.atoms(), getBondBetweenAtoms, the property system,
  bookmarks, conformers
- Python boundary: the boost::python wrappers for Mol accept MacroMol* for
  free

Code/GraphMol/testMacroMol.cpp
```diff
@@ -0,0 +1,86 @@
...
+  auto m2_atom_1 = macro_mol_2->addMacroAtom(MonomerClass::AA, "C");
+  auto m2_atom_2 = macro_mol_2->addMacroAtom(MonomerClass::AA, "D");
+  auto m2_atom_3 = macro_mol_2->addMacroAtom(MonomerClass::AA, "A");
+  macro_mol_2->addMacroBond(m2_atom_1, m2_atom_2, 2, 1);
+  macro_mol_2->addMacroBond(m2_atom_3, m2_atom_1, 2, 1);
+
+  std::string m1_smiles = MolToSmiles(*macro_mol_1);
+  std::string m2_smiles = MolToSmiles(*macro_mol_2);
+  CHECK(m1_smiles == m2_smiles);
+}
```
rachelnwalker: General comment - I think this is a really important basis for a creating a MacroMol hash. Since addMacroAtom sets an isotope that hashes the monomer class and name, the constructed SMILES is a canonical representation of the connectivity of the monomers. To get a full hash including the attachment points, you would just need to add a layer that includes the used attachment points.

What we currently do looks something like this:
```
helm1 = "PEPTIDE1{G.P}$$$$V2.0"
mm = rdkit_extensions.to_rdkit(helm1, Format.HELM)
layers1 = _get_monomer_mol_layers(mm)

helm2 = "PEPTIDE1{P.G}$$$$V2.0"
mm = rdkit_extensions.to_rdkit(helm2, Format.HELM)
layers2 = _get_monomer_mol_layers(mm)

print("difference: ")
print(layers1)
print(layers2)

difference: 
{<MonomerMolHashLayer.CANONICAL_SMILES: 1>: '[24362G][50588P]', <MonomerMolHashLayer.LINKAGES: 2>: 'R2-R1', <MonomerMolHashLayer.SGROUP_DATA: 3>: '[]'}
{<MonomerMolHashLayer.CANONICAL_SMILES: 1>: '[24362G][50588P]', <MonomerMolHashLayer.LINKAGES: 2>: 'R1-R2', <MonomerMolHashLayer.SGROUP_DATA: 3>: '[]'}
```

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,120 @@
...
+                       std::optional<Bond::BondType> bondType) {
+  PRECONDITION(!this->getAtomWithIdx(beginAtomIdx)
+                    ->hasProp(common_properties::isMacroAtom),
+               "begin atom is a macro atom");
+  PRECONDITION(!this->getAtomWithIdx(endAtomIdx)
+                    ->hasProp(common_properties::isMacroAtom),
+               "end atom is a macro atom");
+
+  auto bt = bondType.value_or(Bond::BondType::SINGLE);
+  RWMol::addBond(beginAtomIdx, endAtomIdx, bt);
```
rachelnwalker: I agree that blocking the ability to add a regular bond (i.e., not set the right properties) between two macroatoms is important so that we ensure users don't accidentally create a corrupt state. Can you add a test that shows this isn't allowed?

Code/GraphMol/MacroMol.cpp
tadhurst-cdd: The following was resolved with the BondInfo construct:

I had already changed ROMOL to allow more than one  bondsbetween two macro atoms.   This is required in the example you have below, and also when two RNA bases have two or three hydrogen bonds between them.  This change was allowed in ROMol.cpp.

Also, the routine getBondsBetweenAtoms returns a vector of bonds between two atoms.

I think these changes must be incorporated here

tadhurst-cdd: I had already allowed more than one bond between macro atoms - this code was in getBondsBetweenAtoms.   This is required for RNA bases that have two or three hbonds between them.

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,120 @@
...
+    return MonomerClass::OTHER;
+  }
+}
+
+unsigned int MacroMol::addMacroAtom(MonomerClass monomerClass,
+                                    std::string templateName) {
+  auto className = monomerClassToString(monomerClass);
+  auto atom = new Atom(0);
+
+  atom->setProp(common_properties::isMacroAtom, true);
```
tadhurst-cdd: It seems superfluous to have "isMacroAtom".  If the atom has a templateName and a class Name, then it is a macro atom, is it not?

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,120 @@
...
+  atom->setProp(common_properties::dummyLabel, templateName);
+  atom->setProp(common_properties::molAtomClass, className);
+  atom->setProp(common_properties::smilesSymbol, templateName);
```
tadhurst-cdd: I think it is hard to define the MacroAtom without any definition of the MacroTemplate.  It seems that the symbol used in the smiles generation should be determined by the MacroTemplate definition.   Each Template must allow multiple templateNames, and any of those names should be allowed in the addMacroAtom call.  The canonical representation should be independent of the name chosen for this call.

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,120 @@
...
+                            bool isDirectional, Bond::BondType bondType) {
+  auto bond = this->getBondBetweenAtoms(beginAtomIdx, endAtomIdx);
+  if (!bond) {
+    auto bondIdx =
+        this->addBond(beginAtomIdx, endAtomIdx, Bond::BondType::ZERO) - 1;
+    bond = this->getBondWithIdx(bondIdx);
+  }
+  MacroBondProps props;
+  props.beginAttachPt = beginAttachPt;
+  props.endAttachPt = endAttachPt;
```
tadhurst-cdd: Having the attachment points be integers might be problematic.   Althouigh the MacroTemplate is not defined here,  the attachment points were previously defined, and I think it should be defined, using the existing RDKIT SUP group AttachPoint list.   These refer to the attachment points by named strings.

We could translate the integer attach point IDs to string as "1","2", "3", so the integers being used in the MacroMol could work.  

Lets keep this in mind as we go forward.

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,120 @@
...
+  props.bondType = bondType;
+  props.isDirectional = isDirectional;
```
tadhurst-cdd: I do not think we should label the bonds as directional.   It is just a bond betweeen two atoms.   Any directionality comes from the atoms it attachs. If it goes from attachment point "1" to "2" we know what the directionality is.

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,70 @@
...
+RDKIT_GRAPHMOL_EXPORT std::string monomerClassToString(
+    MonomerClass monomerClass);
+
+RDKIT_GRAPHMOL_EXPORT MonomerClass
+stringToMonomerClass(const std::string &className);
+
+struct RDKIT_GRAPHMOL_EXPORT MacroBondProps {
+  int beginAttachPt;
+  int endAttachPt;
+  Bond::BondType bondType;
```
tadhurst-cdd: The bond type should not be stored here - it is a property of the bond already.

Code/GraphMol/MacroMol.cpp
tadhurst-cdd: When SCSR blocks are parsed, there will be regular bonds between to macro atoms (hydrogen bonds).  As part of the parsing process, these are converted to MacroAtom bonds that have proper attachment point designations.  Lets make sure we don't block this intermediate state.

Code/GraphMol/MacroMol.cpp
emilyrrhodes: In an effort to keep the changes in this PR focused and small, I am going to remove the ability to store information about multiple bonds (between the same two atoms) here. We can discuss this implementation further in a future PR that builds off of this one.

tadhurst-cdd:
I realize that the intent to make small definition and changes, but it is hard to define the macro atom without defining the template and template library. 

Canonicalizing the MacroAtom with canonizing the Templates used does not ensure equally of macro atoms.   So I think the MacroAtomTemplate and MacroAtomTemplateLibrary need to be the next part added.

tadhurst-cdd:
Lets not remove that ability - it is required for Hbonds between RNA bases.

emilyrrhodes:
While I know this is a very limited PR, I think it is simply the foundation for other pieces.

Code/GraphMol/MacroMol.cpp
tadhurst-cdd: Lets not lose sight of the end-goal.   Multiple bonds between two macro atoms is required, and we know how to do it.

greglandrum: > I had already changed ROMOL to allow more than one bondsbetween two macro atoms. This is required in the example you have below, and also when two RNA bases have two or three hydrogen bonds between them. This change was allowed in ROMol.cpp.
> 
> Also, the routine getBondsBetweenAtoms returns a vector of bonds between two atoms.
> 
> I think these changes must be incorporated here

I understand that it's essential to support something like this, but I believe that this change is likely to cause all kinds of problems in other parts of the RDKit - all of the other code has been written in a world where atoms only have one bond between them.

It seems like we can do an implementation where at the level of the ROMol having a bond between two macroatoms indicates that there is, at the atomic level, at least one bond between the two residues/monomers. Information about the nature of those bonds could be stored as properties (or even data members) on the macroatoms themselves.

But I agree with @emilyrrhodes: we can explore/decide this in a future PR.

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,108 @@
...
+//  The contents are covered by the terms of the BSD license
+//  which is included in the file license.txt, found at the root
+//  of the RDKit source tree.
+//
+#include "MacroMol.h"
+#include "Atom.h"
+
+namespace RDKit {
+
+std::string monomerClassToString(MonomerClass monomerClass) {
```
greglandrum: We started using better_enum to make handling enum <-> string logic easier and more robust. I'd suggest using it here too.
A good place to see how it works is in `RGroupDecompParams.h` and `RGroupDecomJSONParsers.cpp`

If we decide not to go that way, I would suggest at least using two `std::unordered_map` for this instead of standalone functions.

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,108 @@
...
+unsigned int MacroMol::addMacroAtom(MonomerClass monomerClass,
+                                    std::string symbol) {
+  auto className = monomerClassToString(monomerClass);
+  auto atom = new Atom(0);
+
+  atom->setProp(common_properties::isMacroAtom, true);
+  atom->setProp(common_properties::dummyLabel, symbol);
+  atom->setProp(common_properties::molAtomClass, className);
+
+  return this->addAtom(atom, false, true);
```
greglandrum: This is more readable later if you provide names for the extra arguments:

```suggestion
  bool updateLabel = false;
  bool takeOwnership = true;
  return this->addAtom(atom, updateLabel, takeOwnership);
```

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,64 @@
...
+  bool isDirectional;
+};
+
+class RDKIT_GRAPHMOL_EXPORT MacroMol : public RWMol {
+ public:
+  using RWMol::addBond;
+
+  unsigned int addMacroAtom(MonomerClass monomerClass, std::string symbol);
+
+  unsigned int addMacroBond(unsigned int beginAtomIdx, unsigned int endAtomIdx,
```
greglandrum, line 46: I think this function and its parameters here need to be documented

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,64 @@
...
+                            int beginAttachPt, int endAttachPt,
+                            Bond::BondType bondType = Bond::BondType::SINGLE);
+
+  unsigned int addAtomToMacroAtomBond(
```
greglandrum, line 62: Same comment about documentation

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,64 @@
...
+      unsigned int beginAtomIdx, unsigned int endMacroAtomIdx, int endAttachPt,
+      Bond::BondType bondType = Bond::BondType::SINGLE);
+
+  unsigned int addMacroAtomToAtomBond(
```
greglandrum, line 78: Same comment about documentation

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,64 @@
...
+      unsigned int beginMacroAtomIdx, unsigned int endAtomIdx,
+      int beginAttachPt, Bond::BondType bondType = Bond::BondType::SINGLE);
+
+  unsigned int addBond(unsigned int beginAtomIdx, unsigned int endAtomIdx,
```
greglandrum, line 94: Should be documented that this doesn't work between macro atoms

greglandrum (commented):
For discussion:
Since accessing/using properties can be expensive and not particularly memory efficient, how about either:
1. adding a class to store necessary information about macro atoms - `MacroAtomInfo` or some such thing - and adding a unique pointer to the `Atom` class for this. Then actual macro atoms will have the pointer set and the values can be looked up. This is analogous to what we currently do with MonomerInfo.
2. Add a MacroAtom subclass of Atom that has space to store the additional values.

I prefer the first option since I think it would be nice to eventually be able to get rid of virtual functions on Atoms (would require getting rid of the `QueryAtom` class, but that's completely unconnected to this PR).

(I guess we could also think about adding the additional info to the MonomerInfo class... but that may be a step too far)

Code/GraphMol/MacroMol.cpp
tadhurst-cdd: The change I made only allows multiple bonds between the same. two atoms for MacroAtoms.  All tests pass.  The only troubleing place I found an issue is where the call is made to return THE bond between two atoms, which I fixed to work is there is only one bond, and make a new call to return all bonds between two atoms.

I think I mis-read what you suggested about a MacroAtom class.   I thought you wanted to have a pointerer to an Atom in a new MacroAtom class, but you actuallly suggested a pointer to a MacroInfo class in each Atom, which could be null for regular atoms.  That seems OK.  It does make each Atom a little bigger.

Code/GraphMol/MacroMol.h
tadhurst-cdd: This struct is apparently NOT used.    Howerever, if we adopt Greg L's suggestion that the bond's existence indicates that there is at least one bond between the two atoms, -  then this should be used and could be an array or list of these - thus addressing the need for more than one bond between two Macro atoms.
This gets kind of ugly if one atom is a real atom and the other is a macro atom - then there must be one and only one item in that list.

tadhurst-cdd:
> For discussion: Since accessing/using properties can be expensive and not particularly memory efficient, how about either:
> 
> 1. adding a class to store necessary information about macro atoms - `MacroAtomInfo` or some such thing - and adding a unique pointer to the `Atom` class for this. Then actual macro atoms will have the pointer set and the values can be looked up. This is analogous to what we currently do with MonomerInfo.
> 2. Add a MacroAtom subclass of Atom that has space to store the additional values.
> 
> I prefer the first option since I think it would be nice to eventually be able to get rid of virtual functions on Atoms (would require getting rid of the `QueryAtom` class, but that's completely unconnected to this PR).
> 
> (I guess we could also think about adding the additional info to the MonomerInfo class... but that may be a step too far)

I like the first option too.

Code/GraphMol/MacroMol.cpp
emilyrrhodes: @rachelnwalker @tadhurst-cdd @greglandrum Would something like "MacroBondInfo" work for this? https://github.com/emilyrrhodes/rdkit/pull/13

greglandrum: > @rachelnwalker @tadhurst-cdd @greglandrum Would something like "MacroBondInfo" work for this? [emilyrrhodes#13](https://github.com/emilyrrhodes/rdkit/pull/13)

I think that's a good solution to the "we need multiple bonds between macro atoms" problem. And putting it on the bond instead of the atoms, as I suggested above, makes a lot more sense.

tadhurst-cdd: We/you already have MacroBondProps.   I like MacroBondInfo.

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,105 @@
...
+
+    \param beginAtomIdx  index of the atom where the bond begins
+    \param endAtomIdx    index of the atom where the bond ends
+    \param beginAttachPt the attachment point on the begin (macro) atom
+    \param endAttachPt   the attachment point on the end (macro) atom
+    \param bondType      the type of the bond
+
+    \return the index of the newly added bond
+  */
+  unsigned int addMacroBond(unsigned int beginAtomIdx, unsigned int endAtomIdx,
```
i-tub: It seems confusing that `addMacroBond()` returns the bond index while `addBond()` returns the number of bonds. Of course, `addAtom()` returns the atom index, so there's already a pre-existing inconsistency between `addAtom()` and `addBond()`, but that ship has sailed...

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,105 @@
...
+    macro atoms.
+
+    \param beginAtomIdx index of the atom where the bond begins
+    \param endAtomIdx   index of the atom where the bond ends
+    \param bondType     the type of the bond
+
+    \return the index of the newly added bond
+  */
+  unsigned int addBond(unsigned int beginAtomIdx, unsigned int endAtomIdx,
+                       Bond::BondType bondType = Bond::BondType::SINGLE);
```
i-tub: 1. `RWMol::addBond()` returns the number of bonds, not the index of the new bond, and so does this function, since it delegates to the former. (And that's good; an override should preserve the semantics of the parent implementation.)
2. We should probably override the overload that takes atom pointers too, because otherwise calls to that overload will bypass the validation.

Code/GraphMol/MacroAtomInfo.h
```diff
@@ -0,0 +1,76 @@
...
+  const std::string &getSymbol() const { return d_symbol; }
+
+  //! Sets the macro atom symbol.
+  /*!
+    \param symbol the symbol used to identify the monomer
+  */
+  void setSymbol(const std::string &symbol) { d_symbol = symbol; }
+
+  //! Returns the macro atom monomer class.
+  const std::string &getMonomerClass() const { return d_monomerClass; }
```
i-tub: Would it make sense for the monomer class to be an enum rather than a string?

Code/GraphMol/MacroMol.h
emilyrrhodes: Added

Code/GraphMol/MacroAtomInfo.h
emilyrrhodes: Changed

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,98 @@
...
+unsigned int MacroMol::addAtomToMacroAtomBond(unsigned int beginAtomIdx,
+                                              unsigned int endMacroAtomIdx,
+                                              int endAttachPt,
+                                              Bond::BondType bondType) {
+  PRECONDITION(isMacroAtom(this->getAtomWithIdx(endMacroAtomIdx)),
+               "end atom is not a macro atom");
+  PRECONDITION(!isMacroAtom(this->getAtomWithIdx(beginAtomIdx)),
+               "begin atom is a macro atom");
+  return addMacroBond(beginAtomIdx, endMacroAtomIdx, -1, endAttachPt, bondType);
+}
```
rachelnwalker, line 89: Couldn't users still accidentally make a macro bond between an atom and a monomer just using addMacroBond? I think you may want to abstract some of this into a different function so that addMarcoBond requires both atoms to be MacroAtoms

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,96 @@
...
+    \param endAtomIdx    index of the atom where the bond ends
+    \param beginAttachPt the attachment point on the begin (macro) atom
+    \param endAttachPt   the attachment point on the end (macro) atom
+    \param bondType      the type of the bond
+
+    \return the new number of bonds
+  */
+  unsigned int addMacroBond(unsigned int beginAtomIdx, unsigned int endAtomIdx,
+                            int beginAttachPt, int endAttachPt,
+                            Bond::BondType bondType = Bond::UNSPECIFIED);
```
rachelnwalker: Should we make the default just SINGLE? I imagine that is what people will use 90% of the time and wouldn't want someone to accidentally make bonds that are unspecificed is they aren't supported

Code/GraphMol/MacroMol.cpp
emilyrrhodes: I merged the MacroBond PR into this one. Multiple bonds between a single pair of MacroAtoms is now supported.

Code/GraphMol/MacroMol.cpp
emilyrrhodes: Fixed

Code/GraphMol/MacroMol.h
emilyrrhodes: I copied the pattern here: 
https://github.com/rdkit/rdkit/blob/810e19700386e85756d7ef209aa82bac381ae813/Code/GraphMol/RWMol.h#L129. If SINGLE is better, happy to change.

greglandrum: I agree that using SINGLE here is the right thing to do

emilyrrhodes: I am concerned that this would be unexpected behavior if the default for `addBond` is different than the default for `addMacroBond`. 

For instance, if someone was building a MacroMol they could do something like: 
```
auto macro_mol = std::make_unique<MacroMol>();
auto regular_atom_1 = new Atom(6);
auto regular_atom_idx_1 = macro_mol->addAtom(regular_atom_1, false, true);
auto regular_atom_2 = new Atom(6);
auto regular_atom_idx_2 = macro_mol->addAtom(regular_atom_2, false, true);
auto macro_atom_idx = macro_mol->addMacroAtom(MonomerClass::AA, "C");
auto num_bonds1 = macro_mol->addAtomToMacroAtomBond(regular_atom_idx_1, macro_atom_idx, 1); // default = SINGLE
auto num_bonds2 = macro_mol->addBond(regular_atom_idx_1, regular_atom_idx_2); // default = UNDEFINED
```
And might not expect the bond types to be different.

Additionally, if we changed the default for `addBond` to resolve this issue (making it SINGLE as well), then it could be confusing why doing something like: 
```
auto macro_mol = std::make_unique<MacroMol>();
auto regular_atom_1 = new Atom(6);
auto regular_atom_idx_1 = macro_mol->addAtom(regular_atom_1, false, true);
auto regular_atom_2 = new Atom(6);
auto regular_atom_idx_2 = macro_mol->addAtom(regular_atom_2, false, true);
auto num_bonds = macro_mol->addBond(regular_atom_idx_1, regular_atom_idx_2); // default = SINGLE
```
is different from something like: 
```
auto mol = std::make_unique<RWMol>();
auto regular_atom_1 = new Atom(6);
auto regular_atom_idx_1 = mol->addAtom(regular_atom_1, false, true);
auto regular_atom_2 = new Atom(6);
auto regular_atom_idx_2 = mol->addAtom(regular_atom_2, false, true);
auto num_bonds = mol->addBond(regular_atom_idx_1, regular_atom_idx_2); // default = UNDEFINED
```

I think default consistency would make this the most intuitive (everything having a default of UNDEFINED) when adding bonds.

rachelnwalker: It would be a little strange to override the default for the addBond, but I think it makes sense for the addMacroBond to default to single since that is what will be used in the majority of cases

emilyrrhodes: Changed the default bondType to SINGLE.

Code/GraphMol/MacroAtomInfo.cpp
```diff
@@ -0,0 +1,50 @@
...
+    }};
+}  // namespace
+
+const char *monomerClassToString(MonomerClass monomerClass) {
+  for (const auto &[value, name] : monomerClassNames) {
+    if (value == monomerClass) {
+      return name;
+    }
+  }
+  PRECONDITION(false, "unknown monomer class");
```
greglandrum:
```suggestion
  POSTCONDITION(false, "unknown monomer class");
```

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "Atom.h"
+
+#include <RDGeneral/Invariant.h>
+
+#include <utility>
+
+namespace RDKit {
+namespace {
+bool isMacroAtom(const Atom *atom) {
+  return atom->getMacroAtomInfo() != nullptr;
```
greglandrum, line 21:
```suggestion
  PRECONDITION(atom, "atom is null");
  return atom->getMacroAtomInfo() != nullptr;
```

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,128 @@
...
+               "end atom is a macro atom");
+
+  return RWMol::addBond(beginAtomIdx, endAtomIdx, bondType);
+}
+
+unsigned int MacroMol::addBond(Atom *beginAtom, Atom *endAtom,
+                               Bond::BondType bondType) {
+  PRECONDITION(beginAtom && endAtom, "NULL atom passed in");
+  return addBond(beginAtom->getIdx(), endAtom->getIdx(), bondType);
+}
```
greglandrum, line 116: It seems like it's simpler to do these two functions as:

```suggestion
unsigned int MacroMol::addBond(unsigned int beginAtomIdx,
                               unsigned int endAtomIdx,
                               Bond::BondType bondType) {
  return addBond(this->getAtomWithIdx(beginAtomIdx),
                 this->getAtomWithIdx(endAtomIdx), bondType);
}

unsigned int MacroMol::addBond(Atom *beginAtom, Atom *endAtom,
                               Bond::BondType bondType) {
  PRECONDITION(beginAtom && endAtom, "NULL atom passed in");
  PRECONDITION(!isMacroAtom(beginAtom), "begin atom is a macro atom");
  PRECONDITION(!isMacroAtom(endAtom), "end atom is a macro atom");
  return RWMol::addBond(beginAtom->getIdx(), endAtom->getIdx(), bondType);
}
```

greglandrum (changes requested):
I've made a few small suggestions, but I think this looks good and would be happy merging it after those are made.

greglandrum (approved):
Assuming the CI builds pass, the changes LGTM

greglandrum:
@tadhurst-cdd : are you 👍 on merging this as a first step towards having a working MacroMol?

Code/GraphMol/MacroAtomInfo.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <array>
+#include <utility>
+
+namespace RDKit {
+namespace {
+const std::array<std::pair<MonomerClass, const char *>, 4> monomerClassNames = {
+    {
+        {MonomerClass::AA, "AA"},
+        {MonomerClass::NA, "NA"},
+        {MonomerClass::CHEM, "CHEM"},
```
tadhurst-cdd: The abbreviation NA for Nucleic Acid look a lot like "Not Available".  Is there a better choice than NA?

Code/GraphMol/MacroMol.cpp
```diff
@@ -0,0 +1,127 @@
...
+namespace RDKit {
+namespace {
+bool isMacroAtom(const Atom *atom) {
+  PRECONDITION(atom, "atom is null");
+  return atom->getMacroAtomInfo() != nullptr;
+}
+}  // namespace
+
+unsigned int MacroMol::addMacroAtom(MonomerClass monomerClass,
+                                    std::string symbol) {
```
tadhurst-cdd: It might be nice to swap the order of monomerClass and symbol.  These are in the other order in setMacroAtomInfo.

Code/GraphMol/MacroMol.h
```diff
@@ -0,0 +1,110 @@
...
+    \param monomerClass the class of monomer the macro atom represents
+    \param symbol       the symbol (dummy label) used to identify the monomer
+
+    \return the index of the newly added atom
+  */
+  unsigned int addMacroAtom(MonomerClass monomerClass, std::string symbol);
+
+  //! Adds a bond between two macro atoms.
+  /*!
+    At least one of the two atoms must be a macro atom.
```
tadhurst-cdd: The comment "At least one of ..."  I think is incorrect - both must be macro atoms for this call.

tadhurst-cdd:
> @tadhurst-cdd : are you 👍 on merging this as a first step towards having a working MacroMol?

The one major thing I think is missing is the internal template library for the MacroMol.    If this does not exist, software that parses SCSRs blocks/files will have to maintain the relationship between a MacroMol and its defining Template library.  I guess we could have yet another class (ScsrMol?)  that derived from MacroMol and has an internal Template library, but that is ugly.

I think there should also be validation for a MacroMol that checks that every macro atom is found in the defining template library.

Other than that, I think this looks REALLY good.

tadhurst-cdd (changes requested):
this looks really good!  

I left a comment about having an internal Template Library and one about validation of the MacroMol against the template library.

Those could be done after this PR, or maybe done differently.   I am anxious to use this to test all the SCSR stuff I have, and I kinda need to know what the plan for internal Template libraries is.

Code/GraphMol/MacroAtomInfo.cpp
emilyrrhodes: Changed these names to be clearer.

Code/GraphMol/MacroMol.cpp
emilyrrhodes: Switched

Code/GraphMol/MacroMol.h
emilyrrhodes: Updated this comment

i-tub (approved):
The changes since my last review LGTM.

rachelnwalker (approved):
This looks good as an initial implementation - thanks @emilyrrhodes!

@tadhurst-cdd - if we go ahead and merge this and there are future small changes we have to make to the interface in order to get the template library working as needed, I think that is ok (assuming @greglandrum is also ok with that). There is probably more discussion we need to have on the template library so it is nice to separate it a little from the discussion on the general MacroMol interface

tadhurst-cdd (approved):
Look good.  

In a future PR, we need to link a MacroMol with its defining Template library - perhaps by having a Template library internal to the MacroMol

greglandrum (approved):
LGTM

greglandrum:
Looks like we all agree that this one is ready to go. I will merge after the last two CI builds go through (there were failures unconnected to the PR).

Really exciting!

PR #9377, opened 2026-06-30
fix rendering of mols with multiple highlights from PandasTools

greglandrum:
The `Draw._moltoimg()` function that PandasTools calls wasn't handling mulitple highlights correctly.

Thanks to Bernd Rohde for reporting this.

### DIFF
--- a/rdkit/Chem/Draw/__init__.py
+++ b/rdkit/Chem/Draw/__init__.py
@@ -147,8 +147,7 @@ def MolToFile(mol, filename, size=(300, 300), kekulize=True, wedgeBonds=True, im
     outf.close()
 
 
-def MolToSVG(mol, size=(300, 300), kekulize=True, wedgeBonds=True, drawOptions=None,
-              **kwargs):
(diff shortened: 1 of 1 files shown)


rdkit/Chem/Draw/__init__.py
```diff
@@ -274,6 +272,13 @@ def _moltoimg(mol, sz, highlights, legend, returnPNG=False, drawOptions=None, **
   # we already prepared the molecule:
   d2d.drawOptions().prepareMolsBeforeDrawing = False
   bondHighlights = kwargs.get('highlightBonds', None)
+  if highlights:
+    if isinstance(highlights[0], (list, tuple)):
+      # multiple highlights
+      tmp = []
+      for sublist in highlights:
+        tmp.extend(sublist)
+      highlights = tmp
```
ricrogz: This could also use https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable:

```
      # don't forget to import itertools !
      highlights = list(itertools.chain.from_iterable(highlights))
```

ricrogz (approved):
LGTM.

PR #9378, opened 2026-06-30
add new clash resolution approach: open angles of the clashing chain

ZontaNicola:
#### Any other comments?

Here's a presentation that Codex put together after running on the whole of Chmbl with a 100 atoms cutoff
[chembl34_path_angle_expansion_current_clash_solver_under100.pdf](https://github.com/user-attachments/files/31638923/chembl34_path_angle_expansion_current_clash_solver_under100.pdf)

EDIT: this is a newer version, which includes tweaks to clash detection. With this change we fix 25% of problematic structures (up from 20% without the fix) for an average slowdown of approximately 2.4%

I ran the whole CHmbl to identify clashing structures. 
here's. the first 1200, comparison with and without the approach. Note that I turned of the bond flipping so the two columns are comparable. Click the filter checkbox to only see the molecules that get fixed (around 15%).
https://zontanicola.github.io/sketcher/galleries/chembl_angle_expansion_1200_final.html


Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1591,7 +1591,20 @@ void EmbeddedFrag::randomSampleFlipsAndPermutations(
     const DOUBLE_SMART_PTR *dmat, double mimicDmatWt, bool permuteDeg4Nodes) {
   PRECONDITION(dp_mol, "");
 
-  const auto &rotBonds = getAllRotatableBonds(*dp_mol);
```
ZontaNicola, line 1598: this came up in my tests: if the smiles is fragmented (contains ".") this was crashing because getAllRotatabelBonds would get bonds that were not in the current fragment

Code/GraphMol/Depictor/Wrap/rdDepictor.cpp
rachelnwalker, line 370: Please add a python docstring for this.

You will also need to include this in the nanobind wrapper as well.

Code/GraphMol/Depictor/RDDepictor.cpp
```diff
@@ -618,6 +618,11 @@ unsigned int compute2DCoords(RDKit::ROMol &mol,
     // if there are any remaining collisions
     eri.removeCollisionsOpenAngles();
     eri.removeCollisionsShortenBonds();
+
+    // Path angle expansion (off by default)
+    if (params.usePathAngleExpansion) {
+      eri.removeCollisionsPathAngleExpansion();
+    }
```
rachelnwalker, line 625: It seems like this new approach may include moving coordinate mapped templates (greptile pointed this out) - is there a way to ensure that mapped atoms are not moved, or should we just disable the option if a mapping is specified?

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -2308,4 +2326,281 @@ void EmbeddedFrag::removeCollisionsShortenBonds() {
...
+  // Choose the smaller side to rotate (less disruptive)
+  bool rotateSide1 = atomsSide1.size() < atomsSide2.size();
+  auto &atomsToMove = rotateSide1 ? atomsSide1 : atomsSide2;
+
+  // Use cross product to determine orientation
+  // cross = v1.x * v2.y - v1.y * v2.x
+  // If cross > 0: CCW orientation (prevAtom -> center -> nextAtom)
+  // If cross < 0: CW orientation
+  double cross = v1.x * v2.y - v1.y * v2.x;
+
```
rachelnwalker: could something weird happen here is cross = 0 (so the bonds are at a 180 deg angle)? Seems unlikely but worth catching just in case

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -2308,4 +2326,281 @@ void EmbeddedFrag::removeCollisionsShortenBonds() {
...
+    }
+
+    // OPTIMIZATION: Only recompute distance matrix after atom positions changed
+    dmat = RDKit::MolOps::getDistanceMat(*dp_mol);
+
+    // Check if expansion helped or made things worse
+    colls = this->findCollisions(dmat, 0);
+
+    // accept if we have fewer or same collisions, otherwise revert
+    if (colls.size() > prevCollisionCount) {
```
rachelnwalker: Is there a reason to accept when there are the same # of collisions? I would probably try to keep the standard depiction if nothing is fixed

rachelnwalker (commented):
This looks pretty good to me, just a few comments. Please add some tests that exercise this option!

Code/GraphMol/Depictor/Wrap/rdDepictor.cpp
ZontaNicola: done

Code/GraphMol/Depictor/RDDepictor.cpp
ZontaNicola: yeah, I had codex fix this and move only the sub portion without fixed atoms (if any exists)

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: we probably shouldn't move them in that case (there's nothing to open). codex pointed out that if we are close to 180, like 179, we could end up overcorrecting, so it proposed a trial and error approach: try one direction and if it opens the angle good, otherwise go for the other. It's simple and performance wise should be very close so I agreed

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: yes, my idea was that most of the time we take little steps that just push the collision outwards, towards the perimeter of the molecule, before we can actually solve it. but I had codex test both ways and it agreed that this fixes more cases. details follow.

I tested 10,000 structures sampled across five regions of ChEMBL, comparing:

- **Equal allowed:** revert only when collisions increase.
- **Strict:** revert unless collisions immediately decrease.

For molecules with fewer than 100 atoms:

| Result | Equal allowed | Strict |
|---|---:|---:|
| Structures without errors | 9,324 / 9,643 (96.69%) | 9,273 / 9,643 (96.16%) |
| Structures with errors | 319 | 370 |
| Bond crossings | 275 | 304 |
| Atom clashes | 140 | 186 |

Allowing equal counts produced:

- **13.8% fewer molecules with geometry problems**
- **9.5% fewer bond-crossing reports**
- **24.7% fewer atom-clash reports**
- 51 molecules clean with equal allowed but problematic with strict
- 0 molecules clean with strict but problematic with equal allowed

Across all 10,000 molecules, equal-count continuation produced 113 clean-vs-problematic wins, while strict produced only one.

The flag-off controls had only three clean/problematic disagreements between runs, so nondeterminism is far too small to explain this result.

Timing was effectively indistinguishable:

- Median enabled/control ratio with equal allowed: **1.0193**
- Median enabled/control ratio with strict: **1.0205**

Therefore, the extra iterations did not produce a measurable performance penalty in this sample.

The likely mechanism is that the first increment may move a collision along the path without reducing the total collision count. A subsequent increment can then move it far enough outward to resolve it. Strict acceptance aborts that sequence after the first step.

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -1730,46 +1809,44 @@ std::vector<PAIR_I_I> EmbeddedFrag::findCollisions(const double *dmat,
...
+            if (_segmentsIntersect(d_eatoms[beg1].loc, d_eatoms[end1].loc,
+                                   d_eatoms[beg2].loc,
+                                   d_eatoms[end2].loc)) {
+              // Choose the outermost endpoints of the crossed bonds. The
+              // downstream getRotatableBonds() call deliberately skips the
+              // two outer bonds on this path, so they will not themselves be
+              // flipped; their adjacent bonds remain available as candidates.
+              auto cAids =
+                  _findFarthestPair(beg1, end1, beg2, end2, *dp_mol, dmat);
+              res.push_back(cAids);
```
rachelnwalker, line 1849: Isn't this technically a change to the default depiction code, since it is in `findCollisions`? Are you able to tell if this is a positive change?

Code/GraphMol/Depictor/EmbeddedFrag.cpp
rachelnwalker: Ok, that makes sense!

rachelnwalker (commented):
Aside from one comment I think this looks good! I'm excited about these depictions!

@greglandrum this is ready for you to take a look at

Code/GraphMol/Depictor/RDDepictor.h
```diff
@@ -90,6 +90,8 @@ struct RDKIT_DEPICTOR_EXPORT Compute2DCoordParameters {
                                   //!< preferCoordGen is set to true
   bool useRingTemplates = false;  //!< whether to use ring system templates for
                                   //!< generating initial coordinates
+  bool usePathAngleExpansion = false;  //!< use path angle expansion for
```
greglandrum, line 93: I think we should default this to `true`.
This solves a lot of bad depictions with a negligible cost.

Code/GraphMol/Depictor/Wrap/rdDepictor.cpp
```diff
@@ -50,10 +51,21 @@ unsigned int Compute2DCoords(RDKit::ROMol &mol, bool canonOrient,
...
+  params.coordMap = cMap.empty() ? nullptr : &cMap;
+  params.canonOrient = canonOrient;
+  params.clearConfs = clearConfs;
+  params.nFlipsPerSample = nFlipsPerSample;
+  params.nSamples = nSamples;
+  params.sampleSeed = sampleSeed;
+  params.permuteDeg4Nodes = permuteDeg4Nodes;
+  params.forceRDKit = forceRDKit;
+  params.useRingTemplates = useRingTemplates;
+  params.usePathAngleExpansion = usePathAngleExpansion;
```
greglandrum, line 65: Maybe use initialization instead of assignment:
```suggestion
  Compute2DCoordParameters params{
      .coordMap = cMap.empty() ? nullptr : &cMap,
      .canonOrient = canonOrient,
      .clearConfs = clearConfs,
      .nFlipsPerSample = nFlipsPerSample,
      .nSamples = nSamples,
      .sampleSeed = sampleSeed,
      .permuteDeg4Nodes = permuteDeg4Nodes,
      .forceRDKit = forceRDKit,
      .useRingTemplates = useRingTemplates,
      .usePathAngleExpansion = usePathAngleExpansion};
```

Code/GraphMol/Depictor/nbWrap/rdDepictor.cpp
```diff
@@ -57,10 +59,18 @@ static unsigned int compute2DCoordsHelper(RDKit::ROMol &mol, bool canonOrient,
...
+  params.canonOrient = canonOrient;
+  params.clearConfs = clearConfs;
+  params.nFlipsPerSample = nFlipsPerSample;
+  params.nSamples = nSamples;
+  params.sampleSeed = sampleSeed;
+  params.permuteDeg4Nodes = permuteDeg4Nodes;
+  params.forceRDKit = forceRDKit;
+  params.useRingTemplates = useRingTemplates;
+  params.usePathAngleExpansion = usePathAngleExpansion;
+  unsigned int res = RDDepict::compute2DCoords(mol, params);
```
greglandrum, line 73: Maybe use initialization instead of assignment:
```suggestion
  Compute2DCoordParameters params{
      .coordMap = cMap.empty() ? nullptr : &cMap,
      .canonOrient = canonOrient,
      .clearConfs = clearConfs,
      .nFlipsPerSample = nFlipsPerSample,
      .nSamples = nSamples,
      .sampleSeed = sampleSeed,
      .permuteDeg4Nodes = static_cast<bool>(permuteDeg4Nodes),
      .forceRDKit = forceRDKit,
      .useRingTemplates = useRingTemplates,
      .usePathAngleExpansion = usePathAngleExpansion};
```

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -2308,4 +2385,361 @@ void EmbeddedFrag::removeCollisionsShortenBonds() {
...
+  //   - This preserves ring geometry while allowing ring-chain junction
+  //   expansion
+  //   - Maximum 15° expansion per angle (prevents unrealistic geometries)
+  //   - 5° increments (small steps to detect new collisions early)
+  //   - Revert if expansion makes things worse
+  //
+
+  // OPTIMIZATION: Pre-compute ring membership for each atom
+  // Build a map: atom_id -> set of ring indices containing that atom
+  std::map<int, std::set<int>> atomRings;
```
greglandrum, line 2566: I think we probably want to consider coming back here at some point and looking at how many of the maps and sets in this function can be converted to the unordered versions

greglandrum (changes requested):
Small suggestions

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: yes this a bug fix that allows the code to consider clashes on terminal bonds

Code/GraphMol/Depictor/RDDepictor.h
ZontaNicola: agreed

Code/GraphMol/Depictor/Wrap/rdDepictor.cpp
ZontaNicola: implemented this

Code/GraphMol/Depictor/nbWrap/rdDepictor.cpp
ZontaNicola: implemented this

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: I had codex take a crack at this and it converted 8 containers to unordered

ZontaNicola:
I implemented Greg's comments, this should be ready to go

rachelnwalker (approved):

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -2590,7 +2593,14 @@ void EmbeddedFrag::removeCollisionsPathAngleExpansion() {
     return PAIR_I_I(std::min(collision.first, collision.second),
                     std::max(collision.first, collision.second));
   };
-  std::set<PAIR_I_I> skippedCollisions;
+  struct CollisionHash {
+    std::size_t operator()(const PAIR_I_I &collision) const noexcept {
+      const auto firstHash = std::hash<int>{}(collision.first);
+      const auto secondHash = std::hash<int>{}(collision.second);
+      return firstHash ^ (secondHash << 1);
```
greglandrum: I think it's probably better to do this with `boost::hash_combine`:
add `#include <boost/functional/hash.hpp>` at the top of the file and then replace the body of this function with:
```
      size_t res = 0;
      boost::hash_combine(res, collision.first);
      boost::hash_combine(res, collision.second);
      return res;

```

greglandrum (changes requested):
I'd suggest changing the hash function, but otherwise this looks good

PR #9384, opened 2026-07-08
Add ROMol name accessors

i-tub:
Added `ROMol::setName()` and `ROMol::getName()` as convenience methods for the molecule title stored in the `_Name` property. The aim is to make the name feel more like a first-class property of the molecule instead of just an internal convention, while keeping backward compatibility.

The new methods are callable from Python as `Mol.SetName()` and `Mol.GetName()`.

#### Notes / questions

I asked Greg many moons ago about this idea and he said something like "sure, send a PR". Here it is, very belatedly! I'm not sure if I should file an issue for it or if going straight to PR is OK.

Question: what do you think about `getName()` returning empty string by default instead of throwing?

### DIFF
--- a/Code/GraphMol/ROMol.cpp
+++ b/Code/GraphMol/ROMol.cpp
@@ -589,6 +589,14 @@ ROMol::ConstBondIterator ROMol::endBonds() const {
   return ConstBondIterator(this, end);
 }
 
+void ROMol::setName(const std::string &name) const {
+  setProp(common_properties::_Name, name);
+}
+
+std::string ROMol::getName() const {
+  return getProp<std::string>(common_properties::_Name);
+}
+
 void ROMol::clearComputedProps(bool includeRings) const {
   // the SSSR information:
   if (includeRings) {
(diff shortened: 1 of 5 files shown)


DavidACosgrove:
Catching an exception in C++ is relatively expensive, so should be reserved for things that are exceptional, or so I read recently.  A molecule not having a name is fairly common in my experience so I would go with an empty string.

DavidACosgrove:
The function is a great idea, I think.

greglandrum:
> #### Notes / questions
> I asked Greg many moons ago about this idea and he said something like "sure, send a PR". Here it is, very belatedly! I'm not sure if I should file an issue for it or if going straight to PR is OK.

Straight to PR is great.

> Question: what do you think about `getName()` returning empty string by default instead of throwing?

I think returning the empty string is better than throwing. As @DavidACosgrove said: dealing with exceptions can have a performance impact.

i-tub:
> I think returning the empty string is better than throwing. As @DavidACosgrove said: dealing with exceptions can have a performance impact.

SGTM; I've updated it accordingly.

greglandrum (approved):
LGTM

PR #9388, opened 2026-07-10
Speedup testDGeomHelpers

nmaeder:
#### Reference Issue
Following the discussion in #9292 

Extracts one particular test that takes 2x~50 seconds into a seperate test file.
The two tests can be called in parallel using the [catch2 shard utility](https://catch2-temp.readthedocs.io/en/latest/command-line.html#test-sharding).

#### Any other comments?

There is a [macro defined by catch2](https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md) that one could use for that, but I figured the change here is small enough to not use it.
Happy to hear suggestions & comments.

### DIFF
--- a/Code/GraphMol/DistGeomHelpers/CMakeLists.txt
+++ b/Code/GraphMol/DistGeomHelpers/CMakeLists.txt
@@ -9,18 +9,13 @@ rdkit_headers(BoundsMatrixBuilder.h
 
 rdkit_catch_test(testDistGeomHelpers testDgeomHelpers.cpp
            LINK_LIBRARIES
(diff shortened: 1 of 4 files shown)


ricrogz (approved):
LGTM. Thanks!

greglandrum:
@nmaeder I did a PR against your branch with a suggested change to go ahead and use the ctest/cmake sharding integration

nmaeder:
@greglandrum nice, thanks, is merged.

greglandrum (approved):
LGTM

greglandrum:
I kind of fear that the windows CI failures are going to be a recurring thing, but let's see what happens.

PR #9389, opened 2026-07-10
Removed redundant code

kabu00002:
Removed some checks in the 1-4 distance for rings (`BoundsMatrixBuilder.cpp`) that either are redundant or introduce unintended inconsistencies.


greglandrum:
@kabu00002 the CI failures here are real (caused by the fact that we do `-Werror` in the CI builds).
You can either just remove the names of the unused arguments from the function definitions or use `RDUNUSED_PARAM()` for them

greglandrum (approved):
LGTM

PR #9393, opened 2026-07-10
Fixes #9381

### LINKED ISSUE #9381: EmbedMolecule leaves RDKit SIGINT handler installed after successful embedding
Disclaimer: the below issue was posted by GPT-5.5 on my behalf, but I first noticed the issue and verified the reproduction script myself. I first noticed the issue when using rdkit alongside fasthtml/uvicorn, as mentioned below.

---

**Describe the bug**

`AllChem.EmbedMolecule()` appears to leave RDKit's native `SIGINT` handler installed after a successful embedding call. This prevents Python applications which own signal handling from receiving later `SIGINT`s.

The externally visible symptom is that a Python signal handler still appears to be installed according to `signal.getsignal(signal.SIGINT)`, but a real `SIGINT` no longer reaches it after `EmbedMolecule()` has returned.

This breaks host applications such as uvicorn/FastAPI/Starlette servers: after serving a request that calls `EmbedMolecule()`, pressing Ctrl-C no longer triggers uvicorn's graceful shutdown handler.

Looking at the source, the relevant path seems to be:

- `Code/GraphMol/DistGeomHelpers/Wrap/rdDistGeom.cpp`
  - `EmbedMolecule2()` releases the GIL and calls `DGeomHelpers::EmbedMolecule(mol, params)`.
- `Code/GraphMol/DistGeomHelpers/Embedder.cpp`
  - `EmbedMolecule()` calls `ControlCHandler::reset()`.
- `Code/RDGeneral/ControlCHandler.h`
  - `reset()` does:
    ```cpp
    d_gotSignal = false;
    std::signal(SIGINT, signalHandler);
    ```
  - This installs RDKit's C-level SIGINT handler without preserving/restoring the previously installed Python-level handler.

The `ControlCHandler` constructor/destructor appears to preserve and restore the previous handler, but the static `reset()` path does not.

**To Reproduce**

```python
import os
import signal
import time

from rdkit import Chem

seen = []


def handler(sig, frame):
    print("python handler ran")
    seen.append(sig)


signal.signal(signal.SIGINT, handler)

print("before:", signal.getsignal(signal.SIGINT))

mol = Chem.AddHs(Chem.MolFromSmiles("C[C@H](O)c1ccccc1"))
params = AllChem.ETKDGv3()
params.randomSeed = 0xF00D

result = AllChem.EmbedMolecule(mol, params)
print("embed result:", result)

print("after:", signal.getsignal(signal.SIGINT))

os.kill(os.getpid(), signal.SIGINT)
time.sleep(0.2)

print("seen:", seen)
```

Output:

```text
before: <function handler at ...>
embed result: 0
after: <function handler at ...>
seen: []
```

So Python still reports that `handler` is installed, but the process-level SIGINT is no longer delivered to it.

A second probe shows that re-registering the same Python handler after `EmbedMolecule()` makes delivery work again:

```python
signal.signal(signal.SIGINT, signal.getsignal(signal.SIGINT))
os.kill(os.getpid(), signal.SIGINT)
```

After this, the Python handler runs.

**Expected behavior**

After `EmbedMolecule()` returns normally, the previous process-level SIGINT handler should be restored.

If Python/uvicorn/a CLI/test runner installed a SIGINT handler before calling `EmbedMolecule()`, that handler should continue to receive SIGINT after embedding completes.

**Screenshots**

Not applicable.

 - RDKit version: reproduced on 2024.09.6, 2025.03.3, and 2026.03.3
 - OS: macOS 27.0 arm64
 - Python version (if relevant): reproduced on CPython 3.11, 3.12, 3.13, and 3.14 where wheels are available
 - Are you using conda? No
 - If you are not using conda: installed from PyPI wheels in isolated `uv run --with rdkit==...` environments

**Additional context**

This was first observed in a local uvicorn/FastHTML app.

Before any request calls `EmbedMolecule()`, Ctrl-C shuts uvicorn down cleanly:

```text
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
```

After a request calls `EmbedMolecule()`, Ctrl-C no longer reaches uvicorn's signal handler, so the process remains running.

I reduced this outside uvicorn to the minimal signal reproduction above, so the issue does not depend on uvicorn, uv, FastHTML, sockets, or an embedding call still running. `EmbedMolecule()` has already returned successfully when the later SIGINT is lost.

I also did a coarse matrix sweep on macOS arm64 using isolated `uv run --no-project --isolated --with rdkit==... --python ...` environments:

| RDKit | Python 3.11 | Python 3.12 | Python 3.13 | Python 3.14 |
|---|---:|---:|---:|---:|
| 2024.09.6 | BROKEN | BROKEN | BROKEN | no cp314 wheel |
| 2025.03.3 | BROKEN | BROKEN | BROKEN | no cp314 wheel |
| 2026.03.3 | BROKEN | BROKEN | BROKEN | BROKEN |

This seems related to #8576 in that both involve `ControlCHandler::reset()`, but this report is specifically about a successful `EmbedMolecule()` call leaving the process-level SIGINT handler changed afterward.

greglandrum:
When we introduced the `ControlCHandler()` in #8153 we added RAII features, but then we only actually used the handler's static functions. The net result of this is that we never reset the handler to what it was originally.

This remedies that. Hopefully.

It's still unlikely to solve complex multi-threaded use cases, but it will hopefully work most of the time.

### DIFF
--- a/Code/GraphMol/DistGeomHelpers/Embedder.cpp
+++ b/Code/GraphMol/DistGeomHelpers/Embedder.cpp
@@ -372,7 +372,7 @@ bool _checkKTerms(RDGeom::Point3DPtrVect &positions,
       std::ranges::count_if(eargs.etkdgDetails->angles,
                             [](const auto &angle) { return angle[3]; }) +
       eargs.etkdgDetails->improperAtoms.size();
-  if (totalEnergy > (nCenters * planarityTolerance)){
+  if (totalEnergy > (nCenters * planarityTolerance)) {
     return false;
   }
   for (const auto &contrib : field->contribs()) {
@@ -1353,8 +1353,8 @@ void findChiralSets(const ROMol &mol, DistGeom::VECT_CHIRALSET &chiralCenters,
           }
         }
       }  // if block -chirality check
-    }    // if block - heavy atom check
-  }      // for loop over atoms
+    }  // if block - heavy atom check
+  }  // for loop over atoms
 
   // now do atropisomers
   for (const auto &bond : mol.bonds()) {
@@ -1840,7 +1840,8 @@ void EmbedMultipleConfs(ROMol &mol, INT_VECT &res, unsigned int numConfs,
     }
     int numThreads = getNumThreadsToUse(params.numThreads);
 
-    ControlCHandler::reset();
+    ControlCHandler hdlr;
+    // ControlCHandler::reset();
 
     // do the embedding, using multiple threads if requested
     detail::EmbedArgs eargs = {&confsOk,        fourD,
@@ -1874,7 +1875,7 @@ void EmbedMultipleConfs(ROMol &mol, INT_VECT &res, unsigned int numConfs,
       res.push_back(-1);
       return;
     }
-    if (ControlCHandler::getGotSignal()) {
+    if (hdlr.getGotSignal()) {
       BOOST_LOG(rdWarningLog) << INTERRUPT_MESSAGE << std::endl;
       return;
     }
--- a/Code/GraphMol/DistGeomHelpers/Embedder.cpp
+++ b/Code/GraphMol/DistGeomHelpers/Embedder.cpp
@@ -1841,7 +1841,6 @@ void EmbedMultipleConfs(ROMol &mol, INT_VECT &res, unsigned int numConfs,
     int numThreads = getNumThreadsToUse(params.numThreads);
 
     ControlCHandler hdlr;
-    // ControlCHandler::reset();
 
     // do the embedding, using multiple threads if requested
     detail::EmbedArgs eargs = {&confsOk,        fourD,
--- a/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
+++ b/Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
@@ -121,13 +121,14 @@ SearchResults SynthonSpaceSearcher::search() {
   }
   bool timedOut = false;
   std::uint64_t totHits = 0;
+  ControlCHandler hdlr;
   auto allHits = assembleHitSets(endTime, timedOut, totHits);
-  if (!timedOut && !ControlCHandler::getGotSignal() && d_params.buildHits) {
+  if (!timedOut && !hdlr.getGotSignal() && d_params.buildHits) {
     buildHits(allHits, endTime, timedOut, results);
   }
 
   return SearchResults{std::move(results), totHits, timedOut,
-                       ControlCHandler::getGotSignal()};
+                       hdlr.getGotSignal()};
 }
 
 std::unique_ptr<ROMol> SynthonSpaceSearcher::buildAndVerifyHit(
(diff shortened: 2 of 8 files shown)


Code/RDGeneral/ControlCHandler.h
```diff
@@ -14,6 +14,7 @@
 #include <atomic>
 #include <csignal>
 #include <stdexcept>
+#include <iostream>
```
ricrogz: This is a rather heavy header, that will be included transitively in everything that includes this header. Can we avoid it ?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
```diff
@@ -386,15 +387,14 @@ void sortAndUniquifyToTry(
...
-  toTry.erase(
-      std::remove_if(toTry.begin(), toTry.end(),
-                     [&seen](const auto &entry) {
-                       return !seen
-                                   .insert(details::buildProductHash(
-                                       entry.first, entry.second))
-                                   .second;
-                     }),
-      toTry.end());
+  toTry.erase(std::remove_if(toTry.begin(), toTry.end(),
```
ricrogz, line 390: C++ 20 has `std::erase_if`!  😄

Code/GraphMol/DistGeomHelpers/Embedder.cpp
```diff
@@ -1840,7 +1840,8 @@ void EmbedMultipleConfs(ROMol &mol, INT_VECT &res, unsigned int numConfs,
     }
     int numThreads = getNumThreadsToUse(params.numThreads);
 
-    ControlCHandler::reset();
+    ControlCHandler hdlr;
+    // ControlCHandler::reset();
```
ricrogz: Maybe remove this commented out code?

Code/GraphMol/SynthonSpaceSearch/SynthonSpaceSearcher.cpp
DavidACosgrove: Probably not one for this PR, though?  I could take a sweep through (there are 5 remove_if's in SynthonSpaceSearch) as a separate thing once my outstanding PR is merged.

Code/RDGeneral/ControlCHandler.h
DavidACosgrove: It certainly doesn't seem to be used in the header.  CLion also thinks <atomic> and <RDGeneral/export.h> aren't used either which seems correct though with the latter I never know!

greglandrum: Yeah, it's definitely not needed

DavidACosgrove (approved):
LGTM, with @ricrogz's comments addressed.  I've tried it on the synthon space search by running testSynthonSpaceLONG_TEST and banging control-C at random points and it seemed to work ok.

ricrogz (approved):
LGTM

PR #9396, opened 2026-07-13
Only use ignoreAtomMapNumbers if canonicalising.

### LINKED ISSUE #9225: ignoreAtomMapNumbers=True removes atom map numbers even when canonical=False, contradicting documentation
I'm using RDKit's API to generate SMILES strings and noticed an issue. According to the RDKit documentation, the `ignoreAtomMapNumbers` parameter should only affect whether atom map numbers are considered during canonicalization. However, when I use `canonical=False, ignoreAtomMapNumbers=True`, I get a SMILES string **with all atom map numbers removed**. This contradicts the documented behavior and is quite confusing. I'd like to know if this is a bug.

**Code to reproduce:**
```python
import rdkit
from rdkit import Chem
from rdkit.Chem import SmilesParserParams

x = "[H:1][C:2]([Br:3])([Cl:4])[H:5]"
para = SmilesParserParams()
para.removeHs = False
mol = Chem.MolFromSmiles(x, para)

print(Chem.MolToSmiles(mol, canonical=False, ignoreAtomMapNumbers=True))
print(Chem.MolToSmiles(mol, canonical=True, ignoreAtomMapNumbers=True))
```

**Actual output:**
```
[H]C(Br)(Cl)[H]
[H:1][C:2]([H:5])([Cl:4])[Br:3]
```

**Expected behavior (based on documentation):**  
When `canonical=False`, `ignoreAtomMapNumbers=True` should **not** remove atom map numbers, because the parameter is only supposed to affect the canonicalization process.

**Actual behavior:**  
Atom map numbers are removed even when `canonical=False` (no `:digit` in the first output line).

DavidACosgrove:
#### Reference Issue
Fixes #9225

As the issue says, the atom map numbers should not be removed  if not canonicalising.

### DIFF
--- a/Code/GraphMol/SmilesParse/catch_tests.cpp
+++ b/Code/GraphMol/SmilesParse/catch_tests.cpp
@@ -3027,8 +3027,8 @@ TEST_CASE("Ignore atom map numbers") {
   // if ignoreAtomMapNumbers is true as the latter should only apply if
   // canonicalising. (Github 9225).
   auto m3 = "c1ccc([NH2:1])cc1"_smiles;
-  CHECK(MolToSmiles(*m3, true, false, -1, false, false, false, false, true) ==
-        "c1ccc([NH2:1])cc1");
+  SmilesWriteParams params2{.canonical = false, .ignoreAtomMapNumbers = true};
+  CHECK(MolToSmiles(*m3, params2) == "c1ccc([NH2:1])cc1");
 }
 
 TEST_CASE("Github #7340", "[Reaction][CX][CXSmiles]") {
--- a/Code/GraphMol/SmilesParse/catch_tests.cpp
+++ b/Code/GraphMol/SmilesParse/catch_tests.cpp
@@ -3022,6 +3022,13 @@ TEST_CASE("Ignore atom map numbers") {
   CHECK(MolToSmiles(*m1, params) == MolToSmiles(*m2, params));
   CHECK(MolToSmiles(*m1, true, false, -1, true, false, false, false, true) ==
         MolToSmiles(*m2, true, false, -1, true, false, false, false, true));
+
+  // If not doing canonical SMILES, the map numbers should not be removed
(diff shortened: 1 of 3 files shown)


Code/GraphMol/SmilesParse/catch_tests.cpp
```diff
@@ -3022,6 +3022,13 @@ TEST_CASE("Ignore atom map numbers") {
   CHECK(MolToSmiles(*m1, params) == MolToSmiles(*m2, params));
   CHECK(MolToSmiles(*m1, true, false, -1, true, false, false, false, true) ==
         MolToSmiles(*m2, true, false, -1, true, false, false, false, true));
+
+  // If not doing canonical SMILES, the map numbers should not be removed
+  // if ignoreAtomMapNumbers is true as the latter should only apply if
+  // canonicalising. (Github 9225).
+  auto m3 = "c1ccc([NH2:1])cc1"_smiles;
+  CHECK(MolToSmiles(*m3, true, false, -1, false, false, false, false, true) ==
```
greglandrum: I know the current tests all use the unreadable stream of arguments, but there's no reason to continue that in newly added tests.
How about this instead?

```suggestion
  SmilesWriteParams params2{.canonical = false, .ignoreAtomMapNumbers = true};
  CHECK(MolToSmiles(*m1, params2) == "c1ccc([NH2:1])cc1");
```

greglandrum (changes requested):
Small suggested change to the test.
Since this is a definite change in behavior (I agree that it's the right thing to do), please add a comment to the Backwards Incompatible part of the release notes too.

Code/GraphMol/SmilesParse/catch_tests.cpp
DavidACosgrove: Force of habit and lazy cut-n-paste I guess, although CLion helpfully labels the parameters with the name in the prototype, so what I see is 
image
But clearly that is a luxury that I pay for.

greglandrum (approved):
LGTM

PR #9402, opened 2026-07-16
make the distgeom CI tests more verbose

greglandrum:
This is a temporary one to try and help track down the intermittent failure we're seeing with the new sharded tests


greglandrum:
I'm doing a self-merge on this one because there are no code differences and I hope the data it produces lets us track down and solve the intermittent failures.

PR #9409, opened 2026-07-17
make sure embedding runs in tests have RNG seeds set

greglandrum:
There were some instances in the tests where we were not setting the RNG seeds and this led to intermittent test failures in CI


greglandrum:
Apologies that code-formatting changes ended up making the diffs on this complicated.
I'm going to do a self-merge since this shouldn't be a controversial one.

PR #9411, opened 2026-07-17
Fix 64-bit fingerprint bit ID truncation in AdditionalOutput on 32-bit targets

gedeck:
We use RDKit in the browser as a wasm. In a recent change, I started to use `bitPaths`. The initial development was done in a 64-bit environment, but when moving the code changes into the WASM code, there were differences in the bits between the fingerprints and the bitPaths. Analysis showed that the differences are caused by converting the 64-bit fingerprint bits to 32-bit. This PR describes the issue and provides a solution. While there are no tests for this, I can confirm that applying these changes in our code base, resolved the problem. 

Widens the `bitId` parameter of `AtomEnvironment::updateAdditionalOutput` (the
virtual hook in `FingerprintGenerator.h`) and all four overrides — `AtomPair`,
`TopologicalTorsion`, `Morgan`, and `RDKitFP` — from `size_t` to
`std::uint64_t`.

On 32-bit targets (e.g. Emscripten/wasm32, any ILP32 build) `size_t` is 32-bit,
so 64-bit fingerprint bit IDs are truncated when routed into `AdditionalOutput`.
The `AdditionalOutput` containers key bit IDs as `std::uint64_t`
(`bitPathsType` / `bitInfoMapType` / `atomsPerBitType`, and the `atomToBits`
value type), and the fingerprint vector (`SparseIntVect<std::uint64_t>`, via
`res->setVal`) keeps the full 64-bit ID. The two therefore disagree: bit IDs
`>= 2^32` (e.g. `TopologicalTorsion<std::uint64_t>` torsions of length `>= 4`)
collide/shift in `bitPaths` et al. while `getNonzeroElements` keeps the full
value.

The call sites in `FingerprintGenerator.cpp` already pass a `std::uint64_t`
(`OutputType`); they previously narrowed into `size_t`. Native LP64 builds are
unaffected (there `size_t` is already 64-bit).

#### Any other comments?

- **No unit test is included**, because the bug is only observable on ILP32
  builds. On the 64-bit CI, `size_t` and `std::uint64_t` are identical, so the
  code path is bit-for-bit unchanged and any regression test would pass both
  before and after the fix. Happy to add a test if the maintainers can point me
  at an existing 32-bit / wasm test lane where it would be meaningful.
- The change is signature-only (9 lines across 6 files) and ABI-neutral on
  LP64.
- **AI tool acknowledgment:** this change was prepared with the help of Claude
  Code (Claude Opus). All code was reviewed by a human before submission.

### DIFF
--- a/Code/GraphMol/Fingerprints/AtomPairGenerator.cpp
+++ b/Code/GraphMol/Fingerprints/AtomPairGenerator.cpp
@@ -106,7 +106,7 @@ void AtomPairArguments::fromJSON(const boost::property_tree::ptree &pt) {
 
 template <typename OutputType>
 void AtomPairAtomEnv<OutputType>::updateAdditionalOutput(
-    AdditionalOutput *additionalOutput, size_t bitId) const {
+    AdditionalOutput *additionalOutput, std::uint64_t bitId) const {
   PRECONDITION(additionalOutput, "bad output pointer");
   if (additionalOutput->bitInfoMap) {
     (*additionalOutput->bitInfoMap)[bitId].emplace_back(d_atomIdFirst,
--- a/Code/GraphMol/Fingerprints/AtomPairGenerator.h
+++ b/Code/GraphMol/Fingerprints/AtomPairGenerator.h
@@ -108,7 +108,7 @@ class RDKIT_FINGERPRINTS_EXPORT AtomPairAtomEnv
       const std::uint64_t fpSize = 0  // unused
   ) const override;
   void updateAdditionalOutput(AdditionalOutput *output,
-                              size_t bitId) const override;
+                              std::uint64_t bitId) const override;
 
   /*!
     \brief construct a new AtomPairAtomEnv object
--- a/Code/GraphMol/Fingerprints/FingerprintGenerator.h
+++ b/Code/GraphMol/Fingerprints/FingerprintGenerator.h
@@ -148,7 +148,7 @@ class RDKIT_FINGERPRINTS_EXPORT AtomEnvironment : private boost::noncopyable {
                               const bool hashResults = false,
                               const std::uint64_t fpSize = 0) const = 0;
   virtual void updateAdditionalOutput(AdditionalOutput *AdditionalOutput,
-                                      size_t bitId) const = 0;
+                                      std::uint64_t bitId) const = 0;
 
   virtual ~AtomEnvironment() {}
 };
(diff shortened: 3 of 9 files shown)


greglandrum (approved):
LGTM

greglandrum:
I would not be terribly surprised if there were other bugs lurking on 32bit systems.

gedeck:
I used Claude to go through the code, it didn't find anything else.

# No other instances of this bug exist
The `size_t` truncation you fixed was the only real occurrence. Everywhere else a fingerprint bit ID crosses a width boundary, it's either already full-width `std::uint64_t`, or a value that's provably bounded below 2³² by design.

# The genuine 64-bit paths — all correctly full-width
- `SparseIntVect<uint64_t>` core (DataStructs/SparseIntVect.h) — keyed by `IndexType` throughout (storage, `get/setVal`, `getNonzeroElements`, Dice/Tversky similarity). No narrowing.
- `getUnfoldedRDKFingerprintMol` (Fingerprints.cpp:524-567) — uses `FingerprintGenerator<uint64_t>`, assigns `atomBits/bitInfo` as `uint64` → `uint64`.
- Python bindings (FingerprintGeneratorWrapper.cpp, SparseIntVect.cpp) and Java `SparseIntVect64` — boost/SWIG convert the `uint64_t` key to a native `int` without narrowing.
- Legacy topological-torsion (AtomPairs.cpp:254-256) — `getTopologicalTorsionCode` returns a real `uint64_t` (can exceed 2³²) and stores it into `SparseIntVect<int64_t>` with no cast.

# The narrow-typed sites — all bounded / by-design (NOT bugs)
- Folded fingerprints — `ExplicitBitVect/SparseBitVect`, `RDKFingerprintMol`, `LayeredFingerprint`, `PatternFingerprint` all fold bit `IDs % fpSize` (< 2³²) before storing. unsigned int indexing is correct here.
- Fingerprints.cpp:223/232 — worth a note: after your fix, `atomToBits`/`bitPaths` are now `uint64_t`, and these lines narrow them to `uint32_t`. But the source generator here is `FingerprintGenerator<std::uint32_t>` and the public API params are `uint32_t`, so the values are folded and the narrowing is lossless. Not a bug.
- Atom-pair codes (~23-27 bits), Morgan invariants (32-bit by design), MHFP MinHash (32-bit), atom indices, distances, popcounts — none are 64-bit IDs.
- 
# The wasm target specifically (where the bug bites)
MinimalLib (Code/MinimalLib/common.h) instantiates only `FingerprintGenerator<std::uint32_t>` and returns folded `ExplicitBitVect`, so no 64-bit ID is ever created there. Your fix matters for any native ILP32 build (or a wasm build that instantiates the uint64_t generator and reads AdditionalOutput), not the shipped MinimalLib.

PR #9413, opened 2026-07-18
test: increase DistGeomHelpers compareConfs tolerance to 0.1

### LINKED ISSUE #9406: RDKit 2026.03.4: testDistGeomHelpers fails
**Describe the problem**

While preparing an EasyBuild recipe for RDKit 2026.03.4, we see deterministic failures in `testDistGeomHelpers` on multiple HPC systems using a native, non-conda build with GCC 15.2.0.

The failures are in DistGeom/ETKDG regression checks. They appear to be platform-sensitive numerical/regression-test issues rather than general RDKit functionality failures: the installed RDKit distance-geometry API passes basic smoke tests, but the C++ regression test `testDistGeomHelpers` fails reproducibly on some CPU/toolchain combinations.

This looks related to earlier platform-sensitive DistGeom test behavior reported in #8790, but the affected test in RDKit 2026.03.4 is now `testDistGeomHelpers`. It may also be connected to the DistGeom/ETKDG test changes in #9320 and the platform-specific DistGeom tolerance work in #9328.

In our RDKit 2026.03.3 testing, `testDistGeomHelpers` passed in the same EasyBuild stack. The failure started showing up after updating to RDKit 2026.03.4.

**To Reproduce**

Configure and build RDKit 2026.03.4 in Release mode with GCC 15.2.0 and run:

```bash
cd <RDKit build directory>

export LD_LIBRARY_PATH=$PWD/lib:${LD_LIBRARY_PATH}
export RDBASE=$PWD
export PYTHONPATH=$PWD:$PYTHONPATH

ctest -R '^testDistGeomHelpers$' --output-on-failure -j 1
```

The failure is deterministic on affected systems. Re-running the same test serially gives the same failure pattern, so this does not appear to be caused by parallel `ctest` execution.

On one system, the failure is in `testEmbedParameters / predefined ETKDGv3 - macrocycle`:

```text
Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp:138: FAILED:
  CHECK( (pt1i - pt2i).length() < 0.05 )
with expansion:
  0.05164320170871522 < 0.05
```

On another system, the same check fails with a slightly larger deviation:

```text
Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp:138: FAILED:
  CHECK( (pt1i - pt2i).length() < 0.05 )
with expansion:
  0.06032537082286997 < 0.05
```

On AMD EPYC Milan / Zen3, we observed additional failures in the same `testDistGeomHelpers` executable:

```text
testSymmetryPruningAIO
Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp:1806: FAILED:
  CHECK( cids.size() == 8 )
with expansion:
  7 == 8
```

and:

```text
testGithub1990
  Reported Problem
Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp:1590: FAILED:
  CHECK( cid >= 0 )
with expansion:
  -1 >= 0
```

The ETKDGv3 macrocycle check also failed on the same AMD EPYC Milan / Zen3 system:

```text
Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp:138: FAILED:
  CHECK( (pt1i - pt2i).length() < 0.05 )
with expansion:
  0.05045033916696297 < 0.05
```

and:

```text
Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp:138: FAILED:
  CHECK( (pt1i - pt2i).length() < 0.05 )
with expansion:
  0.07004503389124658 < 0.05
```

Test reports from the EasyBuild PR are available here:

* https://gist.github.com/boegelbot/0194c4c19d9bcde682fd6658e957fdc4
* https://gist.github.com/pavelToman/1203ee51444ffb86b7af807e80008d9f

The EasyBuild easyconfig is here:

* https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/r/RDKit/RDKit-2026.03.4-foss-2026.1.eb

**Expected behavior**

`testDistGeomHelpers` should either pass across supported CPU/compiler combinations, or the affected checks should use tolerances/expectations that are robust to expected platform-dependent floating-point differences in optimized Release builds.

**Configuration**

* RDKit version: `2026.03.4`
* Build system: EasyBuild
* Toolchain: `foss/2026.1`
* Compiler: GCC `15.2.0`
* Python: `3.14.2`
* CMake: `4.2.1`
* Eigen: `3.4.1`
* Boost: `1.90.0`
* Build type: Release
* Conda: No, native build

The EasyBuild toolchain provides `-O2` and architecture-specific optimization flags such as `-march=native`.

We also tried overriding CMake/RDKit’s default Release `-O3` flags so that only `-DNDEBUG` is added by `CMAKE_{C,CXX}_FLAGS_RELEASE`:

```cmake
-DCMAKE_C_FLAGS_RELEASE="-DNDEBUG"
-DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG"
```

This removes the extra `-O3` added after EasyBuild’s `-O2`, but the failure is still reproducible on some CPUs.

**Additional context**

This appears to be a regression-test portability issue rather than a generally broken RDKit installation. The installed public API passes smoke tests for distance geometry, for example embedding a simple molecule with `AllChem.ETKDG()` succeeds.

For packaging in EasyBuild, we are currently planning to skip `testDistGeomHelpers` because it fails deterministically on some platforms.

younesouzza:
### Fix : #9406

This PR fixes the deterministic test failures in testDistGeomHelpers, specifically the predefined ETKDGv3 - macrocycle test, when RDKit is built with compiler optimizations such as -march=native.

Root cause

The compareConfs() helper compares generated 3D coordinates with a reference .mol file. When RDKit is compiled with -march=native, different compiler and CPU optimizations can introduce small floating-point differences in the generated coordinates. In my testing, the largest difference I observed was around 0.07 .

The current tolerance of 0.05 is slightly too strict, so these harmless numerical differences can cause the test to fail even though the generated conformations are still correct.

Change

This PR increases the tolerance in compareConfs() (in Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp) from 0.05 to 0.1.

A tolerance of 0.1 is enough to account for these small platform-dependent differences while still being strict enough to catch real regressions.

Testing

I reproduced the failure on Ubuntu 24.04 using GCC 13.3.0 with:

-DCMAKE_CXX_FLAGS_RELEASE="-O2 -march=native -DNDEBUG"

After increasing the tolerance to 0.1, all four testDistGeomHelpers shards pass locally.

### DIFF
--- a/Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp
+++ b/Code/GraphMol/DistGeomHelpers/testDgeomHelpers.cpp
@@ -135,7 +135,13 @@ void compareConfs(const RWMol *m, const RWMol *expected, int molConfId = -1,
 
     RDGeom::Point3D pt1i = conf1.getAtomPos(i);
     RDGeom::Point3D pt2i = conf2.getAtomPos(i);
-    CHECK((pt1i - pt2i).length() < 0.05);
+
+
+  // Increased tolerance from 0.05 to 0.1. 
+  // This prevents false test failures on systems using -march=native, 
+  // where compiler optimizations cause tiny, harmless math differences.
+  // (See issue #[9406])
+    CHECK((pt1i - pt2i).length() < 0.1);
   }
 }
 }  // namespace


greglandrum (approved):
LGTM

PR #9416, opened 2026-07-21
Crdgen 363 branch prioritisation

ZontaNicola:
#### Reference Issue

This introduces a new flag, on by default, to prioritise paths that extend chains in layout. When this is on, the random flipping to improve density is skipped, because the structure is already rationally-laid out. We then procede to solving clashes, if there are any, by flipping bonds along the clashing path. The quality of structures looks improved, with far fewer clashes. The total performance is better for average and even large molecules, but gets worse on huge ones. This is because the clash solution is not cheap. I think I can optimise it further, but will do so in a separate PR.

here's a comparison on the first 600 structures of ChmBl that were giving problems. The db is diverse, but a lot of structures that had problems due to chains get fixed or get much better in shorter or at least reasonable time
https://zontanicola.github.io/sketcher/galleries/chembl_34_no_branch_depth_failures_comparison.html


greglandrum:
I just looked about the first 100 structures and agree that these depictions are generally a lot better. The fact that the embedding is often quicker is an unexpected and very welcome bonus!

Something for future optimization efforts: you probably noticed this already, but it looks like big macrocycles that end up being rendered as circles tend to result in slowdowns relative to the current code.

greglandrum:
@ZontaNicola it looks like the test failures are real

ZontaNicola:
> @ZontaNicola it looks like the test failures are real

yes thank you, will address them and ping when ready for code review

External/ChemDraw/CMakeLists.txt
```diff
@@ -2,152 +2,114 @@ add_custom_target(chemdraw_support ALL)
 include(CMakePrintHelpers)
```
rachelnwalker, line 2: I think you can drop this file change it may be causing some of your CI failures

Code/GraphMol/Depictor/EmbeddedFrag.cpp
```diff
@@ -266,15 +294,76 @@ void EmbeddedFrag::updateNewNeighs(
...
+                       // Sort by score descending (highest score first)
+                       if (scoreA != scoreB) {
+                         return scoreA > scoreB;
+                       }
+
+                       // Final tie-break with CIP rank if available
+                       unsigned int rankA = 0, rankB = 0;
+                       dp_mol->getAtomWithIdx(a)->getPropIfPresent(RDKit::common_properties::_CIPRank, rankA);
+                       dp_mol->getAtomWithIdx(b)->getPropIfPresent(RDKit::common_properties::_CIPRank, rankB);
+                       return rankA < rankB;
```
rachelnwalker: _CIPRank is not set by modern stereo perception, so I don’t think we should rely on it for tie-breaking here. it looks like rankAtomsByRank() uses _ChiralAtomRank as a fallback, which I think should work. Maybe do a pre-sort with `rankAtomsByRank()` then just remove this final tie-break?

Code/GraphMol/Depictor/DepictUtils.cpp
```diff
@@ -622,6 +687,97 @@ template RDKit::INT_LIST rankAtomsByRank(const RDKit::ROMol &mol,
...
+  }
+
+  unsigned int count = 1;
+  visited.set(currentAtom);
+
+  const auto *atom = mol.getAtomWithIdx(currentAtom);
+  for (const auto *nbr : mol.atomNeighbors(atom)) {
+    unsigned int nbrIdx = nbr->getIdx();
+    if (!visited[nbrIdx]) {
+      count += countAcyclicSubtree(nbrIdx, visited, mol);
```
rachelnwalker: I am a little confused by this because you are counting subtrees here, but on line 729 you add this up as 'depth' - I think these are different, are you meaning to measure depth?

Code/GraphMol/Depictor/EmbeddedFrag.h
```diff
@@ -67,6 +68,7 @@ class RDKIT_DEPICTOR_EXPORT EmbeddedAtom {
     ccw = other.ccw;
     neighs = other.neighs;
     d_density = other.d_density;
+    d_distanceFromStart = other.d_distanceFromStart;
```
rachelnwalker, line 71: Should you also copy `d_isBalancedBifurcation`?

Code/GraphMol/Depictor/RDDepictor.cpp
rachelnwalker: isn't this what you mean to skip when `useBranchDepthPrioritization` is on?

Code/GraphMol/Depictor/DepictUtils.cpp
```diff
@@ -622,6 +687,97 @@ template RDKit::INT_LIST rankAtomsByRank(const RDKit::ROMol &mol,
...
+
+      unsigned int depth = countAcyclicSubtree(nbrIdx, visited, mol);
+      depthMap[std::make_pair(atomIdx, nbrIdx)] = depth;
+    }
+  }
+
+  return depthMap;
+}
+
+RDKit::INT_VECT sortNeighborsByDepth(const BRANCH_DEPTH_MAP &depthMap,
```
rachelnwalker: Is this function used?

rachelnwalker (changes requested):
I like that this promotes a linear molecule layout, which is something I think people would like for peptides. But I don't think we'd be able to turn this flag on by default without a ton of testing since it obsoletes an existing coordinate generation option and changes essentially all molecule layouts.

Code/GraphMol/Depictor/EmbeddedFrag.h
ZontaNicola: yes, thanks!

External/ChemDraw/CMakeLists.txt
ZontaNicola: yep, gone

Code/GraphMol/Depictor/EmbeddedFrag.cpp
ZontaNicola: sounds good to me

Code/GraphMol/Depictor/DepictUtils.cpp
ZontaNicola: removed unused function

Code/GraphMol/Depictor/DepictUtils.cpp
ZontaNicola: I think you are right

ZontaNicola: so does codex

Code/GraphMol/Depictor/RDDepictor.cpp
ZontaNicola: yep, skipping

PR #9105, opened 2026-02-11
Change SSSR finding algorithm (fixes #9064)

ricrogz:
This is a work in progress for a patch that updates the SSSR finding algorithm that fixes finding some extra rings which are being missed by the current algoritm.

An example of the issue with the current algorithm can be summarized in the following:

We want to find not just an SSSR, but also any extra rings with the same sizes as the ones in the SSSR. This is relevant to both symmetrizing the SSSR (for queries and such; this is the issue in #9064), but also for aromaticity (which causes some canonicalization issues). 

Some of these issues come from the fact that the modified Figueras algorithm we use to find rings starting from a degree 2  (D2) node can only store one parent for each atom in the graph. For example, for the following structure:

image

If we start searching for the smallest rings starting from atom 8, one path will go left, towards atom 7, and the other will go towards atom 9. The "left" branch will split in 2 at atom 6, and the right one will do the same at 10. Then, one of the 6 branch will traverse to 11, and the other one to 46. Then, the right branch will do something similar from 10, and both branches will meet at 11 (8-7-6-11 and 8-9-10-11, closing the ring), and the other will do at 46 (8-7-4-46 and 8-9-10-46).

But if the exploration starts from atom 7 (something similar would happen if we started from 9), the steps in the exploration would be:
7-6 (L); 7-8 (R); 6-11 (L1) and 6-46 (L2); 8-9 (R); 11-10 (L1); 46-10 (L2)

At this point L2 hits atom 10, which has been marked as already visited (GREY), as L1 saw it before, so we build the ring by backtracking the parents of the atoms in both paths. But the ring will fail the "Is the least common ancestor not the root?"  check in the  "backwards path" (see line 679 in 
https://github.com/rdkit/rdkit/blob/1159026a5dfdef2b5057901f9e2317c9bc723c4c/Code/GraphMol/FindRings.cpp), because atom 6 is common to both L1 and L2, so the ring containing atom 46 will be discarded, and the branch pruned. After that, the algorithm will continue, and L1 will meet R, and the ring through 10 will be found.

This patch aims to fixing this by 
both remembering more than one parent for each atom and by only considering rings that are formed by the collision of one "left" branch with a "right" one (otherwise, since starting point is D2, the ring won't contain it!).

See further explanations in the comments I left in the code.

Note that I'm still working on it, so there's a bunch of debug and testing code in there :)

PR #9422, opened 2026-07-22
Mols with lots of rings may trigger a stack overflow (and segfault)

ricrogz:
After https://github.com/rdkit/rdkit/pull/9105 `SymmSSSR() ` can find more rings than with the previous algorithm. In some pathological cases like the ones in some tests in `graphmolMolOpsTest`, we can see thousands of rings. This, besides making aromaticity detection way slower, may lead to a stack overflow failure due to the recursive algorithm used in `pickFusedRings()` due to the increased depth of the recursion caused by the many rings.

This is actually something that is currently happening in the Windows CI runners (Windows has a smaller stack size than linux and Mac, so it's easier to overflow). The failure is currently hidden by https://github.com/rdkit/rdkit/pull/9402 (the second `ctest` calls added there hide the return code of the first one, making the CI pass if the shard tests don't fail).

This patch refactors `pickFusedRings()` to use a queue-based approach instead of a recursive one, effectively avoiding the stack overflows.

The patch was first generated by AI, and then edited by me for clarity and fixing a few typos.

### DIFF
--- a/Code/GraphMol/Aromaticity.cpp
+++ b/Code/GraphMol/Aromaticity.cpp
@@ -77,7 +77,7 @@ void pickFusedRings(int curr, const INT_INT_VECT_MAP &neighMap, INT_VECT &res,
 
     // We haven't seen this ring yet, so add it to the stack
     pos = neighMap.find(neigh);
-    PRECONDITION(pos != neighMap.end(), "bad argument");
+    CHECK_INVARIANT(pos != neighMap.end(), "neighboring ring not found");
     done[neigh] = 1;
     res.push_back(neigh);
     stack.emplace_back(neigh, 0);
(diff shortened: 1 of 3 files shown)


Code/GraphMol/Aromaticity.cpp
```diff
@@ -41,17 +41,46 @@ namespace RingUtils {
...
+    // and that we haven't popped the stack!)
+    ++nextNeighbor;
+
+    if (done[neigh]) {
+      continue;
+    }
+
+    // We haven't seen this ring yet, so add it to the stack
+    pos = neighMap.find(neigh);
+    PRECONDITION(pos != neighMap.end(), "bad argument");
```
greglandrum:
```suggestion
    CHECK_INVARIANT(pos != neighMap.end(), "neighboring ring not found");
```

greglandrum (changes requested):
One minor suggestion

greglandrum (approved):
LGTM

PR #9423, opened 2026-07-23
split shard tests into separate block

greglandrum:
@ricrogz noticed that having two ctest runs (added in #9402) in a single block obscured failures in the first block.
This pulls the shard tests into a separate test block and removes the verbosity flag added to them.


ricrogz (approved):
LGTM

PR #9424, opened 2026-07-23
Fix RascalMCES crash

### LINKED ISSUE #9418: Rascal FindMCES crash when time out reached and maxFragSeparation set
**Describe the bug**

`FindMCES` crashes with a segmentation fault inside `RascalResult::applyMaxFragSep` when the clique search times out and `maxFragSeparation` is set to a positive value.

The constructor calls `applyMaxFragSep()` unconditionally whenever `d_maxFragSep != -1`, even when `d_timedOut == true`. With a partial clique from a timed-out search, the fragment count or atom indexing can be inconsistent between mol1 and mol2, leading to an out-of-bounds access or null dereference inside `applyMaxFragSep`.

Stack trace from the crash (signal 11):
```
libRDKitRascalMCES.so.1  RDKit::RascalMCES::RascalResult::applyMaxFragSep()+0x1a4
libRDKitRascalMCES.so.1  RDKit::RascalMCES::RascalResult::RascalResult(...)+0x278
libRDKitRascalMCES.so.1  RDKit::RascalMCES::findMCES(...)+0x1f7
libRDKitRascalMCES.so.1  RDKit::RascalMCES::rascalMCES(...)+0x41
rdRascalMCES.so          RDKit::findMCESWrapper(...)+0x199
```

Immediately before the crash the log prints:
```
[HH:MM:SS] Timed out after 61 seconds.
```

**To Reproduce**

This has proved difficult to reproduce outside the workflow I've created that uses this MCS search protocol. The crash is only encountered when the timeout is reached. I provided an example below for 2 large macrocyclic peptides that reproduce crash when the CPU load is high. However, I cannot reproduce this in an independent script.

```python
from rdkit.Chem import MolFromSmiles
from rdkit.Chem.rdRascalMCES import FindMCES, RascalOptions

mol1 = MolFromSmiles(
    "CC[C@H](C)[C@@H]1NC(=O)[C@H](C)N(C)C(=O)C[C@@H](C(=O)N2CCCC2)NC(=O)"
    "[C@H](C(C)C)N(C)C(=O)[C@H](Cc2ccccc2)N(C)C(=O)[C@H](Cc2c[nH]c3ccccc23)"
    "NC(=O)[C@H](Cc2cccc(Cl)c2)NC(=O)CN(C)C(=O)[C@H](Cc2ccc(Cl)cc2)N(C)C(=O)"
    "CN(C)C(=O)CN(C)C1=O"
)
mol2 = MolFromSmiles(
    "CC[C@H](C)[C@@H]1NC(=O)[C@H](C)N(C)C(=O)C[C@@H](C(=O)N2CCCC2)NC(=O)"
    "[C@H](CC(C)C)N(C)C(=O)[C@H](Cc2ccccc2)N(C)C(=O)[C@H](CC(C)C)NC(=O)"
    "[C@H](Cc2ccccc2)NC(=O)CN(C)C(=O)[C@H](Cc2ccc(Cl)cc2)N(C)C(=O)CN(C)"
    "C(=O)CN(C)C1=O"
)

opts = RascalOptions()
opts.timeout = 60            # seconds
opts.maxBondMatchPairs = 5000
opts.maxFragSeparation = 2   # required — crash does not occur when this is unset
opts.similarityThreshold = 0.5
opts.ignoreAtomAromaticity = True

result = FindMCES(mol1, mol2, opts)  # segfaults if the search times out
```

**Expected behavior**

`FindMCES` should return gracefully with a timed-out result (as it does when `maxFragSeparation` is not set), not crash. A timed-out clique is already incomplete, so skipping `applyMaxFragSep` on it is both safe and semantically correct.

**Screenshots**

N/A

**Configuration:**

- RDKit version: 2026.03.3
- OS: Ubuntu 22.04, Linux x86_64
- Python version: 3.11
- Are you using conda? No (installed as part of Schrödinger Suite)

**Additional context**

The bug is in `RascalResult.cpp`. The constructor calls `applyMaxFragSep()`
unconditionally:

```cpp
details::extractClique(clique, vtx_pairs, swapped, d_bondMatches);
matchCliqueAtoms(*mol1AdjMatrix);
if (d_maxFragSep != -1) {
    applyMaxFragSep();   // called even when d_timedOut == true
}
```

Proposed fix — guard on `d_timedOut`:

```cpp
if (!d_timedOut && d_maxFragSep != -1) {
    applyMaxFragSep();
}
```

DavidACosgrove:
#### Reference Issue
Fixes #9418

When checking maxFragSep, elements of frags1 and frags2 are reset, but the test for null pointers in subsequent loops was in the wrong place.

#### Any other comments?
The test uses different molecules from the original report, that were known to give a timeout.  Changes in LONG_TEST_mces_catch.cpp are not really relevant to the bug but seemed sensible in passing.

### DIFF
--- a/Code/GraphMol/RascalMCES/LONG_TEST_mces_catch.cpp
+++ b/Code/GraphMol/RascalMCES/LONG_TEST_mces_catch.cpp
@@ -115,7 +115,7 @@ TEST_CASE("timeout") {
     opts.maxBondMatchPairs = 2000;
     auto res = rascalMCES(*m1, *m2, opts);
     REQUIRE(res.size() == 1);
-    REQUIRE(res.front().getBondMatches().size() >= 39);
+    CHECK(res.front().getBondMatches().size() >= 39);
     check_smarts_ok(*m1, *m2, res.front());
   }
   {
@@ -124,7 +124,7 @@ TEST_CASE("timeout") {
     opts.maxBondMatchPairs = 2000;
     auto res = rascalMCES(*m1, *m2, opts);
     REQUIRE(res.size() == 1);
-    REQUIRE(res.front().getBondMatches().size() >= 44);
+    CHECK(res.front().getBondMatches().size() >= 44);
     check_smarts_ok(*m1, *m2, res.front());
   }
   {
@@ -133,7 +133,7 @@ TEST_CASE("timeout") {
     opts.maxBondMatchPairs = 2000;
     auto res = rascalMCES(*m1, *m2, opts);
     REQUIRE(res.size() == 1);
-    REQUIRE(res.front().getBondMatches().size() >= 44);
+    CHECK(res.front().getBondMatches().size() >= 44);
     check_smarts_ok(*m1, *m2, res.front());
   }
 }
--- a/Code/GraphMol/RascalMCES/RascalResult.cpp
+++ b/Code/GraphMol/RascalMCES/RascalResult.cpp
@@ -384,8 +384,8 @@ void RascalResult::matchCliqueAtoms(
 }
 
 void RascalResult::applyMaxFragSep() {
-  std::unique_ptr<RDKit::ROMol> mol1_frags(makeMolFrags(1));
-  auto frags1 = RDKit::MolOps::getMolFrags(*mol1_frags, false);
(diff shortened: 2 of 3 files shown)


greglandrum (approved):
LGTM

PR #9425, opened 2026-07-23
eased tolerances for DistGeomHelpers catch_tests

tadhurst-cdd:
#### Reference Issue
build of master fails for DistGeomHelpers/catch_test.cpp.  Two comparisons are slightly out of the expected range.
This local build was done on a Mac using VSCODE and build in a docker container.  
The container was spun up with:
NAME="Ubuntu"
VERSION_ID="24.04"

The failed test occur in the test "Github #9143: ETKDGv3 generating twisted amides"

The tolereances were loosened slightly

### DIFF
--- a/Code/GraphMol/DistGeomHelpers/catch_tests.cpp
+++ b/Code/GraphMol/DistGeomHelpers/catch_tests.cpp
@@ -1724,10 +1724,10 @@ TEST_CASE("Github #9143: ETKDGv3 generating twisted amides") {
                Catch::Matchers::WithinAbs(-180, 10) ||
                    Catch::Matchers::WithinAbs(180, 10));
     CHECK_THAT(MolTransforms::getDihedralDeg(conf, 31, 30, 28, 29),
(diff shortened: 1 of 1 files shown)


greglandrum:
> This local build was done on a Mac using VSCODE and build in a docker container. The container was spun up with: NAME="Ubuntu" VERSION_ID="24.04"

@tadhurst-cdd does this mean that the build and tests were running on an ARM CPU?
Can you please let me know which compiler was used?

tadhurst-cdd:
> > This local build was done on a Mac using VSCODE and build in a docker container. The container was spun up with: NAME="Ubuntu" VERSION_ID="24.04"
> 
> @tadhurst-cdd does this mean that the build and tests were running on an ARM CPU? Can you please let me know which compiler was used?

building is done in a docker container.
The container was spun up with:
NAME="Ubuntu"
VERSION_ID="24.04"

gcc is: 
gcc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0

